PR #635, opened 2011-01-31
When updating a document, $unset nil fields.

chanks:
When updating a document where some fields have been nullified, this uses the $unset operator to remove nil fields entirely. This ensures that fields with no content don't take up space unnecessarily in the db.

Also, there were some failures in the spec suite that were present both before and after my changes - I put them in a gist in case you were unaware of them:

https://gist.github.com/803439

Thoughts and changes are of course welcome :)

Thanks!

### DIFF
--- a/lib/mongoid/atomicity.rb
+++ b/lib/mongoid/atomicity.rb
@@ -55,6 +55,18 @@ module Mongoid #:nodoc:
           end
         end
         updates
+      end.tap do |updates|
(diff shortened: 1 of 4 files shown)


nofxx:
+1 this as a config or a plugin

```
never_write_null: true
```

tbh this should be default, null and 'doesnt exists' creates a different 'state' not needed (or worst: that you are unaware of)

PR #644, opened 2011-02-01
RE: Auto generated _id for embedded documents

hmorales:
Added new config option 'embedded_object_id' to allow disabling of auto generated object_id for all embedded documents. There are several use cases where there is no need to lookup an embedded object by _id. Having no way to disable this leads to bloated document size and potential issues integrating with other language ODM's such as Python MongoKit where embedded objects are not auto generated.

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -34,6 +34,7 @@ module Mongoid #:nodoc
       define_method("#{name}?") { send(name) }
     end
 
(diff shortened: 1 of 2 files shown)


ronakjain90:
I see that this issue is still not resolved. Any pointers on how do we get this working? Since I too feels that the _id is unnecessarily being generated, even if its an embeds-one document type. practically, its only increasing the document size, 

There should be an option at document level, to disable it.

PR #714, opened 2011-02-23
Chaining $in criteria intersects arguments

visnup:
I looked around in the current issues on github and dredged up this old thread:

http://groups.google.com/group/mongoid/browse_thread/thread/381e69a9be19f3d0/b82e7f79a79255e4?lnk=gst&q=%24in+intersection#b82e7f79a79255e4

nothing seemed conclusive. so I wanted to bring it up:

```
Person.where(:name.in => [ "Bob", "Mary" ]).where(:name.in => [ "Bob", "Joe" ])
```

to me should really be the AND of the two arrays, which means intersection. currently, mongoid does a union (and there's a test for it). I'm pretty sure this should be an intersection to be consistent with all other chained `.where()` calls, right?

### DIFF
--- a/lib/mongoid/criteria.rb
+++ b/lib/mongoid/criteria.rb
@@ -315,7 +315,9 @@ module Mongoid #:nodoc:
     # Example:
     #
     # <tt>criteria.update_selector({ :field => "value" }, "$in")</tt>
-    def update_selector(attributes, operator)
+    #
+    # @param [ Symbol ] combine The operator to use when combining sets.
+    def update_selector(attributes, operator, combine = :+)
       clone.tap do |crit|
         converted = BSON::ObjectId.convert(klass, attributes || {})
         converted.each do |key, value|
@@ -323,8 +325,7 @@ module Mongoid #:nodoc:
             crit.selector[key] = { operator => value }
           else
             if crit.selector[key].has_key?(operator)
-              op = operator == "$in" ? :& : :+
-              new_value = crit.selector[key].values.first.send(op, value)
+              new_value = crit.selector[key].values.first.send(combine, value)
               crit.selector[key] = { operator => new_value }
             else
               crit.selector[key][operator] = value
--- a/lib/mongoid/criterion/inclusion.rb
+++ b/lib/mongoid/criterion/inclusion.rb
@@ -20,6 +20,22 @@ module Mongoid #:nodoc:
       end
       alias :all_in :all
 
+      # Adds a criterion to the +Criteria+ that specifies values where any can
+      # be matched in order to return results. This is similar to an SQL "IN"
+      # clause. The MongoDB conditional operator that will be used is "$in".
+      # Any previously matching "$in" arrays will be unioned with new
+      # arguments.
+      #
+      # @example Adding the criterion.
+      #   criteria.in(:field => ["value1"]).also_in(:field => ["value2"])
+      #
+      # @param [ Hash ] attributes Name/value pairs any can match.
+      #
+      # @return [ Criteria ] A new criteria with the added selector.
+      def also_in(attributes = {})
+        update_selector(attributes, "$in")
+      end
+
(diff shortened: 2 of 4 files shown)


calfzhou:
Sorry to ping on such an old issue, but why with latest version, `Person.where(:name.in => [ "Bob", "Mary" ]).where(:name.in => [ "Bob", "Joe" ])` will get a selector `{"name"=>{"$in"=>["Bob", "Joe"]}}`? All former `where` on the same filed are overwritten.

calfzhou:
```ruby
[16] pry(main)> Person.where(:name.in => [ "Bob", "Mary" ]).where(:name.in => [ "Bob", "Joe" ])
=> #<Mongoid::Criteria
  selector: {"deleted_at"=>nil, "name"=>{"$in"=>["Bob", "Joe"]}}
  options:  {}
  class:    Person
  embedded: false>
```

cblage:
@durran what happened to the `also_in` operator? I can't seem to find it anymore :(

estolfo:
@cblag I'm not sure why exactly this was merged in or when it was taken out, but it hasn't been available for a long time. It wasn't even in Mongoid 3.0 (which used Origin >= 1.0.3: https://github.com/mongoid/origin/blob/v1.0.3/lib/origin/selectable.rb) 

Can you tell me more about the kind of query you're trying to do and what version of Mongoid you're using so I can help construct it using the current API?

cblage:
@estolfo so basically I have this query:

```
#<Mongoid::Criteria
  selector: {"is_hidden"=>{"$in"=>[false, nil]}, "advisor_id"=>BSON::ObjectId('5817d88f1961400001aad219')}
  options:  {}
  class:    Post
  embedded: false>
```

When I use directly `query.any_in(:is_hidden.in => [true, false, nil])`, I get the result I expect:
```
#<Mongoid::Criteria
  selector: {"is_hidden"=>{"$in"=>[true, false, nil]}, "advisor_id"=>BSON::ObjectId('5818a974834c8600012e7f80')}
  options:  {}
  class:    Post
  embedded: false>
```

But if I use a scope `query.visible_or_hidden`, with `scope :visible_or_hidden, -> { any_in(:is_hidden.in => [true, false, nil]) }`:

```
#<Mongoid::Criteria
  selector: {"is_hidden"=>{"$in"=>[false, nil]}, "advisor_id"=>BSON::ObjectId('5818a974834c8600012e7f80')}
  options:  {}
  class:    Post
  embedded: false>
```

The `is_hidden` condition isn't getting overridden when using the scope, but it does get overriden when using directly the `any_in` call. I need the latter behavior.

Help :)

cblage:
Also, I'm on Mongoid 5.1.5

estolfo:
Hi @cblage 

What version of Origin are you using?
Can you also provide a simple example of a Model with the scopes, how you are chaining them, and your expected result?

Chaining criteria with $in selectors uses the intersection of the $in lists. See this ticket: https://jira.mongodb.org/browse/MONGOID-4310

Thanks!
Emily

cblage:
I'm on Origin 2.2.2. The example would be too complex. Is there any way around using the intersections? (A la `also_in`?)

cblage:
Missed the @estolfo

estolfo:
Hi @cblage

You could tell the criteria to use the Union strategy for the selector merge:

```ruby
> crit = Band.in(name: ['test1','test2'])
=> #<Mongoid::Criteria
  selector: {"name"=>{"$in"=>["test1", "test2"]}}
  options:  {}
  class:    Band
  embedded: false>

> crit.union.in(name: ['test3','test4'])
=> #<Mongoid::Criteria
  selector: {"name"=>{"$in"=>["test1", "test2", "test3", "test4"]}}
  options:  {}
  class:    Band
  embedded: false>
```
Doe that work for your use case ?

cblage:
I'll try that out :)

PR #3468, opened 2013-12-22
Ensure timestamps of parent documents get updated

sinisterchipmunk:
As discussed in #1551 and #2852 (and possibly others), currently Mongoid won't update timestamps of parent documents if embedded documents are changed. I believe that if this is not a bug, at the very least it violates the [principle of least surprise](http://en.wikipedia.org/wiki/Principle_of_least_astonishment), because embedded documents are, by definition, part of the document that contains them. Updates to an embedded document are literally updates to the root document.

The current functionality presents a challenge for applications that use `updated_at` for cache expiry. This has presented logical issues for me in 3 separate Rails apps, and I had to hack around the real problem each time. The only place I could make a change, and not feel dirty, is in Mongoid itself.

This pull request also exposes and corrects (5cf9be1) a bug in `touch` which prior to this pull request would cause modified documents to be saved in their entirety when `touch` promises only to affect the `updated_at` field. The problem came down to the regular expression used to touch documents; when the optional field was `nil`, the regular expression became `/updated_at|/`, which is guaranteed to match all attribute names. Thus, `touch` would save all modified attributes and not just the `updated_at` field.

This PR ensures that parent documents get their timestamps updated whenever an embedded document is created, updated, deleted or touched. It also affects parents of parents, all the way up to the root document. The changes I made did not affect other tests, and they are all passing for me. Further, using this PR in my most recent Rails app has solved all of the issues I've been fighting with for several days.


bencrouse:
+1 would love to see this change. is there any way to help?

rpechayr:
+#

damnitjim:
+1 could really use this change too!

jwliechty:
So I solved this problem a while ago. I will share.

https://gist.github.com/jwliechty/eb7c94657bc6d6bf79f5#file-touch_parents_on_modify-rb

damnitjim:
Thanks for sharing jwliechty !

jonhyman:
@sinisterchipmunk can you fix the merge conflicts so we can ping arthurnn about merging this?

sinisterchipmunk:
@jonhyman Pulled from master, 1 conflict resolved. Hope we get to see this soon!

jonhyman:
Cool, @arthurnn thoughts on this? Seems like you agreed to it 15 months ago but then it got stale.

calfzhou:
@jwliechty your touch_parents_on_modify seems has a problem that even if embedded document validation failed, the invalid document will be saved into db.

johnnyshields:
I don't think that doing a separate atomic set is the right solution here. The root document timestamp should be modified in the same UPDATE query in which the embedded docs are modified.

johnnyshields:
@estolfo @durran I will take over this PR and get it to the finish line. I'd like to confirm the desired spec:

1) `belongs_to` with `touch: true` should use atomic `set` to update the timestamp on the parent.

2) `embeds_one/many` will have a new option `:touch`.
   - The default value will be `true`
   - The timestamp of the parent doc will be updated in a non-atomic fashion, because the entire doc needs to be saved in order to save the embedded doc. 
   - In case of nested embeds, the touch action will bubble up and touch every document along the way. For example, `parent embeds_many kids` and `kid embeds_many grandkids`, modifying a `grandkid` will touch both the `kid` and the `parent` in which it is embedded.

If these look good I will do #1 and #2 as two separate PRs.

johnnyshields:
Upon further investigation, #1 in my previous comment has been fixed already in this commit https://github.com/mongodb/mongoid/commit/16ad56e1009725bf466f303c2361a6be159ed5ce and specs have been added:

```
describe Mongoid::Relations::Touchable do
    ...
    context "when other document attributes have been changed" do
```

I will raise a PR for #2.

johnnyshields:
...and #2 is also supported as of Mongoid 4.0.0 in this commit https://github.com/mongodb/mongoid/commit/a94c2f43573e58f973913c881ad9d11d62bf857c

You need to explicitly define touch: true on your `embedded_in` call, which is probably good to make it off by default to avoid circular touching etc.

I'm going to backport it to Mongoid 3 in my [mongoid-monkey gem](https://github.com/johnnyshields/mongoid_monkey) for those interested.

@estolfo @durran please close this PR.

johnnyshields:
@estolfo please close this.

PR #3641, opened 2014-05-12
Removed everything related to making entire documents readonly...

edejong:
...based on selected fields. Perhaps I'm missing something here but the ReadonlyAttribute error already prevents us from making mistakes with partially loaded documents?


estolfo:
We've altered the behavior for when a document is read-only
See https://jira.mongodb.org/browse/MONGOID-3535

PR #3645, opened 2014-05-14
Add dry-run db:mongoid:list_indexes and db:mongoid:list_undefined_indexes Rake tasks

johnnyshields:
Useful when you want to check state of indices without modifying them.

Prints to console (via logger). Output format looks like:

```
ModelNameA
  _id_
  my_index
ModelNameB
  (none)
ModelNameC
  _id_
  my_other_index
```


estolfo:
Closing this, as there hasn't been discussion on it since 2014, nor any additional requests.

johnnyshields:
@estolfo I think we should definitely have this. Index migrations are a huge pain because they lock the DB... we need to see what is going to change first.

estolfo:
Hi @johnnyshields 
I don't see anything similar in Rails to this task. Do you know of anything?
This is pretty simple to write as a custom rake task and there aren't any thumbs up from other users so I would hesitate to add it to the codebase.

johnnyshields:
ActiveRecord and Mongoid migrations work a bit differently. Rails has `rake db:migrate:status` which will show which migrations are pending, based on your "migrations" table. What I'm proposing would be the Mongoid equivalent--since there's no "migrations" table, you need to do a diff between the indexes specified in your model code and the indexes which are actually in MongoDB.

estolfo:
Hi @johnnyshields 

Ok, I don't see the harm with putting this in then, as Rails has the db:migrate:status task. Would you mind updating the documentation to say it has been included in 6.0.0, and rebasing against master so the travis tests can run?

Thanks!
Emily

johnnyshields:
@estolfo I'll rebase this PR soon.

estolfo:
thanks @johnnyshields

estolfo:
@johnnyshields 
Would you like to rebase this? You can update documentation too to say since 5.2.0

johnnyshields:
Changes done, waiting for CI

johnnyshields:
Hmmm looking at this PR I think I want to change the implementation... what we really need is an index DIFF task which diffs indexes defined in mongoid with whats in the DB. Please hold off on merging I'll do the change soon.

johnnyshields:
Replaced by #4386

PR #3646, opened 2014-05-15
Fix for ticket # 3644

washu:
sorry for 2 files, pushed it up befor ei took out the comment. This will fix the above ticket issue.

### DIFF
--- a/spec/app/models/package.rb
+++ b/spec/app/models/package.rb
@@ -1,0 +1,6 @@
+# New Class created to isolate the test case of embedding with many referenced.
+class Package
+  include Mongoid::Document
+  embeds_one :package_item, class_name: 'PackageItem', inverse_of: :testable
+  has_and_belongs_to_many :package_containers
+end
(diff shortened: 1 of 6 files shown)


washu:
any idea if this wil be merged in?

spec/app/models/package.rb
```diff
@@ -0,0 +1,5 @@
+class Package
```
arthurnn: do we need these new classes? 
Can we use an existent model?

spec/mongoid/relations/referenced/many_spec.rb
```diff
@@ -3443,7 +3443,27 @@
       end
     end
   end
-
+  
+  context "when a document has embedded and references it saves correctly" do
+    let(:container) do
```
arthurnn: can you follow the spacing patterns from this test file?
thanks

estolfo:
no longer an issue: https://jira.mongodb.org/browse/MONGOID-3644

PR #3697, opened 2014-06-20
Make Hash.mongoize return a new Hash rather than modifying its argument

misfo:
This bug was causing confusing behavior that would prevent mongoid setter methods from being able to accept frozen Hashes:

    some_mongoid_model.my_attr = {a: 1}.freeze
    # kablammo! "Can't modify frozen Hash"


estolfo:
Fixed: https://jira.mongodb.org/browse/MONGOID-3697
https://github.com/mongodb/mongoid/blob/master/lib/mongoid/extensions/hash.rb#L201

PR #3709, opened 2014-06-26
(uninitialized constant for polymorphic) Mongoid does not specify the target class when using nested attributes

TheNaoX:
Hi guys!

I've been experiencing this issue when using nested attributes and polymorphic associations, the issue is described as follows:

``` ruby
class Product
  include Mongoid::Document
  belongs_to :shop, polymorphic: true
  accepts_nested_attributes_for :shop, autosave: true
end

class ShopOne
  include Mongoid::Document

  field :something, type: String

  has_one :product, as: :shop
end

class ShopTwo
  include Mongoid::Document

  field :another_thing, type: Integer

  has_one :product, as: :shop
end
```

And when I do this
``` ruby
Product.new(shop_type: "ShopOne", shop_attributes: { something: 'Foo' })
```

It gives me this
```ruby
uninitialized constant Shop
{... path}/mongoid/relations/builders/nested_attributes/one.rb:32
```

The solution I'm proposing is to override the class_name attribute inside of the metadata class only if the association you're trying to store is polymorphic (See code in the pull request).

### DIFF
--- a/lib/mongoid/attributes/nested.rb
+++ b/lib/mongoid/attributes/nested.rb
@@ -53,7 +53,7 @@ module Mongoid
             autosave_nested_attributes(metadata)
             re_define_method(meth) do |attrs|
               _assigning do
-                if metadata.polymorphic?
+                if metadata.polymorphic? and metadata.inverse_type
                   metadata.class_name = self.send(metadata.inverse_type)
                 end
                 metadata.nested_builder(attrs, options).build(self)
--- a/lib/mongoid/attributes/nested.rb
+++ b/lib/mongoid/attributes/nested.rb
@@ -53,6 +53,9 @@ module Mongoid
             autosave_nested_attributes(metadata)
             re_define_method(meth) do |attrs|
               _assigning do
+                if metadata.polymorphic?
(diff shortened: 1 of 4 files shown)


estolfo:
Thanks for this pull request but we went with another way of solving this. You can see the pull request here: https://github.com/mongodb/mongoid/pull/4190/files
It is preferable because it leaves the @ klass variable immutable and doesn't create a setter for the class_name on metadata.

TheNaoX:
That's great! I'm glad that it got fixed one way or another, I'm upgrading my mongoid gem to use this new code instead, on which version should I expect this change?

estolfo:
You can expect this to be in version 5.0.2, which is planned for release at the end of January. Here is the corresponding jira ticket, for your reference: https://jira.mongodb.org/browse/MONGOID-3709
We've just released Mongoid 5.0.1 earlier this week that has a number of bug fixes, if you want to try that out.
Thanks again!
Emily

TheNaoX:
Cool, I appreciate you letting me know.

Kudos for all the great work!

estolfo:
= )

PR #3733, opened 2014-07-11
Add find_by!

guilleiguaran:
This is an alternative version for find_by that raises error regardless of the value of Mongoid.raise_not_found_error

/cc @arthurnn


arthurnn:
thanks guys! sorry for the delay.

guilleiguaran:
Thanks @arthurnn

PR #3799, opened 2014-09-09
Move MONGODB_VERSION to version.rb

seuros:
Why do we even have this constant ? Should i remove it ?


estolfo:
MONGODB_VERSION is the minimum MongoDB version supported.
https://github.com/mongodb/mongoid/blob/master/lib/mongoid.rb#L39-L40

PR #3801, opened 2014-09-10
fix deleting document data when using atomically without any operations

xronos-i-am:
see example from my code. There is no one "record.set" some times. In these cases record's data is completely destroyed with "update( {} )"

```
atomically do |record|
  fields_to_update.each do |field, value|
    next unless value_should_be_changed?(value)
    record.set(field, value)
  end
end
```


estolfo:
Already changed: https://github.com/mongodb/mongoid/blob/master/lib/mongoid/persistable.rb#L210

PR #3831, opened 2014-10-13
Updated the Range extension to allow for String to be passed and parsed ...

mxaddict:
...as a RANGE


johnnyshields:
@estolfo I think this behavior is outside the scope of what Mongoid should do. Suggest to close this PR.

estolfo:
You can write your own #mongoize method for this.

PR #3864, opened 2014-11-18
Added limit functionality to geo_near.

vanboom:
geo_near would not support the limit function, so it would always return 100 matches.  This change allows you to chain the limit() function and specify the number of results for geo_near.


durran:
This functionality is already present.

```ruby
Bar.limit(200).geo_near([ 12, 50 ])
```

PR #3879, opened 2014-12-05
update process_attributes to be more clear as to what is happening

kyledecot:
I've made some changes to the `process_attributes` method that I feel make it a little more understandable. It is functionally the same (I think...could use a second pair of eyes though).


estolfo:
It has been a while since this PR so I'll close it. Feel free to rebase against master and open it again if you would still like the code to be updated.

estolfo:
no longer relevant

PR #3882, opened 2014-12-11
checks added for :update_attribute for dynamic attributes refs #3875

Shwetakale:
When we call process_attribute, it does all the checks for ensuring that
dynamic attributes are allowed.

**Note** This impacts the `:counter_cache` option as it requires dynamic fields.
When using the counter_cache, the model must include
`Mongoid::Attributes::Dynamic`

refs #3875

### DIFF
--- a/lib/mongoid/persistable/updatable.rb
+++ b/lib/mongoid/persistable/updatable.rb
@@ -29,11 +29,7 @@ module Mongoid
           raise Errors::ReadonlyAttribute.new(normalized, value)
         end
         setter = "#{normalized}="
-        if respond_to?(setter)
-          send(setter, value)
-        else
-          write_attribute(database_field_name(normalized), value)
-        end
+        process_attribute(normalized, value)
         save(validate: false)
       end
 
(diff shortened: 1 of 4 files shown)


Shwetakale:
@durran Any feedback on this PR? This bug can still reproduced against master.

estolfo:
@Shwetakale we still need to figure out how to allow counter_cache to continue to work with this change. Including `Mongoid::Attributes::Dynamic` is not a good idea as it'll have unexpected behavior.

Shwetakale:
@estolfo As per [this](https://github.com/mongoid/mongoid-site/issues/214), In case of counter cache field needs to be manually added in parent document and here as field is not added in parent_document I have added ```Mongoid::Attributes::Dynamic``` so should we add field in parent document instead of ```Mongoid::Attributes::Dynamic``` ?

estolfo:
Yes, we should add the attribute on the parent document and not rely on `Mongoid::Attributes::Dynamic`

Shwetakale:
@estolfo Ok I will update PR.

estolfo:
thanks!

Shwetakale:
@estolfo Closing this PR as created new one with updated code from mongoid master branch.

PR #3883, opened 2014-12-11
Updating mongoid.yml template

niedfelj:
Adding in all of the Moped/session configuration options...was having a hard time tracking these down.


arthurnn:
thanks

Issue #3889, opened 2014-12-16
set doesn't allow to set value of non existing field when included Mongoid::Attributes::Dynamic


Shwetakale:
My Game model is -

    class Game
       include Mongoid::Document
       include Mongoid::Attributes::Dynamic
       field :name, type: String
    end

Here i have included Attributes::Dynamic and when i try to set value of non existing field

    Game.last.set(no_of_players: 11)

It gives NoMethodError

    NoMethodError: undefined method `no_of_players=' for #<Game _id: 548fcea67368776fcf010000, name: nil>
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/attributes/dynamic.rb:136:in `method_missing'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable/settable.rb:25:in `block (2 levels) in set'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable.rb:172:in `block in process_atomic_operations'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable.rb:167:in `each'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable.rb:167:in `process_atomic_operations'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable/settable.rb:24:in `block in set'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable.rb:145:in `prepare_atomic_operation'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable/settable.rb:23:in `set'
	from (irb):1
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/railties-4.1.1/lib/rails/commands/console.rb:90:in `start'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/railties-4.1.1/lib/rails/commands/console.rb:9:in `start'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/railties-4.1.1/lib/rails/commands/commands_tasks.rb:69:in `console'
[...]

Issue #3888, opened 2014-12-16
set gives NoMethodError for non existing field instead of Mongoid::Errors::UnknownAttribute


Shwetakale:
My Game model is -

    class Game
      include Mongoid::Document
      field :name
    end

When i am trying to set value of non existing field 

    Game.last.set(no_of_players: 11)

It gives 'NoMethodError'

    NoMethodError: undefined method `no_of_players=' for #<Game _id: 548fcea67368776fcf010000, name: nil>
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable/settable.rb:25:in `block (2 levels) in set'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable.rb:172:in `block in process_atomic_operations'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable.rb:167:in `each'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable.rb:167:in `process_atomic_operations'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable/settable.rb:24:in `block in set'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable.rb:145:in `prepare_atomic_operation'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/mongoid-4.0.0/lib/mongoid/persistable/settable.rb:23:in `set'
	from (irb):5
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/railties-4.1.1/lib/rails/commands/console.rb:90:in `start'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/railties-4.1.1/lib/rails/commands/console.rb:9:in `start'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/railties-4.1.1/lib/rails/commands/commands_tasks.rb:69:in `console'
	from /home/shweta/.rvm/gems/ruby-2.1.0/gems/railties-4.1.1/lib/rails/commands/commands_tasks.rb:40:in `run_command!'
[...]

PR #3890, opened 2014-12-16
refs #3889 #3888 checks added for :set for dynamic attributes and error ...

Shwetakale:
...on non-existing field

When we call process_attribute, it does all the checks for ensuring that
dynamic attributes are allowed and it also gives 'Mongoid::Errors::UnknownAttribute' when Mongoid::Attributes::Dynamic is not included in model.

refs #3889 #3889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ For instructions on upgrading to newer versions, visit
 
 ### Resolved Issues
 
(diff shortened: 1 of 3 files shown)


arthurnn:
merged via https://github.com/mongoid/mongoid/commit/35aea60b285afce7a1d1f22602b14b21e91c08fa .
thanks

PR #3897, opened 2014-12-26
changes done as per ruby style guide

narutosanjiv:
1) Hash#has_key? has been depreciated in favour of Hash#key?
2) &block is unused variable , it would be better to name it starting with _.


arthurnn:
why is https://github.com/narutosanjiv/mongoid/commit/d0c30f4f382ba1f0a2867a0977892ecd8a5e847c necessary ?

narutosanjiv:
@arthurnn  It was necessary earlier as rspec was failing due to locale issue but something send same pull request so it got fixed.

PR #3901, opened 2014-12-30
Create the dependent behaviour restrict_with_error

maurogeorge:
Causes an error to be added to the owner if there are any associated
objects.

It is the same behaviour we have on [AR](http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many).


johnnyshields:
@maurogeorge can you please rebase this on latest master? It looks good to merge otherwise, just been almost two years since you raised.

johnnyshields:
@estolfo please close this because it's covered by #4402

PR #3914, opened 2015-01-10
Fix relations named "parent".

### LINKED ISSUE #3911: belongs_to relation with name "parent" causes update to wrong collection
We are upgrading to Mongoid 4.0.0 from Mongoid 3.1.4 (in the process of upgrading to Rails 4.1) and have spent over a day tracking down this extremely peculiar bug. It turns out that our existing relationship (named "parent") on one of our models, results in updates to that model being applied to the wrong collection (specifically the parent collection).

I pulled down Mongoid and wrote a spec test (forgive me if this is thrown together):

```ruby
  describe "child" do

    context "updating" do
      let(:parent) do
        TestParent.create(title: "parent")
      end

      let(:child) do
        TestChild.create(parent_id: parent.id, title: "original")
      end

      it "succeeds" do
        Mongoid.logger.level = Moped.logger.level = Logger::DEBUG

        # Pull the child from the db
        c = TestChild.find(child.id)

        # Access belongs_to association on the child
        # BUG: This relation access causes the problem
        temp = c.parent

        # Update attributes on the child
        params = { "title" => "updated" }
        c.attributes = params

        puts "child class = #{c.class}"
        puts "child collection = #{c.collection.name}"

        puts "calling c.save!"
        c.save!

        expect(c.title).to eq("updated")

        reloaded = TestChild.find(child.id)
        expect(reloaded.title).to eq("updated")
      end
    end
  end
```
test_parent.rb
```ruby
class TestParent
  include Mongoid::Document

  field :title, type: String
end
```
test_child.rb
```ruby
class TestChild
  include Mongoid::Document

  belongs_to :parent, class_name: "TestParent", index: true
  field :title, type: String
end
```

You can see in the output of this spec test, that the update is being applied to the parent collection instead of the collection the child is actually associated with:

```
D, [2015-01-08T11:21:32.039067 #28756] DEBUG -- :   MOPED: 127.0.0.1:27017 UPDATE       database=mongoid_test collection=test_parents selector={"_id"=>BSON::ObjectId('54aed8bc6a61637054010000')} update={"$set"=>{"title"=>"updated"}} flags=[]
```

It should be run against the "test_childs" collection instead. What is curious here is that the bug only occurs if you have accessed the relationship; if you comment out the line, things work fine for the update.

nkriege:
The @_parent variable defined in Traversable was colliding with the "@_#{name}" variables defined by Relations. Added an extra underscore to the name to protect against this.

Fixes #3911

### DIFF
--- a/lib/mongoid/traversable.rb
+++ b/lib/mongoid/traversable.rb
@@ -7,8 +7,12 @@ module Mongoid
   module Traversable
     extend ActiveSupport::Concern
 
-    included do
-      attr_accessor :_parent
+    def _parent
+      @__parent
+    end
+
+    def _parent=(p)
+      @__parent = p
     end
 
     # Get all child +Documents+ to this +Document+, going n levels deep if
--- a/spec/mongoid/relations/referenced/many_spec.rb
+++ b/spec/mongoid/relations/referenced/many_spec.rb
@@ -3555,4 +3555,27 @@ describe Mongoid::Relations::Referenced::Many do
       expect(person.posts.open).to eq([ post ])
     end
   end
+
+  context "when accessing a relation named parent" do
+    let!(:parent) do
+      Odd.create(name: "odd parent")
+    end
+
+    let(:child) do
+      Even.create(parent_id: parent.id, name: "original even child")
+    end
+
+    it "updates the child after accessing the parent" do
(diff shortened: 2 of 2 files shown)


arthurnn:
thanks

PR #3919, opened 2015-01-16
fix #symbolize_keys called on BSON::Document

topac:


arthurnn:
Why do you need this change? can you add a regression test? thanks

arthurnn:
This was fixed already on master.

PR #3922, opened 2015-01-19
Require rails only if Rails variable is present

swistaczek:
This change will allow to load this rake tasks outside Rails enviroment :).


arthurnn:
thanks

PR #3923, opened 2015-01-19
Converts preloaded models to their actual filename when loading them

benedikt:
According to the documentation `preload_models` is either a Boolean or an Array of models to preload. Unfortunatelly specifing the model name (`User`) would not cause proper preloading on case sensitive file systems.

This commit fixes this and also adds support for preloading camel cased model names by calling `.underscore` on the given model name.


arthurnn:
Can you add a changelog entry for this? thanks

PR #3935, opened 2015-02-01
fix contextual memory multi field sort spec

chamnap:
There is a bug on contextual memory when sorting multiple fields as mentioned in this issue, https://github.com/mongoid/mongoid/issues/3782.

This spec used to be fixed on https://github.com/mongoid/mongoid/pull/2664. However, it still got problem. Looking at the specs, `context.sort(number: 1, street: 1)` should sort by number first and only then by street. I just added a new address "Kampuchea Krom" to the test with a conflicting house number to verify the bug, and here is the result before this fix expects:

Street |Number
------------ | -------------
friedel|1
kampuchea-krom|5
pfluger|5
lenau|5
hobrecht|9

The expected result after the fix is then:

Street |Number
------------ | -------------
friedel|1
kampuchea-krom|5
lenau|5
pfluger|5
hobrecht|9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ For instructions on upgrading to newer versions, visit
 
 ### Resolved Issues
 
+* \#3935 fix multiple fields sorting on contextual memory. (chamnap)
+
 * \#3888 raise UnknownAttributeError when 'set' is called on non existing field and Mongoid::Attributes::Dynamic is not included in model. (Shweta Kale)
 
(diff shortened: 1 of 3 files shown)


arthurnn:
Thanks for the fix!
Can you add a changelog entry for it?

arthurnn:
thanks! 
merged via 44c711a1faa39b28508e0f06dfe6dcbf85be6970

chamnap:
:+1:

uchsarath:
Cheer, 1+

PR #3948, opened 2015-02-13
Fix remove_undefined_indexes.

amw:
Currently the task removes all of the indexes from Mongoid.models.
BSON::Document does not allow symbolic keys so we need to convert to
hash first.

### DIFF
--- a/lib/mongoid/tasks/database.rb
+++ b/lib/mongoid/tasks/database.rb
@@ -45,7 +45,8 @@ module Mongoid
             model.collection.indexes.each do |index|
               # ignore default index
               unless index['name'] == '_id_'
-                key = index['key'].symbolize_keys
+                # to_h because BSON::Document does not allow symbolic keys
+                key = index['key'].to_h.symbolize_keys
                 spec = model.index_specification(key)
                 unless spec
                   # index not specified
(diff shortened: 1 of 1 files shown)


bnorton:
+1

arthurnn:
I dont understand.
`collection.indexes` should return a list of `Moped::Indexes` , which are Hashes already.
```
[6] pry(main)> Problem.collection.indexes.first
=> {"v"=>1,
 "key"=>{"_id"=>1},
 "ns"=>"errbit_development.problems",
 "name"=>"_id_"}
[3] pry(main)> Problem.collection.indexes.first['key'].symbolize_keys
=> {:_id=>1}
```

bnorton:
This is not the full story here though I do think that this will fix it.

The issue is with getting the latest version (2.3.0) of BSON. It includes conversions that make a BSON::Document act like a indifferent access hash and `symbolize_keys` doesn't really conver the keys which makes the comparison to a BSON::Document with a Hash fail.

in `Mongoid::Tasks::Database.undefined_indexes`

```ruby
# With bson 2.2.5
    key = index['key'].symbolize_keys
    puts "Key class #{key.class} - Original class #{index['key'].class}"
# You get Hash and BSON::Document
```

```ruby
# With bson 2.3.0
    key = index['key'].symbolize_keys
    puts "Key class #{key.class} - Original class #{index['key'].class}"
# You get BSON::Document and BSON::Document
```

arthurnn:
I understand the problem now..
thanks for looking in this guys! :heart:

amw:
There is a test for remove_undefined_indexes that does not verify that defined indexes were not removed along with those deemed undefined (what happened to me because of this bug). I think we could enhance that test. Here's a link to the incomplete test:

https://github.com/mongoid/mongoid/blob/1f8e12c5b03da1c9bd49bab12d86bef969868fb3/spec/mongoid/tasks/database_spec.rb#L133

It should probably also say something along "expect defined_indexes to be present".

arthurnn:
Sounds good @amw , feel free to open a new PR with the tests.

Thanks

PR #3959, opened 2015-02-23
Remove Object::ClassMethods

akicho8:
https://github.com/mongoid/mongoid/issues/3951

### DIFF
--- a/lib/mongoid/extensions/object.rb
+++ b/lib/mongoid/extensions/object.rb
@@ -2,220 +2,221 @@
 module Mongoid
   module Extensions
     module Object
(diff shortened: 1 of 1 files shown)


arthurnn:
Can we just use `include ActiveSupport::Concern` and only include the module?

akicho8:
ActiveSupport::Concern is used in the following manner.

```ruby
require 'active_support/concern'

module M
  extend ActiveSupport::Concern
  class_methods do
    ...
  end
end
```

Since my english is not good, sorry if I have been a wrong interpretation.

durran:
Have #3990 now

PR #3962, opened 2015-02-24
don't try to update counter_cache on frozen records

zarqman:
Sometimes actions on nested records with counter caches can be triggered on already destroyed records. When records are frozen, avoid attempting to update the counter_cache column.

Without this, counter_cache.rb:119 causes RuntimeError: can't modify frozen Hash.

### DIFF
--- a/lib/mongoid/relations/counter_cache.rb
+++ b/lib/mongoid/relations/counter_cache.rb
@@ -116,7 +116,7 @@ module Mongoid
 
           before_destroy do
             if record = __send__(name)
-              record[cache_column] = (record[cache_column] || 0) - 1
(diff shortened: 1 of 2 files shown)


lib/mongoid/relations/counter_cache.rb
```diff
@@ -116,7 +116,7 @@ def add_counter_cache_callbacks(meta)
 
           before_destroy do
             if record = __send__(name)
-              record[cache_column] = (record[cache_column] || 0) - 1
+              record[cache_column] = (record[cache_column] || 0) - 1 unless record.frozen?
```
arthurnn: Maybe this `if` should be on line 118?

arthurnn:
Can you add a changelog entry for this change.

Issue #3937, opened 2015-02-02
Option to not use fallbacks for some specific localized field.


Exoth:
I think, that it would be great, if there is an option to disable fallbacks for some localized field, becuase sometimes I really need to get a value only if it's defined for current locale, and for other fields I need standard fallback behaviour. It could be like `field :foo, localize: true, fallbacks: false`

PR #3969, opened 2015-03-02
Add option to localized fields to disable fallbacks. #3937

brauliomartinezlm:
This PR addresses #3937 by adding the fallbacks options to localized fields.

The default behavior will stay the same, so if fallbacks is not specified, I18n.fallbacks will be used as usual.


brauliomartinezlm:
Build failed. Locally I don't have any failures though. As far as I understand the failure is not related to changes.

johnnyshields:
+1

estolfo:
I didn't see that this pull request was still here. I've merged in a new pull request, which is basically the same as this one. Thank you!

PR #3985, opened 2015-03-13
Return nil when using {upsert: true} in find_and_modify

Intrepidd:
I guess this is related to mongoid/moped#83

When using ``{upsert: true}``, ``result`` will return ``{}`` as seen [here](https://github.com/mongoid/moped/blob/b9ab9e0adc9c83cef1eb08e88fae067346e2324b/lib/moped/query.rb#L210)

This PR assesses this behavior.

Without it, when using upsert, even if no document was present before, Mongoid will return an empty document.

Cheers


arthurnn:
Can you add a changelog entry for this?
thanks

Intrepidd:
@arthurnn Just pushed it !

arthurnn:
thanks

PR #3990, opened 2015-03-17
Bugfix conflict active_support 4.2.0 concern class_methods

akicho8:
And tests to reproduce the bug, I wrote the code to fix the bug.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,8 @@ source "https://rubygems.org"
 gemspec
 
 gem "rake"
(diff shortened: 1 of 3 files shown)


akicho8:
Due to the origin gem 2.1.1, Travis CI failed.
What can I do. :disappointed_relieved:

akicho8:
If the https://github.com/mongoid/origin/pull/107 to origin not have my merge, it is not this bug is healed.

PR #3992, opened 2015-03-17
Fixes #3991

### LINKED ISSUE #3991: You can't add multiple embedded documents that are in embedded documents
```ruby
class Test1
  include Mongoid::Document
  embeds_many :test2
end

class Test2
  include Mongoid::Document
  embedded_in :test1
  embeds_many :test3
end

class Test3
  include Mongoid::Document
  embedded_in :test2
end

# This all works fine
t = Test1.create
t.test2 << Test2.new
t.test2.first.test3.new
t.changed?
# => true
t.save

# This will now do nothing
t.test2.first.test3.new
t.changed?
# => false
t.test2.first.changed?
# => false
t.save
# returns true, yet it does nothing -- this is extremely dangerous/bad
t.test2.first.test3.last.persisted?
# => false
```

jonhyman:
@arthurnn can you review soon and release 4.0.3? This is a really bad issue.

### DIFF
--- a/lib/mongoid/traversable.rb
+++ b/lib/mongoid/traversable.rb
@@ -27,7 +27,7 @@ module Mongoid
     #
     # @return [ Array<Document> ] All child documents in the hierarchy.
     def _children
-      collect_children
+      @__children ||= collect_children
     end
 
     # Collect all the children of this document.
@@ -124,7 +124,6 @@ module Mongoid
         child.move_changes
         child.new_record = false
       end
-      @__children = nil
     end
 
     # Return the root document in the object graph. If the current document
--- a/lib/mongoid/traversable.rb
+++ b/lib/mongoid/traversable.rb
@@ -27,7 +27,7 @@ module Mongoid
     #
     # @return [ Array<Document> ] All child documents in the hierarchy.
     def _children
-      @__children ||= collect_children
+      collect_children
     end
 
     # Collect all the children of this document.
--- a/spec/mongoid/changeable_spec.rb
+++ b/spec/mongoid/changeable_spec.rb
@@ -816,6 +816,30 @@ describe Mongoid::Changeable do
       end
     end
 
+    context "when changed? has been called before and a deeply embedded child adds extra elements" do
+
+      let(:person) do
+        Person.create
+      end
+
+      let(:address) do
+        person.addresses.create(street: "hobrecht")
+      end
+
+      let!(:location) do
+        address.locations.create(name: "home")
+      end
+
+      before do
+        person.changed?
+        address.locations.new
+      end
+
+      it "returns true" do
+        expect(person).to be_changed
+      end
+    end
+
     context "when a child is new" do
 
       let(:person) do


arthurnn:
tests are broken..

jonhyman:
@arthurnn, tests all pass for me locally. Might those specs just be fragile? I have a hard time imagining t this change caused the test `Mongoid::Criteria::Modifiable#update when updating a relation when the relation is a references many when updating the relation directly does not update the last document` to fail.

jonhyman:
@arthurnn I just ran the spec that Travis said failed across multiple versions of Ruby, all passes. Can you kick the build on this branch or something? Or, run it yourself and you'll see that they pass, then merge?

jonhyman:
@arthurnn can you kick the tests and/or inspect yourself? As I explained, I really don't think those test failures are caused by this commit. I ran the specs across 5 different versions of Ruby and it all passes, and the spec which failed seems extremely unrelated and I bet it's just fragility, as some of the test runs were entirely green.

arthurnn:
I am wondering if it could be a mongodb version thing, which mongodb version are you running the tests?

jonhyman:
I've run on both 2.4.10 and 2.6.5

lib/mongoid/traversable.rb
```diff
@@ -27,7 +27,7 @@ def _parent=(p)
     #
     # @return [ Array<Document> ] All child documents in the hierarchy.
     def _children
-      @__children ||= collect_children
```
arthurnn: I am not sure if we can remove the memoization , see the comment
> This operation can get expensive in domains with large hierarchies.

arthurnn: see https://github.com/mongoid/mongoid/commit/1602bbc713d0f5461f5cce3497df5545e70671de

jonhyman: The memoization is what is causing the problems, though. Did you read the issue at https://github.com/mongoid/mongoid/issues/3991? There are myriad situations where the memoization can screw you, and the downside is horrible: Mongoid won't generate an UPDATE command and will silently fail to save documents, while returning `true` to the `save` command.

jonhyman: The performance improvement is not worth the danger of the example in #3991. If you think the memoization is highly important, I'd recommend merging this, releasing the bugfix, and then adding memoization back in an improved way as a performance improvement dot release. As it is now, silently failing is terrible.

arthurnn: Yes, thats indeed a problem, however I am not sure removing the memoization is the right solution.
We can clear that once we do a new change on the root document.

jonhyman: Yeah, that can work, you just need to make sure you think through all the cases of where this can change (new/create/delete -- does this need to happen for foreign key relations as well or just embedded, etc.) and I imagine clear the memoization all the way up to the root. I'm not familiar enough with the internals to sketch out all the scenarios in which this can happen and feel comfortable that I've addressed them all, though.

jonhyman:
@arthurnn so what's the plan here? Can you attempt to fix in the "proper" way by clearing the cache? I think it's negligent to keep this bug outstanding for a long time, it's rather dangerous.

PR #3994, opened 2015-03-17
Fixes #3993

### LINKED ISSUE #3993: dup fails for embedded documents that use store_as without using Mongoid::Attributes::Dynamic
```ruby
class Test1
  include Mongoid::Document
  embeds_one :test2, :store_as => :t
  field :name
end

class Test2
  include Mongoid::Document
  embedded_in :test1
  field :name
end

t = Test1.new(:name => "hi")
t.build_test2(:name => "there")
t.save

# Will throw an error
t.dup()

# Mongoid::Errors::UnknownAttribute:
# Problem:
#  Attempted to set a value for 't' which is not allowed on the model Test1.
# Summary:
#  Without including Mongoid::Attributes::Dynamic in your model and the attribute does not already exist in the attributes hash, attempting to call Test1#t= for it is not allowed. This is also triggered by passing the attribute to any method that accepts an attributes hash, and is raised instead of getting a NoMethodError.
# Resolution:
#  You can include Mongoid::Attributes::Dynamic if you expect to be writing values for undefined fields often.
```

jonhyman:


jonhyman:
@arthurnn I can't get this test to fail locally, either, but Travis says it failed a spec on one of the builds (`Mongoid::Scopable.default_scope when there is more then one default_scope delays execution of the merge until called`).

jonhyman:
@arthurnn thoughts on merging this in?

spec/mongoid/copyable_spec.rb
```diff
@@ -54,6 +54,30 @@
         end
       end
 
+      context "when using store_as" do
+        context "and dynamic attributes are not set" do
```
arthurnn: can you add space here and in the next line, to keep the file style consistent ?

spec/mongoid/copyable_spec.rb
```diff
@@ -54,6 +54,30 @@
...
+            class StoreAsDupTest2
+              include Mongoid::Document
+              embedded_in :store_as_dup_test1
+              field :name
+            end
+
+            t = StoreAsDupTest1.new(:name => "hi")
+            t.build_store_as_dup_test2(:name => "there")
+            t.save
+            copy = t.__send__(method)
```
arthurnn: can you use a normal `send` call here? instead of `__send__`

spec/mongoid/copyable_spec.rb
```diff
@@ -54,6 +54,30 @@
         end
       end
 
+      context "when using store_as" do
+        context "and dynamic attributes are not set" do
+          it "clones" do
+            class StoreAsDupTest1
```
arthurnn: I think it is better to define this classes outside this test, as this is in a loop.

arthurnn:
Can you also amend commit a changelog entry for this bug fix?
thanks

spec/mongoid/copyable_spec.rb
jonhyman: Sure

spec/mongoid/copyable_spec.rb
jonhyman: Will do, I'll move to their own files.

spec/mongoid/copyable_spec.rb
jonhyman: Yep.

jonhyman:
@arthurnn please check now and let me know your comments. Thanks!

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ For instructions on upgrading to newer versions, visit
 
 * \#3985 Return nil when using `{upsert: true}` in `find_and_modify` (Adrien Siami)
 
+### Resolved Issues
+
+* \##3993 dup/clone fails for embedded documents that use store_as without using Mongoid::Atributes::Dynamic
```
arthurnn: There is an extra `#` .

The changelog, probably should say:
>  Fix an issue where `dup/clone` fails....

jonhyman: Done

jonhyman:
@arthurnn can this be merged?

arthurnn:
thanks

PR #3997, opened 2015-03-18
add ignore_type_attribute to configuration

l8nite:
Setting `Mongoid::Configuration.ignore_type_attribute=true` will cause Mongoid to ignore the _type
attribute on queries where normal inheritance rules would apply that criteria, and will also ignore _type when marshaling objects from the database (preferring to use the calling class instead)

This fixes a situation I encountered wherein I was versioning document models by incrementing the module namespace they were a part of, but I wanted my existing documents to continue being fetched/queried by the new V2 document model.

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -18,6 +18,7 @@ module Mongoid
 
     option :include_root_in_json, default: false
     option :include_type_for_serialization, default: false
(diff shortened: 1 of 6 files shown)


arthurnn:
Not a big fan of options. I think this is a 10-20% use , so I :-1: on it.
thanks anyways.

l8nite:
Bummer.

PR #4020, opened 2015-04-01
UnknownAttribute Exception when cloning documents containing deprecated attributes

kmewhort:
I'm not certain if this is newly intended behaviour, but now an UnknownAttribute exception is thrown whenever you call `.clone` on a document that contains deprecated data (that is, it was assigned data under an old schema containing that attribute, but the attribute has since been removed from the model).

I find one of the major advantages of Mongo is the flexible data model that allows for easy changes to the schema while retaining deprecated data, so hopefully this isn't the intended behaviour!  Explicitly including Dynamic::Attributes would solve the issue, but this would undesirably throw-away enforcement of the schema on new documents.  IMO, when cloning old documents, the deprecated attributes should be copied or possibly silently ignored.

This PR fixes the issue by copying them, but let me know if a different approach would be favourable...

### DIFF
--- a/lib/mongoid/attributes/processing.rb
+++ b/lib/mongoid/attributes/processing.rb
@@ -16,13 +16,13 @@ module Mongoid
       # @param [ Hash ] attrs The attributes to set.
       #
       # @since 2.0.0.rc.7
-      def process_attributes(attrs = nil)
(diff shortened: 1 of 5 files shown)


estolfo:
fixed as of https://github.com/mongodb/mongoid/pull/4278

PR #4025, opened 2015-04-05
Fix regression I caused in 9e5a279f306db9905f7cf331c6915a4ff24518f9

jonhyman:
@arthurnn this is the cause of my problems I emailed out at https://groups.google.com/forum/#!topic/mongoid/Wby7De2Ug0Q.

### DIFF
--- a/lib/mongoid/attributes/processing.rb
+++ b/lib/mongoid/attributes/processing.rb
@@ -92,7 +92,8 @@ module Mongoid
       #
       # @since 2.0.0.rc.7
       def process_attribute(name, value)
-        if store_as = aliased_fields.invert[name.to_s]
+        responds = respond_to?("#{name}=")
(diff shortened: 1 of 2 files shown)


lib/mongoid/attributes/processing.rb
```diff
@@ -92,7 +92,8 @@ def pending_nested
       #
       # @since 2.0.0.rc.7
       def process_attribute(name, value)
-        if store_as = aliased_fields.invert[name.to_s]
+        responds = respond_to?("#{name}=")
+        if !responds && store_as = aliased_fields.invert[name.to_s]
```
arthurnn: why not call `respond_to?("#{name}=")` from the `if` line?

jonhyman: good call, updated

arthurnn:
thanks

PR #4046, opened 2015-04-23
Typo in Mongoid::Errors::Validations doc block

maximeg:


durran:
That's not a typo, the class takes a document.

maximeg:
Yup, the `@example` is wrong.

PR #4058, opened 2015-05-04
ensure .heriditary? returns true or false, not nil.

ifightcrime:
`.hereditary?` returns nil for the parent. Small, but caused an issue for me. Fixed tests to ensure strict true/false checking.


durran:
Thanks.

PR #4081, opened 2015-05-28
Updates `mongoid.yml` generator for mongo driver

nofxx:
Updates the generator for mongo driver options:
Adds 'mode' on 'read' (maybe also add a commented :tag_sets later?)
Keeps yml structure on 'write'.


durran:
Thanks!

PR #4084, opened 2015-05-31
Don't throw exception on empty localized fields when calling valid?

cammellos:
I have came across an issue when trying to persists localized documents:

I have a model with: 

 `field :name, type: String, localize: true`
 `validates_uniqueness_of :name`

And when I run Model.new.valid?

I get: 

```NoMethodError: undefined method `inject' for nil:NilClass` in `lib/mongoid/validatable/uniqueness.rb:126:in `criterion'````

I have wrote a small test for it and changed the behaviour of the method so that `valid?` would return false in this case. Feel free to merge if it's any useful. 
Thanks for the library.
cheers

### DIFF
--- a/lib/mongoid/validatable/uniqueness.rb
+++ b/lib/mongoid/validatable/uniqueness.rb
@@ -123,7 +123,7 @@ module Mongoid
         field = document.database_field_name(attribute)
 
         if localized?(document, field)
-          conditions = value.inject([]) { |acc, (k,v)| acc << { "#{field}.#{k}" => filter(v) } }
+          conditions = (value || {}).inject([]) { |acc, (k,v)| acc << { "#{field}.#{k}" => filter(v) } }
           selector = { "$or" => conditions }
         else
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by johnnyshields** [comment]
:+1:

**Comment by estolfo** [comment]
I'm not finding that it throws an error anymore:
```ruby
class Band
  include Mongoid::Document

  field :name, type: String, localize: true
  validates_uniqueness_of :name
end

[4] pry(main)> Band.new.valid?
D, [2016-04-26T16:48:18.894681 #99196] DEBUG -- : MONGODB | localhost:27017 | mongoid_test.find | STARTED | {"find"=>"bands", "filter"=>{"name.en"=>nil}, "projection"=>{"_id"=>1}, "limit"=>1}
D, [2016-04-26T16:48:18.895689 #99196] DEBUG -- : MONGODB | localhost:27017 | mongoid_test.find | SUCCEEDED | 0.0007880000000000001s
=> false
```

PR #4110, opened 2015-06-30
Fix created_at being set to nil during a non-timeless call

kcdragon:
If an existing document has `timeless.save` called on it and then a document in that collection is created with `Collection.create`, the second document will not get a `created_at` timestamp.

The original implementation assumed that `clear_timeless_option` would be called once for each timestamp module it included. However, it is only called once during an update since the create callback does not get called.

This implementation adds an alternate `clear_timeless_option` that will get called if the document is already persisted and decrements the counter for each timestamp module that is included since it will only be called once (by the update callback).

### DIFF
--- a/lib/mongoid/timestamps/created.rb
+++ b/lib/mongoid/timestamps/created.rb
@@ -27,7 +27,7 @@ module Mongoid
           self.created_at = time
         end
 
-        self.class.clear_timeless_option
+        clear_timeless_option
       end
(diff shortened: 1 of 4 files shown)


estolfo:
Incorporated this fix as of https://github.com/mongodb/mongoid/commit/095433fa53fc5f770fedb8d28b57fa33b2b2c4dc

PR #4116, opened 2015-07-04
Relations validations bugfix 2015 07 04

EhsanYousefi:
This bugfix is related to this issue: https://github.com/mongoid/mongoid/issues/4114.


estolfo:
I don't see what bug this is fixing. Please reopen if you can provide more context for why this is needed.

PR #4144, opened 2015-08-18
Backport "Remove all uses of atomic $ operator in updates." to 4.0.0-stable

### LINKED ISSUE #4033: Saving embedded grandchildren creates additional embedded documents
I ran into an issue where I have a model that `embeds_many` of another model, that has it's own `embeds_many`.

For example:
```yaml
parent:
  - child_1:
  - child_2:
    - grandchild
```

The issue comes up when you update one of the grandchildren. It saves the grandchild just fine but also creates a additional grandchild record. The number of records that get created is directly tied to how many children there are. Three children will create three grandchild records. If there is only one child record, there is no problem.

I created a repo with specs do demonstrate this: https://github.com/adamcrown/grandchild-save-bug

The best workaround I've found is to call save on the child instead of the grandchild.

_For some reason I can't get Travis CI to reproduce the bug, but I've reproduced it on multiple computers. Please clone the repo and let me know if it passes for you. There should be one failed and one passed spec._

simi:
The operator regressed in MongoDB 2.6 and higher and simply does not
work when used in conjunction with embedded documents more than one
level deep. This removes our use of it and replaces it with the common
index positions.

- Fix #4033

Conflicts:
	CHANGELOG.md
	lib/mongoid/contextual/memory.rb
	lib/mongoid/persistable.rb
	lib/mongoid/persistable/creatable.rb
	lib/mongoid/persistable/deletable.rb
	lib/mongoid/persistable/updatable.rb
	lib/mongoid/relations/embedded/batchable.rb
	lib/mongoid/relations/touchable.rb


muichkine:
:+1:

PR #4146, opened 2015-09-03
Provide read-access to problem, summary and resolution fields for Mongoid::Errors::MongoidError

rpocklin:
In some cases, I want to catch Mongoid::Errors::MongoidError and include the **problem** field in my own exception, without the **summary** or **exception** fields.  Currently i'm using `problem = exception.instance_variable_get(:@problem)` and I can't think of a reason why these fields shouldn't be readable from outside the class.

### DIFF
--- a/lib/mongoid/errors/mongoid_error.rb
+++ b/lib/mongoid/errors/mongoid_error.rb
@@ -7,6 +7,8 @@ module Mongoid
     # translating the messages.
     class MongoidError < StandardError
 
+      attr_reader :problem, :summary, :resolution
+
       BASE_KEY = "mongoid.errors.messages"
 
       # Compose the message.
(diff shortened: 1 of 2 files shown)


rpocklin:
Build failures are because master is currently broken (unrelated to my changes).  Will rebase once it is green.

durran:
Can you change the names of the methods back to `problem`, `summary` and `resolution` please? They don't actually set anything on the instance.

rpocklin:
Ruby doesn't support method overloading AFAIK so the methods needed renaming otherwise they would clash with the (added) public reader methods.  I renamed `set_*` to `translate_*` which makes more sense.

durran:
Thanks.

PR #4147, opened 2015-09-10
Fix Fields::Localized#lookup to work with boolean and empty keys

fernandokosh:
When the translated has a key, but with empty value, mongo will return it empty, brooking the fallback.

Eg.
p = Post.find 'some_id'
p.title_translations
=> {"de"=>"", "en"=>"Bibliography", "es"=>"", "fr"=>"", "it"=>"", "pt-BR"=>""}

If set I18n.locale = :de will return p.title_translations['de'] = blank

### DIFF
--- a/lib/mongoid/fields/localized.rb
+++ b/lib/mongoid/fields/localized.rb
@@ -63,7 +63,13 @@ module Mongoid
       def lookup(object)
         locale = ::I18n.locale
         if ::I18n.respond_to?(:fallbacks)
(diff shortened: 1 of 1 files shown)


durran:
Could you add a test for this as well please?

johnnyshields:
I think the problem here is that locales are being persisted as empty strings.

I would propose an alternative fix whereby we remove empty string keys when persisting the localized document. That way would save space in the database.

estolfo:
Addressed by this ticket: https://jira.mongodb.org/browse/MONGOID-3490

PR #4150, opened 2015-09-15
MONGOID-3365 Prove that querying by Time field works as designed

estolfo:


spec/app/models/account.rb
```diff
@@ -25,4 +25,12 @@ class Account
   def overridden
     self[:overridden] = "not recommended"
   end
+
+  # MONGOID-3365
```
simi: Is this comment needed?

estolfo: These models are used in multiple tests and each field has different meanings for each test it's used in.

This is a comment and thus is completely harmless. It helps any future reader of the code know what that field or model behavior is meant to test. Otherwise, the reader will have to do git blame or look at the file history to figure out why this field/behavior is included in this Model.

How distracting is it? I personally find it helpful but if it makes the code difficult to read, I'm happy to take it out.

simi: I think someone can be confused by ```MONGOID-3365``` comment easily, since there's no chance to understand it without knowledge of some context (for example I know this is JIRA id).

If you want to keep reference for this change, it will be better to add direct link to JIRA or a simple text comment like ```searching by a Time value```.

But this is just my opinion.

PR #4152, opened 2015-09-15
MONGOID-3214 Ensure that update is not called multiple times for an embedded doc

estolfo:


spec/mongoid/persistable/savable_spec.rb
```diff
@@ -83,6 +83,27 @@
             end
           end
         end
+
+        context 'when the embedded document is unchanged' do
+
+          before do
+            allow(Kangaroo.collection).to receive(:insert).once
```
simi: I think it will be better to move expectations and stubs into the actual test block.

durran: I agree here as well, since the expectation is the actual test here, not that the call to save is true.

estolfo: ok, updated it.

durran:
:+1:

spec/mongoid/persistable/savable_spec.rb
```diff
@@ -83,6 +83,24 @@
...
         end
+
+        context 'when the embedded document is unchanged' do
+
+          let(:kangaroo) do
+            Kangaroo.new
+          end
+
+          after do
+            Kangaroo.destroy_all
```
simi: :(

PR #4157, opened 2015-09-17
MONGOID-4125 Make sure none scopes referenced in procs are applied

estolfo:


spec/mongoid/scopable_spec.rb
```diff
@@ -398,6 +398,41 @@ class << Band
         end
       end
 
+      context 'when the block is an none scope' do
+
+        context 'when there is no default scope' do
+
+          before do
+            Band.scope(:nothing, ->{ none })
```
simi: This will leak nothing scope for all following tests. Maybe it will be better to move this scope to model since it is used twice.

estolfo: Good point, thanks

PR #4159, opened 2015-09-17
MONGOID-4119 Ensure that criteria selector becomes pipeline operator value

estolfo:
Need to figure out a better way to test, i.e. not test a private method specifically.

Change code to apply two $match operators, first will apply the $exists selector, the second will apply the user-defined selector.

Test the situation in which the selector refers to a field that is not present to make sure calculations don't fail.

### DIFF
--- a/lib/mongoid/contextual/aggregable/mongo.rb
+++ b/lib/mongoid/contextual/aggregable/mongo.rb
@@ -124,8 +124,8 @@ module Mongoid
         def pipeline(field)
           db_field = "$#{database_field_name(field)}"
           pipeline = []
-          pipeline << { "$match" => { field => { :$exists => true }  } }
-          pipeline << { "$match" => { field => criteria.selector[field.to_s] } }
+          pipeline << { "$match" => criteria.selector }
+          pipeline << { "$match" => { field => { :$exists => true } } }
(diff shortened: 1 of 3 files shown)


lib/mongoid/contextual/aggregable/mongo.rb
```diff
@@ -124,7 +124,7 @@ def sum(field = nil)
         def pipeline(field)
           db_field = "$#{database_field_name(field)}"
           pipeline = []
-          pipeline << { "$match" => criteria.nin(field => nil).selector }
+          pipeline << { "$match" => { field => criteria.selector[field.to_s] } }
```
estolfo: @durran Shouldn't we be setting the $match operator to the criteria selector, not to a selector finding all documents in which the field exists?

In other words, wouldn't this:

```ruby
account.campaigns.where(name: /Untitled (\d+)/).max(:name)
```

return the max of name values of campaigns matching the where criteria, not the max name value across all documents?
I took the example from the JIRA ticket: https://jira.mongodb.org/browse/MONGOID-4119

durran: In your case you are guaranteed the field is there, but in this case you are not:

```ruby
account.campaigns.where(name: '...').max(:number)
```

So we just need to make sure that when the field does not exist, the calculations don't fail.

PR #4160, opened 2015-09-21
MONGOID-3296 Update callback for counter cache

estolfo:


lib/mongoid/relations/counter_cache.rb
```diff
@@ -103,6 +103,24 @@ def add_counter_cache_callbacks(meta)
           name = meta.name
           cache_column = meta.counter_cache_column_name.to_sym
 
+          after_update do
+            if record = __send__(name)
+
+              if __send__("#{name}_id_changed?".to_sym)
+                original, current = __send__("#{name}_id_change".to_sym)
```
estolfo: @durran are there other methods I can use instead of
__send__("#{name}_id_changed?".to_sym) 
and
__send__("#{name}_id_change".to_sym)

spec/mongoid/relations/counter_cache_spec.rb
```diff
@@ -290,27 +290,72 @@
...
+        let!(:drug) do
+          person.drugs.create
+        end
+
+        before do
+          drug.destroy
+        end
+
+        it "updates the counter cache" do
+          expect(person.reload.drugs_count).to eq(0)
```
estolfo: todo - shouldn't be necessary to call #reload on the object

lib/mongoid/relations/counter_cache.rb
estolfo: nevermind, I found some more explicit methods to use.

PR #4166, opened 2015-10-03
RUBY-1046. Minor performance optimization for getting a collection in Mongoid

jonhyman:
Test:

```
puts(Benchmark.measure do
  1000.times do
    coll = MyModel.collection
  end
end)
```

Before:

```
0.040000   0.000000   0.040000 (  0.044849)
```

After:

```
0.040000   0.000000   0.040000 (  0.036465)
```

### DIFF
--- a/lib/mongoid/clients.rb
+++ b/lib/mongoid/clients.rb
@@ -117,8 +117,11 @@ module Mongoid
       def mongo_client
         name = client_name
         client = Clients.with_name(name)
-        client = client.use(database_name)
-        client.with(self.persistence_options)
+        if self.persistence_options
+          client.with(self.persistence_options.merge(database: database_name))
+        else
+          client.use(database_name)
+        end
       end
       alias :mongo_session :mongo_client
       deprecate :mongo_session, :mongo_client, 2015, 12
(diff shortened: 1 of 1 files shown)


lib/mongoid/clients.rb
```diff
@@ -117,8 +117,11 @@ module ClassMethods
       def mongo_client
         name = client_name
         client = Clients.with_name(name)
-        client = client.use(database_name)
-        client.with(self.persistence_options)
+        if self.persistence_options
+          client.with(self.persistence_options.merge(database: database_name))
+        else
+          client.use(database_name)
```
estolfo: Mongo::Client#use actually just calls Mongo::Client#with
https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/client.rb#L207

so do you want to change this to the following? It might end up being a tiny bit faster.
```ruby
client.with(database: database_name)
```

jonhyman: I thought about it but didn't think it'd matter much. I'll change it though.

jonhyman:
@estolfo updtaed

estolfo:
Thank you!

PR #4167, opened 2015-10-06
add missing index options

ArikMaor:
specifically I was missing 'language_override' in my project and had to create the index manually
I've also add 'storage_engine' although I'm not sure what it does

based on
http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/


estolfo:
Thanks @ArikMaor

PR #4168, opened 2015-10-06
Adapt index option mappings to new driver

Nielsomat:
See [Mongo::Index::View::OPTIONS](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/index/view.rb#L48), [Mongo::Index::View#create_one](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/index/view.rb#L95) and
http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/#ensureindex-options.

### DIFF
--- a/lib/mongoid/indexable/specification.rb
+++ b/lib/mongoid/indexable/specification.rb
@@ -12,9 +12,7 @@ module Mongoid
       #
       # @since 4.0.0
       MAPPINGS = {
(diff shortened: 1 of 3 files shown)


Nielsomat:
Closes https://jira.mongodb.org/browse/MONGOID-4162.

PR #4169, opened 2015-10-08
MONGOID-4157 Ensure that new clusters are not created when new client options are set

estolfo:
This pull request solves a few problems:

- If #with is called on a Model, the database name is passed as an option to the new client as a Symbol. In the driver, the database name is a string, so [this line](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/client.rb#L324) will return false and a new cluster will be created. A new cluster means more monitoring threads.

- Arbitrary options passed to Mongo::Client#with, creating new clusters (and monitoring threads); if an option is passed to Mongo::Client#with that isn't even a valid option, a new cluster will be created because the option won't be a member of [this list](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/client.rb#L28).

- Mongoid now never calls #with on Mongo::Client with options that would create a new cluster. If Mongoid were to create clusters, it would have to keep track of them and stop their monitoring threads from Mongoid#disconnect_clients.

Update: with [this change](https://github.com/mongodb/mongo-ruby-driver/pull/701) to the driver, Mongoid won't have to whitelist the options it send to #with on the client.

### DIFF
--- a/lib/mongoid/clients.rb
+++ b/lib/mongoid/clients.rb
@@ -13,11 +13,6 @@ module Mongoid
     include ThreadOptions
     include Options
 
-    # Options that a new client should be created with.
-    #
-    # @since 5.0.1
-    NEW_CLIENT_OPTS = [ :read, :write ]
-
     class << self
 
       # Clear all clients from the current thread.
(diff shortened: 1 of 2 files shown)


durran:
Why not just expand the behaviour already in place in the core driver?

https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/client.rb#L319

PR #4171, opened 2015-10-14
MONGOID-3969 adding spec to prove this ticket is no longer a problem

DiannaHohensee:


spec/mongoid/query_cache_spec.rb
```diff
@@ -123,6 +123,27 @@
         end
       end
     end
+
+    context "when cache is enabled and second batch of items is requested" do
+
+      before do
+        Mongoid::QueryCache.enabled = true
```
simi: this should be restored to initial value after test

spec/mongoid/query_cache_spec.rb
```diff
@@ -123,6 +123,27 @@
...
+          Band.create(views: i % 10)
+        end
+      end
+
+      it "no error raised getting next batch from cache" do
+        expect {
+          Band.batch_size(4).where(:views.gte => 0).each do |doc|
+            doc.set(likes: Random.rand(100))
+          end
+          Band.delete_all
```
simi: Is this `delete_all` needed in this expect block?

durran: Not it's not needed here.

spec/mongoid/query_cache_spec.rb
durran: Same.

spec/mongoid/query_cache_spec.rb
simi: Then it can be removed.

spec/mongoid/query_cache_spec.rb
```diff
@@ -123,6 +123,27 @@
         end
       end
     end
+
+    context "when cache is enabled and second batch of items is requested" do
```
estolfo: You can call this context:
'when query caching is enabled and the batch_size is set'

spec/mongoid/query_cache_spec.rb
estolfo: @DiannaHohensee you can put a block to unset this like:

```ruby
after do
  Mongoid::QueryCache.enabled = false
end
```

spec/mongoid/query_cache_spec.rb
estolfo: @DiannaHohensee You can delete this line, it won't affect the test expectation.

spec/mongoid/query_cache_spec.rb
```diff
@@ -123,6 +123,27 @@
...
+
+      before do
+        Mongoid::QueryCache.enabled = true
+
+        5.times do |i|
+          Band.create(views: i % 10)
+        end
+      end
+
+      it "no error raised getting next batch from cache" do
```
estolfo: @DiannaHohensee You can change this string to 

'does not raise an error when requesting the second batch'

spec/mongoid/query_cache_spec.rb
simi: I think `around` will be better here, because you can't be sure what's the initial value.

```ruby
around(:each) do |example|
  query_cache_enabled = Mongoid::QueryCache.enabled
  Mongoid::QueryCache.enabled = false
  example.run
  Mongoid::QueryCache.enabled = query_cache_enabled
end
```

DiannaHohensee: Will the 'before do ... end' changes persist after the context scope is left? It doesn't seem to have any effect outside the context in which it is defined when I set it to false -- the default seems to be true from doing 'QueryCache.enabled?' elsewhere, so setting it to false would be a change.

It would probably be cleaner to add the after block, though capturing and reseting the original state seems awkward: can't just set it without knowing what the original setting was.

estolfo: The problem here is that setting ``Mongoid::QueryCache`` to enabled is a global setting, regardless of where you change it. So yes, in this case, the changes in the ``before`` block will persist after the context scope is left.

It's up to you whether you use around or before and after.

DiannaHohensee:
Updated the code.

estolfo:
thanks @DiannaHohensee !

PR #4176, opened 2015-10-17
MONGOID-3074 Considering time zones while returning values of distinct query

sivagollapalli:


durran:
The problem here is that if we are converting values returned from distinct, it cannot be *just* time values, it needs to be all values that have fields defined in the model. This will also simplify the code - one would just need to grab the field object and demongoize from there for each value in the array.

sivagollapalli:
@durran Updated the PR as per your suggestions. Could you please review?

sivagollapalli:
@durran Could you please let me know your comments?

durran:
Looks great just needs to wait because it is a behaviour change and cannot go into a minor release or we break semver.

PR #4178, opened 2015-10-20
MONGOID-3213 adding spec to prove this ticket is not an existing bug

DiannaHohensee:


spec/mongoid/attributes/nested_spec.rb
```diff
@@ -4249,6 +4249,43 @@ class BandWithAllowDestroyedRecords < Band
...
+            party.guests.to_a
+            party
+          end
+
+          let(:party_timestamps) do
+            party = Party.create(name: 'Ruby Conference', guests_attributes: [{ name: 'attendee 1'}])
+            party.reload.guests
+            party.guests.to_a
+            time_one = party.updated_at
+            sleep 2
```
simi: You can use [travel helpers from ActiveSupport](http://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html) instead of sleep to reduce time overhead.

spec/mongoid/attributes/nested_spec.rb
```diff
@@ -4249,6 +4249,43 @@ class BandWithAllowDestroyedRecords < Band
...
+            party.update_attributes(name: 'Ruby Conference', guests_attributes: [{ id: party.guests.first, name: 'attendee 1', _destroy: '1'}])
+            party.reload.guests
+            party.guests.to_a
+            time_two = party.updated_at
+            return time_one, time_two
+          end
+
+          it "updated_at timestamp is updated on destruction of sub-document" do
+            expect(party_timestamps[0]).to be < party_timestamps[1]
```
simi: It will be better IMHO to move almost whole let into the actual spec (it block).

simi:
I had an idea how to refactor this. See the code bellow. That code is not tested and it is still just an idea. I'm not sure why

```ruby
party.reload.guests
party.guests.to_a
```

is needed and I skipped it in my example code.

```ruby

require 'active_support/testing/time_helpers'
include ActiveSupport::Testing::TimeHelpers

let(:party) { Party.create(name: 'Ruby Conference', guests_attributes: [{ name: 'attendee 1'}]) }
let(:guest) { Guest.create(name: 'attendee 1', party: party }

def remove_guest_from_party(guest, party)
  party.update_attributes(name: party.name, guests_attributes: [{ id: guests.first, name: guest.name, _destroy: '1'}])
end

it "sub-document cache counter is decremented on destruction of sub-document" do
  remove_guest_from_party(guest, party)
  expect(party.guests_count).to eq(0)
end

it "updated_at timestamp is updated on destruction of sub-document" do
  expect{
    travel(2.seconds) { remove_guest_from_party(guest, party) }
  }.to change{party.updated_at}
end
```

spec/app/models/guest.rb
```diff
@@ -0,0 +1,7 @@
+class Guest
+  include Mongoid::Document
+
+  field :name, type: String, default: ''
+
+  belongs_to :party, counter_cache: true, touch: true
+end
```
estolfo: add a newline here

spec/app/models/party.rb
```diff
@@ -0,0 +1,10 @@
+class Party
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  field :name, type: String, default: ''
+  field :guests_count, type: Integer, default: 0
+
+  has_many :guests, dependent: :destroy
+  accepts_nested_attributes_for :guests, reject_if: :all_blank, allow_destroy: true
+end
```
estolfo: add a newline here

PR #4179, opened 2015-10-22
MONGOID 4159 fix chain scoping

DiannaHohensee:
scopes are now stored by class rather than thread.

### DIFF
--- a/lib/mongoid/threaded.rb
+++ b/lib/mongoid/threaded.rb
@@ -197,13 +197,22 @@ module Mongoid
     # Get the current Mongoid scope.
     #
     # @example Get the scope.
+    #   Threaded.current_scope(klass)
     #   Threaded.current_scope
+    #
+    # @param [ Klass ] klass The class type of the scope.
     #
     # @return [ Criteria ] The scope.
     #
     # @since 5.0.0
-    def current_scope
-      Thread.current[CURRENT_SCOPE_KEY]
+    def current_scope(klass = nil)
+      if klass && Thread.current[CURRENT_SCOPE_KEY].respond_to?(:keys)
+        Thread.current[CURRENT_SCOPE_KEY][
+            Thread.current[CURRENT_SCOPE_KEY].keys.find { |k| k <= klass }
+        ]
+      else
+        Thread.current[CURRENT_SCOPE_KEY]
+      end
     end
 
     # Set the current Mongoid scope.
@@ -218,6 +227,29 @@ module Mongoid
     # @since 5.0.0
     def current_scope=(scope)
       Thread.current[CURRENT_SCOPE_KEY] = scope
+    end
+
+    # Set the current Mongoid scope. Safe for multi-model scope chaining.
+    #
+    # @example Set the scope.
+    #   Threaded.current_scope(scope, klass)
+    #
+    # @param [ Criteria ] scope The current scope.
+    # @param [ Class ] klass The current model class.
+    #
+    # @return [ Criteria ] The scope.
+    #
+    # @since 5.0.1
+    def set_current_scope(scope, klass)
+      if scope.nil?
+        if Thread.current[CURRENT_SCOPE_KEY]
+          Thread.current[CURRENT_SCOPE_KEY].delete(klass)
+          Thread.current[CURRENT_SCOPE_KEY] = nil if Thread.current[CURRENT_SCOPE_KEY].empty?
+        end
+      else
+        Thread.current[CURRENT_SCOPE_KEY] ||= {}
+        Thread.current[CURRENT_SCOPE_KEY][klass] = scope
+      end
     end
 
     # Is the document autosaved on the current thread?
--- a/spec/mongoid/scopable_spec.rb
+++ b/spec/mongoid/scopable_spec.rb
@@ -143,16 +143,34 @@ describe Mongoid::Scopable do
         Band.where(name: "Depeche Mode")
       end
 
-      before do
-        Mongoid::Threaded.current_scope = criteria
+      context "when using #current_scope=scope" do
+
+        before do
+          Mongoid::Threaded.current_scope = criteria
+        end
+
+        after do
+          Mongoid::Threaded.current_scope = nil
+        end
+
+        it "returns the criteria on the stack" do
+          expect(Band.queryable).to eq(criteria)
+        end
       end
 
-      after do
-        Mongoid::Threaded.current_scope = nil
-      end
+      context "when using #set_current_scope(scope, klass)" do
 
-      it "returns the criteria on the stack" do
-        expect(Band.queryable).to eq(criteria)
+        before do
+          Mongoid::Threaded.set_current_scope(criteria, Band)
+        end
+
+        after do
+          Mongoid::Threaded.set_current_scope(nil, Band)
+        end
+
+        it "returns the criteria on the stack" do
+          expect(Band.queryable).to eq(criteria)
+        end
       end
     end
   end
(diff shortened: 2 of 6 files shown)


lib/mongoid/threaded.rb
```diff
@@ -197,27 +197,37 @@ def client_override=(name)
...
     # @example Get the scope.
-    #   Threaded.current_scope
+    #   Threaded.current_scope(klass)
     #
     # @return [ Criteria ] The scope.
     #
     # @since 5.0.0
-    def current_scope
-      Thread.current[CURRENT_SCOPE_KEY]
+    def current_scope(klass)
```
estolfo: We should have the argument be (klass = nil) so that we don't break the API. This allows you to call the method as before 
```ruby
Threaded.current_scope
```
without an argument and no argument error will be thrown. This is kind of like overloading, though not exactly.

Then in the method body, you should check if klass is nil. If so, you return Thread.current[CURRENT_SCOPE_KEY], otherwise you can run the code you've written for finding the scope by class.

lib/mongoid/threaded.rb
```diff
@@ -197,27 +197,37 @@ def client_override=(name)
...
+      if Thread.current[CURRENT_SCOPE_KEY]
+        klass_key = klass
+        Thread.current[CURRENT_SCOPE_KEY].each_key { |key| klass_key = key if key <= klass }
+        Thread.current[CURRENT_SCOPE_KEY][klass_key]
+      end
```
estolfo: Another, more Ruby-like way of doing this logic would be using the #find method. http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-find

You could do:

```ruby
Thread.current[CURRENT_SCOPE_KEY].find do |k, v|
  k <= klass
end
```

This will test each key and value pair and return the first pair for which k <= klass returns true.

lib/mongoid/threaded.rb
```diff
@@ -197,27 +197,37 @@ def client_override=(name)
...
-    #   Threaded.current_scope = scope
+    #   Threaded.current_scope(scope, klass)
     #
     # @param [ Criteria ] scope The current scope.
     #
     # @return [ Criteria ] The scope.
     #
     # @since 5.0.0
-    def current_scope=(scope)
-      Thread.current[CURRENT_SCOPE_KEY] = scope
```
estolfo: We have to keep the current_scope= method for backwards compatibility. You can define the new method though and use it internally instead of current_scope=

DiannaHohensee: Should I duplicate the old tests, one test as it was with the old methods and another new test with the new methods?

estolfo: Yes, if we are leaving that method in, we should leave the tests in as well.

DiannaHohensee:
Updated code, now is backwards compatible if someone chooses to call the methods directly.

lib/mongoid/threaded.rb
```diff
@@ -197,13 +197,17 @@ def client_override=(name)
...
+    #
     # @return [ Criteria ] The scope.
     #
     # @since 5.0.0
-    def current_scope
-      Thread.current[CURRENT_SCOPE_KEY]
+    def current_scope(klass = nil)
+      return Thread.current[CURRENT_SCOPE_KEY] unless klass && Thread.current[CURRENT_SCOPE_KEY] && Thread.current[CURRENT_SCOPE_KEY].is_a?(Hash)
+      Thread.current[CURRENT_SCOPE_KEY][ Thread.current[CURRENT_SCOPE_KEY].keys.find { |k| k <= klass } ]
     end
```
estolfo: We can split this into two conditions:
1) is klass not nil and is Thread.current[CURRENT_SCOPE_KEY] a hash? -> return line 210
2) return Thread.current[CURRENT_SCOPE_KEY]

So another way of writing this would be

```ruby
    def current_scope(klass = nil)
      if klass && Thread.current[CURRENT_SCOPE_KEY].respond_to?(:keys)
        Thread.current[CURRENT_SCOPE_KEY][
            Thread.current[CURRENT_SCOPE_KEY].keys.find { |k| k <= klass }
        ]
      else
        Thread.current[CURRENT_SCOPE_KEY]
      end
    end
```

lib/mongoid/threaded.rb
```diff
@@ -220,6 +224,29 @@ def current_scope=(scope)
       Thread.current[CURRENT_SCOPE_KEY] = scope
     end
 
+    # Set the current Mongoid scope. Safe for multi-model scope chaining.
+    #
+    # @example Set the scope.
+    #   Threaded.current_scope(scope, klass)
+    #
+    # @param [ Criteria ] scope The current scope.
+    # @param [ Klass ] klass The current model class.
```
estolfo: This is actually type Class

lib/mongoid/threaded.rb
```diff
@@ -220,6 +224,29 @@ def current_scope=(scope)
...
+    #
+    # @return [ Criteria ] The scope.
+    #
+    # @since 5.0.1
+    def set_current_scope(scope, klass)
+      if scope.nil?
+        if Thread.current[CURRENT_SCOPE_KEY]
+          Thread.current[CURRENT_SCOPE_KEY].delete(klass)
+          Thread.current[CURRENT_SCOPE_KEY] = nil if Thread.current[CURRENT_SCOPE_KEY].length == 0
```
estolfo: You can call #empty? here instead, as it makes more sense than calling #length on a hash:

```ruby
Thread.current[CURRENT_SCOPE_KEY].empty?
```

DiannaHohensee:
Fixed up again.

lib/mongoid/threaded.rb
```diff
@@ -220,6 +224,29 @@ def current_scope=(scope)
...
+        end
+      else
+        Thread.current[CURRENT_SCOPE_KEY] = {} unless Thread.current[CURRENT_SCOPE_KEY]
```
estolfo: One very ruby-like thing you could do here is

```ruby
Thread.current[CURRENT_SCOPE_KEY] ||= {}
```

This line means set Thread.current[CURRENT_SCOPE_KEY] to {} if it's false (or nil).

It means this: 

```ruby
Thread.current[CURRENT_SCOPE_KEY] || Thread.current[CURRENT_SCOPE_KEY] = {}
```

DiannaHohensee:
Fixed

tricknotes:
It works fine for me :+1: 
I can hardly wait for it landed :-)

spec/mongoid/scopable_spec.rb
```diff
@@ -548,6 +566,66 @@ class << Band
...
+            end
+
+            context "when calling another model's scope from within a scope" do
+
+              let(:authoredCount) do
+                authoredCount = 0
+                Article.authored.each do |doc|
+                  authoredCount += 1
+                end
+                authoredCount
```
estolfo: I think you can just do

```ruby
Article.authored.size
```

spec/mongoid/scopable_spec.rb
```diff
@@ -548,6 +566,66 @@ class << Band
...
+              Article._declared_scopes.clear
+              class << Author
+                undef_method :author
+              end
+              Author._declared_scopes.clear
+            end
+
+            context "when calling another model's scope from within a scope" do
+
+              let(:authoredCount) do
```
estolfo: We snake-case everything in Ruby so this variable would be:

```ruby
authored_count
```

spec/mongoid/scopable_spec.rb
```diff
@@ -548,6 +566,66 @@ class << Band
...
+              end
+
+              it "correct documents returned" do
+                expect(authoredCount).to eq(3)
+              end
+            end
+
+            context "when calling another model's scope from within a chained scope" do
+              let(:isPublicAuthoredCount) do
+                isPublicAuthoredCount = 0
```
estolfo: Same here, this variable would be

```ruby
is_public_count
```

spec/mongoid/scopable_spec.rb
```diff
@@ -548,6 +566,66 @@ class << Band
...
+                Article.is_public.authored.each do |doc|
+                  isPublicAuthoredCount += 1
+                end
+                isPublicAuthoredCount
```
estolfo: And then you can do

```ruby
Article.is_public.size
```

spec/mongoid/scopable_spec.rb
```diff
@@ -823,17 +901,38 @@ class << Circle
...
+          end
+
+          it "clears default scope" do
+            expect(unscoped.selector).to eq({ "active" => true })
+          end
         end
 
-        it "clears default scope" do
-          expect(unscoped.selector).to eq({ "active" => true })
+        context "when scope is created on the fly" do
```
estolfo: You can call this context, 'when the scope is created dynamically'

spec/mongoid/scopable_spec.rb
```diff
@@ -938,9 +1037,20 @@ class << Band
       end
     end
 
-    it "pops the criteria off the stack" do
-      Band.with_scope(criteria) {}
-      expect(Mongoid::Threaded.current_scope).to be_nil
+    context "when using #current_scope" do
+
+      it "pops the criteria off the stack" do
+        Band.with_scope(criteria) {}
```
estolfo: A more 'ruby-like' way of doing this would be:

```ruby
Band.with_scope(criteria) do;end
```

estolfo:
@DiannaHohensee I've added a few more comments on the specs. After they're updated we can merge it. Thanks!

DiannaHohensee:
@estolfo All set, I think.

spec/mongoid/scopable_spec.rb
```diff
@@ -548,6 +566,58 @@ class << Band
...
+              Author._declared_scopes.clear
+            end
+
+            context "when calling another model's scope from within a scope" do
+
+              let(:authored_count) do
+                Article.authored.size
+              end
+
+              it "correct documents returned" do
```
estolfo: It'd be great if this read like a sentence:

it 'returns the correct documents' do

estolfo:
just one tiny spec sentence comment, other than that, looks good to me

DiannaHohensee:
@estolfo K, added.

estolfo:
ok, looks good. Can you rebase against master so I can merge it?

PR #4180, opened 2015-10-22
MONGOID-4078 Fix #last on criteria

estolfo:


durran:
The automatic sort here cannot be applied. It's a huge performance hit and the removal of it was intentional in Mongoid 5. See https://jira.mongodb.org/browse/MONGOID-3768

Should close the Jira issue and reference the above ticket.

estolfo:
Ok, thanks. I saw the specs were present but marked as pending so I thought it was an oversight.
Do we document this anywhere?

durran:
Yeah the specs should have been removed - was my oversight. We documented it in the CHANGELOG (https://github.com/mongodb/mongoid/blob/master/CHANGELOG.md) under backwards incompatible changes and it's also in the docs under upgrading (https://docs.mongodb.org/ecosystem/tutorial/ruby-mongoid-tutorial/#upgrading-to-5-0-0)

estolfo:
Cool, thanks. I'll just remove the specs then so no one makes the same mistake I did in the future = )

PR #4181, opened 2015-10-23
MONGOID-3267 Using relations as a default values

sivagollapalli:


lib/mongoid/fields/standard.rb
```diff
@@ -231,7 +231,11 @@ def included?(fields)
       # @since 2.4.4
       def evaluated_default(doc)
         if default_val.respond_to?(:call)
-          evaluate_default_proc(doc)
+          begin
+            evaluate_default_proc(doc)
+          rescue NoMethodError => e
+            return nil
+          end
```
estolfo: Thanks for the pull request! 
I think instead of catching a specific error here, we would want to do a more general change so that we could solve the problem noted here as well: https://github.com/mongodb/mongoid/blob/master/lib/mongoid/document.rb#L113-L114

sivagollapalli: @estolfo Do you mean we should move exceptional block to `initialize` method in `lib/mongoid/document.rb` ?

estolfo: We actually need to redesign a little so that methods referencing an association aren't called until after the relation is set on the new object. The initialize method of Document determines the order in which this setup happens but changing the order wouldn't be enough-- we need to change fields.rb to account for a dependency of a certain field on an association.

sivagollapalli: I have done little investigation on this. As per my knowledge, `append(document)` in `lib/mongoid/relations/referenced/many.rb` sets the relationship. So I think we need to call `append(document)` prior to `apply_post_processed_defaults` in `initialize` method in `lib/mongoid/document.rb`

Apart from this I thought some other such that we could avoid changing order of `initialize`  method. When we call `project.rules.create(name: 'test rule')` why couldn't we send parent object as well which means build attributes would be modified into `{ name: 'test rule', project: project }`. Thoughts please?

johnnyshields:
@estolfo this PR should be closed as the proposed implementation does not solve the issue listed in MONGOID-3267. The author can re-open a new PR with a different implementation if needed.

estolfo:
This needs to be fixed as part of a larger refactor.

PR #4182, opened 2015-10-26
MONGOID-3737 adding a spec to prove this issue no longer exists

DiannaHohensee:


spec/mongoid/relations/eager/has_many_spec.rb
```diff
@@ -177,6 +177,43 @@
...
         end
       end
+
+      context "when the child has a default scope" do
+
+        let(:criteria) do
+          Exhibitor.where(:status.ne => "removed")
+        end
+
+        let(:exhibitorPresent) do
```
simi: *it is common to use snake case naming in Ruby*

```exhibitorPresent => exhibitor_present```

PR #4184, opened 2015-10-29
MONGOID-3858 MONGOID-3696 adding spec to prove this is no longer an issue

DiannaHohensee:


spec/mongoid/relations/embedded/many_spec.rb
```diff
@@ -2748,26 +2748,47 @@ class TrackingIdValidationHistory
...
-          expect(address.locations.first.name).to eq("home")
+        before do
+          location.number = 7
+          location.save
         end
 
-        it "overwrites the existing documents" do
-          expect(address.locations.count).to eq(1)
+        it "persists the change" do
+          expect(person.reload.addresses.first.locations.find(location.id).number).to eq(7)
```
estolfo: I would put this in a let block so you would have something like:

```ruby
let(:updated_attribute) do
  person.reload.addresses.first.locations.find(location.id).number
end
```

and then 
```ruby
expect(updated_attribute).to eq(7)
```

DiannaHohensee:
@estolfo Fixed.

PR #4185, opened 2015-10-30
MONGOID-4172: Use positional operator only on 1 level deep nesting.

durran:


lib/mongoid/positional.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def process_updates(keys, update, updates = {})
+      update.each_pair do |position, value|
+        updates[replace_index(keys, position)] = value
+      end
+      updates
+    end
+
+    def replace_index(keys, position)
+      # replace index with $ only if that key is on the selector and it is only
```
estolfo: key is *in* the selector

spec/mongoid/positional_spec.rb
```diff
@@ -0,0 +1,221 @@
...
+
+        let(:selector) do
+          { "_id" => 1, "children._id" => nil }
+        end
+
+        let(:expected) do
+          {
+            "$set" => {
+              "field" => "value",
+              "children.0.field" => "value",
```
estolfo: shouldn't this be 

```ruby
"children.$.field" => "value",
```

durran: No because the _id is nil on the second child.

estolfo: ok so shouldn't the test say it doesn't do a replacement?

durran: Yeah - I updated the test wording.

estolfo: ok makes sense now = )

PR #4186, opened 2015-11-03
MONGOID-4113 adding a spec to prove this is no longer an issue

DiannaHohensee:


spec/mongoid/relations/embedded/one_spec.rb
```diff
@@ -1019,4 +1019,18 @@ class << person
...
+
+    let(:building) do
+      building = Building.create
+      building.building_address = BuildingAddress.new
+      building.save
+      building.reload
+    end
+
+    it "parent successfully embeds an invalid child" do
+      expect(building.building_address).to be
```
estolfo: I believe you're missing the comparison after 'be'

DiannaHohensee: I'm trying to test that building.building_address is set, i.e. not nil. I could maybe do

expect(building.building_address).to_not be_nil             <- guessing on syntax, haven't tested

'.to be' seems to do what I want?

estolfo: You could do

```ruby
expect(building.building_address).to be_a(BuildingAddress)
```

Generally, you want the tests you are doing to be as explicit as possible. Not just testing 'not nil' or 'true'

DiannaHohensee:
@estolfo Fixed

PR #4190, opened 2015-11-06
MONGOID-3709 Support nested polymorphic association assignment

estolfo:


TheNaoX:
:+1:

spec/app/models/post.rb
```diff
@@ -15,6 +15,9 @@ class Post
   has_many :videos, validate: false
   has_many :roles, validate: false
 
+  belongs_to :posteable, polymorphic: true
```
simi: Maybe `postable` :question:

PR #4191, opened 2015-11-10
MONGOID-4181 adding spec to prove this is no longer an issue

DiannaHohensee:


estolfo:
hey @DiannaHohensee can you rebase against master so the tests can run on travis ?

jsperezg:
This still happens when the belongs_to has a touch: true

estolfo:
Hi @jsperezg 

Can you please share what version of Mongoid you are using? I'm not seeing that it's an issue with the latest 6.2.0.

If it's indeed still an issue, can you please reopen the [mongoid ticket](https://jira.mongodb.org/browse/MONGOID-4181) and put steps to reproduce along with the version you are using?

Thanks!

Emily

jsperezg:
Hi, 

You can safely ignore it. I've been experiencing lots of weird issues like 'after_update' triggered after document creation, and regressions like this but my conclusion is that it's a side effect from this gem:

https://github.com/mrrooijen/mongoid-paperclip

PR #4164, opened 2015-09-29
Improve BigDecimal mongoize behavior

johnnyshields:
Recently I attempted to migrate a number of `Float` fields to `BigDecimal`, and I found that the behavior is inconsistent between the two. This PR rectifies that situation. Note that `BigDecimal` is stored in MongoDB as a string, as there is no native BSON type for it.

- `demongoize` should always return a BigDecimal or else nil, never an empty string (this mirrors how Floats and Integers behave).
- `mongoize` should always return a String which can be demongoized to a BigDecimal, otherwise nil
- Support storing BigDecimal values `"Infinity"` and `"-Infinity"` in the database.
- The strings `"Infinity"` and `"-Infinity"` now evaluate as numeric? true.
- `BigDecimal`s now evaluate as numeric? true.
- Additional tests for BigDecimal covering all edge cases.

### Demongoize

| Value to Demongoize | Current            | New | Behavior Changed? |
| ---         | ---                | --- | --- |
| "0"         | BigDecimal 0.0     | BigDecimal 0.0 |   |
| "-0"        | BigDecimal -0.0    | BigDecimal -0.0 |   |
| "1"         | BigDecimal 1       | BigDecimal 1 |   |
| "1.1"       | BigDecimal 1.1     | BigDecimal 1.1 |   |
| nil         | String ""          | nil | **Y** |
| false       | nil                | nil |   |
| true        | true (TrueClass) | nil | **Y** |
| ""          | String ""          | nil | **Y** |
| "other1"    | String "other"     | nil | **Y** |
| "NaN"       | BigDecimal NaN     | BigDecimal NaN |   |
| "Infinity"  | String "Infinity"  | BigDecimal Infinity  | **Y** |
| "-Infinity" | String "-Infinity" | BigDecimal -Infinity | **Y** |

### Mongoize

| Value to Mongoize | Current        | New | Behavior Changed? |
| ---            | ---            | --- | --- |
| BigDecimal 0   | String "0.0"   | String "0.0"  |   |
| BigDecimal -0  | String "-0.0"  | String "-0.0" |   |
| BigDecimal 1   | String "1"     | String "1"    |   |
| BigDecimal 1.1 | String "1.1"   | String "1.1"  |   |
| nil            | String ""      | nil | **Y** |
| false          | false (FalseClass) | nil | **Y** |
| true           | String "true"  | nil | **Y** |
| ""             | String ""      | nil | **Y** |
| "1"            | String "1"     | String "1"   |   |
[...]

PR #4194, opened 2015-11-13
6.0.0: Use Refinements instead of Extensions in Global Namespace

durran:


simi:
You'll need to update JRuby to 9000 on CI to test refinements there.

johnnyshields:
Please consider merging https://github.com/mongodb/mongoid/pull/4164 first as it's likely to conflict with this.

durran:
@johnnyshields Ah thanks for reminding me of that. Will do.

estolfo:
We're not going to use Refinements in Mongoid, they're not fully supported by RRuby.

PR #4197, opened 2015-11-18
MONGOID-4190 Disconnect new clusters to avoid leaving monitor threads running in calls to Model#with

estolfo:
This goes in 5.1.0 so don't merge until later.

### DIFF
--- a/lib/mongoid/clients/options.rb
+++ b/lib/mongoid/clients/options.rb
@@ -76,13 +76,27 @@ module Mongoid
           Thread.current["[mongoid][#{klass}]:persistence-options"]
         end
 
+        # Get the client with special options for the current thread.
+        #
+        # @example Get the client with options.
+        #   Threaded.client_with_options(Band)
+        #
+        # @param [ Class ] klass The model class.
+        #
+        # @return [ Mongo::Client ] The client.
+        #
+        # @since 5.1.0
+        def client_with_options(klass = self)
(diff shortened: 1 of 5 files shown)


lib/mongoid/clients/options.rb
```diff
@@ -93,6 +107,97 @@ def persistence_options(klass = self)
...
+        # @example Unset the client with options.
+        #   Threaded.unset_client_with_options(Band)
+        #
+        # @param [ Class ] klass The model class.
+        #
+        # @return [ nil ] nil.
+        #
+        # @since 5.1.0
+        def unset_client_with_options(klass)
+          if Thread.current["[mongoid][#{klass}]:mongo-client"]
```
durran: A bit nit-picky but you could do this:

```ruby
if client = Thread.current["[mongoid][#{klass}]:mongo-client"]
  client.close
  Thread.current["[mongoid][#{klass}]:mongo-client"] = nil
end
```

estolfo: sure
Otherwise, do you have another idea of how to address this? I think the solution could be cleaner so will continue thinking about it / iterating on it.

durran: Well we need to get rid of the thread locals in the near future - will clean this up once we are passing a "context" object of sorts through the chain.

estolfo: ok makes sense

PR #4199, opened 2015-11-25
RUBY-1072 MONGOID-4192 Add limit of 1 to Model#first and Model#last to avoid leaving cursors open on the server

estolfo:


durran:
:+1:

PR #4202, opened 2015-12-03
MONGOID-3407 FIX# Can't use dot notation with atomic updates on the document instance level

sivagollapalli:


estolfo:
Thanks for this fix @sivagollapalli !

PR #4203, opened 2015-12-09
Support partial indexes.

amw:
Partial indexes were added in MongoDB 3.2 and ruby driver 2.2.


amw:
Alternative approach here: amw@0c7fe5222fbda513043cb72a63ab690c743d95ac

Let me know if you prefer that then I will send a new pull request.

estolfo:
great, thanks @amw !
We can use this approach as it's more explicit.

PR #4204, opened 2015-12-11
MONGOID-3246 Added spec such that 3246 is no longer a bug

sivagollapalli:
#3246

### DIFF
--- a/spec/mongoid/relations/referenced/many_to_many_spec.rb
+++ b/spec/mongoid/relations/referenced/many_to_many_spec.rb
@@ -3535,4 +3535,39 @@ describe Mongoid::Relations::Referenced::ManyToMany do
       end
     end
   end
+
+  example "HABTM should update refereces from two sides" do
+    class Band
+      include Mongoid::Document
+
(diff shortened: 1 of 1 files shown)


estolfo:
Thanks for this @sivagollapalli 
The whole test suite is not passing though because you've redefined the Band class in the test, which has the consequence of affecting further tests that rely on the original definition. Can you update the test to use an existing model in the spec/app/models folder or to define a completely separate class? Thanks

sivagollapalli:
@estolfo Update PR with new models.

estolfo:
Thanks @sivagollapalli 
I'm wondering why you've created an rspec example and not used a context block? If there's any way you could convert it to a context block with a before block, that'd be great. The rest of the file uses describe and context blocks, so it would be nice to keep it consistent. Thanks!

sivagollapalli:
@estolfo As it is just single test case so I thought don't need to use context block. Now I modified the test case as per your suggestions.

estolfo:
thanks @sivagollapalli !

PR #4207, opened 2015-12-19
MONGOID-3833 Matchable throws error for lt, lte, gt and gte on strings

sivagollapalli:


durran:
Actuall this should not be raising an error... If we want to fix this behaviour properly it should follow the same comparison order as the server, and not raise an exception.

https://docs.mongodb.org/v3.0/reference/bson-types/#bson-types-comparison-order

If you'd like to make a pull request with that - it would be great! (Although it's considerably more work)

PR #4208, opened 2015-12-21
checks added for :update_attribute for dynamic attributes refs #3875 …

Shwetakale:
…#3882

When we call process_attribute, it does all the checks for ensuring that dynamic attributes are allowed.

Note This impacts the :counter_cache option as it requires field defined on parent.


johnnyshields:
@Shwetakale can you write a test case for this which uses an aliased field name, e.g.

```
field :foo, as: :my_field_name
```

and test that both `foo` and `my_field_name` can be set?

Shwetakale:
@johnnyshields Sorry, my bad I missed this comment. I have updated pull request to check aliased field as well. Value of an aliased field is also set. Let me know if any other changes are required.

johnnyshields:
@Shwetakale the PR is in an unmergable state. Can you please resolve conflicts?

Shwetakale:
@johnnyshields Yes noticed that. I am resolving them.

Shwetakale:
@johnnyshields Resolved conficts.

lib/mongoid/persistable/updatable.rb
```diff
@@ -27,11 +27,7 @@ def update_attribute(name, value)
         as_writable_attribute!(name, value) do |access|
           normalized = name.to_s
           setter = "#{normalized}="
-          if respond_to?(setter)
-            send(setter, value)
-          else
-            write_attribute(access, value)
-          end
+          process_attribute(normalized, value)
```
johnnyshields: The line `setter = "#{normalized}="` looks like its no longer needed here.

johnnyshields:
@Shwetakale I don't see any tests for aliased fields in your latest PR.

johnnyshields:
Ignore above comment I see it.

lib/mongoid/persistable/updatable.rb
Shwetakale: @johnnyshields Yes you are right. the line ```setter = "#{normalized}="``` is no longer needed here. My bad. I have removed that line and updated PR. Please check this now.

PR #4210, opened 2015-12-23
MONGOID-4173 nested eager loading

estolfo:
TODO: add more specific tests

### DIFF
--- a/lib/mongoid/criteria.rb
+++ b/lib/mongoid/criteria.rb
@@ -1,5 +1,6 @@
 # encoding: utf-8
 require "mongoid/criteria/findable"
+require "mongoid/criteria/includable"
 require "mongoid/criteria/inspectable"
 require "mongoid/criteria/marshalable"
 require "mongoid/criteria/modifiable"
@@ -19,6 +20,7 @@ module Mongoid
     include Origin::Queryable
     include Findable
     include Inspectable
+    include Includable
     include Marshalable
     include Modifiable
     include Scopable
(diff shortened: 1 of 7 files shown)


lib/mongoid/criteria.rb
```diff
@@ -216,10 +218,12 @@ def initialize(klass)
     #
     # @since 2.2.0
     def includes(*relations)
```
johnnyshields: `#includes` method needs doc updated for `relations` param to be Array<Symbol **or Hash**>

johnnyshields: Also consider moving `#includes`, `#inclusions`, and `#inclusions=` to `Includable` module

lib/mongoid/criteria/includable.rb
```diff
@@ -0,0 +1,82 @@
...
+      # @example Add an inclusion.
+      #   criteria.add_inclusion(Person, :posts)
+      #
+      # @param [ Class, String, Symbol ] k The class or string/symbol of the class name.
+      # @param [ Symbol ] relation The relation.
+      #
+      # @raise [ Errors::InvalidIncludes ] If no relation is found.
+      #
+      # @since 5.1.0
+      def add_inclusion(k, relation)
```
johnnyshields: Style point--better to use `key` instead of `k`, and in other places `relation` instead of `r`, etc. which seems to be the standard throughout this repo.

estolfo:
@johnnyshields thanks for your review, I've updated the code. Would you mind testing this out as well, as you've participated in requesting the feature? Thanks!

johnnyshields:
I don't have a Mongoid 5 project but will set up a dummy one to do some light testing.

lib/mongoid/relations/eager.rb
```diff
@@ -34,11 +34,18 @@ def eager_load(docs)
       end
 
       def preload(relations, docs)
-
-        relations.group_by do |metadata|
-          metadata.relation
-        end.each do |relation, associations|
-          relation.eager_load_klass.new(associations, docs).run
+        rs = relations.group_by do |metadata|
```
johnnyshields: Whenever I see a variable `rs` I keep thinking "replica set" or "record set". Better to call this `_relations` IMHO (or `grouped_relations`)

estolfo: ok, thanks for pointing that out. I've updated the code.

estolfo:
great, thanks @johnnyshields ! Let me know if you have any further comments.

PR #4214, opened 2016-01-05
[WIP] MONGOID-4103 Support findAndModify method for embedded documents

estolfo:
Todo:
* Tests for when parent documents have identical embedded documents
* Figure out what to about not being able to only update/replace/delete one embedded document.
* Match error initialization structure
* Documentation for methods


estolfo:
Closing until we have a way to truly implement findAndModify commands on embedded documents. Any implementation now would not have the same guarantees as a FAM command on a top-level document.

PR #4220, opened 2016-01-08
Prevent proxy's method_missing from running intentionally hidden methods by calling public_send

khall:
Calling send directly allows for protected/private methods to be invoked, breaking encapsulation. Most other method_missing methods in mongoid either use respond_to? to prevent calling send or use public_send. I think public_send here is more appropriate. mongoid/relations/targets/enumerable's method_missing might also need this change as well.

### DIFF
--- a/lib/mongoid/relations/proxy.rb
+++ b/lib/mongoid/relations/proxy.rb
@@ -147,7 +147,7 @@ module Mongoid
       # @param [ String, Symbol ] name The name of the method.
       # @param [ Array ] *args The arguments passed to the method.
       def method_missing(name, *args, &block)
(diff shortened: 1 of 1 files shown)


estolfo:
Hi @khall with the way the design is right now, we have to use ``send`` here because some private methods like ``cascadable_children`` need to be called on the target. This is the reason the tests are failing in travis. Thanks for pointing this out though - we'll keep it in mind when refactoring some of this code to make sure that private methods can't be called on a proxy. If you'd like to open a ticket in the [JIRA project for MONGOID](https://jira.mongodb.org/browse/MONGOID) to track this issue, that'd be great.

PR #4221, opened 2016-01-11
Mongoid::Changeable#changes returns a hash with indifferent access.

rinmu:
Mongoid::Changeable#changes should return a HashWithIndifferentAccess.
Or whether there is a reason to return the Hash it 's not HashWithIndifferentAccess?

### DIFF
--- a/lib/mongoid/changeable.rb
+++ b/lib/mongoid/changeable.rb
@@ -64,12 +64,12 @@ module Mongoid
     #
     # @since 2.4.0
     def changes
-      _changes = ActiveSupport::HashWithIndifferentAccess.new
+      _changes = {}
       changed.each do |attr|
         change = attribute_change(attr)
         _changes[attr] = change if change
       end
-      _changes
+      _changes.with_indifferent_access
     end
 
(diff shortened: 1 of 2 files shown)


estolfo:
Thanks @rinmu 
It looks like ``#changes`` should having been returning a HashWithIndifferentAccess, based on the rspec test but the test wasn't actually verifying it properly. I don't see any reason why the returned Hash shouldn't be HashWithIndifferentAccess.

lib/mongoid/changeable.rb
```diff
@@ -64,7 +64,7 @@ def changed_attributes
     #
     # @since 2.4.0
     def changes
-      _changes = {}
+      _changes = ActiveSupport::HashWithIndifferentAccess.new
```
estolfo: why don't we instead do the following on line 72:

```ruby
_changes.with_indifferent_access
```
to be consistent with the rest of the codebase?

rinmu: thanks @estolfo 
I will rewrite it as has been pointed out to you.

estolfo:
thanks!

PR #4226, opened 2016-01-13
[MONGOID-3158] Returns nil if string is invalid for date and datetime

sivagollapalli:


estolfo:
Hi @sivagollapalli, thanks for the pull request. This will need to wait until Mongoid 6.0 as it's a behavior change. I also think we should leave the EPOCH constant in.

sivagollapalli:
@estolfo Updated PR with `EPOCH` constants

migueldiab:
:+1: Nice

@estolfo, why do you think we should leave the EPOCH constant? An argument error should either return nil or throw an error, but returning a valid date I think is counterintuitive.

ch0ke:
I actually agree with @estolfo on this one.  There are many reasons to return epoch constant instead of a nil or error.  epoch is still an 'error', but it's a non-breaking error.  While in most cases I'd agree with you on returning a nil, that agreement ceases when dates and times are returned in that anytime we're dealing with dates it should be a valid object.

migueldiab:
@ch0ke how do you distinguish between Jan 1st 1970 and epoch?

ch0ke:
Well pretty simple I'd think with a good test case.  Figure if your application needs to support back to 1970 ( say for birthdates ), you run it.  It'll still work perfectly for all dates, however if there is an error, it's not likely that it's going to be a single date with the problem - it'll be every date.  Which the developers application test-cases should test against.  When I call `mongoize`, I'm looking to get an instance of the class passed in ( DateTime, Time, etc ) returning nil violates that expectation, which means we now force developers to have to account for that error instead of providing a fall-through that's easy enough to test against.

Looking at plenty of other languages and frameworks, most default to epoch because it's really what makes sense.

estolfo:
fixed as of https://github.com/mongodb/mongoid/pull/4279

PR #4228, opened 2016-01-14
MONGOID-3924 Add validation error messages of associations to document's errors

estolfo:
This maybe should wait for Mongoid 6.0. Need to double check if we're fixing behavior to be consistent with AR or if this is a behavior change that could break users' code.

### DIFF
--- a/lib/mongoid/validatable/associated.rb
+++ b/lib/mongoid/validatable/associated.rb
@@ -29,19 +29,14 @@ module Mongoid
       #
       # @since 2.0.0
       def validate_each(document, attribute, value)
-        begin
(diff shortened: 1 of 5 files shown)


durran:
If the behaviour matches AR then I think we can put in 5.1 - this has been requested for a while.

estolfo:
The current behavior of Mongoid matches AR 4.2.4. Both show the association in the errors with the message: "is invalid"

I'll check AR in Rails 5 to see if it has changed in this respect.

estolfo:
AR 5 has not changed its behavior regarding this so I'll close this PR.

PR #4231, opened 2016-01-26
MONGOID-3606 Don't evolve foreign keys of polymorphic relations

estolfo:


ElMassimo:
@estolfo Isn't this breaking the desired behaviour? The filed bug was:

`Polymorphic foreign keys are not converted to ObjectIds`

but this change will ensure that keeps happening, rather than preventing it.

Looks like that bug had previously been fixed (from what we have experienced over years with Mongoid), but this pull request changes the behaviour again.

It's desirable for polymorphic foreign keys to be evolved to Object Ids, because it's more consistent with the behaviour for non-polymorphic associations. Could we please revert this?

estolfo:
In master, it does evolve the foreign key: https://github.com/mongodb/mongoid/blob/master/spec/mongoid/fields/foreign_key_spec.rb#L467

ElMassimo:
Oh, https://github.com/mongodb/mongoid/pull/4416, excellent! Thanks @estolfo 😃

PR #4232, opened 2016-01-27
The $and and $or Matcher should recurse on Document#matches? in order to handle $not properly.

jonhyman:
MONGOID-4228


estolfo:
Thanks @jonhyman

PR #4236, opened 2016-02-01
MONGOID-4175 Put #last functinality back in

estolfo:


durran:
Why did this go back in?

https://jira.mongodb.org/browse/MONGOID-1336
https://jira.mongodb.org/browse/MONGOID-1739

By adding the sort we run the risk of missing compound indexes, and we don't have a foolproof way to check if the indexes exist on the client side that's not expensive. I would rather us removed the #last method completely, and let Enumerable handle it so if someone *really* wants the last document, they can add a sort themselves and then add a limit of 1.

estolfo:
My mistake, I'll take it out again. Sorry!

estolfo:
@durran, should we take out the #last and #first methods from the Enumerable module entirely to avoid confusion? In that case, in order to add a sort, a user would have to do model.items.where(..).sort(_id: -1).limit(1) to get the object.

ch0ke:
@estolfo from my experience ( which admittedly isn't much ), I actually had no clue that first and last was in Enumerable and have done the sort as per the common sort & limit methods

durran:
@estolfo It's a tricky one and one could argue either way but I think we need to keep #first/#last there with our addition of the limit to not cause huge performance problems to users upgrading by letting it delegate to Enumerable. They probably wouldn't notice the problem until it hit production as well, which I think would make this change worse. 

So pretty much I would leave this functionality as-is. We should document this a bit better on why our #first/#last by default just returns a single document with no order guarantees in the doc site.

estolfo:
Ok. I also looked into sorting by $natural but it still can't use indexes so that doesn't help. I'll update the documentation to make it more clear about sorting guarantees.

PR #4238, opened 2016-02-02
MONGOID-4225 Merge Origin into Mongoid's codebase

estolfo:
Origin is a separate gem but we're merging it into the Mongoid codebase for easier maintenance and versioning.


durran:
I think the next step of this is to remove the `Origin` namespace, and stick this functionality under `Mongoid::Criteria` where it makes more sense.

estolfo:
ok

PR #4241, opened 2016-02-04
Move the Origin namespace to Mongoid::Criteria::Queryable

estolfo:


estolfo:
hey @durran, I've removed the Origin namespace and put everything previously under it, under ``Mongoid::Criteria::Queryable``. When Origin was a separate gem, the querying DSL was included via ``include Origin::Queryable``. If people still want to achieve that functionality, it now can be included via ``include Mongoid::Criteria::Queryable``.
Does that make sense?

PR #4243, opened 2016-02-08
Method names containing hyphens are not valid

tdonia:
Dynamic attributes relies on this regex to filter method name candidates that are valid to def: https://github.com/mongodb/mongoid/blob/master/lib/mongoid/attributes/dynamic.rb#L39

### DIFF
--- a/lib/mongoid/extensions/string.rb
+++ b/lib/mongoid/extensions/string.rb
@@ -117,7 +117,7 @@ module Mongoid
       #
       # @since 3.0.15
       def valid_method_name?
-        /[@$"]/ !~ self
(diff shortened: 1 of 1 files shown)


tdonia:
cc @smitchsmith for finding this one.

tdonia:
pre 35c1b99ca70fc433b9b10d2df32725a135c956d1 this relied on a `to_sym` that solves my use case by had other trade offs: [MONGOID-3041] (https://jira.mongodb.org/browse/MONGOID-3041)

http://ruby-doc.org/core-2.3.0/doc/syntax/methods_rdoc.html#label-Method+Names

PR #4245, opened 2016-02-10
Update @since attributes for Mongoid client methods

NobodysNightmare:
Those methods had the word session instead of client in versions prior mongoid 5.0.0.
Thus it is irritating to say that the method default_client is available since version 3.0.0, since I can't call it in versions 3.0.0 or 4.0.0.


estolfo:
thanks @NobodysNightmare

PR #4250, opened 2016-02-22
MONGOID-4241 Fix touch to only update timestamp and optional passed field

connerfritz:
https://jira.mongodb.org/browse/MONGOID-4241


estolfo:
Thanks @connerfritz

PR #4253, opened 2016-02-25
MONGOID-4246 Properly convert objects related via polymorphic assoc in queries

estolfo:


maxjacobson:
:+1: Fix works great!

I wasn't able to point my Gemfile at your branch because a few other gems in my Gemfile depend explicitly on a mongoid that is less than version 6.0 (e.g. [this one](https://github.com/ElMassimo/mongoid_includes/blob/9011459550c95aae545c99f235c04f1a4aef9361/mongoid_includes.gemspec#L25)) but I was able to modify the version of the gem that was already installed on my system to use your change and it works great.

Is there a process for backporting fixes like this to 5.1.x or do you think we should plan to wait for 6.0?

Thanks again!

estolfo:
Seeing as 6.0 won't come out for a few more months, it would make sense to do a 5.1.x release. I'll update the Jira ticket and make sure this fix gets applied to the 5.1.0-stable branch.

maxjacobson:
That would be very much appreciated! :smile:

PR #4254, opened 2016-02-26
[WIP] MONGOID-4251 Adding lookup to aggregate pipeline

kayakyakr:
I wanted to use the mongo 3.2 lookup pipeline stage in an app and ran into some things that needed to be done. I'm very unfamiliar with mongoid's codebase, so looking for some suggestions on where some of this code would more appropriately go.

This actually manages to pass all of the important tests, but that, I think, is because I don't see the tests ever actually running aggregation queries. Very strange. The `Mongo::View::Aggregation` that is returned by collection.aggregate is not completely compatible with the `Mongo::View` that is returned by collection.find, which is also problematic.

I'd really like to see a `lookup(:relation)` construct that would make things much easier, including instantiating related documents as if they were embedded. I've worked around it in my own app by overriding the relation helpers, instantiating new objects if my keyword is present or manually building the criteria if not.

Ignore the version downgrade, that was something to make this work with an unrelated gem, would be removed before this is ready to actually merge (alternatively someone much more familiar with this codebase could take over implementing this and I can leave it behind)

### DIFF
--- a/lib/mongoid/contextual/mongo.rb
+++ b/lib/mongoid/contextual/mongo.rb
@@ -312,8 +312,12 @@ module Mongoid
         @criteria, @klass, @cache = criteria, criteria.klass, criteria.options[:cache]
         @collection = @klass.with(criteria.persistence_options || {}).collection
         criteria.send(:merge_type_selection)
-        @view = collection.find(criteria.selector)
-        apply_options
+        if criteria.aggregating?
+          @view = collection.aggregate(criteria.pipeline)
+        else
+          @view = collection.find(criteria.selector)
+          apply_options
+        end
       end
 
       delegate(:database_field_name, to: :@klass)
--- a/lib/mongoid/criteria/queryable/aggregable.rb
+++ b/lib/mongoid/criteria/queryable/aggregable.rb
@@ -71,6 +71,22 @@ module Mongoid
           end
         end
 
+        # Add a lookup ($lookUp) to the aggregation pipeline.
+        #
+        # @example Add a lookup to the pipeline.
+        #   aggregable.lookup(from: 'rightCollection', localField: 'leftVal', foreignField: 'rightVal', as: 'embeddedData')
+        #
+        # @param [ Hash ] criterion The lookup to make.
+        #
+        # @return [ Aggregable ] The aggregable.
+        #
+        # @since 3.2.0
+        def lookup(operation)
+          aggregation(operation) do |pipeline|
+            pipeline.lookup(operation)
+          end
+        end
+
         # Add an unwind ($unwind) to the aggregation pipeline.
         #
         # @example Add an unwind to the pipeline.
--- a/lib/mongoid/criteria/queryable/pipeline.rb
+++ b/lib/mongoid/criteria/queryable/pipeline.rb
@@ -68,6 +68,20 @@ module Mongoid
           push("$project" => evolve(entry))
         end
 
+        # Add the $lookup entry to the pipeline.
+        #
+        # @example Add the lookup.
+        #   pipeline.lookup(entry)
+        #
+        # @param [ Hash ] entry The left outer join.
+        #
+        # @return [ Pipeline ] The pipeline.
+        #
+        # @since 3.2.0
+        def lookup(entry)
+          push("$lookup" => evolve(entry))
+        end
+
         # Add the $unwind entry to the pipeline.
         #
         # @example Add the unwind.
--- a/lib/mongoid/version.rb
+++ b/lib/mongoid/version.rb
@@ -1,4 +1,4 @@
 # encoding: utf-8
 module Mongoid
-  VERSION = "6.0.0"
+  VERSION = "5.1.2"
 end


estolfo:
Hi @kayakyakr 

Thanks for the pull request. I'd rather not add helpers to Mongoid for each pipeline operator, as the driver can be used to do an aggregation. The aggregation API is intentionally flexible so that operators can be chained and new ones can be implemented in the future. Adding a helper in Mongoid would mean more maintenance for something that's pretty straightforward to do directly in the driver. [Here](https://docs.mongodb.org/ecosystem/tutorial/ruby-driver-tutorial/#aggregation) is the documentation for it.
Would it be possible for you to use the aggregation method on the driver instead?

You could do:
```ruby
Model.collection.aggregate([....])
```
Let me know if there's a reason you'd like it supported in Mongoid via a helper, I'm open to being convinced = )

kayakyakr:
I have done that in the past, but that does not instantiate models at all, so you're going to have to do a second pass, manually, if you want something that's usable in Rails. Otherwise, you're just working with hashes. While I can loop my results to create my model objects, the ODM can probably do that more efficiently. With a bit of help, it could also instantiate related documents that have already been loaded through this new operator. That's the biggest reason why I'd like to see Mongoid have some support for aggregations.

If no aggregation support is indeed the decision you want to go with, then I would counter that everything in `queryable/aggregable` and `queryable/pipeline` should be removed. For one thing, some of it is not actually functional code (very little of what is added to the pipeline array actually gets run via the criteria which opts, instead, to do its own thing with methods on the Mongo driver). So you have a number of misleading methods in the codebase in the first place.

If, instead, you want to very carefully pick and choose what aggregations you support in Mongoid, I think that there is a strong case for lookup, at least, so be included in that list as it increases your feature parity with ActiveRecord, potentially acting as both join and include. I'd love to be able to run a lookup to pre-load related documents. It's a cool new feature in Mongo, one that I would hope would be exposed in a more Ruby-esque way in Mongoid.

I'd go so far as to argue that the point of an ODM is to provide a framework-ready interface into the database layer, and that Mongoid _should_ endeavor to reach parity with all of the pipeline stages that Mongo provides.

I don't mind if full aggregation support is moved to a gem, and I wouldn't mind building that gem, but as I think I've demonstrated with my artful _hitting your codebase with a hammer_ technique, I'd need a bit of help to get that started. OTOH, Mongoid's interface has been a bit unstable lately (big changes that I've seen in 4 -> 5 -> 6), so gem maintenance would be much more difficult.

estolfo:
Hi @kayakyakr 

Thanks for the explanation, I understand why you would want a helper for this. I'm going to give a little more thought to this and how it might work / what the API would look like and get back to you.
Thanks!
Emily

estolfo:
Hello again, I've made a [jira ticket](https://jira.mongodb.org/browse/MONGOID-4251), as we'll most likely add a helper for the $lookup operator. It makes sense for relations, as you've pointed out.

As for the other pipeline operators, I'm hesitating making helpers because the ordering of them is really important for the aggregation you're doing. Adding helpers would allow people to construct a pipeline without thinking about the actual ordering of the operators. I think it's best to do this explicitly using the driver.

kayakyakr:
What if Mongoid provided its own aggregate operator to allow people to define arbitrary aggregations that would then be instantiated into proper Mongoid documents? This would give you best of both worlds: an open aggregation framework (pretty much straight against the database), that would still return models and not a simple hash?

estolfo:
The problem with that approach is the aggregation framework pipeline operators manipulate documents. They can change the name of fields, the value types, or even the structure of the document. I would hesitate to provide such an open functionality in Mongoid because I think there is a huge potential for errors. If a model is instantiated from an aggregation result and the field names or value types have changed, there will be many errors.

kayakyakr:
If you compare to ActiveRecord, you can write a custom SQL select that does the same thing. AR still loads it. I think it even gives you access methods for those random fields, not dissimilar to how mongoid dynamic fields works.

johnnyshields:
:+1: +1 for more ways to hook custom Javascript in like AR allows custom SQL. While it's nice that Mongoid "Ruby-ifies" most of the API, there will always be edge cases and we should not shy away from allowing Mongoid users to roll-up their sleeves and add custom JS when needed.

johnnyshields:
@kayakyakr @estolfo what's the status of this PR?

estolfo:
@johnnyshields It's a non-trivial amount of work and I didn't want it to hold up 6.0 so I marked it for 6.1. You can watch the jira ticket, if you'd like: https://jira.mongodb.org/browse/MONGOID-4251

johnnyshields:
@estolfo understood. It would be useful to tag this an other PRs with the appropriate milestone. I see many PRs tagged as 5.x.x which should be updated.

estolfo:
We're using JIRA to manage the Mongoid project and I'd like to centralize all the management there, instead of relying on github issues/milestones (and having to update both)

Your point is good though - I'll update the PR description to note the JIRA ticket number for tracking. The JIRA ticket has a link to this PR already.

estolfo:
Implementation of this will be via https://jira.mongodb.org/browse/MONGOID-4251

kayakyakr:
Update for anyone who is interested: the mongoid team has determined that they will not implement the feature. 

I am very disappointed by this decision, naturally, as getting mongoid to create instances of associated records from lookup queries is an essential component of my framework and effectively orphans me onto my hacked PR. Using the pipeline directly is a possibility, but it's effectively saying not to use mongoid at all.

estolfo:
Hi @kayakyakr 

I'm really sorry that our decision to not implement this feature has discouraged you from using Mongoid.

We did originally plan to incorporate it into the codebase but in looking at how it could actually be supported, it became clear it would cause many issues for users. Pulling documents from another collection via $lookup in an aggregation allows you to rename fields, change the value types, and create new fields in the resulting documents. Creating a new field and renaming a field would cause errors in Mongoid when instantiating the model object. Changing the type of a value wouldn't necessarily raise an error right away, but could cause a slew of issues either immediately or later on in users' code.

Perhaps more importantly though, providing the ability to do "joins" across collections allows users to treat MongoDB like a relational database and will cause performance problems over time. For example, aggregations can't always take advantage of indexes.

$lookup also only works on non-sharded collections, which would probably be a significant limitation for some Mongoid users in production.

For these same reasons (with the exception of the sharding limitation), we don't try to turn general aggregation query results into model objects and encourage users to write custom methods instead.

We consulted with the Python team when making this decision and and confirmed that they have the same approach to supporting aggregations and the $lookup operator in their ODM, pymodm.

All of that said, it is still possible to write a custom method for fetching related model objects via $lookup. I would recommend writing a method on a Model that performs an aggregation using $lookup, and then that instantiates the instances of the related Model directly. Have you tried this approach? If it hasn't work for you, can you share why? I'm happy to help in any way.

Emily

johnnyshields:
@kayakyakr is there any reason this functionality couldn't be done as a gem?

kayakyakr:
@estolfo I could take that approach, but run into the same issues that I did when requesting assistance with this PR: if the data for the event has already been looked up for the model, I want to be able to use that relation without it performing another query. That caching code is buried deep, deep inside mongoid and I couldn't find a good point at which to hack into it. I'm open to pointers.

@johnnyshields I believe a gem-based approach that uses the driver directly would need to duplicate a lot of what mongoid already does re: querying and instantiating. Unless you know of a way to take a mongoid query and convert it to an aggregation on the base driver that we can add a lookup pipeline to, plus a hook to get that into the relation cache and/or hand that back to the base instantiation code.

Unless I monkey patch it in the gem. That sounds less than ideal.

johnnyshields:
@kayakyakr why not start it as a gem, there are quite few gems which essentially monkey-patch mongoid's code, moreover you'd probably be hooking into relatively stable methods, and with good test coverage you should be able to avoid issues.

estolfo:
Hi @kayakyakr 

I'm happy to help review code if you'd like to implement lookup support in a separate gem. It's definitely a feature I'd like to be available to users, I'm just hesitant to add it directly into the Mongoid codebase.

If I were implementing lookup, I'd maybe put a method on a model and criteria called "join" or "lookup" that passes in the name of the related object and foreign key. You could use inclusions as inspiration.
Or maybe make use of the "foreign_key" attribute of the metadata object describing the models...
What kind of API did you have in mind?

PR #4256, opened 2016-03-10
MONGOID-4167 Use persistence context object to encapsulate all persistence logic

estolfo:


spec/mongoid/validatable/uniqueness_spec.rb
```diff
@@ -19,34 +19,42 @@
...
+            Dictionary.with(collection: "dicts") do |klass|
+              klass.create(name: "websters")
+            end
           end
 
           context "when the document is not valid" do
 
             let(:websters) do
-              Dictionary.with(collection: "dicts").new(name: "websters")
+              valid = nil
```
durran: Why did this test need to change? Shouldn't the block be returning the result of the yield?

estolfo: Because if you call #valid? on the object outside of the block, the alternative collection will not be used, thus returning true for valid? instead of the expected false.

I need to call #valid? on the object in the block and access the return value as well as the object itself in the spec so that I can check the errors list.

PR #4257, opened 2016-03-10
Support ids array input in Selectable#in

oss92:
Support ids array inputs for `in` method.

Example:

`User.in(['fake_id', 'fake_id_2'])` will now return an array of users with these ids instead of `undefined method 'each_pair' for []:Array`


johnnyshields:
@oss92 @estolfo I would assert that this is contrary to Mongoid's design. `User.where('fake_id')` is not valid, and `.in` is meant to behave like `.where`.

It's very easy to get the same result by doing:

```
User.in(_id: ['fake_id', 'fake_id_2'])
```

johnnyshields:
@estolfo can you please close this PR?

PR #4261, opened 2016-04-04
Demonstrate .first_or_create failure when scope includes special ($) operators

niels:
https://jira.mongodb.org/browse/MONGOID-3887 works well for "normal" scopes but fails when the scope includes special operators; i.e. query fragments such as `{ "$in" => … }`. Depending on configuration, this leads to data corruption (i.e. the field ends up containing `{ "$in" => … }`) or exceptions (when the field is configured not to allow `Hash` values).

This PR only demonstrates the problem to get the conversation started. I am unsure what the desired fix would be. See https://jira.mongodb.org/browse/MONGOID-4261 for discussion.


estolfo:
see https://jira.mongodb.org/browse/MONGOID-4052
for discussion and pull request to fix

PR #4263, opened 2016-04-12
Fix Findable#last behavior

oss92:
* Fix `Findable#last` behavior by specifying an `:_id` sort order.
* Added tests for `Findable#first` and `Findable#last`

### DIFF
--- a/lib/mongoid/findable.rb
+++ b/lib/mongoid/findable.rb
@@ -156,7 +156,7 @@ module Mongoid
     #
     # @return [ Document ] The last matching document.
     def last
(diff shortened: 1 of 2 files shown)


estolfo:
Thanks for the pull request, but we cannot add a sort on _id because of reasons explained in this comment:
https://jira.mongodb.org/browse/MONGOID-4175?focusedCommentId=1162135&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1162135

I would recommend adding the sort yourself and a limit of -1 to get the last document.

PR #4267, opened 2016-04-15
MONGOID-4260 Fix regression with blank fields

tagliala:
MONGOID-4260 introduced a regression when checking for presence on blank
fields. This PR switches back to the old lookup method and adds a spec
to prevent this issue in the future.

/cc @estolfo 

Please review my PR

### DIFF
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -350,7 +350,7 @@ module Mongoid
       if localized_fields.has_key?(name)
         value = localized_fields[name].send(:lookup, value)
       end
-      !!value
+      value.present?
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/mongoid/attributes.rb
```diff
@@ -350,7 +350,7 @@ def lookup_attribute_presence(name, value)
       if localized_fields.has_key?(name)
         value = localized_fields[name].send(:lookup, value)
       end
-      !!value
+      value == true || value.present?
```
tagliala: @estolfo 

Can I use `value.present?` It should have the same effect. All specs pass

tagliala:
Does [this failure](https://travis-ci.org/mongodb/mongoid/jobs/123371295) depend on my PR? I don't think so. Where can I find the test seed to check on my dev machine?

Similar failure here: https://travis-ci.org/mongodb/mongoid/jobs/120102307

estolfo:
Those failures are not related, but if you're looking to run the tests locally, start a mongod on the default host and port and then just run rspec in the mongoid directory.
In any case, this passes for me locally and on travis so I'll merge it. Thanks so much!

tagliala:
You're welcome 👍

PR #4268, opened 2016-04-18
Model.first(User.first) and Model.last returns first and last record respectively

mayurkhatri:
If we do Model.first (say User.first) or Model.last (say User.last) it returns the same record.
In this fix Model.first and Model.last return first and last records respectively

### DIFF
--- a/lib/mongoid/findable.rb
+++ b/lib/mongoid/findable.rb
@@ -145,7 +145,7 @@ module Mongoid
     #
     # @return [ Document ] The first matching document.
     def first
-      with_default_scope.first
+      with_default_scope.to_a.first
     end
     alias :one :first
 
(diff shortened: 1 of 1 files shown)


estolfo:
Thanks for the pull request, but we cannot add a sort on _id because of reasons explained in this comment:
https://jira.mongodb.org/browse/MONGOID-4175?focusedCommentId=1162135&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1162135

I would recommend adding the sort yourself and a limit of -1 to get the last document.

As a sidenote, this implementation would not be a desirable solution as it would end up fetching every single document matching the query on the client side, only to get the first or last.

mayurkhatri:
@estolfo thanks :)

PR #4269, opened 2016-04-18
MONGOID-4264 Require a belongs_to association by default and accept :required and :optional relation options

estolfo:
ActiveRecord by default requires an association for belongs_to relations. This pull request implements that behavior in Mongoid as well.

Commit in Rails: https://github.com/rails/rails/pull/18937/files


simi:
Awesome!

estolfo:
ah, ok. Thanks for the clarification.

PR #4275, opened 2016-04-25
MONGOID-2882 set common base object for has many relations

estolfo:


spec/mongoid/query_cache_spec.rb
```diff
@@ -7,94 +7,94 @@
...
+  #
+  #       it 'does not query for access to the base' do
+  #         expect(server).to receive(:context).exactly(0).times.and_call_original
+  #         relations.each do |object|
+  #           object.person
+  #         end
+  #       end
+  #     end
+  #   end
+  # end
```
johnnyshields: if these tests are intended to be removed, better to delete them than to comment out.

estolfo: Work on this has been paused because it can only truly be resolved with a relations refactor.
You can follow the ticket here, if you'd like: https://jira.mongodb.org/browse/MONGOID-4276

PR #4277, opened 2016-04-26
MONGOID-4052 Add option to return all translations when calling #only for a field

estolfo:


johnnyshields:
@estolfo on second thought here, why not make localize: false the behavior always. Is there any real meaning to have localize true?

If a user really wants a particular locale they should be able to do `field.en` for example.

estolfo:
I think users expect localizations to be applied, if a field is localized and there is a locale set. I'd like to keep it with { localize: true } as the default and consider { localize: false } the exception. This is consistent with accessing the attribute after doing a simple #where:

```ruby
I18n.locale = :de
d = Dictionary.first
d.description
=> "deutsch-text"
```

johnnyshields:
I really don't think so, because 90%+ of users using I18n have fallbacks set, i.e. if key is missing for `:de`, fallback to the `:en` word. The major problem with Mongoid here is that it doesn't work with fallbacks, so unless a key was set for the specific locale it does work.

Fallbacks are especially critical for db-stored (e.g. user-entered) data, because users are lazy and don't enter all locales.

johnnyshields:
Also, note that if we set `localize = false` your example will still work exactly as you have it in the only case:

```
I18n.locale = :de
d = Dictionary.only(:description, localize: false).first
d.description
=> "deutsch-text"
I18n.locale = :en
d.description
=> "english-text"
```

johnnyshields:
The only use case for `localize: true` is the very rare case where you only want a specific locale and want to ingore all other locales to reduce your data payload. As mentioned you can achieve the same result by simply doing `.only("description.de")` or `.only("description.#{I18n.locale}")`. Going by the principle of least surprise the default should definitely be `localize: false`, i.e. pull all locales from the database.

johnnyshields:
(apologies for bombarding with text but one more point...) `localize: true/false` is a misnomer here. In fact `localize: false` really means "**load all locales** and display the correct one on the frontend", where as `localize: true` means "**load only currently locale**, and the frontend will return nil if you switch locales"

estolfo:
Hi @johnnyshields 

We have two questions here:

What should the default be?
I don't think we should load all translations from the database by default. Users will use #only in order to do a projection -- i.e. select the exact fields fetched for each document. The list of translations could potentially be long and I'd rather not load all of them, without the user explicitly asking to do so. In using #only, users will be aware that they are only pulling a single field and I think it's appropriate that users will have to opt-in for loading all translations.

What should the option be called?
I agree the option name is misleading. The behavior difference between localize: true and localize: false is whether fallbacks are possible, so maybe we can call it "fallbacks". The behavior would then be:

```ruby
I18n.locale = :de
x = Dictionary.only(:description).first
x.description_translations
=> {de: "deutsch-text"}

I18n.locale = :de
d = Dictionary.only(:description, fallbacks: true).first
d.description
=> "deutsch-text"
I18n.locale = :en
d.description
=> "english-text"
d.description_translations
=> {en: "english-text", de: "deutsch-text"} 

I18n.locale = :de
x = Dictionary.only(:description, fallbacks: false).first
x.description_translations
=> {de: "deutsch-text"}
```
What do you think?

johnnyshields:
@estolfo I think:
* `Dictionary.only(:description)` should load all translations. This is the most intuitive; if you think about the database structure, you would expect this to get the full hash without additional magic.
* If you want a particular locale, you can do `Dictionary.only(:description.de)`
* There does not need to be a `fallback` option.

johnnyshields:
@estolfo any thoughts on the above?

johnnyshields:
@estolfo pinging on this

johnnyshields:
@estolfo pinging again...

estolfo:
Hi @johnnyshields 
I'm just getting back to this and will have an update soon. Thanks for your patience!

johnnyshields:
@estolfo any update on this?

estolfo:
Hi @johnnyshields 
I've updated the code to return all translations by default but also allow a :localize option passed to #only for achieving the previous behavior.
Let me know what you think, I believe this is the behavior you've been advocating for.

johnnyshields:
@estolfo firstly thank you for your support on this. I agree with your implementation.

Two comments:

**1)** My ideal would be that we do not need "localize" at all. The following two should be equivalent:

```
Dictionary.only(:description, localize: true) ==
Dictionary.only("description.#{I18n.locale}")
```

This is quite a rare case so in my opinion the second way is OK; we don't need to further complicate Mongoid code.

**2)** How does `.pluck` behave? In my opinion it should return the raw translation hash, e.g.:

```
Dictionary.only(:description)   #=> { en: "Very good!", de: "Sehr gut!", ja: "すばらしい！" }
```

estolfo:
Hi @johnnyshields 

```ruby
Dictionary.only(:description, localize: true)
Dictionary.only("description.#{I18n.locale}")
```
are equivalent with this PR. Have you tested and found otherwise?

We need the localize option to be passed around internally anyway, as the code is written to apply the localization if the field is localized. I decided I might as well expose the option in the public API for #only, as I'm going to have to pass it through the code path anyway.

I haven't looked at #pluck yet. I'll do that now.

johnnyshields:
@estolfo ok agreed on #1 then. Fine to have `localize` option.

estolfo:
#pluck returns all translations

lib/mongoid/criteria/queryable/smash.rb
```diff
@@ -74,8 +74,12 @@ def [](key)
         # @return [ String ] The normalized key.
         #
         # @since 1.0.0
-        def normalized_key(name, serializer)
-          serializer && serializer.localized? ? "#{name}.#{::I18n.locale}" : name
+        def normalized_key(name, serializer, opts = {})
+          if serializer && serializer.localized? && opts[:localize] != false
```
johnnyshields: shouldn't this be `&& opts[:localize]` (i.e. both nil and false skip localization?)

lib/mongoid/criteria/queryable/optional.rb
```diff
@@ -139,16 +139,30 @@ def no_timeout
...
         # @since 1.0.0
         def only(*args)
-          args = args.flatten
+          if args[-1].is_a?(Hash)
+           opts = args[-1]
+           args = args[0...-1].flatten
+          else
+            opts = {}
+            args = args.flatten
+          end
```
johnnyshields: this could be refactored to use `extract_options` from active support

estolfo: I didn't know about that method, thanks!

lib/mongoid/criteria/queryable/smash.rb
estolfo: yes, thanks

johnnyshields: I strongly believe we should change the default behavior, and announce it as a breaking change. It can be a separate PR if needed. The `:localize` option is almost never what anyone would want to use, you'd only want to use it in explicit performance optimization scenarios.

johnnyshields: ...sigh. This goes back to my previous statement that localize itself is a misnomer. It should really be called "only_current_locale: true" or something like that, because without it the field **will still be localized**--in fact more so because it can use fallbacks.

johnnyshields: Reading my earlier comments. Rather than `localize: true` we should call it `i18n_fallbacks: false`

estolfo:
Would you like to open a pull request implementing this how you'd like? That might end up being more efficient than going back and forth = )

johnnyshields:
Sure, I'd be glad to!

estolfo:
Awesome, that seems like a much easier option and I want to make sure this works how you've requested.

johnnyshields:
Thanks! I'm a heavy user of Mongoid's I18n functionality.

simi:
@estolfo is this last blocker for 6.0 release? Do you need help here?

estolfo:
Hi @simi 
[These](https://jira.mongodb.org/browse/MONGOID-4052?jql=project%20%3D%20MONGOID%20AND%20fixVersion%20%3D%206.0.0%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20due%20ASC%2C%20priority%20DESC%2C%20created%20ASC) are the remaining tickets holding up the release of 6.0.

I've been in touch with the JRuby team to figure out why performance has decreased so drastically on JRuby. [Here](https://github.com/jruby/jruby/issues/4012) is my ticket with them.

I'd welcome help on implementing this PR, as I've had to turn my attention to the driver in the last few weeks and a PR would help speed up the release of 6.0. Also, if you'd like to help investigate why performance is so bad on JRuby, that'd be great as well!

Thanks = )

johnnyshields:
@simi @estolfo I will help with this ticket over the weekend and get it done

estolfo:
awesome! I'll make some time for it over the next two days as well. Sorry for the delay, but we really need to figure out the JRuby issue too...

johnnyshields:
@estolfo @simi I've raised #4299

PR #4282, opened 2016-04-27
MONGOID-4160 Don't save the foreign key of a parent if it fails validation

estolfo:
This is not a good solution. Relations need to be refactored so that related objects are attempted to be saved first (as AR does)
This should not be merged, it just serves as an example of what you could do to make this work...


estolfo:
will be resolved with a relations refactor.

PR #4284, opened 2016-05-12
Don't assign empty documents to HasAndBelongsToMany association

ElMassimo:
Because [`values_at`](http://ruby-doc.org/core-2.1.0/Hash.html#method-i-values_at) will return an `Array` of values corresponding to each key, and it's possible for a document to be removed from the database without updating the inverse document in a `has_and_belongs_to_many` association, using `includes` to eager load the association can assign `nil` in the list of documents.

Since that is not the behaviour when accessing a `has_and_belongs_to_many` association without eager loading (only the existing documents are loaded, no `nil`s), we should be consistent and do the same when eager loading.

I'd be interested in having this fix in _5.1_, let me know and I can make a PR 😃

### DIFF
--- a/lib/mongoid/relations/eager/has_and_belongs_to_many.rb
+++ b/lib/mongoid/relations/eager/has_and_belongs_to_many.rb
@@ -17,7 +17,7 @@ module Mongoid
 
           @docs.each do |d|
             keys = d.send(group_by_key)
-            docs = entries.values_at(*keys)
+            docs = entries.values_at(*keys).compact
             set_relation(d, docs)
           end
         end
--- a/spec/mongoid/relations/eager/has_and_belongs_to_many_spec.rb
+++ b/spec/mongoid/relations/eager/has_and_belongs_to_many_spec.rb
@@ -113,5 +113,21 @@ describe Mongoid::Relations::Eager::HasAndBelongsToMany do
         end
       end
     end
+
+    context "when some related documents no longer exist" do
+      before do
+        # Deleting the first one to meet Builders::Referenced::ManyToMany#query?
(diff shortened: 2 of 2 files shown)


adammiribyan:
👍

ziolmar:
👍

estolfo:
Hi @ElMassimo 
Thanks for the pull request! Can you please add a test for it? Thanks

ElMassimo:
@estolfo Adding the spec, I realized you fixed the problem a few months ago 😄 

The relevant change is [__here__](https://github.com/mongodb/mongoid/commit/934bc4ae1d377faac7b181257adb650f85618687), and it prevents letting `nil` documents from being loaded into an association. Thanks!

ElMassimo:
Yet, found an interesting corner case, where the algorithm to detect if an association is eager loaded is incorrect. Using `compact` does fix the issue, but let me know if you want to revisit the implementation of the [`query?`](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/relations/builders/referenced/many_to_many.rb#L32-L34) method in  [`Mongoid::Relations::Builders::Referenced::ManyToMany`](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/relations/builders/referenced/many_to_many.rb).

ElMassimo:
@estolfo The spec I added covers that case, to check that we avoid an extra query because of `nils` in the documents array which causes the `query?` method to return `true` incorrectly. Let me know if there's anything else we should do to before getting this one merged 😃

johnnyshields:
@estolfo please merge for 6.0.0?

estolfo:
thanks @ElMassimo 
and thanks @johnnyshields for your review

PR #4285, opened 2016-05-16
MONGOID-4218 Update rails5 to rc1

karmaQ:


estolfo:
The gemspec already depends on rc1 of Rails 5: https://github.com/mongodb/mongoid/blob/master/mongoid.gemspec

PR #4286, opened 2016-05-18
Add failing self referencing habtm spec

rmachielse:
Hi @estolfo 

This pull request adds a failing test case related to this issue:
https://jira.mongodb.org/browse/MONGOID-4273

### DIFF
--- a/spec/mongoid/relations/referenced/many_to_many_spec.rb
+++ b/spec/mongoid/relations/referenced/many_to_many_spec.rb
@@ -814,6 +814,82 @@ describe Mongoid::Relations::Referenced::ManyToMany do
     end
   end
 
+  describe "when self-referencing" do
+    let!(:parent) do
+      Vertex.create
+    end
+
+    let!(:child) do
+      Vertex.create(parents: [parent])
+    end
+
+    before do
+      parent.reload
+      child.reload
+    end
+
+    it "sets the parent" do
+      expect(child.parents).to include(parent)
+    end
+
+    it "sets the parent id" do
+      expect(child.parent_ids).to include(parent.id)
+    end
+
+    it "sets the child" do
+      expect(parent.children).to include(child)
+    end
+
+    it "sets the child id" do
+      expect(parent.child_ids).to include(child.id)
+    end
+
+    describe "#update" do
+      before do
+        child.update(parents: [parent])
+      end
+
+      it "sets the parent" do
+        expect(child.parents).to include(parent)
+      end
+
+      it "sets the parent id" do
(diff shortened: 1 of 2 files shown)


johnnyshields:
@estolfo we should get a fix here before 6.0

rmachielse:
@estolfo @johnnyshields I have rebased this branch on the latest master, unfortunately this is still a problem. Do you think it is a valid problem? This pr has been open for a long time. Otherwise I will close it.

johnnyshields:
@durran @estolfo please help with a fix

johnnyshields:
@rmachielse can you check if https://github.com/mongodb/mongoid/pull/4383 fixes this issue?

rmachielse:
@johnnyshields I have rebased my branch on those changes here and run the tests locally:

https://github.com/mongodb/mongoid/compare/master...feedbackfruits:spec/self-referencing-habtm-on-refactor

Unforturtonately it doesn't make a difference

spec/mongoid/relations/referenced/many_to_many_spec.rb
```diff
@@ -814,6 +814,82 @@
     end
   end
 
+  describe "when self-referincing" do
```
johnnyshields: typo: should be self-referencing

johnnyshields:
@estolfo this issue is worthy of investigation... basically it seems that `reload` is clearing the reference value here... very strange

spec/mongoid/relations/referenced/many_to_many_spec.rb
rmachielse: fixed!

rmachielse:
@johnnyshields well that means it's not persisted right?

johnnyshields:
@rmachielse i guess so--did you inspect the values in MongoDB itself?

rmachielse:
@johnnyshields yes, we had a workaround to persist the values again manually. However we are not using mongoid anymore now.

johnnyshields:
So basically the issue here is that `update` of a self-referencing-habtm with the same value as it originally had seems to clear the value of its related partner.

One hunch is that we're doing an atomic `push` then a `pull`, which would set the same value then remove it.

rmachielse:
@johnnyshields yes indeed.

Our workaround looked like this:

```ruby
after_save do
  parents.each do |parent|
    parent.add_to_set(child_ids: id)
  end
end
```

estolfo:
Fixed in this commit: https://github.com/mongodb/mongoid/commit/8ccd33185158b53a79319a8d670907a744d7ecc9

PR #4287, opened 2016-06-09
Update MongoDB versions, target on Rails 5.0.0.rc1

Fudoshiki:


.travis.yml
```diff
@@ -6,23 +6,22 @@ bundler_args: --without development
...
 rvm:
   - 2.2.2
-  - 2.3.0
+  - 2.3.1
   - jruby-9.0.5.0
 
 gemfile:
   - Gemfile
   - gemfiles/rails5master.gemfile
-  - gemfiles/rails5rc1.gemfile
```
estolfo: Thanks for doing these updates, but why did you delete the rails5rc1 gemfile? I'd like to confirm in travis that our tests pass on rc1. If you could add it back in, that'd be great.

Also, we can update jruby to jruby-9.1.1.0 in travis.yml now as well.

Thanks!

Fudoshiki: @estolfo Can You cancel travis builds except last commit?

PR #4290, opened 2016-06-23
Mongo/Criteria map should use pluck, not only

Benjamin-Dobell:
It's not safe to use `only` in a __lot__ (most?) cases. Back in the good ol' days (Mongoid 3) lazy-loading attributes worked as expected, so this would have been fine. However, `only` has essentially been producing invalid ORM objects since Mongoid 4. `map`'s behaviour needs to be changed to accomodate this.

Example crash caused by the existing `map` implementation:
```
> criteria.map :position
ActiveModel::MissingAttributeError: Missing attribute: 'type'.
	from /app/vendor/bundle/ruby/2.3.0/gems/mongoid-5.1.3/lib/mongoid/attributes.rb:97:in `read_attribute'
	from /app/vendor/bundle/ruby/2.3.0/gems/mongoid-5.1.3/lib/mongoid/fields.rb:414:in `block (2 levels) in create_field_getter'
	from /app/vendor/bundle/ruby/2.3.0/gems/enumerize-1.1.1/lib/enumerize/attribute.rb:86:in `type'
	from /app/vendor/bundle/ruby/2.3.0/gems/enumerize-1.1.1/lib/enumerize/base.rb:97:in `public_send'
	from /app/vendor/bundle/ruby/2.3.0/gems/enumerize-1.1.1/lib/enumerize/base.rb:97:in `block in _set_default_value_for_enumerized_attributes'
	from /app/vendor/bundle/ruby/2.3.0/gems/enumerize-1.1.1/lib/enumerize/attribute_map.rb:23:in `block in each'
	from /app/vendor/bundle/ruby/2.3.0/gems/enumerize-1.1.1/lib/enumerize/attribute_map.rb:22:in `each_pair'
	from /app/vendor/bundle/ruby/2.3.0/gems/enumerize-1.1.1/lib/enumerize/attribute_map.rb:22:in `each'
	from /app/vendor/bundle/ruby/2.3.0/gems/enumerize-1.1.1/lib/enumerize/base.rb:95:in `_set_default_value_for_enumerized_attributes'
	from /app/vendor/bundle/ruby/2.3.0/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:432:in `block in make_lambda'
	from /app/vendor/bundle/ruby/2.3.0/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:228:in `block in halting_and_conditional'
[...]

PR #4291, opened 2016-06-23
Mongo/Criteria map should use pluck, not only

Benjamin-Dobell:
Back-ported https://github.com/mongodb/mongoid/pull/4290 to 5.1.x.


johnnyshields:
@estolfo this is fine to backport to 5.1.0-stable, let's merge it.

PR #4292, opened 2016-06-23
MONGOID-4289 Fix cloning with embedded translations

eric-pigeon:
when cloning process_localized_fields only works from the parent
document causing embedded documents with translations to get
set to a hash instead of the proper translation


estolfo:
thanks @eric-pigeon !

PR #4295, opened 2016-07-12
Add sorting options to the cache key

jefmathiot:
The `sort` option is not added to the cache key, leading to wrong results when querying a collection:

```ruby
Mongoid::QueryCache.enabled = true

Band.asc(:name).pluck(:name)
=> ["AC DC", "ZZ Top"]
Band.desc(:name).pluck(:name)
=> ["AC DC", "ZZ Top"]
```

This PR adds the hash returned by the `Mongoid::Criteria::Queryable::Options#sort` method to the cache key.

### DIFF
--- a/lib/mongoid/query_cache.rb
+++ b/lib/mongoid/query_cache.rb
@@ -223,7 +223,7 @@ module Mongoid
 
       def cached_cursor
         if limit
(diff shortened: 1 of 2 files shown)


estolfo:
thanks!

PR #4296, opened 2016-07-12
MONGOID-4296, fixes regression introduced in 5.1.2 with #only and…

jonhyman:
…embedded document fields accessed via []


estolfo:
thanks @jonhyman !

PR #4297, opened 2016-07-16
Make Proxy objects #equal? to themselves

marshalium:
This makes #object_id and #equals? consistent with each other so they are both always called against the Proxy.

Another possibility would be to pass through #object_id and #equals? to the target so that the Proxy and its target would look like the exact same object. This seems more likely to cause confusion/problems since they aren't actually the exact same object, and the code has being calling #object_id on the Proxy for a long time. 

The first commit rewrites the regex slightly. It now always does an exact match except for '__' prefixes. I'm happy to undo that refactoring if there is a reason the old way was better or more correct.

### DIFF
--- a/spec/mongoid/relations/proxy_spec.rb
+++ b/spec/mongoid/relations/proxy_spec.rb
@@ -97,4 +97,28 @@ describe Mongoid::Relations::Proxy do
       end
     end
   end
+
+  describe "equality" do
+    let(:messages) do
+      Person.create.messages
+    end
+
+    it "is #equal? to itself" do
+      expect(messages.equal?(messages)).to eq(true)
+    end
+
+    it "is == to itself" do
+      expect(messages == messages).to eq(true)
+    end
+
+    it "is not #equal? to its target" do
+      expect(messages.equal?(messages.target)).to eq(false)
+      expect(messages.target.equal?(messages)).to eq(false)
+    end
+
+    it "is == to its target" do
+      expect(messages == messages.target).to eq(true)
+      expect(messages.target == messages).to eq(true)
+    end
+  end
 end
--- a/lib/mongoid/relations/proxy.rb
+++ b/lib/mongoid/relations/proxy.rb
@@ -12,7 +12,7 @@ module Mongoid
       # We undefine most methods to get them sent through to the target.
       instance_methods.each do |method|
         undef_method(method) unless
-          method =~ /(^__|^send|^object_id|^respond_to|^tap|^public_send|extend_proxy|extend_proxies)/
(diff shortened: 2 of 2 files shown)


spec/mongoid/relations/proxy_spec.rb
```diff
@@ -97,4 +97,28 @@ module Testable
...
+
+    it "is not #equal? to its target" do
+      expect(messages).to_not equal(messages.target)
+      expect(messages.target).to_not equal(messages)
+    end
+
+    it "is == to its target" do
+      expect(messages).to eq(messages.target)
+      expect(messages.target).to eq(messages)
+    end
```
johnnyshields: To increase clarity I would suggest to use the operator within the `expect()` call, for instance:

```
expect(messages == messages.target).to eq true
```

johnnyshields:
@marshalium according to [the ruby documentation for Object](http://ruby-doc.org/core-2.3.1/Object.html#method-i-eql-3F), `equal?` should never be overridden, as it tests if two Objects are pointers for the same address in Ruby's memory.

I think you should override `eql?` here instead to get the desired behavior.

My proposal:

| Method | messages X messages | messages X messages.target |
| --- | --- | --- |
| == | true | true |
| eql? | true | false |
| equal? | false | false |

spec/mongoid/relations/proxy_spec.rb
marshalium: Thanks for the feedback! I agree. I'll make that change.

marshalium:
@johnnyshields I'm not overriding #equal?, in fact this fix causes the code to leave the original #equal? alone instead of undefining it. The code in Proxy was calling `undef_method :equal?` and so #equal? was getting proxied through to target by #method_missing instead of getting called on the Proxy.

The docs make it sound like normally #eql? should just be the same as #==. They do mention that Numeric handles type conversion differently between #eql? and #==, but I don't think that makes sense for Proxy. Proxy#class will return the target's class so to a user there isn't any type conversion, the class of the proxy and the class of the target will look exactly the same.

johnnyshields:
@marshalium ah, that makes sense. Your way is correct. Then the matrix would become:

| Method | messages X messages | messages X messages.target |
| --- | --- | --- |
| == | true | true |
| eql? | true | true |
| equal? | true | false |

johnnyshields:
@marshalium please touchup the test cases then this looks good to merge.

marshalium:
Great! I've updated the specs. Should be good to go.

estolfo:
thanks @marshalium and @johnnyshields

PR #4298, opened 2016-07-18
Embedded field changes should be persisted by set

dongintercom:
Related ticket: https://jira.mongodb.org/browse/MONGOID-4300

Not sure this is a valid fix. This PR is for discussion for now.

When using mongoid in our project, we use `set` to persist certain embedded fields that are of one-to-one relationship. I found that in `settable.rb`, changes to these fields are ignored since `attributes[field]` won't have the updated value for embedded fields, instead `nil` is assigned to these fields.

This PR might have some hidden issues that I am not aware of, please take a look and let me know what you think. Thanks @estolfo


dongintercom:
Resolved in https://github.com/mongodb/mongoid/pull/4302

PR #4300, opened 2016-07-21
Add support for $elemMatch in Matchable#matches?

jonhyman:


lib/mongoid/matchable/elem_match.rb
```diff
@@ -0,0 +1,37 @@
...
+      #
+      # @param [ Hash ] value The values to check.
+      #
+      # @return [ true, false ] If the values match.
+      def matches?(value)
+        if !@attribute.is_a?(Array) || !value.kind_of?(Hash) || !value["$elemMatch"].kind_of?(Hash)
+          return false
+        end
+
+        @attribute.each do |sub_document|
```
estolfo: Can we just change this to @attribute.all? as well as the line below (line 22)? We would need the explicit returns in that case.

estolfo:
Hi @jonhyman 
Thanks for this PR. Can we add a test for when a sub document matches but the keys aren't in the same order? I'm not 100% sure what the server behavior is but we should check consistency with it.
Thanks!
Emily

lib/mongoid/matchable/elem_match.rb
jonhyman: Great idea, it makes the code a lot simpler.

jonhyman:
@estolfo refactor done and test added. The order of the syntax doesn't matter, confirmed here:

```ruby
class Subdoc
  include Mongoid::Document
  embedded_in :foo
  field :a
  field :b
end
class Foo
  include Mongoid::Document
  embeds_many :subdocs
end

f = Foo.new
f.subdocs.new(:a => 1, :b => 1)
f.save()

Foo.where(:subdocs => {:$elemMatch => {:b => 1, :a => 1}}).count
# => 1
```

estolfo:
Great thank you!

I verified that the server doesn't consider order either:

```
> db.test.insert({x: 1, subdocs: [ { a: 1, b: 2 } ] })
WriteResult({ "nInserted" : 1 })
> db.test.find({subdocs: {'$elemMatch': {b: 2, a: 1}}})
{ "_id" : ObjectId("57924adda0e9817964ceed13"), "x" : 1, "subdocs" : [ { "a" : 1, "b" : 2 } ] }
> 
```

lib/mongoid/matchable/elem_match.rb
```diff
@@ -0,0 +1,28 @@
...
+      # @return [ true, false ] If the values match.
+      def matches?(value)
+        if !@attribute.is_a?(Array) || !value.kind_of?(Hash) || !value["$elemMatch"].kind_of?(Hash)
+          return false
+        end
+
+        return @attribute.any? do |sub_document|
+          value["$elemMatch"].all? do |k, v|
+            Matchable.matcher(sub_document, k, v).matches?(v)
+          end
```
estolfo: so much better now = )

PR #4302, opened 2016-08-01
MONGOID-4300 Don't double set a relation

estolfo:


spec/mongoid/persistable/settable_spec.rb
```diff
@@ -134,6 +134,35 @@
...
+        let(:home_phone) do
+          Phone.new(number: "555-555-5555")
+        end
+
+        let(:office_phone) do
+          Phone.new(number: "666-666-6666")
+        end
+
+        it "should persist changes of embeds_one field" do
+          person.set(pet: pet)
```
dongintercom: what is the difference between `set({pet: pet.as_document})` and `set(pet: pet)`?

dongintercom:
Looks great! 👍

spec/mongoid/persistable/settable_spec.rb
estolfo: I would stick with using ``pet`` in this case instead of ``#as_document``.
In general, ``#as_document`` will return the raw attributes hash, as it is in the database. When working with objects, I would stick with the object representation and not its hash representation. Otherwise, you might lose some OO behavior.

PR #4303, opened 2016-08-01
Improve test suite performance and close opened connections

estolfo:


johnnyshields:
@estolfo I'd recommend:
1. run the tests with a memory profiler for JRuby
2. go commit-by-commit here and try to isolate the change which caused the JRuby memory leak.

In my experience doing things like increasing heap size only mask the underlying problem. In other words, if the Mongoid 5.x tests ran with 1GB heap size then the Mongoid 6.x tests should as well.

estolfo:
@johnnyshields 

I don't believe Mongoid is responsible for the JRuby memory leaks.
If you take a look at this 5.1.0-stable build: https://travis-ci.org/mongodb/mongoid/builds/148856303

JRuby 1.7 runs just fine and JRuby 9.1.2.0 takes a massive amount of time and eventually fails (runs out of memory).

Unfortunately I can't test JRuby 1.7 with Mongoid 6 because only JRuby >= 9.0 is supported.

Rail's test suite isn't even passing on JRuby 9.0: https://travis-ci.org/rails/rails

I've tried to run a memory profiler but nothing has stood out to me yet. Let me know if you find anything!

johnnyshields:
@headius we're having issues getting the tests to pass for Mongoid (Rails ORM adapter for MongoDB) JRuby 9.0.0.0. It looks like there's a memory leak in JRuby affecting both our old and new versions.

Any pointers on how we can debug this would be appreciated. You can see the failing tests here: https://travis-ci.org/mongodb/mongoid/builds/148858771

estolfo:
@johnnyshields I already have this ticket open with the JRuby team = )

https://github.com/jruby/jruby/issues/4012

johnnyshields:
oh ok sorry, thanks.

@estolfo I propose we mark the JRuby tests as "allowed failures" until we get a response. I don't think this should hold up the entire release, we can just give a caveat that JRuby users should continue to use Mongoid 5 on JRuby 1.7 until the underlying issue is fixed. (If the issue is indeed JRuby related then JRuby should be able to fix it with no change in Mongoid.)

estolfo:
No worries, I appreciate your suggestion!

I agree with you - I think that's the best way to go. Especially as it looks like JRuby users won't be able to upgrade to Rails 5 anyway; their tests aren't passing on JRuby 9.0 either.

PR #4308, opened 2016-08-10
MONGOID-4173 Fix nested eager loading

estolfo:


sivagollapalli:
Looks like this issue didn't include in `5.1.4` release (https://github.com/mongodb/mongoid/releases/tag/v5.1.4). Will this be available from `6.0.0.` ?

estolfo:
Yes, it'll be in 6.0. It's marked in Jira as such: https://jira.mongodb.org/browse/MONGOID-4173

sivagollapalli:
Looks like JIRA yet to be updated. On JIRA it still shows fix versions as `5.1.0` and `6.0.0`. Could it available on `5.1` as it reduces many performance issues on my app?

estolfo:
The Jira ticket is up-to-date: it was fixed in version 5.1.0 but there was a regression, so the ticket was reopened and the regression fix is going into 6.0.
If we release a version 5.1.5, I'll put the fix in there.

PR #4313, opened 2016-08-19
Add a test to show that _type mismatch is not handled gracefully

dongintercom:


johnnyshields:
@dongintercom IMHO the answer here is to create a new error class `Mongoid::UnknownPolymorphicType` (or something like that) and raise it instead of `NameError`.

It does not seem correct that we should coerce the model to a known type, as strange behavior / data corruption could result when the model is saved. (We could add a `fallback_type:` parameter or something here but I think it's a pretty remove use case...)

estolfo:
Closing this PR, as the error will be updated via this ticket: https://jira.mongodb.org/browse/MONGOID-4354

PR #4314, opened 2016-08-19
Add docs for create/remove indexes called on models

miguelgraz:
Sometimes creating/removing indexes for the entire database can be very expensive, when searching for the same behavior for just one collection I found out that it can be used directly on models, I thought that could be interesting to have mentioned on the docs :smile:


johnnyshields:
It's good to document this. I use it all the time.

PR #4315, opened 2016-08-25
Add typed attributes

Tensho:
I've read many issues about typed attributes absence in Mongoid (particularly, the latest is ["Custom types typecasting problem"](https://jira.mongodb.org/browse/MONGOID-2533), I guess). Despite of `#attributes` mongonized nature, I'd like to add method which allows to retrieve type casted (demongoized ) attributes.
This is my first PR to Mongoid project, so don't hesitate to abandon it or give some advices for the improvements.

### DIFF
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -242,6 +242,16 @@ module Mongoid
         (selection.values.first == 1 && !selection_included?(name, selection, field))
     end
 
+    # Return type casted attributes.
+    #
+    # @example Type casted attributes.
+    #   document.typed_attributes
+    #
+    # @return [ Object ] The hash with keys and values of the type casted attributes.
+    def typed_attributes
+      attribute_names.map { |name| [name, send(name)] }.to_h
+    end
+
     private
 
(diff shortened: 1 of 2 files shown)


estolfo:
Hi @Tensho 
Can you tell me if you know of a ActiveRecord method that does this? If ActiveRecord doesn't provide something simliar, I would hesitate to add it into Mongoid.
Thanks

Tensho:
Hi @estolfo 
Actually `ActiveRecord::Base#attributes` implicitly infers from column info and converts to the expected type. But the difference is in internal representation, I guess, ActiveRecord doesn't have the concept of mongoized objects, so it's hard for me to find something similar. Particularly, if I create date type column in ActiveRecord migration, then I get object of `Date` class in the model for that attribute, but if I declare field type as `Date` inside Mongoid document, then I get object of `Time` (mogoized) class for that attribute.

I'd like to have described behaviour for `Mongoid::Document#attributes` and don't add another one method, but as far as I understand, it's ideologically forbidden in `mongoid`,because requires a lot of internal changes in the code).

PR #4317, opened 2016-08-26
Use right database when override_database + options.

sylvain-8422:
Mongoid would not always use the right database to persist data when persistence options (even an empty hash) are given, while at the same time using the `Mongoid.override_database` feature.

Our application is using `Mongoid.override_database` and although we are not using persistence options explicitly, empty hashes would end up in `Options#mongo_client`, and the method would return a `mongo` client that would use the default database, and not the one given to `override_database`.

We experienced very weird behaviour where one document would be saved in one database, and others in another one.

Here is the code sample that I used to figure out the problem, for your entertainment:

```
class Foo
  include Mongoid::Document
  has_many :bars
  accepts_nested_attributes_for :bars
end

class Bar
  include Mongoid::Document
  field :name
  belongs_to :foo
end

# 'first_database' in mongoid.yml
Mongoid.override_database(:second_database)

foo = Foo.create!
attributes = { '0' => { 'name' => 'name' } }
foo.update_attributes!(bars_attributes: attributes)
```

With the code above, the `Foo` document would be persisted in `second_database` and the new `Bar` document would be create in `first_database`. (Because an empty hash is given to `mongo_client` while persisting the `Bar` document.)

If empty hashes are filtered, the same outcome can be achieved by using something like:

```
foo.with(connect_timeout: 10).update_attributes!(bars_attributes: attributes)
```

### TLDR

Mongoid would not always persist to the right database. This is bad.

See additional comments in the code.

Questions, comments, and merge are welcome! Thank you! :-)

### DIFF
--- a/lib/mongoid/clients/options.rb
+++ b/lib/mongoid/clients/options.rb
@@ -44,12 +44,12 @@ module Mongoid
 
       def mongo_client
         tmp = persistence_options
-        if opts = tmp && tmp.dup
+        if (opts = tmp && !tmp.empty? && tmp.dup)
           if opts[:client]
             client = Clients.with_name(opts[:client])
           else
             client = Clients.with_name(self.class.client_name)
-            client.use(self.class.database_name)
+            client = client.use(self.class.database_name)
           end
           client.with(opts.reject{ |k, v| k == :collection || k == :client })
         end
--- a/spec/mongoid/clients_spec.rb
+++ b/spec/mongoid/clients_spec.rb
@@ -731,27 +731,42 @@ describe Mongoid::Clients do
 
     context "when the override is global" do
 
-      before do
-        Mongoid.override_database(:mongoid_optional)
-      end
+      shared_examples_for "a global database override" do
 
-      after do
-        Band.delete_all
-        Mongoid.override_database(nil)
-      end
+        before do
+          Mongoid.override_database(:mongoid_optional)
+        end
 
-      let!(:band) do
-        Band.create(name: "Tool")
-      end
+        after do
+          Band.delete_all
+          Mongoid.override_database(nil)
+        end
 
-      it "persists to the overridden database" do
-        Band.mongo_client.with(database: :mongoid_optional) do |sess|
-          expect(sess[:bands].find(name: "Tool")).to_not be_nil
+        let!(:band) do
+          klass.create(name: "Tool")
+        end
+
+        it "persists to the overridden database" do
+          mongo_client = Band.mongo_client.with(database: :mongoid_optional)
+          expect(mongo_client[:bands].count(name: "Tool")).to eq(1)
+        end
+
+        it 'uses that database for the model mongo_client' do
+          expect(Band.mongo_client.database.name).to eq('mongoid_optional')
         end
       end
 
-      it 'uses that database for the model mongo_client' do
-        expect(Band.mongo_client.database.name).to eq('mongoid_optional')
+      context "when normal usage" do
+        let(:klass) { Band }
+
+        it_behaves_like "a global database override"
+      end
+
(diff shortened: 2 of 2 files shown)


lib/mongoid/clients/options.rb
```diff
@@ -44,12 +44,12 @@ def persistence_options
 
       def mongo_client
         tmp = persistence_options
-        if opts = tmp && tmp.dup
+        if (opts = tmp && !tmp.empty? && tmp.dup)
```
sylvain-8422: The bug was found because Mongoid code gave empty hashes as `persistence_options`.

The new code handles those empty hashes as if there are no persistence options.

lib/mongoid/clients/options.rb
```diff
@@ -44,12 +44,12 @@ def persistence_options
...
           if opts[:client]
             client = Clients.with_name(opts[:client])
           else
             client = Clients.with_name(self.class.client_name)
-            client.use(self.class.database_name)
+            client = client.use(self.class.database_name)
```
sylvain-8422:
```
def use(name)
  with(database: name)
end
```

Unlike `moped`, `mongo` does not switch database by calling `use`, you need to use the new client.

spec/mongoid/clients_spec.rb
```diff
@@ -731,27 +731,42 @@
...
+        end
+
+        it "persists to the overridden database" do
+          mongo_client = Band.mongo_client.with(database: :mongoid_optional)
+          expect(mongo_client[:bands].count(name: "Tool")).to eq(1)
+        end
 
-      it "persists to the overridden database" do
-        Band.mongo_client.with(database: :mongoid_optional) do |sess|
-          expect(sess[:bands].find(name: "Tool")).to_not be_nil
```
sylvain-8422: This test was not executed. `with` does not take a block as argument, and I guess by the name `sess` that it was an old test that used to work with `moped`, but did not work anymore. It just did nothing at all.

Also, `find` would always return a `Mongo::Collection::View`, which is not `nil`, so the test would never fail!

spec/mongoid/clients_spec.rb
```diff
@@ -731,27 +731,42 @@
...
-        expect(Band.mongo_client.database.name).to eq('mongoid_optional')
+      context "when normal usage" do
+        let(:klass) { Band }
+
+        it_behaves_like "a global database override"
+      end
+
+      context "when overriding the persistence options" do
+
+        let(:klass) { Band.with(connect_timeout: 10) }
```
sylvain-8422: Any options would do here. The point was to trigger the problem in `Mongoid::Clients::Options` by providing a non-empty hash for `@persistence_options`.

sylvain-8422:
Oh, and by the way, the test failure at `spec/mongoid/contextual/mongo_spec.rb:128` was there before my changes. Ain't my fault! ;-)

sylvain-8422:
Is there anything wrong with my _Pull Request_ ?

estolfo:
Hi @sylvain-8422 
Sorry for the delay and thank you for the pull request! I'll get back to you within the next day on this.

sylvain-8422:
Thank you @estolfo !

estolfo:
Thank you @sylvain-8422 !!

estolfo:
I've planned a [5.1.5 release](https://jira.mongodb.org/browse/MONGOID/fixforversion/17486/?selectedTab=com.atlassian.jira.jira-projects-plugin:version-summary-panel) in Jira for two weeks from now. It will have this change and in the meantime, I'll figure out why that test is failing. Thanks again for this!

sylvain-8422:
You are welcome! 5.1.5 will be the best 5.x release ever released! :-)

PR #4318, opened 2016-08-28
Fix metadata classify

Melanitski:
Could be some problems with classify
Example:

```ruby
class Project::Interview < ::Article::Abstract

end
```

```ruby
class Article::Abstract
   include Mongoid::Document
end
```


durran:
Could you write a test please as well that illustrates the problem?

estolfo:
Please reopen this is you're able to provide a test that illustrates the problem.

PR #4325, opened 2016-09-24
MONGOID-4324 handle string value in localized fields

kaizencodes:


simi:
What's the usecase/problem here?

kaizencodes:
If you add localize: true to an existing field then those instances contain a string, not hash. Without this pr those cannot be modified.

simi:
I did this before and I wrote `migration` script for this using driver directly to switch from string to hash notation.

I think I can write some notes with example script somewhere.
Would you mind to merge PR to README with this @estolfo?

estolfo:
Hi @kaizencodes 
Thanks for reporting this and for the pull request. Can you please add a test?

Hi @simi 
I'm not sure I understand what you're requesting. Can you give me an example of what you'd like added to the README?

simi:
@estolfo If I understand this well, it introduces pattern where localized column can be both hash (as before) and string (new in this PR).

Isn't this breaking `find_by`, `where` and similar statements for localized columns where hash is expected?

See this example. I tested only master, but I wrote expected results with this patch in comments. Model `Post` represent old model (with `name` field unlocalized), `LocalizedPost` uses same collection, but is having localized `name` field.

```ruby
require 'mongoid'
require 'minitest/autorun'                                
# Ensure backward compatibility with Minitest 4           
# Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)                        
#   
Mongoid.configure.connect_to("mongoid_test")              
class Post                                                
  include Mongoid::Document
  store_in collection: 'posts'                            
    
  field :name                                             
end  
    
class PostLocalized                                       
  include Mongoid::Document                               
    
  store_in collection: 'posts'                            
  field :name, localize: true                             
end  
    
class BugTest < Minitest::Test                            
  def test_select                                         
    Post.delete_all                                       
    PostLocalized.delete_all                              
    
    Post.create(name: 'test')                             
    PostLocalized.create(name: 'test-localized')          
    
    assert_equal Post.first.name, 'test'                  
    
    # will fail on master                                 
    # will not fail with this PR, because reader is patched to handle this case
    # assert_equal PostLocalized.first.name, 'test'       
    
    # you can find localized post, because name is stored as hash
    assert_equal PostLocalized.where(name: 'test-localized').count, 1                                
    
    # will fail with this PR, because post with name test is storing name as string, but hash is expected
    # generated query is "query"=>{"name.en"=>"test"}}
    # but actual name is stored as {"name" => "test"} because it was created with unlocalized field
    assert_equal PostLocalized.where(name: 'test').count, 1
  end
end                                                                                                       
```

I experienced this migration from unlocalized to localized field before and I wrote script to change all unlocalized fields values from string to localized hash using pure driver (to avoid `mongoid` typing). For example from `"test"` to  `{"name.en" => "test"}`.

Ping me for better demo if there are some problems with my explanation.

estolfo:
Hi @simi 
I understand now what this pull request is doing, thanks for the explanation.

This is not a change I'd like to include in Mongoid. If the schema of a Model is changed (for example, if a field goes from not being localized to localized), a script should be run to convert the existing data. The same goes for changing the type of a field (from String to Integer, for example). As you've mentioned, introducing a change like this would break other Mongoid functionality, such as queries.

I encourage anyone looking to change the type or metadata for a Model field to write a script to convert the existing data. 

Thanks
Emily

simi:
Cool. At first I suggested to write note about this somewhere into docs, so you can localize field, but you need to handle data migration on your own with some simple script like this.

```ruby
Model.all.each do |instance|
  # directly save instance.field_name = {"field_name.i18n_locale" => instance.field_name}
end
```

estolfo:
I understand now what you were suggesting. Pull requests to the docs are welcome as well. They are found in the [/docs](https://github.com/mongodb/mongoid/tree/master/docs) folder.

I can update them unless someone beats me to it ; )

PR #4327, opened 2016-09-30
Localized attribute presence lookup which has a nil value should not fail

nicolasblanco:
From the model Person in the specs...

```ruby
class Person
  # ...
  field :desc, localize: true
  # ...
end
```

```ruby
  p = Person.new
  p.desc = nil
  p.desc? #=> false

  p = Person.new
  p.desc? #=>
  NoMethodError: undefined method `[]' for nil:NilClass
  from /Users/nicolasblanco/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/mongoid-6.0.0/lib/mongoid/fields/localized.rb:80:in `lookup'
```

After patch:
```ruby
  p = Person.new
  p.desc = nil
  p.desc? #=> false

  p = Person.new
  p.desc? #=> false
```

This patch fixes the exception by doing the localized lookup only if there is an actual value and also adds one more test to the tests suite.

### DIFF
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -336,7 +336,7 @@ module Mongoid
     end
 
     def lookup_attribute_presence(name, value)
-      if localized_fields.has_key?(name)
+      if localized_fields.has_key?(name) && value
         value = localized_fields[name].send(:lookup, value)
       end
       value.present?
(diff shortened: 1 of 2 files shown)


estolfo:
thanks @nicolasblanco !

PR #4328, opened 2016-10-03
[6.1.0] add string “on” to list of true boolean values

yads:
HTML forms use "on" to indicate a checked checkbox value. This would provide seamless transition from an html form to a boolean value on a model.


estolfo:
Thanks for this, @yads 

Because it's a new feature, I've opened a ticket and marked it for 6.1.0.
https://jira.mongodb.org/browse/MONGOID-4329
Thanks

johnnyshields:
Regex here is missing start of string specifier. Better one would be:

```
/\A(true|t|yes|y|on|1|1.0)\z/
```

johnnyshields:
@estolfo please merge this. Test failures are not related to this PR.

PR #4329, opened 2016-10-13
Add support for cloning documents that have nil legacy attributes

okoriko:
This fixes cloning documents with legacy schemas in which one of the key is set to `nil`.
Before it would raise an error such as:
```
Mongoid::Errors::UnknownAttribute (
message:
  Attempted to set a value for '...' which is not allowed on the model ...
summary:
  Without including Mongoid::Attributes::Dynamic in your model and the attribute does not already exist in the attributes hash, attempting to call ...= for it is not allowed. This is also triggered by passing the attribute to any method that accepts an attributes hash, and is raised instead of getting a NoMethodError.
resolution:
  You can include Mongoid::Attributes::Dynamic if you expect to be writing values for undefined fields often.):
```

### DIFF
--- a/lib/mongoid/copyable.rb
+++ b/lib/mongoid/copyable.rb
@@ -24,7 +24,10 @@ module Mongoid
       attrs = clone_document.except("_id", "id")
       dynamic_attrs = {}
       attrs.reject! do |attr_name, value|
-        dynamic_attrs[attr_name] = value unless self.attribute_names.include?(attr_name)
+        unless self.attribute_names.include?(attr_name)
+          dynamic_attrs[attr_name] = value
+          true
+        end
       end
       self.class.new(attrs).tap do |object|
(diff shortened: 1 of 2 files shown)


lib/mongoid/copyable.rb
```diff
@@ -24,7 +24,10 @@ def clone
       attrs = clone_document.except("_id", "id")
       dynamic_attrs = {}
       attrs.reject! do |attr_name, value|
-        dynamic_attrs[attr_name] = value unless self.attribute_names.include?(attr_name)
+        unless self.attribute_names.include?(attr_name)
+          dynamic_attrs[attr_name] = value
+          true
+        end
```
estolfo: We could also use #merge!

```ruby
dynamic_attrs.merge!(attr_name => value) unless self.attribute_names.include?(attr_name)
```

PR #4337, opened 2016-10-27
MONGOID-3490 Handle localized boolean fields correctly

estolfo:


johnnyshields:
@estolfo I made a comment in ticket #4147... I think we should also remove empty strings from the localized values. Storing empty strings makes the whole thing more difficult to work with:

```
p.title_translations
# currently
=> {"de"=>"", "en"=>"Bibliography", "es"=>"", "fr"=>"", "it"=>"", "pt-BR"=>""}

# should be
=> {"en"=>"Bibliography"}
```

estolfo:
Hi @johnnyshields Yes, I saw your comment but I need to investigate more to see if that would affect behavior anywhere and would be backwards compatible.

PR #4343, opened 2016-11-11
Update Relations::Reflections.reflect_on_all_associations method

ivanovaleksey:
Hi guys!

I have noticed that there is something wrong with `Mongoid::Relations::Reflections.reflect_on_all_associations` method.

It said in rdoc that method supports `[ Array<String, Symbol> ]` as argument but it works only with symbols because `meta.macro` is `Symbol`.

Now I just fixed it so the method indeed can accept strings.
But maybe we just should left the method as is and simply update rdoc?

Also I've made the method work even without any arguments (in this case it returns all relations).
I saw that `ActiveRecord` [works](https://github.com/rails/rails/blob/ac027338e4a165273607dccee49a3d38bc836794/activerecord/lib/active_record/reflection.rb#L98) this way and it is useful for a task I'm working on.
But I am not sure if it would be useful for others. 

Thank you.

### DIFF
--- a/lib/mongoid/relations/reflections.rb
+++ b/lib/mongoid/relations/reflections.rb
@@ -24,7 +24,7 @@ module Mongoid
       # @example Find multiple relation metadata by macro.
       #   person.reflect_on_all_associations(:embeds_many)
       #
-      # @param [ Array<String, Symbol> ] *macros The relation macros.
+      # @param [ Array<Symbol> ] *macros The relation macros.
       #
       # @return [ Array<Metadata> ] The matching relation metadata.
       def reflect_on_all_associations(*macros)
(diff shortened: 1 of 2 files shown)


estolfo:
Hi @ivanovaleksey

Thanks for this pull request. I'm in favor of updating the behavior to return all associations when provided no argument to be consistent with ActiveRecord. I'm not in favor of supporting the arguments as Symbols, as that's not consistent with ActiveRecord. Would you like to update the pull request to show the correct documentation (i.e. take out the "String" reference) and to return all associations if the argument is not provided?
Thanks
Emily

ivanovaleksey:
Hi @estolfo, you said Symbols are not consistent with ActiveRecord but didn't you mean Strings?
As far as I can see ActiveRecord [works](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/reflection.rb#L97) with Symbols.
Am I missing something?

estolfo:
Sorry I meant Strings!
I've updated my comment.

ivanovaleksey:
I see, thanks! I just want to make sure I didn't miss anything.
Will update the PR.

estolfo:
thank you!

ivanovaleksey:
Emily, I updated PR.

estolfo:
great, thank you!

PR #4344, opened 2016-11-14
Fixed bug in relationships with `inverse_of` field

syntruth:
This fixes a bug in `Mongoid::Relations::Referenced::Many` when it goes
to find the inverse of a relationship within the
`#with_inverse_field_criterion` method. The method was trying to use
`:inverse_of_field` which doesn't exist on the metadata and is, instead,
`:inverse_of', resulting in a method-not-found error.


estolfo:
Thanks for this @syntruth 
Can you please add a test so I can merge it?

syntruth:
Test added. I wasn't _quite_ sure where it should go, so I made a best guess.

estolfo:
thanks @syntruth !

PR #4345, opened 2016-11-14
Fix : map_reduce read mode is forced to "primary"

cesar-tonnoir:
### Context
The read mode is forced to `primary` on a `map_reduce` client, when `inline == 1`, which makes it impossible to apply any other read mode (`secondary`, `secondary_preferred`...) when performing a `map_reduce` operation.

### Confirmation
* MongoDB read replicas remain idle while performing a succession of `map_reduce` operations, even when the read mode is set to `secondary` in Mongoid configuration
* By outputing MongoDB logs in a Rails console, I get confirmation that the requests are sent to the primary node only.

### Proposed Correction
Allow the `map_reduce` operation to use the client's `options` when applying the `command`.
This way, assuming you've configured a client in `secondary` read mode, a query like `criteria.with(client: :secondary).map_reduce(map, reduce).out(inline: 1)` would perform the operation in `secondary` read mode.

### DIFF
--- a/lib/mongoid/contextual/map_reduce.rb
+++ b/lib/mongoid/contextual/map_reduce.rb
@@ -299,7 +299,7 @@ module Mongoid
       # @since 3.0.0
       def results
         raise Errors::NoMapReduceOutput.new(command) unless command[:out]
-        @results ||= __client__.command(command).first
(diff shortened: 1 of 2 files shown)


estolfo:
thanks @cesar-tonnoir !

PR #4346, opened 2016-11-29
Remove Attributes::Readonly#loaded? method

romanlehnert:
Before this change, Attributes::ReadOnly defined a private `#loaded?`
method on every mongoid document.

Since it is a common name for a public method, this may be seen as
namespace pollution. In order to reduce the pollution, this method was
dissolved in its one and only call and is removed with this change.


romanlehnert:
There is a corresponding Jira issue here: https://jira.mongodb.org/browse/MONGOID-4393

estolfo:
superseded by https://github.com/mongodb/mongoid/pull/4347

PR #4350, opened 2016-11-30
Bypass the query cache when reloading a document

eugeneius:
When I call `#reload` on a document, I expect the data to be reloaded from the server, even if query caching is enabled.


estolfo:
thanks @eugeneius

PR #4362, opened 2016-12-13
MONGOID-4398 Synchronize driver client creation

estolfo:


lib/mongoid/clients.rb
```diff
@@ -61,7 +61,12 @@ def disconnect
       #
       # @since 3.0.0
       def with_name(name)
-        clients[name.to_sym] ||= Clients::Factory.create(name)
+        name_as_symbol = name.to_sym
+        return clients[name_as_symbol] if clients[name_as_symbol]
+        CREATE_LOCK.synchronize do
+          return clients[name_as_symbol] if clients[name_as_symbol]
+          clients[name_as_symbol] = Clients::Factory.create(name)
```
ivoanjo: Since you're inside the lock, you can use 
```ruby
CREATE_LOCK.synchronize do
  clients[name_as_symbol] ||= Clients::Factory.create(name)
end
```

as the `||=` is protected by it.

estolfo: But any thread (1) waiting to access a certain client would have to wait for the lock to be released if another thread (2) were creating a client (even if the thread (2) were creating a client different from the one the thread (1) were waiting for).

ivoanjo: Indeed. I meant to just shorten the two lines inside the mutex to one, but I agree with allowing readers to access the hash outside the lock.

So my suggestion was just
```ruby
def with_name(name)
  name_as_symbol = name.to_sym
  return clients[name_as_symbol] if clients[name_as_symbol]

  CREATE_LOCK.synchronize do
    clients[name_as_symbol] ||= Clients::Factory.create(name)
  end
```

:smile:

estolfo: Oh, ok, sorry for misunderstanding! That makes sense. Thanks = )

PR #4363, opened 2016-12-16
MONGOID-4237 Support Converting BigDecimal objects to BSON::Decimal128

estolfo:
Summary of the changes:

* New Mongoid.map_big_decimal_to_decimal128 config option. Default is false

When a field is a BigDecimal type:

BigDecimal#demongoize:
- object is BSON::Decimal128
    - mapping is true
        - object.to_big_decimal
    - mapping is false
        - raise unmapped type error
- object is not BSON::Decimal128
    - previous behavior for converting to BigDecimal

BigDecimal#mongoize:
-  mapping is true
    - object is BigDecimal
        - create Decimal128 from BigDecimal
    - object is numeric
        - create Decimal128 from object.string
    - object is neither BigDecimal nor numeric
        - nil
- mapping is false
    - previous behavior for converting to string

When querying:
- mapping is true
    - object is BigDecimal
       - create BSON::Decimal128 from BigDecimal
    - object is BSON::Decimal128
        - object
- mapping is false
    - (previous behavior) object.to_s


estolfo:
Hi @johnnyshields 

Would you like to take a look at this pull request and let me know what you think?
Thanks
Emily

johnnyshields:
I'll look at the code in a bit but I think there are two changes needed to the spec

1) `BigDecimal#demongoize --> object is BSON::Decimal128`

This should case should call `object.to_big_decimal` regardless of whether mapping is true or false. The reason being that mapping only controls what goes into the DB, not what comes out. Consider the case where you have mapping true and later change it to false, there's no reason you should not be able to demongoize it.

2) `BigDecimal#mongoize --> mapping is true`

Please consider the case when object is a string here

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -67,7 +77,31 @@ def demongoize(object)
...
+          end
+        end
+
+        private
+
+        def demongoize_from_decimal128(object)
+          if Mongoid.map_big_decimal_to_decimal128
+            object.to_big_decimal
+          else
+            raise Mongoid::Errors::UnmappedBSONType.new(object)
```
johnnyshields: We can remove this error raising here; if there's a Decimal128 already in the DB there's no harm to convert it to a BigDecimal if that's the mongoid type.

lib/mongoid/errors/unmapped_bson_type.rb
```diff
@@ -0,0 +1,22 @@
...
+      # Create the unmapped BSON type error.
+      #
+      # @example Create the new unmapped BSON type error.
+      #   EagerLoad.new(:preferences)
+      #
+      # @param [ Symbol ] value The object that cannot be mapped.
+      #
+      # @since 6.1.0
+      def initialize(value)
+        super(compose_message("unmapped_bson_type", { value: value }))
```
johnnyshields: As per my comment below this error class can be deleted.

johnnyshields:
Code looks good but please see comments about spec.

estolfo:
Hi @johnnyshields 

Thank you very much for your feedback!

Addressing #1:

Calling ``object.to_big_decimal`` regardless of whether mapping is true or false could result in loss of precision when roundtripping a BSON::Decimal128. I would like to only convert a BSON::Decimal128 to a BigDecimal if the user opts-in. For example:

This BSON::Decimal128 exists in the db:
```ruby
> decimal128 = BSON::Decimal128.new("2.00")
=> BSON::Decimal128('2.00')
```
The BSON::Decimal128 is demongoized, #to_big_decimal is called on it:
```ruby
> big_decimal = decimal128.to_big_decimal
=> #<BigDecimal:7f8c31b12230,'0.2E1',9(18)>
> big_decimal.to_s
=> "2.0"
```
The BigDecimal is then re-mongoized by creating a BSON::Decimal128: 
```ruby
> remongoized_decimal128 = BSON::Decimal128.new(big_decimal)
=> BSON::Decimal128('2')
> remongoized_decimal128 == decimal128
=> false
```

Furthermore, the user always has the option of setting the field type to BSON::Decimal128 so that all DB objects can be demongoized to Ruby objects. Then, the user can convert to BigDecimal if he/she needs to do operations with that number. You can see my PR [here](https://github.com/mongodb/mongoid/pull/4364) for adding support for a BSON::Decimal128 field type in queries.

Addressing #2:

``BigDecimal#mongoize`` does indeed handle the String case. The behavior for handling a String is identical to that when the mapping is false: if a String is numeric, it is considered a valid value for the field. When mapping is true, the field is used to construct a BSON::Decimal128. See the code [here](https://github.com/mongodb/mongoid/pull/4363/files#diff-67b85eefb8a3b22fca557b8af069986cR103) which uses the [``#numeric?`` extension method on String](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/extensions/string.rb#L84-L86) and the tests [here](https://github.com/mongodb/mongoid/blob/dd5c3f0ed897b0b7f10bccee68b64865513e5a05/spec/mongoid/extensions/big_decimal_spec.rb#L656-L676).

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -67,7 +77,31 @@ def demongoize(object)
...
+          end
+        end
+
+        def mongoize_to_decimal128(object)
+          if object.is_a?(BigDecimal)
+            BSON::Decimal128.new(object)
+          elsif object.numeric?
+            BSON::Decimal128.new(object.to_s)
```
estolfo: If the string is numeric, it will be used to construct a BSON::Decimal128. See my comment [here](https://github.com/mongodb/mongoid/pull/4363#issuecomment-267926616)

lib/mongoid/extensions/big_decimal.rb
estolfo: See my comment [here](https://github.com/mongodb/mongoid/pull/4363#issuecomment-267926616). There is a potential loss of precision if a Decimal128 is roundtripped (and automatically converted to a Ruby BigDecimal)

johnnyshields:
Re #1 considering that the user has explicitly specified type BigDecimal type on the field, they should be aware of the side effect of losing trailing zeroes. I think the code is much simpler here if we remove the extra logic branch which throws the, and behaves much more rationally/flexibly.

estolfo:
Hi @johnnyshields 

I think we need to be very explicit if there is a potential loss of precision. The Decimal128 type was introduced into MongoDB because users needed the extreme level of precision, in particular, financial institutions. There is the ``map_big_decimal_to_decimal128`` config option for automatically converting between BigDecimal and BSON::Decimal128. If this is set to false, the expectation should be that there is no automatic mapping whatsoever.

Given this scenario: a user has a Model field set to the BigDecimal type, the ``map_big_decimal_to_decimal128`` config option set to false/is the default, he/she gets an error when there is a BSON::Decimal128 type in that Model's field in the DB and Mongoid tries to read it. The user has the option at this point to set the ``map_big_decimal_to_decimal128`` to true or to run a script to migrate all fields to BSON::Decimal128 and then change the Model field type to BSON::Decimal128. I would much rather be very explicit here and raise an error than do an silent conversion without the user opting-in.

johnnyshields:
If the user declares `type: BigDecimal` they are effectively saying "I don't care about trailing zeroes, since the type I am using (BigDecimal) drops them." Trailing zeroes are not an issue in 95% of real world use cases--i.e. mostly financial data, where the precision is defined by a base currency unit (e.g. cents)--we are creating a headache to here to cater to a small subset of users who shouldn't be using `type: BigDecimal` at all anyway.

estolfo:
I'm concerned more about the case when a user *already* has a field that is BigDecimal and upgrades. He/she would have the mapping as false (because it's the default) and could be completely unaware of the mapping option. If there are some Decimal128 values in the db from another source, the user has to know about the loss of precision caused by the mapping.

johnnyshields:
@estolfo the case you describe is extremely unlikely to happen...

What is much more likely to happen is that a user turns on `map_big_decimal_to_decimal128`, some data gets persisted as Decimal128 in the DB, then for some reason has to rollback their server version. All of a sudden they will be bombarded with "mapping errors" and the application will come screeching to halt.

Please reconsider this.

durran:
@johnnyshields If I understand your last comment correctly, I don't see how the mapping errors would occur. If you had to downgrade a server from 3.4 to 3.2, then you would have to [manually convert all the `Decimal128`s in the database to another type first](https://docs.mongodb.com/manual/release-notes/3.4-downgrade/#convert-data-of-type-decimal). These would most likely be `String` in the case of the application having `BigDecimal` defined as the type in Mongoid. Then without changing any config Mongoid would read the `String`s in and convert them to `BigDecimal`. No error would be thrown in this case.

The only time an error would get raised would be making the round-trip back, which IMO makes perfect sense. The user has told Mongoid to convert `BigDecimal` to `Decimal128` but the server cannot handle `Decimal128` as it's been rolled back. It simpy cannot persist it.

johnnyshields:
@durran you are not understanding correctly.
1) Upgrade DB to 3.4
2) Upgrade server to use latest Mongoid, but leave `map_big_decimal_to_decimal128 = false` (which is the default setting)
3) In server only, set `map_big_decimal_to_decimal128 = true`. Let it run for awhile, DB now contains some D128 values.
4) Rollback from 3 -> 2
5) ...kaboom!

johnnyshields:
Nobody plans to do a rollback :)

johnnyshields:
Please understand we're talking about cost/benefit here. The benefits of being able to rollback from state 3 to state 2 above FAR outweigh any concerns over trailing zeros, which again I will emphasize that any user who wants trailing zeros should NOT be using BigDecimal type in Ruby, full stop.

durran:
@johnnyshields Of course no one plans to do rollbacks :), and I thought by "server" you meant the MongoDB server not your application.

The scenario you are still proposing is not the norm in my opinion. A configuration setting on a live server that changed the way the data was persisted and then was reverted on the live server is not a safe practive and to be allowed without consequence is also dangerous. Data has been persisted to the database in this window that now does not align with the application's settings.

I agree that no one should be using `BigDecimal` in Ruby if they want to retain precision. However with us trying to be consistent across languages we have to account for some situations, for example a Java developer coming over to Ruby and assuming `BigDecimal` in Ruby retains precision like `BigDecimal` in Java.

The configuration setting is there to help users with a gradual migration over time, but the expectation for most would be for users to never change it and definitely not changing it live. Those who care about percision would change the type in Mongoid to `BSON::Decimal128` and do a manual migration of all the types all at once. Those who don't care about precision would leave the type as `BigDecimal`. This requires zero configuration changes.

estolfo:
@johnnyshields How about we detect if there are trailing zeroes that would be lost when demongoizing from BSON::Decimal128 to BigDecimal and only raise an error if precision would be lost?

johnnyshields:
@durran:

> The scenario you are still proposing is not the norm in my opinion.

It can occur as I have outlined, and when it occurs it will break apps painfully.

> Java developer... assuming BigDecimal in Ruby retains precision like BigDecimal in Java

It is not Mongoid's job to add errors which cover the differences between Ruby in Java or other languages. This is what documentation is for.

@estolfo 

> How about we... only raise an error if precision would be lost?

Let's keep the code simple. If a user declares `type: BigDecimal` in *any Mongoid/MongoDB version, past, present or future* they are saying "I do not care about trailing zeros."

The current implementation introduces a new brittle error path and sacrifices usability and flexibily to cover a single esoteric case which no real-world user will care about.

estolfo:
Hello everyone following this issue,

I've done some tests to investigate all the cases in which a BSON::Decimal128 object won't map correctly to a Ruby BigDecimal. The test failures can be found in [this gist](https://gist.github.com/estolfo/0cd44e71745082ba0d88f7c4bf096215).

The test was: a BSON::Decimal128 (A) is converted to a Ruby BigDecimal (B), then converted back to a BSON::Decimal128 (A')
The specs tested if A == A'. The failures mostly concern either a loss of trailing 0s or special objects such as negative NaN, NaN with a payload.

In looking at the list of failures, they are generally edge cases that wouldn't be frequently used in real-world applications. That said, I'm not going to merge the changes in this pull request because of the confusion it has caused. One concern is how queries would be affected if there are a mix of String and BSON::Decimal128 representations of Ruby BigDecimals in the database.

If someone would like a direct mapping between Ruby BigDecimals and Decimal128s, I would encourage that user to define his/her own custom Model Type, with demongize and mongoize methods. Please see the documentation [here](https://docs.mongodb.com/ruby-driver/master/tutorials/6.0.0/mongoid-documents/) under "Custom Fields" for details on how to do this.

johnnyshields:
This is disappointing. The philosophy of the Mongoid team here--focusing on edge cases (trailing zeros, NAN vs negative NAN, numbers larger than the amount of atoms in the universe)--is really missing the forest for the trees.

lib/mongoid/criteria/queryable/extensions/big_decimal.rb
```diff
@@ -24,7 +24,15 @@ module ClassMethods
             # @since 1.0.0
             def evolve(object)
               __evolve__(object) do |obj|
-                obj ? obj.to_s : obj
+                if obj
+                  if obj.is_a?(::BigDecimal) && Mongoid.map_big_decimal_to_decimal128
+                    BSON::Decimal128.new(obj)
+                  elsif obj.is_a?(BSON::Decimal128) && Mongoid.map_big_decimal_to_decimal128
+                    obj
```
p-mongo, line 31: I think BSON::Decimal128 should always be serialized as Decimal128 regardless of config option.

johnnyshields: I agree

PR #4365, opened 2016-12-19
Don't use unless with else

ivanovaleksey:


lib/mongoid/relations/metadata.rb
```diff
@@ -778,10 +778,10 @@ def store_as
       #
       # @since 2.0.0.rc.1
       def validate?
-        unless self[:validate].nil?
-          self[:validate]
-        else
+        if self[:validate].nil?
```
estolfo: Hi @ivanovaleksey 

```ruby
def validate?
   self[:validate] ||= relation.validation_default
end
```

doesn't handle the case when ``self[:validate]`` is false properly so let's stick with your original change.

ivanovaleksey: Yes, you are right. I thought about it today. Thank you.

PR #4368, opened 2016-12-22
Add filename lineno to class_eval and module_eval calls

ivanovaleksey:


lib/mongoid/criteria/queryable/forwardable.rb
```diff
@@ -46,13 +46,13 @@ def select_with(receiver)
...
         def __forward__(name, receiver)
           if self.class == Module
-            module_eval <<-SEL
+            module_eval <<-SEL, __FILE__, __LINE__ + 1
               def #{name}(*args, &block)
                 #{receiver}.__send__(:#{name}, *args, &block)
               end
             SEL
           else
-            (class << self; self; end).class_eval <<-SEL
```
ivanovaleksey: `(class << self; self; end)` does return singleton class of `self`, doesn't it?

estolfo:
Hi @ivanovaleksey 

Can you tell me what problem you are solving with this PR? Thanks

ivanovaleksey:
Hi @estolfo 

Initially I was looking for where is `or` method defined.
So I tried
```
or_method = User.method(:or)
or_method.source_location
# => ["(eval)", 1]
```

With this PR it would return
```
or_method.source_location
# ["/Users/aleksey/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/origin-2.3.0/lib/origin/forwardable.rb", 48]
```

Of course now I know where all these methods are defined but I thought it could be useful for others.

As you know in case of an error `__FILE__` and `__LINE__` in `module_eval` also show where really an error happened.  Am not sure there could be much errors since these methods are not complicated but I thought it would not be superfluous.

ivanovaleksey:
By the way there is [issue](https://github.com/mongoid/origin/issues/105) for `origin` gem with the same suggestion.

lib/mongoid/criteria/queryable/forwardable.rb
estolfo: yes

estolfo:
Thanks!

PR #4377, opened 2017-01-17
Update the template and readme: local_threshold's default from 5 to 0.015

rakusai:
The default value of local_threshold configuration has been changed from 5 to 0.015 since [mongo-ruby-driver](https://github.com/mongodb/mongo-ruby-driver) ver 2.0.

[See source code and comments here](https://github.com/mongodb/mongo-ruby-driver/blob/8b442d8ff30083ef8dca3bc8dac78d4b924d9ff8/lib/mongo/server_selector.rb#L34)

So we'd better to update the document in the code.
There are no changes to the implementation.

### DIFF
--- a/docs/tutorials/mongoid-installation.txt
+++ b/docs/tutorials/mongoid-installation.txt
@@ -9,7 +9,7 @@ Installation
    :backlinks: none
    :depth: 1
    :class: singlecol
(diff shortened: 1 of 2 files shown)


estolfo:
thanks!

PR #4379, opened 2017-01-30
Update config template

ivanovaleksey:
Hi.

I have noticed that config [template](https://github.com/mongodb/mongoid/blob/master/lib/rails/generators/mongoid/config/templates/mongoid.yml) doesn't contain information about some options such as `duplicate_fields_exception` and `belongs_to_required_by_default`.

I added these options but I am not sure about descriptions.
I took description for `duplicate_fields_exception` option from CHANGELOG:
> #2938 A configuration option duplicate_fields_exception has been added that when set to true will raise an exception when defining a field that will override an existing method. (Arthur Neves)

And I wrote description for `belongs_to_required_by_default` by myself.

So maybe I should reword it somehow?

Thanks.

### DIFF
--- a/lib/rails/generators/mongoid/config/templates/mongoid.yml
+++ b/lib/rails/generators/mongoid/config/templates/mongoid.yml
@@ -121,6 +121,10 @@ development:
     # existing method. (default: false)
     # scope_overwrite_exception: false
 
+    # Raise an error when defining a field with the same name as an
+    # existing method. (default: false)
+    # duplicate_fields_exception: false
+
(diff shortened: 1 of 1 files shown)


estolfo:
Hi @ivanovaleksey

Thanks for this PR. You can take some of the text from here https://github.com/rails/rails/pull/18937/files

and add "trigger a validation error by default if the association is not present"

ivanovaleksey:
Hi @estolfo,
I have updated the description.
Hope it is better now.

estolfo:
thanks!

PR #4383, opened 2017-02-08
Refactor eager preload to avoid stack overflow

Val:
In some case (more than one circular associations included) It is possible to get a `SystemStackError: stack level too deep`.

After some digging I found it localized in `Eager#preload` code. This patch fix it (ec2a1c6), remove a double iteration (3fc2e2b) and make the method body more readable (413613e, 81901fb and 6cefdb5).

### DIFF
--- a/lib/mongoid/relations/eager.rb
+++ b/lib/mongoid/relations/eager.rb
@@ -22,13 +22,9 @@ module Mongoid
       end
 
       def preload(relations, docs)
-        grouped_relations = relations.group_by do |metadata|
-          metadata.inverse_class_name
-        end
+        grouped_relations = relations.group_by(&:inverse_class_name)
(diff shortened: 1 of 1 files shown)


Val:
build job https://travis-ci.org/mongodb/mongoid/jobs/199788604 has failed but because MongoDB was not available during the test :angry:

estolfo:
Hi @Val 

Thanks very much for your pull request. Can you provide a test for it as well?
Thanks!
Emily

Val:
Hi @estolfo 

I work on a project with many circular associations in which the stack overflow has appeared.
I already tried to extract a simple test to reveal the bug without success.

I hope I'll get the time to retry shortly.

johnnyshields:
IMHO this looks good to merge even without a test case reproducing the infinite loop.

PR #4384, opened 2017-02-13
Small improvement to spec for Touchable

johnnyshields:
In specs, when touching a model, assert that changes are not rolled back authomatically.


johnnyshields:
@estolfo failures are not related to this spec, please merge

PR #4388, opened 2017-02-21
MONGOID-4416 Don't run touch callbacks when #touch is called on an object from a relation's callback

estolfo:


lib/mongoid/relations/touchable.rb
```diff
@@ -85,7 +85,7 @@ def define_relation_touch_method(name, extra_field = nil)
             def #{method_name}
               without_autobuild do
                 relation = __send__(:#{name})
-                relation.touch #{":#{extra_field}" if extra_field} if relation
+                relation.touch(#{extra_field ? ":#{extra_field}" : 'nil' }, callbacks: [self]) if relation
```
johnnyshields: do you mean to add the string literal `'nil'` here?

lib/mongoid/interceptable.rb
```diff
@@ -124,15 +124,18 @@ def run_before_callbacks(*kinds)
...
       cascadable_children(kind).each do |child|
-        # This is returning false for some destroy tests on 4.1.0.beta1,
-        # causing them to fail since 4.1.0 expects a block to be passed if the
-        # callbacks for the type are empty. If no block is passed then the nil
-        # return value gets interpreted as false and halts the chain.
-        #
-        # @see https://github.com/rails/rails/blob/master/activesupport/lib/active_support/callbacks.rb#L79
-        if child.run_callbacks(child_callback_type(kind, child), *args) == false
-          return false
+        unless opts && opts[:callbacks].include?(child)
```
johnnyshields: I think `:callers` is a better name than `:callbacks`

lib/mongoid/interceptable.rb
```diff
@@ -124,15 +124,18 @@ def run_before_callbacks(*kinds)
     #
     # @since 2.3.0
     def run_callbacks(kind, *args, &block)
+      opts = args.pop
```
johnnyshields: Prefer to use `opts = args.extract_options!` here

lib/mongoid/interceptable.rb
```diff
@@ -124,15 +124,18 @@ def run_before_callbacks(*kinds)
...
-        # @see https://github.com/rails/rails/blob/master/activesupport/lib/active_support/callbacks.rb#L79
-        if child.run_callbacks(child_callback_type(kind, child), *args) == false
-          return false
+        unless opts && opts[:callbacks].include?(child)
+          # This is returning false for some destroy tests on 4.1.0.beta1,
+          # causing them to fail since 4.1.0 expects a block to be passed if the
+          # callbacks for the type are empty. If no block is passed then the nil
+          # return value gets interpreted as false and halts the chain.
+          #
+          # @see https://github.com/rails/rails/blob/master/activesupport/lib/active_support/callbacks.rb#L79
```
johnnyshields: Unrelated to this PR but I'm pretty sure this behavior has now changed in Rails.

johnnyshields:
@estolfo thanks for this PR. Some comments above, I think this is the correct approach and a very elegant one. Nice work.

lib/mongoid/relations/touchable.rb
estolfo: It's not functionally the string literal 'nil', as the method is being defined dynamically. The method will pass nil to the touch method if extra_field is not defined.

lib/mongoid/interceptable.rb
estolfo: updated

lib/mongoid/interceptable.rb
estolfo: updated

lib/mongoid/interceptable.rb
estolfo: removed comment

johnnyshields:
Looks good to me, nice work!

PR #4390, opened 2017-02-23
Fix sort to id_sort option typos

hartator:
`Mongoid::Criteria#first({sort:none})` doesn't actually work. 

But this works: `Mongoid::Criteria#first({id_sort:none}`.

I've just fixed the typo in the documentation.


estolfo:
thanks @hartator

PR #4391, opened 2017-02-25
Another case where infinite loop can occur with touch callback

johnnyshields:
While testing the fix in #4388 I've determined that a stack overflow can still can occur when there are multiple embedded children. If both children are in a state where they can get the touch callback (e.g. `changed?` is `true`), then they will cause an infinite loop like:

touch Child A --> calls touch on Parent --> calls touch on Child B --> calls touch on Parent --> calls touch on Child A --> ...

The only way I can think to solve this is to track an array of already called children on each object and not call the same child twice. I've made a proof of concept implementation of this. If we go with this approach  we should also rollback the changes from #4388

### DIFF
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -87,6 +87,7 @@ module Mongoid
       kinds.each do |kind|
         run_targeted_callbacks(:after, kind)
       end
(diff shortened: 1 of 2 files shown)


johnnyshields:
Replaced by #4392

PR #4392, opened 2017-02-25
Do not cascade :touch callback from parent to child

johnnyshields:
While testing #4391 I discovered another infinite loop case is still possible with two embedded child docs:

touch Child A --> calls touch on Parent --> calls touch on Child B --> calls touch on Parent --> calls touch on Child A --> ...

I also discovered that **in most cases** the `touch` callback **does not actually cascade from parent to child docs**, it only cascades when the child has been modified.

```ruby
    def in_callback_state?(kind)   # kind == :touch
      [ :create, :destroy ].include?(kind) || new_record? || flagged_for_destroy? || changed?
    end
```

Therefore, I believe the best solution is to never cascade the touch callback from parent -> child, and instead only do child -> parent, which avoids any possibility of loops.

This PR reverts the changes in #4391

### DIFF
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -190,7 +190,7 @@ module Mongoid
     #
     # @since 2.3.0
     def cascadable_child?(kind, child, metadata)
-      return false if kind == :initialize || kind == :find
+      return false if kind == :initialize || kind == :find || kind == :touch
       return false if kind == :validate && metadata.validate?
       child.callback_executable?(kind) ? child.in_callback_state?(kind) : false
     end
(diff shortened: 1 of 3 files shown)


johnnyshields:
Spec failures are not related to this PR; refer to Travis results.

johnnyshields:
@estolfo please review and merge

estolfo:
ok, looks good to me, thanks!

estolfo:
ActiveRecord actually only allows touch to be applied from child to parent. It's only an option on a belongs_to relation, so this pull request makes Mongoid consistent with AR. See the documentation here: http://guides.rubyonrails.org/association_basics.html

johnnyshields:
@estolfo thank you!

PR #4393, opened 2017-03-06
Fix habtm self reference

johnnyshields:
Attempted fix for #4286

The issue is that:
- `#substitute` method of the many-to-many relation is calling `purge` which removes the inverse relationship value
- `#substitute` method then calls `push` to add value to the current doc relation WITHOUT updating the inverse.
- Normally this is not a problem because the `Synchornization` callbacks will detect that the value has changed and update the inverse, EXCEPT if the value is the same (i.e. NOT changed)

Solution is to skip `substitute` logic if no change.

### DIFF
--- a/lib/mongoid/relations/accessors.rb
+++ b/lib/mongoid/relations/accessors.rb
@@ -229,7 +229,7 @@ module Mongoid
             without_autobuild do
               if value = get_relation(name, metadata, object)
                 value = __build__(name, value, metadata) unless value.respond_to?(:substitute)
-                set_relation(name, value.substitute(object.substitutable)) unless value == object.substitutable
+                set_relation(name, value.substitute(object.substitutable))
               else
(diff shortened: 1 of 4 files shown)


johnnyshields:
hmmm... this is not working when i tried a few additional tests. think solution here is NOT to call purge on the inverse relations. needs more work, DO NOT MERGE

johnnyshields:
also, I do not think this issue is specific to self-referencing, think it actually affects all many-to-many relationships. I believe I've seen similar behavior in my own app and have added hacks as a result...

estolfo:
Fixed in this commit: https://github.com/mongodb/mongoid/commit/8ccd33185158b53a79319a8d670907a744d7ecc9

johnnyshields:
@estolfo wonderful, thanks!

PR #4394, opened 2017-03-07
Fix compatibility of Mongoid::Contextual::None with other contexts by adding distinct

ElMassimo:
Currently `Mongoid::Contextual::None` does not implement the `distinct` method.

Since `Mongoid::Contextual::None` is intended to provide a compatible API with the other contexts (to be used when no documents should be returned), it should implement `distinct` as well.

Added a test case that fails before the change, and passes after implementing the method.

#### Example:
```ruby
users = if admin?
  User.where(email: /@mongodb.com/)
else
  User.none
end

ids = users.distinct(:id)
```

### DIFF
--- a/lib/mongoid/contextual/none.rb
+++ b/lib/mongoid/contextual/none.rb
@@ -21,6 +21,18 @@ module Mongoid
         other.is_a?(None)
       end
 
+      # Allow distinct for null context.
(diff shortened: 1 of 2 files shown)


estolfo:
thanks @ElMassimo

PR #4397, opened 2017-03-15
Update mongoid-queries.txt with valid ruby

arbourd:
`|scope|` is on the wrong side of the `do` keyword


estolfo:
thanks @arbourd !

PR #4399, opened 2017-03-24
Failing spec. Setter on hash field failes if the original value is nil.

flanker:
In MONGOID-3407 (commit @bf84543a57) There is a change on hash field
setter, but it fails if the original value is nil.

The spec changed in this commit are failing due to that.

### DIFF
--- a/spec/app/models/plane.rb
+++ b/spec/app/models/plane.rb
@@ -1,0 +1,5 @@
+class Plane
+  include Mongoid::Document
+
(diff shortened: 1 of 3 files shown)


estolfo:
See https://jira.mongodb.org/browse/MONGOID-4426
Fixed by commit: https://github.com/mongodb/mongoid/commit/2e4c640ffb230676a8ee528148726871c30d2ccd

PR #4402, opened 2017-04-04
MONGOID-4276 First step: Refactor Relations::Metadata Hash into Associations objects

estolfo:
- Replaced relations metadata Hash with association objects of different types. Model.relations values will now be Objects of type Mongoid::Associations. Each Mongoid::Associations object has particular behavior, appropriate for its type. Previously, each one of these relations objects was a Hash making its instance methods have messy, confusing, and sometimes incorrect logic. The metadata hash would also respond to methods entirely irrelevant for the type of relationship it was representing.

- Each association has a reference to the Owner class, so that this class doesn't havent to be guessed later on. This will enable future work for relating objects to their parents when, for example, iterating over objects in a has_many relation.

- Defined modules (for example, Accessors, Autosave, CounterCache) that define methods on given objects if their relations have particular options. This replaces definitions of such methods for *all* objects. For example, the #touchable method doesn't need to be defined for all objects. We can instead called a method Touchable#define_relation_touch_method that will set up the touch callbacks for certain objects. This avoids pollution of the method space for models.

- Defined modules Constrainable, Conversions that are included in Association classes so they don't have to be referenced explicitly.
 
- Updated the Dependency options for associations to be consistent with Rails. Preivously, the Dependency types were :delete, :destroy, :nullify, :restrict. Now the types are the same as Rails: :delete_all, :destroy, :nullify, :restrict_with_error, :restrict_with_exception


johnnyshields:
@estolfo what is the reason for changing the name "Relation" to "Association"?

estolfo:
Hi @johnnyshields 
Before these changes, the Relations::Metadata object had the same role and behavior as an ActiveRecord::Association in Rails. In refactoring the Metadata object (which was essentially a Hash) into separate objects with particular behavior, I wanted the design to more closely mimic ActiveRecord's. I also wanted there to be a clear departure from the older design.
I therefore decided to put the objects saving the metadata on associations between Models, under the namespace "Mongoid::Association". For someone coming from Rails, the architecture in Mongoid now will look more familiar.

The documentation for Rails here http://guides.rubyonrails.org/association_basics.html#options-for-belongs-to can pretty much apply to Mongoid now.

estolfo:
Thanks for your support @Fudoshiki = )
Please let me know if you have feedback! This pull requests has an enormous number of changes and the more people who look at it, the better.

durran:
What an epic pull request and a long time coming. Great work here @estolfo 

LGTM. :)

johnnyshields:
@estolfo please check this GIST and give comments. https://gist.github.com/johnnyshields/dd06c089d887acf984f0cfcb67d0b8fd

I think we should revisit how relations work in general.

estolfo:
Hi @johnnyshields 
Thank you for the ideas outlined in the gist. Durran and I were actually talking last week about the has_and_belongs_to_many with inverse_of: nil and what to do about it.

I do think your ideas in the gist are good but I'm afraid completely changing the relation terminology will prevent the majority of users from upgrading to the next version of Mongoid. I think we need to make some significant changes, but not ones that will require everyone rewriting their models.

That said, Durran and I were thinking of replacing has_and_belongs_to_many with inverse_of: nil with the relation references_many or owns_many. has_and_belongs_to_many will always track the list of ids on each side, while the new relation will only track the ids on the *_many side.

johnnyshields:
@estolfo I fell `references_many` and `owns_many` are probably more confusing for users to upgrade... I will put together a matrix later which demonstrates that my syntax proposed is not difficult to upgrade to.

PR #4404, opened 2017-04-07
MONGOID-4425 #as_document continues to return BSON::Document, but use internal api method #as_attributes otherwise

estolfo:
Mongoid's callbacks rely on being able to mutate the ``attributes`` hash of a model object directly, so we replace the use of ``#as_document`` with ``#as_attributes``, and allow the public API method ``#as_document`` to continue returning a ``BSON::Document``.

### DIFF
--- a/lib/mongoid/contextual/memory.rb
+++ b/lib/mongoid/contextual/memory.rb
@@ -44,7 +44,7 @@ module Mongoid
         deleted = count
         removed = map do |doc|
           prepare_remove(doc)
-          doc.as_document
+          doc.send(:as_attributes)
         end
         unless removed.empty?
           collection.find(selector).update_one(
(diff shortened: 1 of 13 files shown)


lib/mongoid/contextual/memory.rb
```diff
@@ -44,7 +44,7 @@ def delete
         deleted = count
         removed = map do |doc|
           prepare_remove(doc)
-          doc.as_document
+          doc.send(:as_attributes)
```
johnnyshields: why not make the method public rather than using send?

estolfo: The fewer methods exposed publicly on a document/model, the better. as_attributes is only ever used internally right now, so I don't want to get stuck not being allowed to change or get rid of it if I need to as part of a bug fix or refactor later on.

PR #4405, opened 2017-04-10
MONGOID-4431 Avoid use of certain names in Synchronization module so not to interfer with attribute names

estolfo:


estolfo:
https://evergreen.mongodb.com/version/58eb664e2fbabe1a3f002b8d

johnnyshields:
LGTM

PR #4407, opened 2017-04-20
MONGOID-2882 Set base object on documents being iterated from criteria and from Association Enumerable

estolfo:


johnnyshields:
👍

jarthod:
🎉

PR #4419, opened 2017-05-03
MONGOID-3982 An orphaned embedded document should be considered a new record

estolfo:


johnnyshields:
👍 LGTM. Conflicts need resolution

johnnyshields:
Guess it should also be raised against master?

PR #4428, opened 2017-06-07
MONGOID-4448 Use "background_indexing" option by default to create new indexes

exAspArk:
Hey,

I wonder if we can add an additional option to create DB indexes in the background by default. 

I think it might be useful in case if somebody added an index on a huge collection and forgot to specify `background: true`. By setting something like:

```yaml
# config/mongoid.yml
production:
  options:
    background_indexing: true
```

And after adding a new index:

```ruby
# app/models/person.rb
class Person
  include Mongoid::Document
  field :ssn
  index ssn: 1 # <=== NEW INDEX
end
```

And running:

```sh
bundle exec rails runner "Person.create_indexes"
# or
bundle exec rake db:mongoid:create_indexes
```

We can avoid human mistakes, which may lead to catastrophic DB blocking:
> By default, creating an index blocks all other operations on a database.
> https://docs.mongodb.com/manual/core/index-creation/

These changes don't break backward compatibility 😉

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -27,6 +27,7 @@ module Mongoid
     option :log_level, default: :info
     option :belongs_to_required_by_default, default: true
     option :app_name, default: nil
+    option :background_indexing, default: false
 
     # Has Mongoid been configured? This is checking that at least a valid
     # client config exists.
--- a/lib/mongoid/indexable.rb
+++ b/lib/mongoid/indexable.rb
@@ -28,8 +28,11 @@ module Mongoid
       # @since 1.0.0
       def create_indexes
         return unless index_specifications
+
+        default_options = {background: Config.background_indexing}
+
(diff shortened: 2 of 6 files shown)


exAspArk:
@estolfo please let me know what you think about it :)

estolfo:
Hi @exAspArk 
This doesn't seem like a bad idea, I'll take a look. Thank you!

estolfo:
Hi @exAspArk 

Would you mind opening this pull request against the [7.0-dev branch](https://github.com/mongodb/mongoid/tree/7.0-dev)? Master is 6.2.0 and I don't plan on releasing a major version of the 6.x series soon. Because this is a new feature, it would have to go into 7.0. Thanks!
Emily

exAspArk:
Hey @estolfo,

I rebased and changed the base branch to `7.0-dev` in this PR 🎉 

Thanks!

PR #4429, opened 2017-06-09
Add pass document parameter to unwind method support since new specif…

chenzww:
…ication in MongoDB version 3.2.

### DIFF
--- a/lib/mongoid/criteria/queryable/pipeline.rb
+++ b/lib/mongoid/criteria/queryable/pipeline.rb
@@ -72,16 +72,26 @@ module Mongoid
         #
         # @example Add the unwind.
         #   pipeline.unwind(:field)
+        #   pipeline.unwind(document)
+        #
+        # @see https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/
         #
         # @param [ String, Symbol ] field The name of the field.
+        # @param [ Hash ]           document.
         #
         # @return [ Pipeline ] The pipeline.
         #
         # @since 2.0.0
         def unwind(field)
-          normalized = field.to_s
-          name = aliases[normalized] || normalized
-          push("$unwind" => name.__mongo_expression__)
+          unless field.respond_to? :keys
+            normalized = field.to_s
+            name = aliases[normalized] || normalized
+            push("$unwind" => name.__mongo_expression__)
+          else
+            permit_keys = ['path', 'includeArrayIndex', 'preserveNullAndEmptyArrays'].freeze
+            normalized = BSON::Document.new(field.select { |k| permit_keys.include?(k.to_s) })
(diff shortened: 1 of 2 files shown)


lib/mongoid/criteria/queryable/pipeline.rb
```diff
@@ -72,16 +72,26 @@ def project(entry)
...
         # @since 2.0.0
         def unwind(field)
-          normalized = field.to_s
-          name = aliases[normalized] || normalized
-          push("$unwind" => name.__mongo_expression__)
+          unless field.respond_to? :keys
+            normalized = field.to_s
+            name = aliases[normalized] || normalized
+            push("$unwind" => name.__mongo_expression__)
+          else
```
estolfo: There's no need to check the keys of the document argument. The MongoDB server may change the names of the keys for $unwind in a future version or add other keys. We can just expect the user will provide a valid document.

In other words, this can just be ``push("$unwind" => field)``

lib/mongoid/criteria/queryable/pipeline.rb
```diff
@@ -72,16 +72,26 @@ def project(entry)
...
+        #
+        # @see https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/
         #
         # @param [ String, Symbol ] field The name of the field.
+        # @param [ Hash ]           document.
         #
         # @return [ Pipeline ] The pipeline.
         #
         # @since 2.0.0
         def unwind(field)
```
estolfo: let's change the name of this argument to ``field_or_doc``, just to be clearer.

lib/mongoid/criteria/queryable/pipeline.rb
chenzww: It's a good idea.

lib/mongoid/criteria/queryable/pipeline.rb
```diff
@@ -72,16 +72,24 @@ def project(entry)
         #
         # @example Add the unwind.
         #   pipeline.unwind(:field)
+        #   pipeline.unwind(document)
+        #
+        # @see https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/
         #
         # @param [ String, Symbol ] field The name of the field.
```
estolfo: This should be updated to be 

```ruby
# @param [ String, Symbol, Hash ] field_or_doc The name of the field or a document.
```

and the line below can be removed.

estolfo: Hi @chenzww 
Would you mind updating the documentation too so we can merge this? Thanks!

chenzww: What is mean by "the line below"?

PR #4436, opened 2017-06-29
Make _destroy alias flagged_for_destroy?

danielfarrell:
We are using cocoon to handle embedded records and have run into issues with the `_destroy` value not persisting through validation cycles.  There is a PR at nathanvda/cocoon#434 that shows a fix on the cocoon side, but it's really working around Mongoid not returning the correct value here.

Thoughts on this?


estolfo:
Hi @danielfarrell 
Would you mind showing me an example using Mongoid models of how the _destroy value is not persisting through validation cycles? I'd like to get better context before merging this. Thank you!

Emily

danielfarrell:
Hey Emily,

Sure.  As I'm sure you know, when you have a nested model with a `accepts_nested_attributes_for` that has an `allows_destroy: true` on it, you can pass in a `_destroy` value that will make the model flagged for destroy.  The issue is that when there is another validation issue in that request(another nested field was missing a required field, for example), then it doesn't actually delete the nested model, and the re-renders the form and tries to use the `_destroy` attribute to set the field.  Because Mongoid always returns false for `_destroy`, that value isn't set properly.

What this has made happen here is best described in our issue title of "After two rounds of validation, things get weird", ha.  Things you said to delete come back as not deleted, causing some confusion.

Because the `_destroy` value sets `@flagged_for_destroy`, it made sense to have this be an alias of that.  This is essentially what Active Record does as well, as seen here: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/nested_attributes.rb#L371-L373

Let me know if you have any more questions.

estolfo:
Hi @danielfarrell 

Thank you very much for the explanation and for the reference to ActiveRecord's behavior. I'll merge this PR and I'll apply it to the 6.1-stable and 6.2-stable branches as well for Rails 5.0 and Rails 5.1 users. Thanks!
Emily

PR #4441, opened 2017-07-14
Update mongoid-relations.txt

Jarvizx:
For this example of storage in has_many, could have square brackets?


estolfo:
Hi @Jarvizx 

No, the member object would have one band_id because it's a "belongs_to" relationship. You would expect an array of ids if the relationship were ``has_and_belongs_to_many``.
Let me know if that makes sense and if you have any other questions!
Emily

PR #4442, opened 2017-07-19
Use valid_options method instead of constant

ebeigarts:
I think this was changed by mistake in https://github.com/mongodb/mongoid/pull/4269/files#diff-8defad7996cfab9e288a4049e3b5c0c6L35

### DIFF
--- a/lib/mongoid/relations/options.rb
+++ b/lib/mongoid/relations/options.rb
@@ -32,7 +32,7 @@ module Mongoid
       #
       # @since 2.1.0
       def validate!(options)
-        valid_options = options[:relation]::VALID_OPTIONS + COMMON
+        valid_options = options[:relation].valid_options + COMMON
         options.keys.each do |key|
           if !valid_options.include?(key)
             raise Errors::InvalidOptions.new(


estolfo:
Hi @ebeigarts 

Can you please tell me how the issue is manifesting itself ? Would you mind providing some sample code that shows a problem you are running into that leads you to think that this was a mistake ?

Thank you
Emily

ebeigarts:
If we leave it to use constant, then `valid_options` methods

```
          def valid_options
            VALID_OPTIONS
          end
```

are useless, because `Options.validate!` was the only place where this method was used.
Also making it use method makes it easier for `mongoid-versioning` and other gems to extend valid options, without redefining constants.

estolfo:
I understand that the method was useless but I left it in because I cannot remove constants unless I deprecate them and then release a major version of Mongoid.

In any case, I'd prefer to make this change for 7.0, as I did a refactor of relations. This is how the constants are used (in the 7.0-dev branch):

https://github.com/mongodb/mongoid/blob/7.0-dev/lib/mongoid/association/referenced/has_many.rb#L24-L43

For mongoid-versioning, would it be easier for there to be a ``valid_options`` method that returns ``(ASSOCIATION_OPTIONS + SHARED_OPTIONS)`` ?

ebeigarts:
It would cool if there could be some public API for gems like `mongoid-versioned` to extend options for associations and fields. `Mongoid.config.custom_options << :custom_option` + `field :x, custom_option: true`

PR #4443, opened 2017-08-16
Fixup the falsey logic of `__nested__`

nolaneo:
Right now, when you have a document with a key that is `false`, the `__nested__` method on hashes will incorrectly return `nil`

For example:

```ruby
# Imagine we have some nested data on our document
my_document["nested_data"]
# => { some_boolean_value: false }

# Looking up this nested value without dot notation returns the correct value
my_document.nested_data["some_boolean_value"]
# => false

#Doing the same thing with `read_attribute` causes the value to be `nil`
my_document.read_attribute("nested_data.some_boolean_value")
# => nil
```

Internally within the nested function we end up doing `false || nil` which ruby evaluates to `nil`.
`false` is a completely valid value in this case and we should return that.

More worryingly is if you have an integer keyed hash:

```ruby
my_document["nested_data"]
# => { 0: "Some value for zero", 1: "Some value for one" }

# attempt to read a string key that doesn't exist
my_document.read_attribute("nested_data.this_doesnt_exist")
# => "Some value for zero"
```

😱 

Internally, there is a `to_i` call on the string key, ruby's madness converts this to `0` which in this case ends up with us returning the incorrect data.

Changes in this PR:
* Only check for integer version of the key if the value is `nil` and the key is definitely just an integer.

### DIFF
--- a/lib/mongoid/extensions/hash.rb
+++ b/lib/mongoid/extensions/hash.rb
@@ -106,8 +106,11 @@ module Mongoid
         value = self
         keys.each do |key|
           return nil if value.nil?
-          nested = value[key] || value[key.to_i]
-          value = nested
+          value_for_key = value[key]
+          if value_for_key.nil? && key.to_i.to_s == key
+            value_for_key = value[key.to_i]
+          end
+          value = value_for_key
         end
         value
       end
--- a/spec/mongoid/extensions/hash_spec.rb
+++ b/spec/mongoid/extensions/hash_spec.rb
@@ -230,10 +230,27 @@ describe Mongoid::Extensions::Hash do
     it "should retrieve a nested value under the provided key" do
       expect(nested).to eq "hundred"
     end
+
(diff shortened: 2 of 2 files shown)


estolfo:
Hi @nolaneo 
Thanks for this pull request. Do you mind taking a look at the tests on travis? https://travis-ci.org/mongodb/mongoid/jobs/265163513
They don't seem to be passing with this change.
Thanks!
Emily

nolaneo:
Hey @estolfo pushed up a tweak to fix that spec 🙂. There's a bunch of `Mongo::Error::NoServerAvailable` failures which I assume are unrelated?

ocfnash:
@estolfo I think @nolaneo is correct: those test failures appear unrelated to the change proposed here.

This issue this PR fixes is pretty frightening; I'd love to see this merged.

estolfo:
Thanks @nolaneo  !

PR #3885, opened 2014-12-12
Mongoid/Moped controller runtime a la ActiveRecord

niedfelj:
Shows the amount of time spent in Mongoid/Moped instrumented calls in the ActionController summary after each request:

Completed 200 OK in 1487ms (Views: 504.2ms | ActiveRecord: 36.6ms | Mongoid: 252.6ms | Solr: 599.9ms)

PR #4446, opened 2017-09-08
Rails controller runtime logging

doooby:
Hi there, I'm bringing back an issue about logging mongoid runtime per rails controller action. 

there was https://github.com/mongodb/mongoid/pull/3885 that failed to get into mongoid, not sure why, but now that code is incompatible.

the log message `Completed 200 OK in 52ms (Views: 0.2ms | ActiveRecord: 0.4ms | MongoDB: 0.1ms)` is nice but marginal feature.

we needed to collect some metrics about our rails application; this allows us to use standard rails notification:

```ruby
ActiveSupport::Notifications.subscribe 'process_action.action_controller' do |*args|
  data = args.extract_options!
  data[:mongo_runtime] # accessible here
end
```

and please, would it be possible to put it in older versions too? (we're currently on mongoid 5.1)

### DIFF
--- a/lib/mongoid/railties/controller_runtime.rb
+++ b/lib/mongoid/railties/controller_runtime.rb
@@ -2,31 +2,45 @@ module Mongoid
   module Railties
     module ControllerRuntime
 
+      # This extension mimics the Rails' internal method to
+      # measure ActiveRecord runtime during request processing.
+      # It appends MongoDB runtime value (`mongoid_runtime`) into payload
+      # of instrumentation event `process_action.action_controller`.
       module ControllerExtension
         extend ActiveSupport::Concern
 
         protected
 
-        attr_internal :mongo_runtime
+        attr_internal :mongoid_runtime
 
+        # reset the runtime before each action
+        def process_action *_
+          Collector.reset_runtime
+          super
+        end
+
+        # override to collect the measurements
         def cleanup_view_runtime
-          mongo_rt_before_render = Instrument.reset_runtime
+          mongo_rt_before_render = Collector.reset_runtime
           runtime = super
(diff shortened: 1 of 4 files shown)


estolfo:
Hi @doooby 
Thank you very much from your pull request! Would you mind adding a few lines of documentation in the code and to [our tutorial](https://github.com/mongodb/mongoid/tree/master/docs/tutorials)?
And a few tests would be great as well.

Thanks!
Emily

doooby:
sure, when i have the time, i'll do.

lib/mongoid/railties/controller_runtime.rb
```diff
@@ -0,0 +1,67 @@
+module Mongoid
+  module Railties
+    module ControllerRuntime
+
+      module ControllerExtension
+        extend ActiveSupport::Concern
+
+        protected
+
+        attr_internal :mongo_runtime
```
johnnyshields: I think this should be either "mongoid_runtime" or "mongodb_runtime"

gnumarcelo:
Any chance this being merged soon?

estolfo:
I'd like some documentation and tests before this is merged. Maybe @doooby has some time?

doooby:
Sorry, I hadn't had the time for this. And for more, the reasons of the previous Travis fail goes over my understanding. But here, I've polished it a bit, added some test that made sense to me and added a mention to the doc.

also merged master from upstream.

p-mongo:
Tracking ticket: https://jira.mongodb.org/browse/MONGOID-4631

p-mongo:
The diff references  `::Mongoid::Railties::ControllerRuntime::Instrument` which does not appear to be defined anywhere, and the code that references doesn't seem to be covered by tests either.

PR #4447, opened 2017-09-12
fix cloning document with polymorphic embedded documents with multiple language field

tomasc:
Currently, localized fields are inferred from the class stored in relation's metadata.

However, in polymorphic embedded relations, each embedded document class can possibly define different localized fields. In the current implementation these are not taken into account and as a result localized fields on embedded polymorphic documents might not be correctly processed. 

I propose this patch, which first attempts to infer localized fields from the class defined in the `_type` attribute, falling back to the class defined on relation's metadata.

Questions:
* is there already a mechanism to convert `_type` attribute to a class, or is the `constantize` method sufficient? Would `safe_constantize` be preferred?
* I have added an extra class to the test suite (`AddressCustomized`), perhaps there is a cleaner way to test the behavior?

### DIFF
--- a/lib/mongoid/copyable.rb
+++ b/lib/mongoid/copyable.rb
@@ -74,8 +74,8 @@ module Mongoid
 
         if metadata.macro == :embeds_many
           attrs[metadata.key].each do |attr|
-            klass = attr.fetch('_type', metadata.class_name).constantize
-            process_localized_attributes(klass, attr)
+            embedded_klass = attr.fetch('_type', metadata.class_name).constantize
+            process_localized_attributes(embedded_klass, attr)
           end
         else
           process_localized_attributes(metadata.klass, attrs[metadata.key])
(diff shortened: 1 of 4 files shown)


tomasc:
PS Not sure how to fix the failing Travis CI test, they all pass locally just fine.

estolfo:
Hi @tomasc 
Thanks for this pull request!
Your code looks fine but the test failures are happening because the class, ``Address`` is assumed to not be subclassed in other parts of the test suite. I would suggest either subclassing a different class in your test or changing the tests that are failing to use another, non-subclassed model. It might be easier to just change your test = )
Thanks again!

simi:
Why not target master branch in this pull request?

johnnyshields:
@tomasc yes please resubmit the PR against master.

tomasc:
@estolfo @simi @johnnyshields updated to latest API and resubmitted against master as #4455

PR #4448, opened 2017-09-21
remove warning: instance variable @__parent not initialized

qbantek:
warning shows when running specs on my application


estolfo:
thanks @qbantek !

PR #4451, opened 2017-11-10
shard key selector issue

intale:
Issue description:
When updating field that is also shard key, a wrong value is generated by `Mongoid::Shardable#shard_key_selector`.
Example:
```
2.4.1 :004 > t=Test.create(name: 'asd', ip: '127.0.0.1')
D, [2017-11-10T20:35:54.749102 #26512] DEBUG -- : MONGODB | localhost:27017 | mongoid_test.insert | STARTED | {"insert"=>"mongoid_tests", "documents"=>[{"_id"=>BSON::ObjectId('5a05f18af06f086790a5abbf'), "name"=>"asd", "ip"=>"127.0.0.1"}], "ordered"=>true}
 => #<Test _id: 5a05f18af06f086790a5abbf, name: "asd", ip: "127.0.0.1"> 
2.4.1 :005 > t.update_attributes(ip: '127.0.0.2')
D, [2017-11-10T20:36:07.373379 #26512] DEBUG -- : MONGODB | localhost:27017 | mongoid_test.update | STARTED | {"update"=>"mongoid_tests", "updates"=>[{"q"=>{"_id"=>BSON::ObjectId('5a05f18af06f086790a5abbf'), "ip"=>"127.0.0.2"}, "u"=>{"$set"=>{"ip"=>"127.0.0.2"}}, "multi"=>false, "upsert"=>false}], "ordered"=>true}
 => true 
2.4.1 :006 > t.reload
D, [2017-11-10T20:36:54.297448 #26512] DEBUG -- : MONGODB | localhost:27017 | mongoid_test.find | STARTED | {"find"=>"mongoid_tests", "filter"=>{"_id"=>BSON::ObjectId('5a05f18af06f086790a5abbf')}}
 => #<Test _id: 5a05f18af06f086790a5abbf, name: "asd", ip: "127.0.0.1"> 
```
As you can see, when updating record, an unpersisted value of `:ip` is used in filter.
After changes in this PR:
```
2.4.1 :005 > t.update_attributes(ip: '127.0.0.2')
D, [2017-11-10T20:40:17.553021 #26611] DEBUG -- : MONGODB | localhost:27017 | mongoid_test.update | STARTED | {"update"=>"mongoid_tests", "updates"=>[{"q"=>{"_id"=>BSON::ObjectId('5a05f18af06f086790a5abbf'), "ip"=>"127.0.0.1"}, "u"=>{"$set"=>{"ip"=>"127.0.0.2"}}, "multi"=>false, "upsert"=>false}], "ordered"=>true}
 => true 
2.4.1 :006 > t.reload
D, [2017-11-10T20:42:05.510011 #26611] DEBUG -- : MONGODB | localhost:27017 | mongoid_test.find | STARTED | {"find"=>"mongoid_tests", "filter"=>{"_id"=>BSON::ObjectId('5a05f18af06f086790a5abbf')}}
 => #<Test _id: 5a05f18af06f086790a5abbf, name: "asd", ip: "127.0.0.2"> 
```

Gist with test class definition https://gist.github.com/intale/e1d0b0877c00268c12f020d9b3cefca4.

P.S. Branched from master and, as a result, inherited some failing tests from there.

### DIFF
--- a/lib/mongoid/shardable.rb
+++ b/lib/mongoid/shardable.rb
@@ -37,7 +37,7 @@ module Mongoid
     def shard_key_selector
       selector = {}
       shard_key_fields.each do |field|
-        selector[field.to_s] = attribute_was(field) || send(field)
+        selector[field.to_s] = new_record? ? send(field) : attribute_was(field)
       end
       selector
     end
--- a/lib/mongoid/shardable.rb
+++ b/lib/mongoid/shardable.rb
@@ -37,7 +37,7 @@ module Mongoid
     def shard_key_selector
       selector = {}
       shard_key_fields.each do |field|
-        selector[field.to_s] = send(field)
+        selector[field.to_s] = attribute_was(field) || send(field)
       end
       selector
     end
--- a/spec/mongoid/shardable_spec.rb
+++ b/spec/mongoid/shardable_spec.rb
@@ -49,23 +49,24 @@ describe Mongoid::Shardable do
     end
   end
 
-  describe "#shard_key_selector" do
+  describe '#shard_key_selector' do
+    subject { instance.shard_key_selector }
+    let(:klass) { Band }
+    let(:instance) { klass.new(name: name) }
+    let(:name) { 'Jo' }
 
-    let(:klass) do
-      Band
-    end
+    before { klass.shard_key(:name) }
 
-    let(:object) do
-      klass.new
-    end
+    it { is_expected.to eq({ 'name' => name }) }
 
-    before do
-      klass.shard_key(:name)
-      object.name = "Jo"
-    end
+    context 'changing shard_key value on persisted record' do
+      let(:instance) { klass.create(name: name) }
 
-    it "returns a hash of shard key names and values" do
-      expect(object.shard_key_selector).to eq({ "name" => "Jo" })
+      before do
+        instance.name = 'Izi4'
+      end
+
+      it { is_expected.to eq({ 'name' => name }) }
     end
   end
 end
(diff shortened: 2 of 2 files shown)


johnnyshields:
To summarize, when you are changing the value of a field which is used by the shard key, in order to find the **existing** document in the DB you need to reference the old value of that field.

There's a potential bug here--if the shard key value was `nil` or `false` and changed to non-nil then you will be referencing the new non-nil value for the lookup.

A better solution might be something like:

```
selector[field.to_s] = new_record? ? send(field) : attribute_was(field) 
```

estolfo:
Hi @intale 

Thanks for this pull request. I think more accurate logic would be the following:

```ruby
if attribute_changed?(field)
  selector[field.to_s] = attribute_was(field) unless attribute_was(field).nil?
else
  selector[field.to_s] = send(field)
end
```

Than we would need to test the following scenarios:

- Shard key field is already set and is changed
  - field is originally set to nil -> not included in the query (we have no way of knowing if it was set to nil or wasn't set)
  - field is originally set to false -> included in the query
  - field is originally set to other -> included in the query

- Shard key field is not already set and is changed
  - field is not included in the query

Would you mind updating the pull request and then I'll review it again. Thanks!
Emily

johnnyshields:
@estolfo `nil` needs to be handled as a valid (possible) value for the shard key doesn't it?

intale:
@johnnyshields yes, indeed. You have the point. Will fix.

intale:
I left variant with checking on `new_record?` - non-persisted records should return proper result of `shard_key_selector` as well.

johnnyshields:
@estolfo @intale this looks good to me.

estolfo:
thanks @intale and @johnnyshields

PR #4453, opened 2017-11-14
add support for using $not in $elemMatch

loganyu:
This fixes using $not in $elemMatch

Ex:
attribute: {["a" => 3]}
selector: "$elemMatch" => {"a" => {"$not" => 4}}

document.matches?(selector) will be true.


estolfo:
thanks @loganyu !

PR #4454, opened 2017-11-17
Fix persistence context clear

benknowles:
If you used Model.with() to specify a client and the name was invalid, it would make any subsequent calls to with() break even if valid.

PersistenceContext.clear(Model) wouldn't properly unset the context because calling on client would raise NoClientConfig.

### DIFF
--- a/lib/mongoid/persistence_context.rb
+++ b/lib/mongoid/persistence_context.rb
@@ -195,10 +195,10 @@ module Mongoid
         Thread.current["[mongoid][#{object.object_id}]:context"]
       end
 
-      # Get the persistence context for a particular class or model instance.
+      # Clear the persistence context for a particular class or model instance.
       #
-      # @example Get the persistence context for a class or model instance.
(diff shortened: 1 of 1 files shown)


johnnyshields:
LGTM 👍

johnnyshields:
@estolfo please fix these failing specs? seems they are coming from driver_master Gemfile.

estolfo:
Hi @benknowles 
Thank you very much for this pull request. Would you mind adding a test please? Thanks

Emily

benknowles:
```
# works as intended
Example.with(client: "default") { |x| puts x.count }

# calling with an invalid client name results in
# Mongoid::Errors::NoClientConfig
Example.with(client: "asd") { |x| puts x.count }

# but now with subsequent usage
# this call also throws Mongoid::Errors::NoClientConfig
# because the invalid persistence context is stuck
Example.with(client: "default") { |x| puts x.count }
```

estolfo:
thanks @benknowles

PR #4455, opened 2017-12-04
fix cloning document with polymorphic embedded documents with multiple language field

tomasc:
Currently, localized fields are inferred from the class stored in relation's metadata.

However, in polymorphic embedded relations, each embedded document class can possibly define different localized fields. In the current implementation these are not taken into account and as a result localized fields on embedded polymorphic documents might not be correctly processed. 

I propose this patch, which first attempts to infer localized fields from the class defined in the `_type` attribute, falling back to the class defined on relation's metadata.

Questions:
* is there already a mechanism to convert `_type` attribute to a class, or is the `constantize` method sufficient? Would `safe_constantize` be preferred?
* I have added an extra class to the test suite (`AddressCustomized`), perhaps there is a cleaner way to test the behavior?

### DIFF
--- a/lib/mongoid/copyable.rb
+++ b/lib/mongoid/copyable.rb
@@ -73,9 +73,8 @@ module Mongoid
         next unless attrs.present? && attrs[association.key].present?
 
         if association.is_a?(Association::Embedded::EmbedsMany)
-          attrs[association.name.to_s].each do |attr|
-            embedded_klass = attr.fetch('_type', association.class_name).constantize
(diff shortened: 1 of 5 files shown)


tomasc:
Hi @estolfo , I have implemented your suggestions.
Please let me know in case there is anything else.
Tomas

estolfo:
Great, thank you so much @tomasc !

tomasc:
Thanks!

PR #4456, opened 2017-12-04
Wrapped query cache cursor with read_with_retry [MONGOID-4305]

mdehoog:
Fixes https://jira.mongodb.org/browse/MONGOID-4305.

We're using the `Mongoid::QueryCache` with great success in production. However, whenever we perform a Mongo failover, we see persistent connections continue to attempt to connect to the secondary, causing a large number of `not master and slaveOk=false` errors.

The Ruby Mongo driver's `Retryable` has logic to scan the cluster when these errors occur: https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/retryable.rb#L48. However, I noticed that `read_with_retry` wasn't being called in the stacktrace of our errors. This is because the `QueryCache`'s cursor doesn't wrap the query with `read_with_retry`.

I monkey-patched this change in production, and it fixed all of our issues. This now allows us to failover our Mongo clusters without the 5-10 minutes of downtime we would previously experience (due to having to re-deploy the application to fix Mongo connections).

This PR wraps the `QueryCache` cursor query with a `read_with_retry`.


mdehoog:
Thanks @estolfo!

PR #4457, opened 2017-12-19
fix elemMatch when symbol is used in query

loganyu:
This fixes an issue when :$elemMatch is used in a query as opposed to "$elemMatch"


estolfo:
thanks!

PR #4458, opened 2018-01-03
[MONGOID-4497] _type could be a hash or a string

pacop:
Solved the problem with https://jira.mongodb.org/browse/MONGOID-4497

### DIFF
--- a/lib/mongoid/factory.rb
+++ b/lib/mongoid/factory.rb
@@ -17,7 +17,8 @@ module Mongoid
     #
     # @return [ Document ] The instantiated document.
     def build(klass, attributes = nil)
-      type = (attributes || {})[TYPE]
+      attributes ||= {}
+      type = attributes[TYPE] || attributes[TYPE.to_sym]
       if type && klass._types.include?(type)
         type.constantize.new(attributes)
       else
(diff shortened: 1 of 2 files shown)


lib/mongoid/factory.rb
```diff
@@ -17,7 +17,7 @@ module Factory
     #
     # @return [ Document ] The instantiated document.
     def build(klass, attributes = nil)
-      type = (attributes || {})[TYPE]
+      type = (attributes || {}).with_indifferent_access[TYPE]
```
johnnyshields: Think it would be better to do:

```
attributes ||= {}
type = attributes[TYPE] || attributes[TYPE.to_sym]
```

pacop: Yeah, sure!. It's done. Thanks

estolfo: @pacop Did you make the change suggested by @johnnyshields ?

pacop: @estolfo Yeah, I forgot to push. Sorry, now it should be.

estolfo: @pacop Can you please correct the syntax error? There shouldn't be a space between the double ``|`` and the ``=``.

pacop: @estolfo done, I don't know why it was there. Sorry again

estolfo: great, thank you

estolfo:
thanks @pacop !

PR #4466, opened 2018-01-19
Remove QueryCache#read_or_server_selector

koraktor:
This is now obsolete as driver version 2.5.0 is required. The additional check can be removed.

See cadffdbba1b9d33dc3fe4134c74f41bccce122d7.


estolfo:
yes, indeed, this check can be removed now. Thanks!

PR #4470, opened 2018-01-23
Adds new atomic method "add_each_to_set" which will do an $addToSet with the $each modifier

johnnyshields:
Replaces #4461

Also, we should consider aliasing/renaming pull_all to "pull_each", or else renaming this method as "add_all_to_set"

### DIFF
--- a/lib/mongoid/atomic/modifiers.rb
+++ b/lib/mongoid/atomic/modifiers.rb
@@ -130,6 +130,27 @@ module Mongoid
         end
       end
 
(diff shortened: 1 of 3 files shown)


johnnyshields:
@estolfo any comment?

johnnyshields:
@estolfo please look at this.

saghm:
Hi @johnnyshields! Sorry for the delay in response; everything looks good on our end, so I'm going to merge this now. Thanks for the contribution!

PR #4473, opened 2018-01-25
MONGOID-4483 Demongoize attributes when accessing via [] or read_attribute

estolfo:
This is a behavior change, but corrects Mongoid's behavior to be consistent with ActiveRecord.


calleo:
👍  I tried it out with tests that were previously failing. Working as expected.

PR #4475, opened 2018-01-30
MONGOID-2882 Set base relation on has_many enumerable objects

estolfo:


jarthod:
\o/ I'll give this a try this weekend ;)

jarthod:
I haven't checked the code yet but at least for my use-case is works great :+1:

PR #4477, opened 2018-02-15
MONGOID-4506 Support sessions

estolfo:


spec/mongoid/clients/sessions_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+        before do
+          Person.with_session do
+            Person.create
+            Person.create
+          end
+        end
+
+        let(:insert_events) do
+          subscriber.started_events.select { |event| event.command['insert'] }
+        end
```
ShaneHarvey: Is this a duplicate of the above `:insert_events`?

spec/mongoid/clients/sessions_spec.rb
```diff
@@ -0,0 +1,297 @@
+require "spec_helper"
```
ShaneHarvey: Should there be any tests that use multiple threads?

spec/mongoid/clients/sessions_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+        end
+
+        let(:insert_events) do
+          subscriber.started_events.select { |event| event.command['insert'] }
+        end
+
+        it 'uses a single session id for all operations on the class' do
+          expect(Person.count).to be(2)
+          expect(insert_events).not_to be_empty
+          expect(insert_events.collect { |event| event.command['lsid'] }.uniq.size).to eq(1)
```
ShaneHarvey: This doesn't actually test that every command sent a session id. I suggest:
```
lsids_sent = insert_events.collect { |event| event.command['lsid'] }
expect(lsids_sent.size).to eq(2)
expect(lsids_sent.uniq.size).to eq(1)
 ```

Same for the other tests. I also suggest adding a test that runs every operation that a Mongoid class can do to make sure we didn't miss adding a session in some place. Please ignore if that's already the case.

spec/mongoid/clients/sessions_spec.rb
estolfo: Yes, indeed. I'll remove this, thanks.

spec/mongoid/clients/sessions_spec.rb
estolfo: OK, good point. I'll change this.

spec/mongoid/clients/sessions_spec.rb
estolfo: Sure, that's a good idea. I'll add a test to make sure a session is not accessible from another thread.

PR #4479, opened 2018-02-20
Allow test rails master

Fudoshiki:
Allow rails master in gemspec
Bump dev ruby version


estolfo:
thank you @Fudoshiki !

PR #4483, opened 2018-03-03
fix query date parser with ActiveSupport timezone

infernalmaster:
Mongoid ignored Active Support timezone when parses time from string inside query object. So I used method from https://github.com/mongodb/mongoid/blob/master/lib/mongoid/extensions/string.rb extension to parse dates from strings.

### DIFF
--- a/lib/mongoid/criteria/queryable/extensions/string.rb
+++ b/lib/mongoid/criteria/queryable/extensions/string.rb
@@ -16,7 +16,7 @@ module Mongoid
           #
           # @since 1.0.0
           def __evolve_date__
-            time = ::Time.parse(self)
+            time = __evolve_time__.utc
             ::Time.utc(time.year, time.month, time.day, 0, 0, 0, 0)
           end
 
@@ -29,7 +29,7 @@ module Mongoid
           #
           # @since 1.0.0
           def __evolve_time__
-            ::Time.parse(self).utc
+            __mongoize_time__.utc
           end
 
           # Get the string as a mongo expression, adding $ to the front.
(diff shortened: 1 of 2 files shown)


estolfo:
Hi @infernalmaster 
Thank you for this pull request. Can you please add a test?
Thanks

infernalmaster:
I wrote some tests and reverted `__evolve_date__` to previous version, because I'm not sure about logic for dates parsing, when we have date shift in different time zones and don't want to break current logic.

For example server is located in Tokyo timezone (+9 hours). Let's imagine right now in Tokyo is `Jun 22 04:00` (`UTC: Jun 21 19:00`). So Tokyo user just created new document and created_at will be `UTC: Jun 21 19:00`. And he tries to search documents, that he created today. So he tries 
```
Document.where(created_at: 'Jan 22 2010')
```
And he will not receive this document. Because:
```
def __evolve_date__
   time = ::Time.parse(self)
  ::Time.utc(time.year, time.month, time.day, 0, 0, 0, 0)
end

Time.parse("Jan 22 2010") =>  2010-01-22 00:00:00 +0900 
=> Time.utc(2010, 1, 22, 0, 0, 0, 0)  => 2010-01-22 00:00:00 UTC
```

This is how it works right now.


But maybe this should be like this:
```
def __evolve_date__
  Time.zone.parse(self).beginning_of_day.utc # Time.zone can be replaced with __mongoize_time__
end

Time.parse("Jan 22 2010").beginning_of_day.utc => 2010-01-21 15:00:00 UTC
```

But this totally breaks current logic, so I didn't do that.

estolfo:
Hi @infernalmaster 
Thank you for adding the tests!

PR #4485, opened 2018-03-09
fix before_remove and after_remove relation callbacks and add shift r…

dks17:
When object of a model has `embeds_many` relation with `before_remove` or `after_remove` relation callbacks. Callback does not work when `shift` method is called on `embeds_many` relation. This pr fix it.

### DIFF
--- a/lib/mongoid/association/embedded/embeds_many/proxy.rb
+++ b/lib/mongoid/association/embedded/embeds_many/proxy.rb
@@ -271,6 +271,29 @@ module Mongoid
             end
           end
 
+          # Shift documents off the relation. This can be a single document or
+          # multiples, and will automatically persist the changes.
+          #
+          # @example Shift a single document.
+          #   relation.shift
+          #
+          # @example Shift multiple documents.
+          #   relation.shift(3)
+          #
+          # @param [ Integer ] count The number of documents to shift, or 1 if not
+          #   provided.
+          #
+          # @return [ Document, Array<Document> ] The shifted document(s).
+          def shift(count = nil)
+            if count
+              if _target.size > 0 && docs = _target[0, count]
+                docs.each { |doc| delete(doc) }
+              end
+            else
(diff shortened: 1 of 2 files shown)


lib/mongoid/association/embedded/embeds_many/proxy.rb
```diff
@@ -271,6 +271,29 @@ def pop(count = nil)
...
+          #
+          # @return [ Document, Array<Document> ] The shifted document(s).
+          def shift(count = nil)
+            if count
+              if _target.size > 0 && docs = _target[0, count]
+                docs.each { |doc| delete(doc) }
+              end
+            else
+              delete(_target[0])
+            end
```
estolfo: We can make this the following, for simplicity:

```ruby
  shifted = _target[0, count || 1].each { |doc| delete(doc) }
  shifted.size <= 1 ? shifted[0] : shifted
```

spec/mongoid/association/embedded/embeds_many/proxy_spec.rb
```diff
@@ -2601,6 +2601,108 @@ class TrackingIdValidationHistory
...
+
+      let!(:address_two) do
+        person.addresses.create(street: "hermannstr")
+      end
+
+      context "when the number is not larger than the relation" do
+
+        let!(:shifted) do
+          # person.addresses.create(street: "kos")
+          # binding.pry
```
estolfo: Can we remove these commented lines please?

lib/mongoid/association/embedded/embeds_many/proxy.rb
dks17: I have added test when the number is zero and this implementation fails the tests.

estolfo: Ok, thank you for adding that edge case test. Your implementation is good then, I'll merge this, thank you!

PR #4486, opened 2018-03-12
fix Model#update_attributes and add Model#update_attributes! document description

dks17:
I hope that it's right place for this.
Sorry for the end of document, I don't know why it was changed.


estolfo:
Great, thank you @dks17 !

PR #4490, opened 2018-04-12
MONGOID-4521 Ensure that QueryCache doesn't cache multi-batch results

saghm:


spec/mongoid/query_cache_spec.rb
```diff
@@ -414,15 +414,15 @@
   context "when querying a very large collection" do
 
     before do
-      123.times { Band.create! }
```
simi: What's the reason for this change?

spec/mongoid/query_cache_spec.rb
```diff
@@ -450,6 +450,19 @@
       Band.collection.indexes.create_one(name: 1)
     end
   end
+
+  context 'when the initial query does not exhaust the results' do
+    before do
+      Mongoid::QueryCache.enabled = true
```
simi: Usually it is good idea to temporarily enable config option and restore original state once test is done (for example using `around` block).

https://github.com/mongodb/mongoid/blob/b3a5918e504f4ac9ae4e750701bb02db78000960/spec/mongoid/query_cache_spec.rb#L308-L313

saghm: Note that at the top of the description block, there is already an `around` block defined:

```ruby
around do |spec|
    Mongoid::QueryCache.clear_cache
    Mongoid::QueryCache.cache { spec.run }
  end
```

The implementation of `QueryCache.cache` already saves the value from before the passed block and restores it after.

spec/mongoid/query_cache_spec.rb
```diff
@@ -452,6 +452,19 @@
...
+      10.times { Band.create! }
+
+      Band.batch_size(4).all.any?
+    end
+
+    it 'does not cache the result' do
+      expect(Band.all.map(&:id).size).to eq(10)
+    end
```
p-mongo: This test is good but there should be another one setting batch size to 4 at query time.

spec/mongoid/query_cache_spec.rb
p-mongo: Is enabling query cache again on this line redundant then?

spec/mongoid/query_cache_spec.rb
saghm: I'm not sure I understand what that has to do with the query cache; is the idea to do a query with the same batch size as the number of results of the previous query?

saghm: Done, as discussed offline

PR #4492, opened 2018-04-13
after_save callback fix for cascaded embedded documents

chaudhary:
after_save callbacks for embedded documents are being called before actually saving the document to the database and if we are doing any processing on the document assuming the document is already saved, code gives error.

When I looked at the inside code, I found that all the callbacks for cascaded embedded documents are being called first and then actual document callbacks are being called using super class where the document is actually being saved.

Ideally, it should have been in such a way that all the before callbacks for cascaded embedded documents are being executed first, root document is being saved with their callbacks then, and after callbacks for cascaded embedded documents are being executed finally.

So, I divided the code for run_callbacks into 2 code blocks, one for before callbacks and another for after callbacks with super called in between these blocks.

Let me know if any change is required in order to merge the change.

### DIFF
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 # encoding: utf-8
 module Mongoid
 
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -126,11 +126,21 @@ module Mongoid
     # @since 2.3.0
     def run_callbacks(kind, *args, &block)
       cascadable_children(kind).each do |child|
-        if child.run_callbacks(child_callback_type(kind, child), *args) == false
+        if child.run_before_callbacks(child_callback_type(kind, child), *args) == false
           return false
         end
       end
-      callback_executable?(kind) ? super(kind, *args, &block) : true
+
+      ret_value = callback_executable?(kind) ? super(kind, *args, &block) : true
+      return false if ret_value == false
+
+      cascadable_children(kind).each do |child|
+        if child.run_after_callbacks(child_callback_type(kind, child), *args) == false
+          return false
+        end
+      end
+
+      ret_value
     end
 
     private
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -124,8 +124,9 @@ module Mongoid
     #
     # @since 2.3.0
     def run_callbacks(kind, *args, &block)
-      cascadable_children(kind).each do |child|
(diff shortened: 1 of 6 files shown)


rubish:
Thanks @chaudhary, this saved us some headache and is working fine on our production system.

p-mongo:
The change looks reasonable, just needs some tests.

johnnyshields:
@chaudhary can you please add test for this so it can be merged?

p-mongo:
Looks like we have a related ticket already: https://jira.mongodb.org/browse/MONGOID-3795

jan-harmsen:
any update on this issue and related issue #4566 would be appreciated...

p-mongo:
@jan-harmsen  Both are in our backlog but not being actively worked on.

jan-harmsen:
@p-mongo: OK, then I will have to look for a workaround, thank you. After upgrading from mongoid 5.4 to 7.0 callbacks for embedded docs are not working as they used to, causing bugs in our system.

johnnyshields:
@p-mongo this is a regression bug and should be prioritized on your end.

johnnyshields:
If you need help with this PR and #4566 I can ask our developers at TableCheck to submit rebased PRs for them, if you can agree to fast-track the review...

p-mongo:
The ticket under which this PR falls, https://jira.mongodb.org/browse/MONGOID-3795, is currently treated as a feature request in the sense that implementing it will involve a behavior change. In particular, such a change likely will not be backported to stable versions but will be released in a feature release (as of now, the next feature release will be 7.1).

There could be a regression in Mongoid from 5 to 7 with respect to callbacks, but at least as far as this PR is concerned, the regression is not described. To have it be considered for its own fix the next step is to create a ticket describing/demonstrating the behavior in Mongoid 5 and the corresponding behavior in Mongoid 7. A PR fixing the regression is welcome as well.

chaudhary:
In order to have a temporary fix, you can add the below code in your application and the code should be loaded on application load.
```module Mongoid::Interceptable
  def run_callbacks(kind, *args, &block)
    cascadable_children(kind).each do |child|
      if child.run_before_callbacks(child_callback_type(kind, child), *args) == false
        return false
      end
    end

    ret_value = callback_executable?(kind) ? super(kind, *args, &block) : true
    return false if ret_value == false

    cascadable_children(kind).each do |child|
      if child.run_after_callbacks(child_callback_type(kind, child), *args) == false
        return false
      end
    end

    ret_value
  end
end
```

jan-harmsen:
@chaudhary: **Thank you very much** for your suggested fix, I had already implemented a different workaround avoiding callbacks. I will check again sometime in the future when there's a mongoid release containing fixes.

PR #4493, opened 2018-04-16
MONGOID-4533 Update documentation and warn to reflect Symbol deprecation

saghm:


docs/tutorials/mongoid-documents.txt
```diff
@@ -415,6 +415,33 @@ For cases when you do not want to have ``BSON::ObjectId`` ids, you can override
...
+A Note about the BSON Symbol type
+*********************************
+
+Because the BSON specification deprecated the BSON symbol type, the `bson` gem will serialize Ruby
+symbols into BSON strings when used on its own. However, in order to maintain backwards
+compatibility with older datasets, `mongo` gem overrides this behavior to serialize Ruby symbols as
+BSON symbols. This is necessary to be able to specify queries for documents which contain BSON
+symbols as fields.
+
+Because of this behavior, new applications should exclusively use the string type in place of the
```
estolfo: We should make it clearer in this paragraph that we are talking about the Model field ``type`` setting. So you can change this to "..new applications should not specify model fields with the ``Symbol`` type and rather the ``String`` type. Existing applications with ``Symbol`` model fields may convert those fields to Strings using one of the following methods:

docs/tutorials/mongoid-documents.txt
```diff
@@ -415,6 +415,33 @@ For cases when you do not want to have ``BSON::ObjectId`` ids, you can override
...
    end
 
+A Note about the BSON Symbol type
+*********************************
+
+Because the BSON specification deprecated the BSON symbol type, the `bson` gem will serialize Ruby
+symbols into BSON strings when used on its own. However, in order to maintain backwards
+compatibility with older datasets, `mongo` gem overrides this behavior to serialize Ruby symbols as
+BSON symbols. This is necessary to be able to specify queries for documents which contain BSON
+symbols as fields.
```
estolfo: Maybe add a sentence here saying that although Mongoid still allows you to define models with Symbol field types, this could present problems when using other mongodb tools that have actually removed support for the type.

docs/tutorials/mongoid-documents.txt
```diff
@@ -415,6 +415,33 @@ For cases when you do not want to have ``BSON::ObjectId`` ids, you can override
...
+
+Because of this behavior, new applications should exclusively use the string type in place of the
+Symbol type in fields. Existing applications with symbol fields will additionally need to convert
+those fields into strings using one of the following methods:
+
+- Eager migration: Eager migration: write a script or rake task that queries each document and
+  updates any symbols it finds to strings. This will take a long time for larger data sets, and
+  unless the application can handle two different schemas existing in the data at once, it will
+  need to be taken offline while the script runs
```
estolfo: please add a .

docs/tutorials/mongoid-documents.txt
```diff
@@ -415,6 +415,33 @@ For cases when you do not want to have ``BSON::ObjectId`` ids, you can override
...
+
+- Lazy migration: rather than updating all of the documents at once, each document received from a
+  query  should be reinserted if it contains a symbol; since the BSON received will have already
+  turned the symbols into strings, this will update the document as necessary. This can be done
+  by simply calling the ``save`` method. Note that this method will only convert documents that
+  are accessed by the application, eventually eager migration may be needed to convert the
+  remaining documents that haven't been accessed
```
estolfo: please add a .

docs/tutorials/mongoid-documents.txt
```diff
@@ -415,6 +415,33 @@ For cases when you do not want to have ``BSON::ObjectId`` ids, you can override
...
+
+For existing applications using
```
estolfo: I don't think you intend for this line to be here.

lib/mongoid/fields/validators/macro.rb
```diff
@@ -104,6 +109,10 @@ def validate_options(klass, name, options)
             if !OPTIONS.include?(option) && !Fields.options.include?(option)
               raise Errors::InvalidFieldOption.new(klass, name, option, OPTIONS)
             end
+
+            if option == :type && options[option] == Symbol
+              Mongoid.logger.warn FIELD_TYPE_IS_SYMBOL
```
estolfo: let's use parenthesis to be consistent with the rest of the style.

spec/mongoid/fields_spec.rb
```diff
@@ -425,6 +425,17 @@
...
+      before do
+        expect(Mongoid.logger).to receive(:warn)
+      end
+
+      it "warns that the BSON symbol type is deprecated" do
+        Band.field :should_warn, type: Symbol
+      end
+    end
+
     context "when the options are not valid" do
```
estolfo: Let's make sure we delete the field off the ``Band`` model after this test.
Unlike the test right above this one, the field is actually added so some cleanup is necessary.

for example:
```ruby
        after do
          Band.fields.delete("acceptable")
        end
```

lib/mongoid/fields/validators/macro.rb
saghm: Sorry about that; I was thinking of `Kernel#warn` when I wrote this, but this shouldn't be written the way I did.

PR #4465, opened 2018-01-18
MONGOID-4499 Replace uses of pushAll with push and each

estolfo:

PR #4494, opened 2018-04-19
Port MONGOID-4499 to 5.2.x

TreyE:
Port fixes from https://github.com/mongodb/mongoid/pull/4465 into 5.2.

### DIFF
--- a/lib/mongoid/atomic/modifiers.rb
+++ b/lib/mongoid/atomic/modifiers.rb
@@ -68,7 +68,7 @@ module Mongoid
         modifications.each_pair do |field, value|
           push_fields[field] = field
           mods = push_conflict?(field) ? conflicting_pushes : pushes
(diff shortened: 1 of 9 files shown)


saghm:
Thanks for the pull request!

In order to change this, we'd need to drop support for server version 2.4, so we'd like this to be released in a new minor version. Could you close this PR and reopen it against the [5.4.0-stable](https://github.com/mongodb/mongoid/tree/5.4.0-stable) branch?

To enforce the dropped support for 2.4, we'd need to update the dependency on the `mongo` gem to 2.5.1, so could you update the gemspec file to reflect this when you reopen the PR? See [here](https://github.com/mongodb/mongoid/blob/cc84f9c60f0f2dd66e7c3cfd28e0271fc85d42b5/mongoid.gemspec#L30) for what it should look like.

TreyE:
Created PR #4495. Closing this PR.

PR #4498, opened 2018-05-10
Demonstrate removal of Mongoid::Document#touch

niels:
This test succeeds in v6.4.1 but starts failing with v7.0.0.beta. The changelogs do not seem to include a deprecation / breaking change notice.

Bug report @ https://jira.mongodb.org/browse/MONGOID-4545.


p-mongo:
Fixed in #4521, thanks.

PR #4499, opened 2018-05-23
Add #one method on has_many associations

p-mongo:
https://jira.mongodb.org/projects/MONGOID/issues/MONGOID-4547?

### DIFF
--- a/lib/mongoid/association/referenced/has_many/enumerable.rb
+++ b/lib/mongoid/association/referenced/has_many/enumerable.rb
@@ -213,6 +213,27 @@ module Mongoid
               end
             end
 
+            # Get an arbitrary document in the enumerable. Will check the persisted
+            # documents first. Does not load the entire enumerable.
+            #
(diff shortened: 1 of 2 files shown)


lib/mongoid/association/referenced/has_many/enumerable.rb
```diff
@@ -213,6 +213,27 @@ def empty?
...
+            #
+            # @note Unlike #first, this does not automatically sort the result set.
+            #
+            # @param [ Hash ] opts The options for the query returning the first document.
+            #
+            # @option opts [ :none ] :id_sort Don't apply a sort on _id.
+            #
+            # @return [ Document ] The first document found.
+            #
+            # @since 2.1.0
```
saghm: This should be 7.1.0, correct?

p-mongo: @saghm Repaired.

PR #4500, opened 2018-06-05
Ensure without_default_scope does not suppress default scopes of other models

brennana:
The method `Mongoid::Scopable.without_default_scope` has the peculiar behavior of suppressing default scopes of models not called upon to do so. For example:

```ruby
Appointment.without_default_scope do
  # the following should be true, obviously
  Appointment.all.selector.empty? #=> true

  # the following should be false, as Audio has a default scope that we are not trying
  # to suppress in this block, given the receiver of `without_default_scope`.
  Audio.all.selector.empty? #=> true
end
```

I have fixed this behavior, and have written specs to ensure it continues exhibiting the appropriate behavior. Additional methods were added to `Mongoid::Threaded` to allow for easily adding and removing classes from the `:without_default_scope` stack of the current thread, which now contains all the classes whose default scopes are being actively suppressed, rather than just containing `true` to denote a global suppression of default scopes.

I have left blank the `@since` doc attribute of the new methods in `Mongoid::Threaded`, since I do not know what version these would be released in. Moreover, this commit should be backportable to mongoid 5.x.

### DIFF
--- a/lib/mongoid/scopable.rb
+++ b/lib/mongoid/scopable.rb
@@ -103,7 +103,7 @@ module Mongoid
       #
       # @since 3.0.0
       def default_scopable?
-        default_scoping? && !Threaded.executing?(:without_default_scope)
+        default_scoping? && !Threaded.without_default_scope?(self)
       end
 
       # Get a queryable, either the last one on the scope stack or a fresh one.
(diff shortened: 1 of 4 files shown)


johnnyshields:
This looks good to me. @p-mongo ?

p-mongo:
I made https://jira.mongodb.org/browse/MONGOID-4630 to track the work here.

PR #4503, opened 2018-06-20
Stop modification of time zones when converting Time objects

LalaDK:
I'm having problems with Mongoid queries modifying time variables. The following example shows how a Time object looses its time zone information after being used in a query where the Time object is used to query a field with the datatype DateTime. 

```
class Person
  include Mongoid::Document
  store_in collection: "persons"
  field :birthday, type: DateTime
end
=> #<Mongoid::Fields::Standard:0 ...

time = Time.new(1970, 1, 1, 0, 0, 0, "+05:00")
=> 1970-01-01 00:00:00 +0500

Person.where(:birthday => time)
=> #<Mongoid::Criteria ...

time
=> 1969-12-31 19:00:00 UTC
```

I understand that Mongo performs conversions to UTC, but it should not modify the variable 'time'. This modification does not happend when using DateTime. The problem occurs at lib/mongoid/criteria/queryable/extensions/time.rb @ 31

The method 'utc' is called on the object which modifies it. The Ruby documentation states: 

> Converts time to UTC (GMT), modifying the receiver.

[https://ruby-doc.org/core-2.2.0/Time.html#method-i-utc](https://ruby-doc.org/core-2.2.0/Time.html#method-i-utc)

I believe the fix is to call 'getutc' instead. 

> Returns a new Time object representing time in UTC.

[https://ruby-doc.org/core-2.2.0/Time.html#method-i-getutc](https://ruby-doc.org/core-2.2.0/Time.html#method-i-getutc)

**Screenshot**
mongodb timezone

This is my first pull request and I couldn't find any guidelines, so bare with me.

### DIFF
--- a/lib/mongoid/criteria/queryable/extensions/time.rb
+++ b/lib/mongoid/criteria/queryable/extensions/time.rb
@@ -28,7 +28,7 @@ module Mongoid
           #
           # @since 1.0.0
           def __evolve_time__
-            utc
+            getutc
           end
 
           module ClassMethods
--- a/Gemfile
+++ b/Gemfile
@@ -10,3 +10,11 @@ group :test do
   gem 'benchmark-ips'
   gem 'rspec', '~> 3.7'
 end
+
+group :development, :testing do
+  platforms :mri do
+    if RUBY_VERSION >= '2.0.0'
+      gem 'byebug'
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


lib/mongoid/criteria/queryable/extensions/time.rb
```diff
@@ -28,7 +28,7 @@ def __evolve_date__
           #
           # @since 1.0.0
           def __evolve_time__
```
saghm: It's interesting that this is called "evolve", which to me implies that the mutation might have originally been intention (although I might just not be realizing that this is an override of something else). That being said, I definitely agree that we shouldn't be mutating the arguments users pass in to queries, so I'm fine with this change.

p-mongo:
I rebased yesterday.

PR #4506, opened 2018-07-05
Fix read preference passing to ruby driver

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4564

### DIFF
--- a/lib/mongoid/persistence_context.rb
+++ b/lib/mongoid/persistence_context.rb
@@ -107,10 +107,6 @@ module Mongoid
     #
     # @since 6.0.0
     def client
-      client_options = send(:client_options)
-      if client_options[:read].is_a?(Symbol)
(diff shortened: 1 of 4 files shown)


lib/mongoid/persistence_context.rb
```diff
@@ -107,6 +107,10 @@ def database_name
     #
     # @since 6.0.0
     def client
+      client_options = send(:client_options)
+      if client_options[:read].is_a?(Symbol)
```
saghm: Do we specifically need to keep the read preference as a symbol for use elsewhere? If not, I think it would be cleaner to just modify the options before caching them in the `client_options` method so that we don't do this check every time. If we do need it a a symbol elsewhere, I'd suggest either putting `return @client if @client` at the top of the method (and changing the assignment below from `||=` to `=`) or moving this within the parenthetical block in the last assignment so that we don't repeat this work.

PR #4507, opened 2018-07-05
Implement transactions

saghm:


saghm:
`driver-stable` tests are failing pending the release of 2.6.1, but everything else is passing, so I think this is ready for review

lib/mongoid/clients/sessions.rb
```diff
@@ -51,6 +51,67 @@ def with_session(options = {})
         Threaded.clear_session
       end
 
+      # Execute a block within the context of a transaction. Note that this can only occur within
+      # a session.
+      #
+      # @example Execute some operations in a transaction.
+      #   band.with_session do
+      #     band.with_transaction do
```
p-mongo: What is preventing Mongoid from automatically executing `with_session` when a user wants a transaction?

saghm: The issue is that sessions can't get nested. If it automatically calls `with_session`, then without extra logic it would raise an error if it were called from within a session block. I'm not big on the idea of having `with_transaction` only sometimes implicitly start an explicit session; a better idea IMO would be to have a separate method for starting a transaction and a session at once (e.g. `with_session_and_transaction`), but I'm not super fond of the cognitive overhead of forcing users to choose between which method to call depending on the situation. I'd be interested to hear your thoughts on the tradeoffs between having `with_transaction`'s behavior be dependent on the context, having a separate method for starting a transaction and a session at once, or just having the user have to call both methods in the case that a session isn't already started.

p-mongo: I would do a github code search for `with_session` to see if there are any instances of people actually using that method.

https://docs.mongodb.com/manual/reference/server-sessions/ says:

> MongoDB’s server sessions, or logical sessions, are the underlying framework used by client sessions to support Causal Consistency and retryable writes.

I would expect that users of Mongoid, being a higher level ODM, would most likely prefer to have implicit sessions created for them all the time, and hence would almost never call `with_session` themselves.

saghm: I'm not sure I accept the premise that a Github code search would give a useful sample of projects; I'd imagine that most real-world applications using Mongoid are not open source.

p-mongo: My impression of the session concept is the entire application should generally be using sessions all the time, or not. Hence I'm thinking whether sessions are used by default should be configurable in mongoid.yml and equivalents.

Is there a performance penalty for having sessions on all the time, and is it quantified anywhere?

saghm: That's a good question! I know that the use of sessions is to facilitate things like transactions, causally consistent reads, and retryable writes. I'm not sure if the general usage pattern is to enable them for the entire lifetime of a client or just sections. The spec doesn't mention anything about a URI option for a session (which is how most things that last the entire lifetime of the client are configured), but that might just be because there isn't any good way to get back that session when creating a client from a URI. Let's check around and see what others on the drivers team think about this.

saghm: Per offline discussion, we'll leave this as-is for now and open a new ticket to make sessions used by default on all Mongoid clients.

saghm:
I've added a new tutorial page for transactions (which lifts heavily from the existing sessions one). It documents the behavior as the code stands now (i.e. needing to call `with_transaction` from within a session, so we'll need to update this when we change sessions to be used by default (although we'd need to update the sessions page anyhow)

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,63 @@
...
+that you would execute a transaction in ActiveRecord as long as the current thread is within a session. Namely, you can
+call a method, ``#with_transaction`` on a model class or on an instance of a model from within a session block and execute
+some operations in a new block. All operations in the inner block will be executed in the context of single transactions. Please
+see the MongoDB Ruby driver documentation for what transaction options are available.
+
+Please note the following limitations of sessions:
+
+- Transactions cannot be shared across threads; transactions are not thread-safe. This is consistent with the Ruby driver's support for transactions.
+
+- Transactions cannot be nested. You cannot called ``#with_transaction`` on a model class or a model instance within the block passed to the ``#with_transaction`` method on another model class or model instance.
```
p-mongo: It is my impression that ActiveRecord allows nesting transactions, and if so, Mongoid should too.

lib/mongoid/clients/sessions.rb
p-mongo: This PR depends on the implicit sessions as discussed.

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,63 @@
...
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Versions 4.0 and higher of the MongoDB server support transactions. You can use sessions with Mongoid in a similar way
+that you would execute a transaction in ActiveRecord as long as the current thread is within a session. Namely, you can
+call a method, ``#with_transaction`` on a model class or on an instance of a model from within a session block and execute
+some operations in a new block. All operations in the inner block will be executed in the context of single transactions. Please
+see the MongoDB Ruby driver documentation for what transaction options are available.
```
p-mongo: This needs to reference either the class/method whose options are accepted by name or have a URL to the same. Same for the session doc where this line originated.

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,63 @@
...
+
+Please note the following limitations of sessions:
+
+- Transactions cannot be shared across threads; transactions are not thread-safe. This is consistent with the Ruby driver's support for transactions.
+
+- Transactions cannot be nested. You cannot called ``#with_transaction`` on a model class or a model instance within the block passed to the ``#with_transaction`` method on another model class or model instance.
+
+- Transactions cannot be used outside of a session. You must be within a session block initiated on the same model class or model instance.
+
+- All model classes and instances used within the transaction block must use the same driver client. For example, if you have specified different ``storage_options`` for another model used in the block than that of the model class or instance on which ``#with_session`` is called, you will get an error.
```
p-mongo: There needs to be guidance provided to the user as to what to do in these situations.

docs/tutorials/mongoid-transactions.txt
saghm: I don't think that's possible; MongoDB only supports one open transaction at a time on a given session

docs/tutorials/mongoid-transactions.txt
saghm: Modifying the existing session documentation in a way that doesn't relate to transactions is a bit out of scope for this PR, but okay

saghm: Actually, it looks like the options are documented in detail within the driver itself; I'll add the reference to `Mongo::Client#start_session`, but we'll need to add the documentation there separately

docs/tutorials/mongoid-transactions.txt
saghm: This is copied directly from the sessions tutorial (mod changing the word "session" to "session"; I'm not sure what else to say beyond "use the same client", which is already stated.

johnnyshields:
Does this PR support transactions across multiple collections at once? If so, please add it to the tutorial. If not, please implement it that way, because that's really the whole point of using transactions.

It would also be ideal if we could use the commit/abort/retry semantics here: https://docs.mongodb.com/manual/core/transactions/

saghm:
@johnnyshields If you want to start/commit/abort manually, you can just use the session which is yielded to the block in a call to `with_session` and call the methods manually. The page you linked to contains examples for how to use the session object to do this.

docs/tutorials/mongoid-transactions.txt
p-mongo: I think @johnnyshields 's comment is in relation to this. What is an example of Mongoid models and configuration which will cause a user to "get an error" as described here?

johnnyshields:
@saghm ok, and it supports transactions for multiple document types at once?

saghm:
@johnnyshields Yes, see the test describing this [here](https://github.com/mongodb/mongoid/pull/4507/files#diff-00981353611a17b5bdc4d214c4b7b25cR99)

docs/tutorials/mongoid-transactions.txt
saghm: As the first sentence states, if you use a model class or instance that has a different underlying client, you'll get an error.

johnnyshields:
Just a comment, I find it slightly counter-intuitive using `Model.with_session`. It seems to me a better way to do it would be something like `Mongoid.default_client.with_session` or `Model.session.with_session` as it's really the session object, not the model, that is being yielded.

docs/tutorials/mongoid-transactions.txt
p-mongo: I doubt a user reading this would know how the underlying clients are constructed without a link to an explanation or an example.

saghm: Considering that all models will use the same client unless explicitly configured not to, I'm not sure how a user could run into issues with this if they don't know how to change clients in the first place.

p-mongo: Exactly. The current wording says a certain usage is disallowed, the doc should make it easy for a user to figure out whether their usage is allowed or not.

saghm: I added a clarification [here](https://github.com/mongodb/mongoid/pull/4507/commits/cb8bc43b65935459c7a1f8aa916abfc47216eb87); do you find this sufficient?

p-mongo:
Docs build produced the following warnings:

<pre>
source/tutorials/mongoid-sessions.txt:1: WARNING: Title overline too short.
=========
Transactions
source/tutorials/mongoid-transactions.txt:32: WARNING: Title underline too short.
Using a session via Model#transaction_session
***************************************
source/tutorials/mongoid-transactions.txt:50: WARNING: Title underline too short.
Using a session via model#with_transaction
</pre>

Transactions are missing in the left sidebar when not on sessions/transactions pages. When on those pages, URL and content are flipped for some reason (the sessions page has transactions url and vice versa).

docs/tutorials/mongoid-sessions.txt
```diff
@@ -1,5 +1,5 @@
 =========
-Sessions
+Transactions
```
p-mongo: Probably because of this.

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,63 @@
+=========
+Sessions
```
p-mongo: And this. The contents and the file name don't match.

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,63 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Versions 4.0 and higher of the MongoDB server support transactions. You can use sessions with Mongoid in a similar way
```
p-mongo: s/sessions/transactions/

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,63 @@
...
+that you would execute a transaction in ActiveRecord as long as the current thread is within a session. Namely, you can
+call a method, ``#with_transaction`` on a model class or on an instance of a model from within a session block and execute
+some operations in a new block. All operations in the inner block will be executed in the context of single transactions.
+Available options are documented in the MongoDB Ruby driver method Mongo::Session#start_transaction.
+
+Please note the following limitations of sessions:
```
p-mongo: s/sessions/transactions/

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,63 @@
...
+
+- Transactions cannot be nested. You cannot called ``#with_transaction`` on a model class or a model instance within the block passed to the ``#with_transaction`` method on another model class or model instance.
+
+- Transactions cannot be used outside of a session. You must be within a session block initiated on the same model class or model instance.
+
+- All model classes and instances used within the transaction block must use the same driver client. For example, if you have specified different ``storage_options`` for another model used in the block than that of the model class or instance on which ``#with_session`` is called, you will get an error. (Models without client options specified will use the same client by default.)
+
+- All connected MongoDB servers must be version 4.0 or higher.
+
+Using a session via Model#transaction_session
```
p-mongo: What is transaction_session?

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,63 @@
...
+     Person.with_transaction do
+        Person.create
+        person = Person.first
+        person.name = "Emily"
+        person.save
+     end
+   end
+
+
+Using a session via model#with_transaction
```
p-mongo: This heading doesn't make much sense. Please proofread the entire file.

p-mongo:
For nested transactions: https://github.com/p-mongo/tests/blob/master/nested-transactions-ar/lib/tasks/test.rake should have a supported equivalent in Mongoid and produce output equivalent to https://github.com/p-mongo/tests/blob/master/nested-transactions-ar/README.md.

saghm:
I strongly disagree that we should model our behavior on nested transactions after what ActiveRecord does. In ActiveRecord, the above example will fail, but without the explicit `fail`, the nested transaction _will_ succeed. Rollback semantics are necessary because the transaction would have succeeded otherwise, so it needs to be rolled back. In Mongoid, a nested transaction will _never_ succeed for _any_ given transaction block. Rolling the transaction back doesn't make sense here because the server is never even going to allow the transaction to start in the first place. Because of this restriction on nested transactions, I think it's extremely misleading to users to imply that the transaction is "rolled back" when the transactions are nested, as it implies both that the transaction started in the first place and that the error occurred somewhere down the line of executing the transaction, neither of which are true. Since nested transactions will not work no matter what users do, we should be as clear about that as possible.

p-mongo:
Per the earlier comment from @johnnyshields , the docs should have two different model classes in a transactions example.

spec/mongoid/clients/transactions_spec.rb
```diff
@@ -0,0 +1,373 @@
...
+
+        context 'when the other class uses the same client' do
+
+          before do
+            Post.with(client: :other) do
+              Person.with_session do
+                Person.with_transaction do
+                  Person.create
+                  Person.create
+                  Post.create
```
p-mongo: There needs to be another test added which:
1. Creates these 3 objects
2. Queries server for these 3 objects, verifies that they are returned.
3. Raises an exception
4. Outside of the transaction block, queries the server again for these 3 objects, verifies that none of them is returned.

p-mongo:
Per #drivers:
> shane.harvey
> Yes a transaction can read/write to multiple collections/databases.

We need to see what is involved in making cross-database transactions work in Mongoid.

johnnyshields:
My two cents is that having the transaction methods as class-level methods on the document models `Post.create` is awkward, even if it's what AR does. I'd much prefer to do something like `Mongoid.with_transaction` or `Mongoid.default_database.with_transaction`

p-mongo:
AR also has them on ActiveRecord::Base, though it's almost always shorter to reference a model.

Unlike AR which talks to a single database per process out of the box, Mongoid supports multiple clients, therefore `Mongoid.with_transaction` doesn't have enough information to know which client to start the transaction on.

`Mongoid.default_database.with_transaction` - I suppose something like `Mongoid.default_client.with_transaction` (however that's spelled) will work, but why do you want to type that much?

johnnyshields:
I would be fine with having `Post.transaction` be an alias/shorthand for `Mongoid.default_client.with_transaction`. The latter makes much more sense to me in terms of how MongoDB actually handles transactions, and we may also need something like that for multi-db support.

johnnyshields:
@saghm do you think MongoDB is likely to support either nested or savepoints in the future, given that it seems to be in a bit of an arms-race with Postgres?

saghm:
I don't know whether it will or not. However, I think the fact that MongoDB might have semantics for how this should behave is also an argument for not trying to implement something now; if the semantics we decide conflict with how MongoDB decides nested transactions or save points should work in some future version, we would be forced to make a behavioral change in order to support this, which I think is something we should try to avoid. We've already run into this issue in the driver with retryable writes (and may run into it in the future if MongoDB implements retryable reads), and it hasn't been trivial to deal with.

saghm:
I'm not convinced that allowing users to express transactions in this way is less error-prone. Given that an inner transaction will not do anything (regardless of options, unlike in ActiveRecord), it seems misleading to allow users to group operations as such when there's no guarantee that the operations will actually complete successfully despite the block existing successfully (since a later error would abort the transaction). While it might be slightly easier for users to be able to start transactions without keeping track of whether a transaction is already started, I think the false implication that the operations have completed successfully after a transaction block has finished without error is more problematic. The cognitive overhead of keeping track of whether a certain point in code is inside a transaction doesn't go away if we allow nested transactions, as they'll still need to know this in order to tell if the inner transactions operations executed or not if afterwards the outer transaction aborts due to an error.

I think part of the disagreement we're having here is about what we consider to be a "transaction" in Mongoid. If I'm understanding correctly, you want to have a concept of a "Mongoid transaction" that's separate from a MongoDB transaction. I'm personally opposed to introducing this right now; this will be the first release supporting MongoDB transactions, and introducing another type of "transaction" to Mongoid at the same time seems likely to cause confusion. I'm not necessarily opposed to introducing some sort of concept like this in the future (although at this point I'd prefer that they be named something distinct from the MongoDB feature), but I don't think that they should just be tacked onto the MongoDB transactions work without having spent any significant amount of time discussing and designing exactly what we want "Mongoid transactions" to be and how they should behave (and no discussion of this would be complete without taking into account the future of what MongoDB transactions might support in the future). To that end, I propose that we table the idea of implementing any type of Mongoid-specific transaction feature pending further discussion (which should include input from the driver leads and the product team pertaining to both the naming of such a feature and the potential future features of MongoDB transactions) and change the documentation in this pull request to clearly indicate that this is an implementation of MongoDB transactions (as opposed to any sort of parallel to the ActiveRecord feature). This would allow us to finally implement and release support for MongoDB 4.0 in Mongoid while leaving our options open to add some sort of Mongoid-specific feature in the future. Thoughts?

p-mongo:
The purpose of Mongoid is to provide a high level interface to MongoDB. An important aspect of Mongoid is API compatibility with ActiveRecord (to the extent practical). Unlike the driver whose priorities are 1) exposing MongoDB functionality and 2) doing so in a way consistent with other language drivers, Mongoid's priorities balance MongoDB functionality, API compatibility with ActiveRecord and ease of use/developer experience.

Mongoid as I understand it provides the option to drop down to the driver and do whatever is desired in a low level fashion that is consistent with other language and with minimum bells and whistles. If this is not clearly documented we should improve the relevant documentation.

Perhaps a good next step would be to document how to use driver-level sessions from an application using Mongoid, for example to retry commits as is possible with the driver.

saghm:
Given that we can't actually implement the feature in the same way as ActiveRecord, I would argue that it is _not_ practical to be compatible here, and structuring the API to imply that we do have parity is misleading. Moreover, I think we've gotten way off from what this ticket/PR was supposed to address, which is adding 4.0 support to Mongoid, not to design an entirely new API for something only tangentially related. If your argument is that we should not provide any sort of direct interface to MongoDB transactions at all, I fundamentally disagree.

Either way, this discussion is not moving forward at all, so I suggest that we take it offline until we can reach a common ground.

johnnyshields:
1) Is it possible to have a literal/strict implementation of transactions in the Ruby MongoDB driver first, then do the "magic" based approach in Mongoid (with a clear documentation of what that magic is?)

2) If Mongoid is attempting to match AR, but there is something like nested transactions which are not supported in Mongoid, then it needs to throw an error "Nested transactions are not supported." It cannot "silently fail"

p-mongo:
@johnnyshields The driver already implements transactions, as of 2.6.0.

As already mentioned in an earlier comment, AR handles databases that support nested transactions, that don't support nested transactions but have save points, and that have neither, using a model and API that as far as I know hasn't changed in 15 years.

p-mongo:
https://docs.mongodb.com/manual/core/transactions/ specifies the following limitations of transactions:

-    You cannot read/write to collections in the config, admin, or local databases.
-    You cannot write to system.* collections.
-    You cannot return the supported operation’s query plan (i.e. explain).
-   For cursors created outside of transactions, you cannot call getMore inside a transaction.

Since Mongoid adds the transaction id when an operation is executed within transaction block, these cases need to be handled, documented and tested. It seems most reasonable to perform the first three operations without specifying a transaction id, and for the last one the getMore should be invoked with the appropriate options matching what its cursor was created with.

saghm:
I think those cases are all handled on the server side; we shouldn't need to have any logic to handle them

saghm:
@p-mongo the changes discussed offline have been made; this is ready for review

p-mongo:
The diff looks good, but what happened to the documentation?

saghm:
The documentation described methods that no longer exist. As Bernie had decided, this pull request should just be testing, and the documentation will be addressed separately.

p-mongo:
Please document the behavior that this diff is testing (start_transaction/end_transaction calls from an application using mongoid).

saghm:
@p-mongo I've re-added the tutorial (borrowing heavily from your pull request on the driver itself)

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,115 @@
+=========
+Transactions
+=========
```
p-mongo: The number of equal signs should equal the length of the word.

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,115 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Version 4.0 of the MongoDB server introduces
+`multi-document transactions <https://docs.mongodb.com/manual/core/transactions/>`_.
+(Updates to multiple fields within a single document are atomic in all
+versions of MongoDB). Mongoid version 7.1.0 adds support for transactions.
```
p-mongo: Going to our earlier discussion about backporting transactions to 6.x, it is my impression that 6.x should support transactions with 2.6.0+ driver.

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,115 @@
...
+
+Using Transactions
+------------------
+
+In order to start a transaction, the application must have a :ref:`session <sessions>`.
```
p-mongo: This link is most likely broken as I had to add the session reference.

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,115 @@
...
+
+  class Post
+    include Mongoid::Document
+  end
+
+  Person.with_session do |s|
+    s.start_transaction
+    Person.create
+    Person.create
+    Post.create
```
p-mongo: Please use the `create!` forms unless the return value is checked.

docs/tutorials/mongoid-transactions.txt
saghm: Good point! I think we should probably at some explicit testing for this, but I'm fine with documenting it for now.

docs/tutorials/mongoid-transactions.txt
```diff
@@ -0,0 +1,115 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Version 4.0 of the MongoDB server introduces
+`multi-document transactions <https://docs.mongodb.com/manual/core/transactions/>`_.
+(Updates to multiple fields within a single document are atomic in all
+versions of MongoDB). Mongoid versions 6.4 and higher support transactions.
```
p-mongo: I think specifying that driver 2.6.0 or higher is required is important here, because using driver 2.6 with older mongoids will make it possible to use transactions. I wouldn't give a mongoid version here at all unless you want to figure out which version the session support was added in, and also add that to the sessions page because it currently doesn't say.

PR #4508, opened 2018-07-09
stop building every mongodb version on travis

saghm:
Given the long running times of our Travis jobs, it has been requested that we only test the latest version of MongoDB on Travis rather than every version (since we already test all versions on Evergreen)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,41 +13,16 @@ before_install:
   - gem update --system
   - gem update bundler
 
-matrix:
-  fast_finish: true
-  exclude:
-    - rvm: ruby-head
-      gemfile: gemfiles/rails_master.gemfile
-    - rvm: jruby-9.1.16.0
-      gemfile: gemfiles/rails_master.gemfile
-    - rvm: jruby-9.1.16.0
-      gemfile: gemfiles/driver_master.gemfile
-    - rvm: jruby-9.1.16.0
-      env: MONGODB=2.6.12
-    - rvm: jruby-9.1.16.0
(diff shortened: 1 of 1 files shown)


p-mongo:
Mongoid doesn't get tested on travis at all, but if it did, the comment in https://github.com/mongodb/mongo-ruby-driver/pull/986 would apply.

saghm:
Ah, forgot that Travis was turned off for Mongoid

saghm:
Actually, it looks like Travis is still running on commits (rather than PRs): https://travis-ci.org/mongodb/mongoid

Do we want to reopen this to change what runs, or should we just turn that off?

p-mongo:
Travis should test a minimal set of configurations, same as for the driver.

PR testing for Mongoid on Travis also needs to be enabled.

saghm:
I'm going to reopen this, then; other drivers are still getting backed up from Mongoid Travis runs and we need to fix this (before enabling Travis runs for even more things)

saghm:
Since Travis isn't enabled on pull requests on this repo, I enabled it for my fork and then opened a PR; it seems to be working: https://travis-ci.org/saghm/mongoid/builds/415657950?utm_source=github_status&utm_medium=notification

saghm:
Bernie has recommended that we just use the default service for MongoDB instead of manually selecting a version. We can discuss it with him if you'd prefer to go another route.

saghm:
@p-mongo Travis is now passing with the agreed upon configuration https://travis-ci.org/saghm/mongoid

PR #4509, opened 2018-07-16
docs: mongoid-installation - remove table column widths

kay-kim:
Remove table column widths in mongoid-installation tutorial doc as the number of column widths and the number of columns are not in sync


p-mongo:
Thanks!

PR #4510, opened 2018-07-17
Fix Settable: allow setting a nested value to a non-string (for 6.4 stable)

mdehoog:
#4487 introduced a regression: one can no longer `$set` a nested hash value to anything that doesn't implement `empty?`.

This PR fixes a couple of bugs with the previous PR:
 - Pull out the `value` from `field_and_value_hash` _before_ checking if it's `empty?`; this ensures we're checking the actual value that is attempting to be set
 - Check that it is actually a `Hash` before checking if it's `empty?`

The reason previous tests passed is because they only checked that nested values could be set to a string, which implements `empty?`.

Relevant stacktrace here:
```ruby
      undefined method `empty?' for 1531794645:Fixnum
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/persistable/settable.rb:29:in `block (2 levels) in set'
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/persistable.rb:166:in `block (2 levels) in process_atomic_operations'
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/attributes/readonly.rb:36:in `as_writable_attribute!'
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/persistable.rb:165:in `block in process_atomic_operations'
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/persistable.rb:164:in `each'
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/persistable.rb:164:in `process_atomic_operations'
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/persistable/settable.rb:24:in `block in set'
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/persistable.rb:142:in `prepare_atomic_operation'
./tmp/bundle/ruby-2.3.7/ruby/2.3.0/gems/mongoid-6.4.1/lib/mongoid/persistable/settable.rb:23:in `set'
```

### DIFF
--- a/lib/mongoid/persistable/settable.rb
+++ b/lib/mongoid/persistable/settable.rb
@@ -25,14 +25,14 @@ module Mongoid
 
             field_and_value_hash = hasherizer(field.split('.'), value)
             field = field_and_value_hash.keys.first.to_s
+            value = field_and_value_hash[field]
 
-            if fields[field] && fields[field].type == Hash && attributes.key?(field) && !value.empty?
+            if fields[field] && fields[field].type == Hash && attributes.key?(field) && Hash === value && !value.empty?
               merger = proc { |key, v1, v2| Hash === v1 && Hash === v2 ? v1.merge(v2, &merger) : v2 }
-              value = (attributes[field] || {}).merge(field_and_value_hash[field], &merger)
-              process_attribute(field.to_s, value)
-            else
-              process_attribute(field.to_s, field_and_value_hash[field])
+              value = (attributes[field] || {}).merge(value, &merger)
             end
+
+            process_attribute(field.to_s, value)
 
             unless relations.include?(field.to_s)
               ops[atomic_attribute_name(field)] = attributes[field]
--- a/spec/mongoid/persistable/settable_spec.rb
+++ b/spec/mongoid/persistable/settable_spec.rb
@@ -329,7 +329,7 @@ describe Mongoid::Persistable::Settable do
           church.set('location.address.city' => 12345)
         end
 
-        it 'does not reset the nested hash' do
+        it 'updates the nested value to the correct value' do
           expect(church.name).to eq('Church1')
           expect(church.location).to eql({'address' => {'city' => 12345, 'street' => 'Yorckstr'}})
         end
--- a/spec/mongoid/persistable/settable_spec.rb
+++ b/spec/mongoid/persistable/settable_spec.rb
@@ -315,6 +315,25 @@ describe Mongoid::Persistable::Settable do
         end
       end
 
+      context 'when a leaf value in the nested hash is updated to a number' do
+
+        let(:church) do
+          Church.new.tap do |a|
+            a.location = {'address' => {'city' => 'Berlin', 'street' => 'Yorckstr'}}
+            a.name = 'Church1'
+            a.save
+          end
+        end
+
+        before do
+          church.set('location.address.city' => 12345)
+        end
+
+        it 'does not reset the nested hash' do
+          expect(church.name).to eq('Church1')
+          expect(church.location).to eql({'address' => {'city' => 12345, 'street' => 'Yorckstr'}})
+        end
+      end
 
       context 'when the nested hash is many levels deep' do
 


mttdffy:
I recently ran into an issue with this regression. Im not sure this PR fixes the whole issue. Won't both of the following `#set` calls fail still?

```ruby
f = Foo.new(details: { 'bar' => 100, 'baz' => true })
f.set('details.bar' => 100)
f.set('details.baz' => false)
```
Neither `Numeric` nor `Boolean` respond to `#empty?`

lib/mongoid/persistable/settable.rb
```diff
@@ -25,15 +25,15 @@ def set(setters)
 
             field_and_value_hash = hasherizer(field.split('.'), value)
             field = field_and_value_hash.keys.first.to_s
+            value = field_and_value_hash[field]
 
-            if fields[field] && fields[field].type == Hash && attributes.key?(field) && !value.empty?
+            if fields[field] && fields[field].type == Hash && attributes.key?(field) && Hash === value && !value.empty?
```
p-mongo: Do you think `respond_to?(:empty?)` would make more sense here?

mdehoog: Not sure. The behavior in this PR seems to make sense: only merge an existing Hash with a new value if that value is a non-empty Hash. The PR that introduced the regression intended to support setting field values to an empty Hash, which remains supported.

gracezhangyaofei:
Hi may I know when is this PR going to be merged? This issue is quite serious though. Many thanks!

p-mongo:
@gracezhangyaofei I need to dig into the code that this PR changes as the change itself is non-obvious. Additionally neither https://docs.mongodb.com/mongoid/master/tutorials/mongoid-persistence/ nor https://docs.mongodb.com/mongoid/master/tutorials/mongoid-nested-attributes/ seem to mention anything that looks like `church.set('location.address.city' => 12345)`, I need to figure out where that is documented.

mdehoog:
Documentation about how `$set` should behave in embedded documents is here: https://docs.mongodb.com/manual/reference/operator/update/set/#set-fields-in-embedded-documents. Currently master will throw an error if `'zzz'` is changed to anything that doesn't implement `empty?`. The current spec missed this because Ruby `String` implements `empty?`.

The new spec should indicate what this PR is attempting to fix. AFAIK there's no existing behavior change here.

spec/mongoid/persistable/settable_spec.rb
```diff
@@ -315,6 +315,25 @@
...
+        let(:church) do
+          Church.new.tap do |a|
+            a.location = {'address' => {'city' => 'Berlin', 'street' => 'Yorckstr'}}
+            a.name = 'Church1'
+            a.save
+          end
+        end
+
+        before do
+          church.set('location.address.city' => 12345)
```
saghm: If I'm understanding correctly, this is where the exception would have occurred before the change? If so, let's put this inside a `let` block instead and invoke inside the `it` block below so that any future regression occurs in a place that's more easily identified.

saghm:
@p-mongo From what I can tell, this change seems to be correct, and I think we should update our documentation to describe this behavior. That being said, I think the documentation you linked to describes a separate feature from this, namely updating both sides of a relation at the same time (which unfortunately is not at all what the name seems to imply). I'm thinking that the best place for the new documentation would be in the ["Fields" section](https://docs.mongodb.com/mongoid/master/tutorials/mongoid-documents/#fields). Thoughts?

p-mongo:
I think the docs should be under persistence/#set, I'm going through the relevant bits in the tutorial to expand them.

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4579 for that, if you want to merge this PR as it is that's ok with me.

spec/mongoid/persistable/settable_spec.rb
mdehoog: Can do, but this way it's more consistent with all the examples around it. I think an exception raised anywhere in the spec is fine for failing a test; it doesn't necessarily need to be raised in the `it` block.

saghm: Fair enough; I don't feel strongly enough to block merging on this, so it's fine to leave as-is.

p-mongo:
@mdehoog Can you provide a reduced failing example that this PR is fixing?

p-mongo:
Or @gracezhangyaofei if you have an example because so far my examples are working.

p-mongo:
Docs are in https://github.com/mongodb/mongoid/pull/4514.

p-mongo:
I see, the bug affects 6.x only. This problem does not exist on master. Have you investigated what makes master work in this case?

mdehoog:
https://github.com/mongodb/mongoid/pull/4487 wasn't merged into master, so it might actually suffer from https://jira.mongodb.org/browse/MONGOID-4525.

p-mongo:
OK, then we need to figure out if 4525 affects master, ensure master has an equivalent test to the one added in this PR, then see if the fix can be backported from master or potentially master needs some version of #4487 and/or this PR.

mdehoog:
Yes the original issue affects master and has not been fixed; verified by running the new specs from #4487 on master. #4515 should fix master and includes this change too.

p-mongo:
After #4515 lands it needs to be cherry picked back into this PR.

mdehoog:
@saghm is it worthwhile getting this fix in to the 6.4.0 branch anyway? It is currently broken on this branch, and #4518 is a different fix that introduces a breaking change, so not sure if it's wise to backport.

saghm:
We don't currently have any bugfix release scheduled for the 6.4.x branch, although we don't have any opposition to doing one. Is this something you need, or would using the 7.0.x branch work for you?

mdehoog:
Sounds good 👍. We're not in a position to move to 7.0.x yet, but we'll put it on the roadmap.

saghm:
After some discussion, we've decided we do in fact want to do a patch release for 6.4, and this seems like a good candidate to be part of that. Unfortunately, it looks like I can't reopen this since the branch is closed. Any chance it's possible to get it pushed back up? (If not, that's totally fine! I can just manually apply the patch, which Github does seem to have cached).

mdehoog:
@saghm I restored the branch, but can't reopen the PR; do you have access to do this?

p-mongo:
Let's cherry-pick this diff on 7.0.0-stable as well.

gracezhangyaofei:
Hi @p-mongo , may I know when this fix can be merged? I saw your last msg was two weeks ago. Keep me posted, many thanks!

p-mongo:
@gracezhangyaofei We are configuring CI internally to get this PR appropriately tested before merging it.

saghm:
Hi @mdehoog! Sorry for the delay in merging this; we had to get our CI set up for the 6.4.0-stable branch, but it's all ready go to now! Is it all right with you if I push a rebase up to your branch with a few backports we made to the 6.4.0 branch to get the CI passing? (Github suggests that I can "add more commits by pushing" to your branch, but I'm not sure about a force push after a rebase, and I don't want to force push to your repo without getting permission first anyhow)

mdehoog:
@saghm completely fine, go for it 👍

saghm:
Just merged this. Thanks again for the contribution!

PR #4511, opened 2018-07-23
Add Ruby 2.5 & rails to support matrix

p-mongo:


docs/tutorials/mongoid-installation.txt
```diff
@@ -32,28 +32,28 @@ Compatibility
...
 Note the following compatibility matrix to determine if Mongoid is
-compatible your Ruby version and MongoDB server. Note that Mongoid
+compatible with your Ruby version and MongoDB server. Note that Mongoid
 is currently compatible with Rails 5 and depends on ActiveModel 5.1 or higher,
 and thus requires Ruby 2.2.2 or higher.
 
 .. list-table::
    :header-rows: 1
 
    * - Ruby Version
```
saghm: Given that each of the columns have the exact same value throughout, I have to wonder if a table is actually the best format to convey this information. Why not just say "Mongoid requires MRI >= 2.2.0 or JRuby 9.1.x"?

PR #4512, opened 2018-07-27
 Test on jruby via chruby toolchain

p-mongo:


saghm:
Is this still blocked? It looks like the ticket you linked to is merged now

p-mongo:
jruby testing vs driver master is busted due to https://github.com/bundler/bundler/issues/6678.

p-mongo:
On top of #4527 and #4531

PR #4515, opened 2018-08-02
MONGOID-4525 Fix ability to reset Hash field to empty Hash (on master)

mdehoog:
Combination of #4487 / #4510: forward-port to master.

### DIFF
--- a/lib/mongoid/persistable/settable.rb
+++ b/lib/mongoid/persistable/settable.rb
@@ -51,16 +51,38 @@ module Mongoid
         prepare_atomic_operation do |ops|
           process_atomic_operations(setters) do |field, value|
 
-            field_and_value_hash = hasherizer(field.split('.'), value)
-            field = field_and_value_hash.keys.first.to_s
-            value = field_and_value_hash[field]
+            field_seq = field.to_s.split('.')
+            field = field_seq.shift
+            if field_seq.length > 0
+              # nested hash path
+              old_value = attributes[field]
 
-            if fields[field] && fields[field].type == Hash && attributes.key?(field) && Hash === value && !value.empty?
-              merger = proc { |key, v1, v2| Hash === v1 && Hash === v2 ? v1.merge(v2, &merger) : v2 }
-              value = (attributes[field] || {}).merge(value, &merger)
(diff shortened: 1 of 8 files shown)


p-mongo:
Thank you for this.

mdehoog:
@johnnyshields I don't believe we do. The original bug fix was to allow setting a value to an empty hash; the root cause was due to the special handling of hashes such that keys are merged. This special handling doesn't apply to arrays, and setting empty arrays should work fine.

p-mongo:
Adding additional test cases, one problem with this implementation is it merges leaf hash values instead of writing them as is (which is what mongo shell does).

I.e. if I have `{'a' => {'b' => 'c'}` and I set `'a' => {'d' => 'e'}`, the result should be `{'a' => {'d' => 'e'}}` but it currently is `{'a' => {'b' => 'c', 'd' => 'e'}}`. (To achieve the merge on the leaf level the operation should be to set `'a.d' => 'e'`.)

mdehoog:
@p-mongo your new tests fail on master also. Perhaps it would be better to track this bug in a separate PR and issue, rather than conflating two fixes into one?

p-mongo:
Should be fixed now. Do you want to make that ticket?

lib/mongoid/persistable/settable.rb
```diff
@@ -23,17 +51,39 @@ def set(setters)
...
-              process_attribute(field.to_s, value)
-            else
-              process_attribute(field.to_s, field_and_value_hash[field])
+              # now we are on the leaf level, perform the set
+              # and overwrite whatever was on this level before
+              cur_value[field_seq.shift] = value
+
+              # and set value to the value of the top level field
+              # because this is what we pass to $set
+              value = old_value
```
mdehoog: now that the bulk of this `set` method consists of logic handling nested hashing, perhaps section should be pulled out into a separate method specifically for merging hashes?

mdehoog:
I went ahead and opened an issue here: https://jira.mongodb.org/browse/MONGOID-4593.

Closing in favor of #4518, which contains the same commits and is specifically a fix for the above issue.

PR #4517, opened 2018-08-10
Add frozen_string_literal pragma

oniofchaos:
This will prevent duplicate strings from getting allocated


oniofchaos:
What does that error mean and how do I resolve it?

p-mongo:
Thank you for the patch @oniofchaos .

PR #4518, opened 2018-08-11
MONGOID-4593 Modify Settable#set hash merge behavior to match $set

mdehoog:
_Breaking change: changes hash merge behavior which has the potential to delete old data._

This PR changes the `Mongoid::Persistable::Settable#set` behavior to match `$set` in MongoDB. It also fixes a bug where `set` cannot be used to set a field to an empty hash.

If we have the following model:
```ruby
class Church
  include Mongoid::Document
  field :location, type: Hash
end
```
and an instance with the value:
```ruby
church = Church.new(location: { 'b' => 'c' })
```
and we call:
```ruby
church.set(location: { 'd' => 'e' })
```

The current behavior merges the hash with the new value:
```ruby
church.location => {"b"=>"c", "d"=>"e"}
```

This PR changes this behavior such that the hash is set to:
```ruby
church.location => {"d"=>"e"}
```

Fixes https://jira.mongodb.org/browse/MONGOID-4593.
Also fixes https://jira.mongodb.org/browse/MONGOID-4525 which hasn't been fixed on master yet.

Overrides #4487 / #4510 / #4515.

### DIFF
--- a/spec/config/mongoid.yml
+++ b/spec/config/mongoid.yml
@@ -3,7 +3,9 @@ test:
     default:
       database: mongoid_test
       hosts:
-        - <%=ENV["MONGOID_SPEC_HOST"]%>:<%=ENV["MONGOID_SPEC_PORT"]%>
+        <% SpecConfig.instance.addresses.each do |address| %>
+          - <%= address %>
+        <% end %>
       options:
         user: "mongoid-user"
         password: "password"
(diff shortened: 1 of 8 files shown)


spec/config/mongoid.yml
```diff
@@ -3,7 +3,9 @@ test:
     default:
       database: mongoid_test
       hosts:
-        - <%=ENV["MONGOID_SPEC_HOST"]%>:<%=ENV["MONGOID_SPEC_PORT"]%>
+        <% SpecConfig.instance.addresses.each do |address| %>
```
saghm: Given that these changes don't seem related, I think a rebase might be necessary?

mdehoog: Agreed 👍. These were not my changes; will revert all the unrelated test environment changes.

PR #4519, opened 2018-08-14
Forking guidance

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4556


docs/tutorials/mongoid-configuration.txt
```diff
@@ -0,0 +1,304 @@
...
+
+.. code-block:: yaml
+
+  development:
+    # Configure available database clients. (required)
+    clients:
+      # Define the default client. (required)
+      default:
+        # A uri may be defined for a client:
+        # uri: 'mongodb://user:password@myhost1.mydomain.com:27017/my_db'
```
saghm: If I recall correctly, a configuration can't have both a URI field and host/options fields. We should document this.

saghm: If I recall correctly, a configuration can't have both a URI field and host/options fields. We should document this.

p-mongo: Feel free to make a ticket about that, this is not in scope of this PR.

PR #4521, opened 2018-08-15
MONGOID-4544 Define #touch on all documents

saghm:


lib/mongoid/touchable.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
```
saghm: This file is identical to the old `lib/mongoid/association/touchable.rb` (sans the `Association` module)

p-mongo: Then why do we have these two identical files in the tree?

spec/mongoid/touchable_spec.rb
```diff
@@ -2,14 +2,31 @@
 
 require "spec_helper"
 
-describe Mongoid::Association::Touchable do
+describe Mongoid::Touchable do
 
   describe "#touch" do
 
+    context "when the document has no relations" do
```
p-mongo: This test invokes `touch` on the document. Whether the document has any *associations* (not relations) doesn't matter.

saghm: I disagree that it doesn't matter; the entire purpose of this ticket is to add the ability to call `touch` on documents without associations, and that's what this is testing.

PR #4522, opened 2018-08-17
Make localized strings able to fallback on :blank? values

asgerb:
I've found that it would be very helpful to fallback through `:blank?` values when working with localized string fields.

This patch makes it so that:
* we fallback on all `:blank?` values
* and only to `:present?` values

Thanks for considering this patch!

### DIFF
--- a/lib/mongoid/fields/localized.rb
+++ b/lib/mongoid/fields/localized.rb
@@ -78,14 +78,28 @@ module Mongoid
       def lookup(object)
         locale = ::I18n.locale
 
-        value = if object.key?(locale.to_s)
+        value = localized_value(object, locale)
+
+        return value if [true, false].include?(value)
+        return value unless value.blank?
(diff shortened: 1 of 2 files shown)


p-mongo:
What use case/problem is this patch attempting to solve please?

asgerb:
Currently localized fields only fall back through locales on nil values. This patch make string fields fall back on empty strings (`""`) as well. I've added a test for it, which should hopefully demonstrate the behaviour.

p-mongo:
I think the documentation is pretty clear:

> When the fallbacks are defined, if a translation is not present, Mongoid will fallback in order of the defined locales.

nil indicates lack of value. An empty string is a value which is present.

`blank?`/`present?` are often used when working with user input where a lack of input will submit an empty string via the browser, but this logic doesn't apply to translations which are internal to the application. If you wish a translation to not be defined please set it to nil.

PR #4523, opened 2018-08-17
Add support for $nor to the Mongoid matcher.

jonhyman:
Lets you use the `$nor` comparison operator for local comparisons: https://docs.mongodb.com/manual/reference/operator/query/nor/

If this is accepted, it would be great to backport to at least the 6.4 branch, too.

### DIFF
--- a/lib/mongoid/matchable/nor.rb
+++ b/lib/mongoid/matchable/nor.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 # encoding: utf-8
 module Mongoid
   module Matchable
--- a/lib/mongoid/matchable/nor.rb
+++ b/lib/mongoid/matchable/nor.rb
@@ -1,0 +1,33 @@
+# encoding: utf-8
+module Mongoid
+  module Matchable
+
+    # Defines behavior for handling $nor expressions in embedded documents.
+    class Nor < Default
+
+      # Does the supplied query match the attribute?
+      #
+      # @example Does this match?
+      #   matcher._matches?("$nor" => [ { field => value } ])
+      #
+      # @param [ Array ] conditions The or expression.
+      #
+      # @return [ true, false ] True if matches, false if not.
+      #
+      # @since 2.0.0.rc.7
+      def _matches?(conditions)
+        conditions.each do |condition|
+          res = true
+          condition.keys.each do |k|
+            key = k
+            value = condition[k]
+            res &&= !document._matches?(key => value)
+            break unless res
+          end
+          return res if res
+        end
+        return false
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


p-mongo:
@jonhyman Thank you for the patch. I changed the empty criteria to not match which I believe is more correct, and reworked the test cases.

spec/mongoid/matchable/nor_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+            expect(matches).to be true
+          end
+        end
+
+        context 'and model has identical value in one of the fields' do
+          let(:target) do
+            Person.new(title: 'Queen', age: 23)
+          end
+
+          it "returns false" do
```
saghm: I don't think this is correct (and by extension, I think the implementation of `_matches?`) is incorrect. The method should return true if none of the queries in the array are true, which is the case since neither `{ title: 'King', age: 100 }` or `{ title: 'Sir', age: 23 }` matches `{ title: 'Queen', age: 100 }`. You can try this out on the shell with the following code and see that the document is actually returned:

```javascript
db.foo.insert({ title: 'Queen', age: 23 })
db.foo.find({ '$nor': [
    { title: 'Sir', age: 23 },
    { title: 'King', age: 100 },
]})
```

See my comment above for how I think `_matches?` should be implemented to fix this.

lib/mongoid/matchable/nor.rb
```diff
@@ -0,0 +1,42 @@
...
+      #
+      # @since 7.1.0
+      def _matches?(conditions)
+        if conditions.length == 0
+          # MongoDB does not allow $nor array to be empty, but
+          # Mongoid accepts an empty array for $or which MongoDB also
+          # prohibits
+          return false
+        end
+        conditions.each do |condition|
```
saghm: I don't think this is correct; I believe it should be this (I also switched the `each`/`returns` to just use array methods that return booleans:

```ruby
conditions.none? do |condition|
  conditions.all? do |key, value|
    document._matches?(key => value)
  end
end
```

See my comment in the tests below for where the current implementation gives the incorrect result.

spec/mongoid/matchable/nor_spec.rb
p-mongo: Right, good catch.

jonhyman: Good find. @p-mongo are you taking on @saghm's comments or should I slot time to make changes?

p-mongo: I fixed it.

p-mongo:
@jonhyman We will backport to 6.4 and 7.0.

PR #4526, opened 2018-08-21
MONGOID-4543 enable autosave for accepts_nested_attributes_for

saghm:


lib/mongoid/attributes/nested.rb
```diff
@@ -45,13 +45,16 @@ module ClassMethods
         # @option *args [ true, false ] :update_only Only update existing docs.
         def accepts_nested_attributes_for(*args)
           options = args.extract_options!
+          options[:autosave] = true if options[:autosave].nil?
```
p-mongo: Do not mutate `options`, and please document `:autosave`.

lib/mongoid/attributes/nested.rb
```diff
@@ -45,13 +45,16 @@ module ClassMethods
...
+
           options[:reject_if] = REJECT_ALL_BLANK_PROC if options[:reject_if] == :all_blank
           args.each do |name|
             meth = "#{name}_attributes="
             self.nested_attributes["#{name}_attributes"] = meth
             association = relations[name.to_s]
             raise Errors::NestedAttributesMetadataNotFound.new(self, name) unless association
-            autosave_nested_attributes(association)
+            autosave_nested_attributes(association) unless !options[:autosave]
```
p-mongo: Please remove the double negative.

lib/mongoid/attributes/nested.rb
```diff
@@ -76,7 +79,8 @@ def accepts_nested_attributes_for(*args)
         #
         # @since 3.1.4
         def autosave_nested_attributes(association)
-          if association.autosave?
+          if association.autosave? || (association.options[:autosave].nil? && !association.embedded?)
+            association.options[:autosave] = true
```
p-mongo: What is the purpose of this assignment?

spec/mongoid/attributes/nested_spec.rb
```diff
@@ -204,7 +215,12 @@
     context "when the relation is a referenced in" do
 
       before do
-        Post.accepts_nested_attributes_for :person
+        Post.accepts_nested_attributes_for :person, autosave: false
+      end
+
+      after do
+        Post.send(:undef_method, :person_attributes=)
+        Post.nested_attributes.clear
```
p-mongo: Please do not mutate models like this. The test suite defines a ton of models, if there isn't a suitable one in it already please add another one but a model's definition should not be mutated in the test suite like this.

saghm: The test suite was already mutating the model (see right above). This is just setting the model back to the way it was before.

lib/mongoid/attributes/nested.rb
saghm: In order for the autosave functionality to work properly, the association needs to be marked as autosave despite the fact that the option isn't present. Because the method `Association#autosave?` (which is used elsewhere in the codebase) is implemented by checking the autosave option, this is the most straightforward way to mark it.

spec/mongoid/attributes/nested_spec.rb
p-mongo: It's bad there too.

lib/mongoid/attributes/nested.rb
p-mongo: Thanks, can you put this entire comment in the source then?

lib/mongoid/attributes/nested.rb
```diff
@@ -44,14 +44,17 @@ module ClassMethods
         # @option *args [ Integer ] :limit The max number to create.
         # @option *args [ true, false ] :update_only Only update existing docs.
         def accepts_nested_attributes_for(*args)
-          options = args.extract_options!
+          options = args.extract_options!.dup
+          options[:autosave] = true if options[:autosave].nil?
```
p-mongo: The autosave option is still not documented.

p-mongo:
I am not asking to refactor the test suite, just don't add to tech debt.

saghm:
As I previously stated, the test case is _already_ modifying the model class; all my change is doing is resetting that change back to the way the model was before the change. This is necessary because after enabling autosave, the change that was already made to the model prior to this pull request breaks one of the sessions tests. Given this, the only options are a) not fixing this bug at all, b) leaving the sessions test broken, c) removing either the sessions test or this test, or d) fixing this test to not permanently modify the model. Options a) and b) are not even worth considering, and option c) is inferior to option d) given that the test already was modifying the model, and fixing it entails resetting it back to the original state.

saghm:
The only thing that's changed in the before block is adding the `autosave` option; no new mutation have been added. I think that the changes that we should make to this test file should be done all at once and more well thought out than just adding extra models ad-hoc, as we already have a huge number of models used for testing; this makes it difficult to find which ones that already exist would apply to a given test, which in turn encourages creating even more models rather than finding existing suitable ones. I'd much rather we schedule work to go over the entire test suite and eliminate all the model mutations as well as attempting to trim down the large number of models that we have than just add a new model every time we make an unrelated change that happens to touch existing test code that uses this pattern.

spec/mongoid/attributes/nested_spec.rb
```diff
@@ -14,20 +14,31 @@
 
       before do
         Person.accepts_nested_attributes_for :favorites
+        Person.accepts_nested_attributes_for :children
```
p-mongo: This is one of the mutations added in this diff.

Instead of mutating `Person` here, add another model and configure it with `children` without autosave and put a note on top of it that it's owned by the `nested_spec` test and not to be used by anyone else. Add another model configured with `autosave: false` with the same note. Use those models in the tests being added in this PR.

saghm:
> I've looked through the models and finding suitable ones in there is indeed difficult, so don't worry about that and make a new one.

That's exactly what I don't want to happen here. We already have over 200 models defined globally for all tests, and I think that's a mistake. If we want models to be global, I think we should try to cut down on the number of them rather than increasing it. If we want "one per test", I don't think we should define them globally at all. I don't necessarily have a strong opinion on which of those two options is preferably, but I very much don't want to keep adding more and more models and then using them all over the place in the test suite; that seems a lot more like "tech debt" to me.

p-mongo:
You can add the models in the spec file rather than `spec/app/models` if you prefer.

spec/mongoid/attributes/nested_spec.rb
saghm: The model is already being mutated in the exact same way, just on a different field. Again, I don't want to add a bunch of random models ad-hoc every time we touch a test that uses a pattern like this. If we want to change this pattern, we should tackle it at once in a coherent unit of work rather than just piggybacking it piecemeal on top of unrelated tickets.

p-mongo: Global mutations fail the test suite in jruby as well as mri (less frequently) and fail when parallelized. They are a constant drain on engineer productivity because they cause failures in unrelated parts of the test suite, as you said yourself in this PR. I am happy to discuss other approaches to testing but global mutations are not good.

saghm: I understand that, and I agree. However, fixing these two individual mutations won't fix the parallel tests though given that it's done throughout this file (and presumably elsewhere in the test suite). That's why I think we should schedule work to fix all of them at once rather than just doing piecemeal work on tickets that aren't related.

lib/mongoid/attributes/nested.rb
```diff
@@ -43,15 +43,20 @@ module ClassMethods
         #   to a class method to reject documents with.
         # @option *args [ Integer ] :limit The max number to create.
         # @option *args [ true, false ] :update_only Only update existing docs.
+        # @options *args [ true, false ] :autosave Whether autosave should be enabled on the
+        #   association.
```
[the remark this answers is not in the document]
saghm: On second thought, I'll just change it so that the `true` option is only set if no key is present rather than if the value is nil.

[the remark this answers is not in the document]
p-mongo: I think that's worse.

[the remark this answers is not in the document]
saghm: I can revert the last commit if you prefer; let me know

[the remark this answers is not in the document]
p-mongo: My preference is to convert nil value to true as originally done.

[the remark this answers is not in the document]
saghm: Understood! Done.

saghm:
@p-mongo Is the current state of the autosave semantics acceptable to you?

p-mongo:
Yes, looks good.

PR #4533, opened 2018-08-27
MONGOID-4542 Fix NameError when polymorphic and touch are specified i…

saghm:
…n a relation

### DIFF
--- a/lib/mongoid/association/relatable.rb
+++ b/lib/mongoid/association/relatable.rb
@@ -323,6 +323,12 @@ module Mongoid
 
       private
 
+      # Gets the model classes with inverse associations of this model. This is used to determine
+      # the classes on the other end of polymorphic relations with models.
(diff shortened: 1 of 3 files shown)


lib/mongoid/association/relatable.rb
```diff
@@ -323,6 +323,12 @@ def validate?
 
       private
 
+      # Gets the model classes with inverse associations of this model. This is used to determine
+      # the classes on the other end of polymorphic relations with models.
```
p-mongo: `@api private` please.

saghm: Just to clarify, do you mean that this should be made public but marked `@api private`, or that I should just add that to the method where it is?

p-mongo: I missed the private access specifier, I suppose this is fine as it is then.

PR #4534, opened 2018-08-28
MONGOID-4549 Correctly track dependents locally to model

saghm:


p-mongo:
Deleting a dependent college user with a dependent report card should delete the report card.

Also since mongoid allows this:

<pre>
class DoubleAssoc
  include Mongoid::Document

  has_many :deps, dependent: :destroy
end

class DoubleAssocOne < DoubleAssoc
  has_many :deps, dependent: :nullify, inverse_of: :double_assoc
end

class DoubleAssocTwo < DoubleAssocOne
  has_many :deps, dependent: :destroy, inverse_of: :double_assoc
end
</pre>

There should be a test that destroying DoubleAssocOne destroys its deps and destroying a DoubleAssocTwo doesn't (on master destroying DoubleAssocOne destroys its deps, if *this* is desired behavior there should be an exception raised when attempting to redefine dependent behavior).

saghm:
> on master destroying DoubleAssocOne destroys its deps, if this is desired behavior there should be an exception raised when attempting to redefine dependent behavior

This is another instance of the same bug. I'll add the two test cases that you suggest.

saghm:
Actually, from looking it over again, I don't think it's the same bug. I believe as implemented on master, dependents are cumulative, so there's both a `destroy` and `nullify` dependent for `deps` on `DoubleAssocOne`. I agree with you that `DoubleAssocOne` should not destroy its dependents and `DoubleAssocTwo` should, so I'll fix that if the test shows it doesn't work like that.

saghm:
@p-mongo do the additional tests look good to you?

p-mongo:
I am not seeing a test that deleting a dependent college user with a dependent report card deletes their report card, is it in the diff?

saghm:
As written before, CollegeUser doesn't even have a  reference to ReportCard, so I added the transitivity test for Student. Can you elaborate on why the test needs to be on CollegeUser?

p-mongo:
The reason why `class_attribute` exists is because simple instance variable assignment on class level, which is what this diff proposes, often times does not behave correctly in the context of inheritance. Hence, inheritance should be explicitly tested.

saghm:
In what way does CollegeUser use inheritance that Student doesn't? They both inherit from the same superclass. I'm really not understanding why you think that testing CollegeUser is better than testing Student when they're literally identical except that Student has the property (a dependent report card) that we actually want to test.

saghm:
Per offline discussion, I added in tests to make sure that the dependent behavior was defined properly across inheritance boundaries. You were correct that subclasses did not inherit the dependents from a superclass, so I added in a manual call to `superclass.dependents` in the initialization of the dependents array. To fix the duplicate dependents defined for a subclass that overrides a given dependent, I changed the `define_dependency!` method to remove existing dependents that are associated with the same class before adding the new one.

lib/mongoid/association/depending.rb
```diff
@@ -10,7 +10,7 @@ module Depending
 
       included do
         def self.dependents
-          @dependents ||= []
+          @dependents ||= superclass.method_defined?(:dependents) ? superclass.dependents.dup : []
```
p-mongo: Wouldn't this then modify superclass' dependents if both super and subclass define dependents?

saghm: I don't think so? The actual method call is `superclass.dependents.dup`, so it isn't reusing the reference to the superclass's array.

saghm: Good point! Let me know what you think of the changes I just pushed

saghm:
Just pushed the changes we made in the pair-coding session

lib/mongoid/association/depending.rb
```diff
@@ -10,7 +10,15 @@ module Depending
 
       included do
         class_attribute :dependents
+        class_attribute :dependents_owner
```
p-mongo: Can this be marked `@api private` please.

lib/mongoid/association/depending.rb
```diff
@@ -10,7 +10,15 @@ module Depending
...
         self.dependents = []
+        self.dependents_owner = self
+      end
+
+      class_methods do
+        def _all_dependents
```
p-mongo: And this also `@api private`.

lib/mongoid/association/depending.rb
```diff
@@ -10,7 +10,15 @@ module Depending
...
+          dependents + (superclass.respond_to?(:_all_dependents) ? superclass._all_dependents : [])
```
p-mongo: This needs to deduplicate the list by association name.

saghm: Oh, good catch!

saghm:
Suggested changes have been made

PR #4536, opened 2018-09-04
MONGOID-4532 Correctly determine module path for associations

saghm:


saghm:
I've changed the word everywhere in the code that I've added (some of the code in the diff uses "relations" in existing method names which we can't change without breaking compatibility).

cesarizu:
This breaks existing apps running mongoid 7.0.1 or lower. See https://jira.mongodb.org/browse/MONGOID-4645

aparmar:
What is the status of this? Seems like associations of classes within a namespace are still broken.

p-mongo:
@aparmar  If you are having an issue with Mongoid please create a jira ticket with the following information:

- a short code snippet illustrating the problem
- what you expected to happen
- what actually happened
- your Mongoid version, Ruby MongoDB driver version, Ruby version, Rails/ActiveSupport version

aparmar:
done. See https://jira.mongodb.org/browse/MONGOID-4834.

PR #4547, opened 2018-09-11
 Don't capture NoMethodError when instantiating STI

malmckay:
NoMethodError is a child of NameError
If a class's instantiation throws a NoMethodError, the instantiation will fail with a Errors::UnknownModel error.
Change to only throw Errors::UnknownModel if the constantized class can not be found, or if the constantized class does not have an #instantiate method

### DIFF
--- a/spec/mongoid/factory_spec.rb
+++ b/spec/mongoid/factory_spec.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 require "spec_helper"
 
 describe Mongoid::Factory do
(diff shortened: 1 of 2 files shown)


spec/mongoid/factory_spec.rb
```diff
@@ -212,5 +210,29 @@
       end
 
     end
+
+    context 'when type is correct but the instantiation throws a NoMethodError' do
+      class BadPerson < Person
+        def self.instantiate(*args)
+          call_some_no_existant_method(*args)
```
p-mongo: existant -> existent

malmckay: Fixed

malmckay:
Oops, the frozen string removal was a copy/paste from the 7.0.1 release code. I've reverted.

PR #4553, opened 2018-09-20
MONGOID-4539 Fix behavior of create_with to match ActiveRecord

saghm:
I ran each of the operations against ActiveRecord to determine how it behaves and modified the tests to assert those results instead. This PR does change the results of a number of operations, but the API  remains the same, so it doesn't require a major version bump.

### DIFF
--- a/lib/mongoid/criteria/modifiable.rb
+++ b/lib/mongoid/criteria/modifiable.rb
@@ -5,6 +5,7 @@ module Mongoid
     module Modifiable
 
       # @attribute [r] create_attrs Additional attributes to add to the Document upon creation.
+      # @api private
       attr_reader :create_attrs
 
       # Build a document given the selector and return it.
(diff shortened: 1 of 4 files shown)


spec/mongoid/criteria/modifiable_spec.rb
```diff
@@ -1683,7 +1687,7 @@
               Person.create_with(attrs).find_or_create_by(query)
             end
 
-            it 'gives the write method args precedence' do
+            it 'gives the find method args precedence' do
```
p-mongo: Can we document `create_with`, possibly within `find_or_create_by` section in the tutorial, and note there that attributes in `find_or_create_by` override those in `create_with`?

lib/mongoid/criteria/modifiable.rb
```diff
@@ -4,6 +4,9 @@ module Mongoid
   class Criteria
     module Modifiable
 
+      # @attribute [r] create_attrs Additional attributes to add to the Document upon creation.
+      attr_reader :create_attrs
```
p-mongo: `# @api private` please.

saghm:
> I would also like to have an integration test which does create_with(...).find_or_create_by(...) and verifies that the correct record is found/created.

I believe that's already what [this](https://github.com/mongodb/mongoid/pull/4553/files#diff-e218443ca35e7bba82b151beb1ddab6dR1691) and [this](https://github.com/mongodb/mongoid/pull/4553/files#diff-e218443ca35e7bba82b151beb1ddab6dR1766) are doing

p-mongo:
:+1:

PR #4556, opened 2018-09-27
Support a dynamic collection name

reidmorrison:
Sometimes we want to create a model and then use that model for different Mongo collections, ideally chosen dynamically at run-time.

For Example, for a static one-to-one mapping between a model and its class:
~~~ruby
Band.create!(name: "Depeche Mode")
band = Band.where(name: "Depeche Mode").first
band.rating = 23
band.save!
~~~

Dynamically changing the collection name at run-time:
~~~ruby
Band.with(collection: 'other_bands').create!(name: "Depeche Mode")

band = Band.with(collection: 'other_bands').where(name: "Depeche Mode").first
band.rating = 23
band.save!
~~~

In Mongoid 5.4, the above kind of worked. It was extremely inefficient and led to all kinds of problems because every call to `with` resulted in a brand new persistence context. Additionally, the `save!` would not save the changed document back to the `other_bands` collection.

In Mongoid 6/7 it uses a block in order to cleanup the persistence context on completion:
~~~ruby
Band.with(collection: 'other_bands') do
  Band.create!(name: "Depeche Mode")
  band = Band.with(collection: 'other_bands').where(name: "Depeche Mode").first
  band.rating = 23
  band.save!
end
~~~

Even in the above case, creating a brand new persistence context when only the `collection_name` has changed is extremely inefficient, and it requires the constant use of the block anytime a modified document is saved.

How about we make `collection` a special case since it does not need a new persistence context every time the collection name is changed. And, if we make the instance of the model remember its collection name then we don't have to complicate any code that attempts to persist any changes to that document.

That would make the following code possible:

~~~ruby
Band.with(collection: 'other_bands').create!(name: "Depeche Mode")

band = Band.with(collection: 'other_bands').first

# Don't have to carry around separately the name of the collection
# that this instance of `band` comes from.
# And avoids yet another persistence context to save the change.
band.rating = 23
band.save!
~~~

Or even allow the collection name to be supplied directly when creating a new model:
~~~ruby
Band.create!(collection_name: 'other_bands', name: "Depeche Mode")
~~~

### DIFF
--- a/lib/mongoid/clients/options.rb
+++ b/lib/mongoid/clients/options.rb
@@ -21,6 +21,12 @@ module Mongoid
       #
       # @since 6.0.0
       def with(options_or_context, &block)
+        # Only changing the collection name does not require the overhead of an entire new persistence context.
+        if options_or_context.is_a?(Hash) && (options_or_context.size == 1) && options_or_context.key?(:collection)
+          self.collection_name = options_or_context[:collection]
+          return block_given? ? yield(self) : self
+        end
+
         original_cluster = persistence_context.cluster
         set_persistence_context(options_or_context)
         yield self
@@ -29,11 +35,15 @@ module Mongoid
       end
 
       def collection(parent = nil)
-        persistence_context.collection(parent)
+        @collection_name ? mongo_client[@collection_name] : persistence_context.collection(parent)
       end
 
       def collection_name
-        persistence_context.collection_name
+        @collection_name || persistence_context.collection_name
+      end
+
+      def collection_name=(collection_name)
+        @collection_name = collection_name.nil? ? nil : collection_name.to_sym
       end
 
       def mongo_client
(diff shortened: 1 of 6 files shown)


p-mongo:
Just to make sure I understand, does this patch propose the following behavior:

<pre>
class Band
  include Mongoid::Document
  store_in collection: "artists"
end

# anywhere in the code
Band.with(collection: 'poets')

# anywhere else in the code at a later time
# queries poets, not artists
Band.all
</pre>

reidmorrison:
Not at all, the persistence context is not modified when `.with` is called with just a collection_name.

`Band.with(collection: 'poets')` would return all Band objects within the `poets` collection.

Calling `Band.all` later on in the code would return all Band documents in the `bands` collection.

p-mongo:
OK, I am playing with this and I am not yet sold on the type of result of #with being argument-dependent:

<pre>
Band.with(collection: foo) # => Mongoid::Criteria

Band.with(collection: foo, database: bar) # => Error: no block given

Band.with(collection: foo, database: bar) { |c| } # => Band, c is also Band
</pre>

p-mongo:
When you say that creating a persistence context is "extremely inefficient", can you describe your use case and performance you are getting in greater detail please?

p-mongo:
Can you elaborate on performance issues you are having with the existing implementation?

reidmorrison:
The purpose of this pull request has nothing to do with performance, it is to obtain the above functionality so as to remove the burden from developers to have to figure out which collection a document came from in order to modify it.

p-mongo:
Okay, then, will a non-block form of `#with` fulfill all of your requirements?

<pre>
# collection change only
band = Band.with(collection: 'other_bands').where(name: "Depeche Mode").first
band.rating = 23
band.save!

# db and collection change
band = Band.with(db: 'foo', collection: 'other_bands').where(name: "Depeche Mode").first
band.rating = 23
band.save!
</pre>

johnnyshields:
I agree we should add a non-block form of `#with` which returns a `Mongoid::Criteria` upon which queries can be chained.

p-mongo:
I created https://jira.mongodb.org/browse/MONGOID-4629 to track the work on a non-block form of #with.

reidmorrison:
Close this PR in favor of the above ticket

PR #4557, opened 2018-10-10
Replace relation with association in relations doc

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4584


p-mongo:
Yes that would be good.

PR #4558, opened 2018-10-10
Document elem_match

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4622

docs/tutorials/mongoid-queries.txt
```diff
@@ -42,6 +42,78 @@ is returned with the new query added. This is so that with scoping
...
+  end
+
+  dm = Band.create!(name: 'Depeche Mode')
+  aerosmith = Band.create!(name: 'Aerosmith')
+  Tour.create!(band: aerosmith, city: 'London', year: 1995)
+  Tour.create!(band: aerosmith, city: 'New York', year: 1999)
+
+  Band.elem_match(tours: {city: 'London'}).to_a # => [aerosmith]
+
+``elem_match`` does not work with non-embedded associations because MongoDB
```
saghm: Out of curiosity, do you think this is something we could implement with the `$lookup` aggregation stage? It seems to me like it might be doable.

p-mongo: Converting a collection criteria to an aggregation pipeline is probably not a viable strategy, but I totally support documenting how to accomplish the equivalent operation with aggregation pipeline.

PR #4562, opened 2018-10-26
MONGOID-4652 Allow atomic persistence context joining

brennana:
I have noticed some peculiar behavior when using nested calls to `Mongoid::Persistable#atomically`. Depending on the order and depth of other nested `atomically` blocks, mongoid may send separate operations to the database for different parts of the total described operation (not quite atomic anymore!).

For example (taking the `Band` class from the specs):
```ruby
class Band
  def my_updates
    atomically do |d|
      d.set(name: "Placebo")
      d.unset(:origin)
    end
  end
end

document = Band.create(member_count: 0, likes: 60, origin: "London")

# The following will make two calls to `update_one` on `Mongo::Collection::View` since the
# `my_updates` call will persist changes immediately after exiting its call to `atomically`,
# and so with the extra operations added after `my_updates`, exiting the outer `atomically`
# block will perform one more `update_one` on this document for the remaining operations.
document.atomically do |doc|
  doc.my_updates
  doc.inc(member_count: 10)
  doc.bit(likes: { and: 13 })
end

# The following will only make a single call to `update_one` due to the ordering of calls.
document.atomically do |doc|
  doc.inc(member_count: 10)
  doc.bit(likes: { and: 13 })
  doc.my_updates
end
```

I have fixed this behavior, and have written specs accordingly. `atomically` now tracks its call depth with a local variable on the document. Now, only the original call to `atomically` will persist the accumulated operations, and any nested calls will skip attempting to persist.

### DIFF
--- a/docs/tutorials/mongoid-persistence.txt
+++ b/docs/tutorials/mongoid-persistence.txt
@@ -508,7 +508,7 @@ Mongoid models are reverted. For example:
 Atomic operations described in this section apply to one document at a time,
 therefore nesting ``#atomically`` blocks invoked on multiple documents does
 not make changes to the different documents be persisted atomically together.
-However, MongoDB offers `multi-document transactions <transactions>`_
+However, MongoDB offers :ref:`multi-document transactions <transactions>`
 as of server version 4.0 which provide atomic persistence across multiple
 documents.
 


p-mongo:
@brennana Thanks for the patch. The problem with this change is that it changes behavior in the following case:

<pre>
document.atomically do |doc|
  doc.my_updates
  raise "oops"
  doc.inc(member_count: 10)
  doc.bit(likes: { and: 13 })
end
</pre>

We had a similar discussion with regard to logical transaction nesting in Mongoid not too long ago. If you have a preference as to how Mongoid should handle the case of an exception being raised in the middle of a set of atomic operations, or transactions, that would be helpful.

brennana:
@p-mongo Personally, I should think that unhandled exceptions in a set of atomic operations (nested or otherwise) should stop the entire operation/transaction, requiring the developer to handle the exception.

Since the operations described/generated in a set of atomically blocks could be dependent on one another, and given the expectation of `atomically` to be all-or-nothing, indeed I think it's reasonable to expect an unhandled exception at any level in a set of atomic operations to stop the entire set (and perhaps roll back in-memory changes to the document as the operations were accumulated).

brennana:
That seems reasonable to preserve the current behavior, though what about also adding some sort of Mongoid configuration variable to set the default of the `join_context` option? I would prefer not having to pass an explicit option to get this behavior, especially if nested operations do not know that they will be nested.

p-mongo:
Yes we can add a global configuration option.

brennana:
I'm curious about another case:

```ruby
document.atomically do |doc|
  doc.set origin: "Paris"
  doc.atomically(join_context: false) do |doc2|
    doc2.inc(member_count: 10)
  end
  doc.inc likes: 1
  raise 'oops'
  # member_count: 10 change is persisted
  # likes: 1 change is not persisted
  # is the origin: "Paris" change persisted?
end
```

Under the original behavior, the `origin` change above would be persisted as well, although it is not part of the "context" of the nested `atomically` block which opted to not join the outer context. Should this be preserved/optioned as well?

If we are trying to perform nested atomic ops in a separate context, it seems that although the inner context would persist independently of the outer context, it would not be very atomic if part of the changes of the outer context were persisted along with the inner, regardless of whether or not the outer context fails (the original behavior would always perform the above `origin` set as part of the nested `atomically` call, rather than the other operations declared in the outer block).

p-mongo:
Paris should not be persisted in that scenario, contexts shouldn't be leaking into one another.

brennana:
I have created the `join_context` option as you described, along with the appropriate specs and documentation. I have also created a global configuration option `atomically_join_context_default`, default false, which sets the default value for this option.

In addition, I wrote specs to ensure that upon exceptions, `atomically` will roll back the failed context's in-memory changes. However, I found that some of the `Mongoid::Persistable` atomic methods did not generate dirty changes for resetting. I wrote specs for all of `Mongoid::Persistable`'s atomic operators to ensure this behavior, and the relevant operations will (only when executing atomically) generate dirty changes via `process_attribute`, which can be reset on failures.

New private methods were added to `Mongoid::Persistable`, and have been documented, but with a blank version doc attribute, since I do not know what version these will be released in.

Naturally, let me know if anything needs tweaking.

lib/mongoid/config.rb
```diff
@@ -25,6 +25,7 @@ module Config
     option :duplicate_fields_exception, default: false
     option :use_activesupport_time_zone, default: true
     option :use_utc, default: false
+    option :atomically_join_context_default, default: false
```
p-mongo: Can this be renamed to just `join_contexts` please so that the same setting will be usable for transactions.

lib/mongoid/persistable.rb
```diff
@@ -44,24 +44,80 @@ module Persistable
...
     # @since 4.0.0
-    def atomically
-      begin
-        @atomic_updates_to_execute = @atomic_updates_to_execute || {}
-        yield(self) if block_given?
-        persist_atomic_operations(@atomic_updates_to_execute)
-        true
-      ensure
-        @atomic_updates_to_execute = nil
+    def atomically(join_context: Mongoid.atomically_join_context_default)
```
p-mongo: The default here should be nil, further defaulted to `Mongoid.atomically_join_context_default` within the method body.

lib/mongoid/persistable.rb
```diff
@@ -44,24 +44,80 @@ module Persistable
...
+        @atomically_depth -= 1
+      end
+
+      if has_own_context
+        persist_atomic_operations @atomically_context
+        remove_atomically_context_changes
+      end
+
+      true
+    rescue StandardError => e
```
p-mongo: This should be `Exception` I think. `StandardError` is more for rescues that handle exceptions, but when we clean up we should do so for all errors.

spec/mongoid/persistable_spec.rb
```diff
@@ -175,24 +175,96 @@
...
+        context "when given join_context: false" do
+
+          let(:run_update) do
+            document.atomically do |doc|
+              doc.set origin: "Paris"
+              doc.atomically(join_context: false) do |doc2|
+                doc2.inc(member_count: 10)
+              end
+              doc.inc likes: 1
+              raise "oops"
```
p-mongo: Can you please make an exception class for this like in https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/integration/heartbeat_events_spec.rb#L26 and rescue this class instead of `RuntimeError` below.

p-mongo:
Thank you, I have this PR queued for review.

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4652

brennana:
Of course! Glad I could help.

lib/mongoid/persistable.rb
```diff
@@ -165,10 +223,70 @@ def process_atomic_operations(operations)
...
         yield(access, value)
-        remove_change(access)
+        remove_change(access) unless executing_atomically?
       end
     end
 
+    # Remove the dirty changes for all fields changed in the current atomic
+    # context.
+    #
+    # @api private
```
saghm: Is there any reason these can't just be private methods instead of documented as `@api private` and then prefixed with `_mongoid_`? They don't seem to be called anywhere except internally.

p-mongo: These methods are already private, however since they are included in all model instances there could be confusion as to whether a model is allowed to call these methods on itself as part of our public API. The private api declaration explicitly says these methods are not part of public API.

PR #4566, opened 2018-11-04
    MONGOID-3795 Add a test of callback invocation order for cascading and non-cascading callbacks

p-mongo:


johnnyshields:
@p-mongo how can we get this fix merged?

p-mongo:
This PR is incomplete. I believe there is a decent amount of work left, and specifically the ActiveRecord callback invocation hierarchy seems to not be directly usable in Mongoid (which is what presently happens) due to Mongoid having embedded associations, for which the order of callback invocation needs to be different from the order were those associations referenced.

PR #4568, opened 2018-11-16
docs: socket_timeout default is nil, not 5

pda:
Update the Mongoid Configuration tutorial to show the default for `socket_timeout` is `nil` not `5`.

From the mongo driver documentation:

> *socket_timeout*
> The number of seconds to wait for an operation to execute on a socket before raising a timeout exception. It should take into account network latency and operation duration. The default is no value; the default is effectively infinity. Please consider using max_time_ms per-operation instead, as the socket_timeout does not stop the operation on the server; a long-running operation will continue to run on the server, beyond a socket timeout being reached.

— https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-create-client/index.html#details-on-timeout-options

Nothing in the driver code appears to set/default `socket_timeout` to anything, so it remains `nil`.

The only place mongoid arguably defaults `socket_timeout` to `5` is in `lib/rails/generators/mongoid/config/templates/mongoid.yml`, however that's a commented out example config line, not a default.

Note, this is also shown as `default: 5` on https://docs.mongodb.com/mongoid/7.0/tutorials/mongoid-installation/#anatomy-of-a-mongoid-config — I'm not sure where that page is built from.

### DIFF
--- a/docs/tutorials/mongoid-configuration.txt
+++ b/docs/tutorials/mongoid-configuration.txt
@@ -153,7 +153,7 @@ various options in the comments above each key:
           connect_timeout: 10
 
           # The timeout to wait to execute operations on a socket before raising an error.
-          # (default: 5)
+          # (default: nil)
           socket_timeout: 5
 
           # The name of the replica set to connect to. Servers provided as seeds that do


pda:
I just noticed https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-create-client/index.html#ruby-options lists `socket_timeout` as default `5` (the HTML layout obscures the default column without some scrolling). That contradicts the “Details on Timeout Options” section, and I can't see the default of 5 implemented anywhere. So I still think it's `nil` by default.

p-mongo:
Thanks for the patch. I think having some timeout is better than having no timeout, to which end I opened https://jira.mongodb.org/browse/RUBY-1593 to set our default timeout to 10 on non-monitoring connections.

p-mongo:
(The tutorial page you mentioned is built from https://github.com/mongodb/mongoid/blob/master/docs/tutorials/mongoid-configuration.txt, driver also has its own tutorials in analogous location.)

johnnyshields:
@p-mongo unrelated but how about taking some time to get the CI to behave reliably? Currently everything is red.

p-mongo:
I will merge this PR since it improves the accuracy of our present documentation and has no issues. Thanks!

PR #4572, opened 2018-11-28
MONGOID-4666 Allow fields of type Regexp to be set to nil.

jakebarnett:
Currently, setting a field of type regexp to nil results in the Ruby type error: `TypeError: no implicit conversion of nil into String` 

This PR nil guards the regexp mongoize extension similar to with the Time extension (mongoid/lib/mongoid/extensions/time.rb:72)

Please let me know if this should be submitted differently.

### DIFF
--- a/lib/mongoid/extensions/regexp.rb
+++ b/lib/mongoid/extensions/regexp.rb
@@ -18,6 +18,7 @@ module Mongoid
         #
         # @since 3.0.0
         def mongoize(object)
+          return nil if object.nil?
           ::Regexp.new(object)
         end
       end
--- a/lib/mongoid/extensions/time.rb
+++ b/lib/mongoid/extensions/time.rb
@@ -49,7 +49,7 @@ module Mongoid
         #
         # @since 3.0.0
         def demongoize(object)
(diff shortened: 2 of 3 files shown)


lib/mongoid/extensions/regexp.rb
```diff
@@ -18,6 +18,7 @@ module ClassMethods
         #
         # @since 3.0.0
         def mongoize(object)
+          return nil if object.blank?
```
p-mongo: This should be `nil?` rather than `blank?` (and this applies to the time extension as well), and can you please add a test that `''` mongoizes to `//` as it currently does?

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4666

jakebarnett:
@p-mongo Another thing I noticed is that there is not demongoize extension for `Regexp`.  Accessing a field of type `Regexp` will return a `BSON::Regexp::Raw`

Was that a design decision?  If not, I'm happy to fix that too.

p-mongo:
Seems like an oversight, a PR fixing that would be great.

PR #4573, opened 2018-11-28
Fix MONGOID-4460 Date.mongoize(String) should not convert string to local time first

p-mongo:


lib/mongoid/extensions/date.rb
```diff
@@ -68,7 +68,12 @@ def demongoize(object)
         def mongoize(object)
           unless object.blank?
             begin
-              time = object.__mongoize_time__
+              if object.is_a?(String)
+                # https://jira.mongodb.org/browse/MONGOID-4460
```
saghm: It might be useful to add a small description of the ticket as well to save future readers some trouble

PR #4596, opened 2019-01-27
Fix MONGOID-4624 scoped text searches wipe all previous criteria

p-mongo:


spec/mongoid/criteria/queryable/selectable_spec.rb
```diff
@@ -3385,16 +3385,34 @@ class TestModel
...
+      let(:selection) do
+        query.text_search("one").text_search('two')
+      end
+
+      # MongoDB server can only handle one text expression at a time,
+      # per https://docs.mongodb.com/manual/reference/operator/query/text/.
+      # Nonetheless we test that the query is built correctly when
+      # a user supplies more than one text condition.
+      it 'merges conditions' do
+        expect(selection.selector).to eq('$and' => [
```
saghm: Should we also test that the warning occurs?

p-mongo: I do not consider this undertaking worthwhile.

saghm: Why not? We do it in many other places in both the driver and mongoid, and it only requires an extra line or two of code

p-mongo: Because I am satisfied with the current implementation and would like to move on to other things. Feel free to file a ticket to test the warning if you find this worthwhile.

p-mongo:
Made the requested changes.

PR #4608, opened 2019-02-04
MONGOID-4524 Combine options when nesting PersistenceContexts on the same object

p-mongo:


lib/mongoid/persistence_context.rb
```diff
@@ -184,9 +187,19 @@ class << self
...
+          existing_context.options
+        else
+          {}
+        end
+        if options_or_context.is_a?(PersistenceContext)
+          options_or_context = options_or_context.options
+        end
+        new_options = existing_options.merge(options_or_context)
+        context = PersistenceContext.new(object, new_options)
+        Thread.current[key] = context
```
saghm: Out of curiosity, will this cause issues if there's more code after the inner block within the outer block? e.g.

```ruby
Band.with(collection: "artists") do |klass|
  Band.with(client: :reports) do |klass|
    klass.new.collection_name
  end

  klass.new.collection_name
end
```ruby

p-mongo: Yes, repaired this as well.

PR #4610, opened 2019-02-07
Fix MONGOID-2681 Using #only with embedded documents sets omitted fields to default values

p-mongo:


p-mongo:
Thinking about it some more, raising for referenced associations is a bad idea as it can break applications. Best to ignore those (https://jira.mongodb.org/browse/MONGOID-4704 is for making projections work through referenced associations) and I added test coverage for good measure.

PR #4614, opened 2019-02-13
MONGOID-4697 Implement Arel-style "and", "or" and "not" combination of scopes


p-mongo:


docs/tutorials/mongoid-queries.txt
```diff
@@ -198,6 +198,93 @@ loaded with ``only`` for those associations to be loaded. For example:
...
+criteria, unless the receiving criteria already has a condition on the
+respective fields, in which case the conditions will be combined with ``$and``.
+
+.. code-block:: ruby
+
+  Band.where(label: 'Trust in Trance').and(name: 'Astral Projection').selector
+  # => {"label"=>"Trust in Trance Records", "name"=>"Astral Projection"}
+
+  Band.where(name: /Best/).and(name: 'Astral Projection').selector
+  # => {"name"=>/Best/, "$and"=>[{"name"=>"Astral Projection"}]}
```
johnnyshields: Why doesn't this do:
```
 {"$and"=>[{"name"=>/Best/}, {"name"=>Astral Projection"}]}
```
See: https://docs.mongodb.com/manual/reference/operator/query/and/

p-mongo: I tried this approach first but it runs into practical complications when chaining more than one time. The current behavior is:

<pre>
irb(main):004:0> Foo.where(bar1: 1).where(bar2: 2).where(bar3: 3)
=> #<Mongoid::Criteria
  selector: {"bar1"=>1, "bar2"=>2, "bar3"=>3}
  options:  {}
  class:    Foo
  embedded: false>
</pre>

Using `$and` between conditions would alter the resulting query for existing applications.

The approach in this PR produces the same behavior for different keys, and adds:

<pre>
irb(main):004:0> Foo.where(bar: 1).where(bar: 2).where(bar: 3)
=> #<Mongoid::Criteria
  selector: {"$and"=>[{"$and"=>[{"bar"=>1}], "bar"=>2}], "bar"=>3}
  options:  {}
  class:    Foo
  embedded: false>
</pre>

johnnyshields: I think we should **not** do this:

```
Foo.where(bar: 1).where(bar: 2).where(bar: 3)
=> selector: {"$and"=>[{"$and"=>[{"bar"=>1}], "bar"=>2}], "bar"=>3}
```

p-mongo: Doing so requires inspecting the contents of conditions, whereas the current implementation can combine conditions without inspecting their contents.

p-mongo: I think it is a noble goal, and certainly one that can be discussed once all of the operators are handled and tests are written. In the interests of making forward progress I focused on attaining the desired behavior with respect to condition combination first and leaving the visuals of constructed queries for later.

johnnyshields: I would not want to accept a PR here that doesn't preserve visual cleanliness of operators.

p-mongo: I agree that, everything else being equal, a more visually appealing output is preferable, but can you please clarify whether you have a use case that depends on the shape of queries generated by Mongoid?

johnnyshields: Yes, absolutely, we have monitoring scripts which depend on the shape of certain queries in the Mongo logs.

In addition, ` selector: {"$and"=>[{"$and"=>[{"bar"=>1}], "bar"=>2}], "bar"=>3}` is very difficult for humans to read compared to `selector: {"bar1"=>1, "bar2"=>2, "bar3"=>3}`; it makes the whole library significantly harder to work with.

Lastly, have you confirmed that the two queries have identical performance, index behavior, etc in MongoDB itself?

johnnyshields: This is a fundamentally breaking change in my opinion which should not be allowed. It is CERTAINLY not worth the "nice-to-have" addition of AREL-style syntax.

p-mongo: I think you are confusing multiple conditions on the same fields and conditions on different fields, because as I mentioned in an earlier comment this PR preserves existing behavior when conditions on different fields are combined.

johnnyshields: Ah, ok I see. But in this case, why not flatten the `$and` into:

```
irb(main):004:0> Foo.where(bar: 1).where(bar: 2).where(bar: 3)
=> #<Mongoid::Criteria
  selector: {"$and"=>[{"bar"=>1}, {"bar"=>2}, {"bar"=>3}]}
```

p-mongo: This was my first thought as well, but again, once more complex queries are considered, plus the multitude of ways Mongoid permits the queries to be constructed, the choice of query shape becomes less obvious. The documentation attached to this PR will explain how the queries are produced.

p-mongo: The query shape differences between where and and will be addressed in https://jira.mongodb.org/browse/MONGOID-4806.

johnnyshields: I think it should be done as part of this PR. Shouldn't introduce new strange behavior.

docs/tutorials/mongoid-queries.txt
```diff
@@ -198,6 +198,195 @@ loaded with ``only`` for those associations to be loaded. For example:
...
+
+The following calls all produce the same query conditions:
+
+.. code-block:: ruby
+
+  # Condition hashes passed to separate and invocations
+  Band.and(name: 'SUN Project').and(member_count: 2)
+  
+  # Multiple condition hashes in the same and invocation
+  Band.and({name: 'SUN Project'}, {member_count: 2})
```
egiurleo: Correct me if I'm wrong here, but it seems like this example is the same as the one below, but with a slightly different syntax. Is that true? Is this also deprecated?

johnnyshields: Isn't the behavior of these two identical?

p-mongo: All of the invocations in this example produce the same result.

PR #4616, opened 2019-02-14
MONGOID-4681 Replace BigDecimal.new with BigDecimal()

HLFH:
To fix warning:

> BigDecimal.new is deprecated; use BigDecimal() method instead

Since:  ruby v2_6_1 & v2_6_0_rc2

Source: https://github.com/ruby/ruby/commit/62b55d9483d72a11345300296190b1e9c0ffd652#diff-cdf5d288e028f232ff6979d98276904a

### DIFF
--- a/lib/mongoid/extensions/big_decimal.rb
+++ b/lib/mongoid/extensions/big_decimal.rb
@@ -54,7 +54,7 @@ module Mongoid
         #
         # @since 3.0.0
         def demongoize(object)
-          object && object.numeric? ? ::BigDecimal.new(object.to_s) : nil
+          object && object.numeric? ? ::BigDecimal(object.to_s) : nil
         end
 
         # Mongoize an object of any type to how it's stored in the db as a String.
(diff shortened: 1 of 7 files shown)


HLFH:
This commit https://github.com/mongodb/mongoid/pull/4616/commits/d1ad6173064446152069f220be02d75e88dc3326 is fixing the remaining issue:

> lib/mongoid/extensions/big_decimal.rb:57: syntax error, unexpected '(', expecting ':'

cgriego:
Might we see this in a release so Mongoid can be used with Ruby 2.6?

p-mongo:
@cgriego BigDecimal.new still works in Ruby 2.6, it just produces a warning.

cgriego:
@p-mongo It functions, but when this is your app's test suite output, it isn't usable.
bash — 233×49 2019-04-29 14-36-22

jonekdahl:
@p-mongo I am in the same boat as @cgriego here, and would very much appreciate a 6.4.3 release of mongoid with this patch included, along with a release of bson-ruby that has similar BigDecimal fixes (see https://github.com/mongodb/bson-ruby/pull/119).

p-mongo:
We will have new releases coming out soon.

PR #4617, opened 2019-02-24
Update copyright notice to 2019

Atul9:


p-mongo:
Thank you, we will do https://github.com/mongodb/mongoid/pull/4620 instead.

PR #4618, opened 2019-02-25
Test against stable branch of the driver

p-mongo:


johnnyshields:
@p-mongo can you please get either Travis CI to be green, or else remove it entirely? It's not good form to have specs constantly failing on all the PRs you merge

PR #4621, opened 2019-04-01
MONGOID-4729 Skip query if all keys from docs are nil

mattruzicka:
Let's say we have a "seats" collection and seats can optionally belong to a user. If we query all the seats and eager load the users, but none of the seats actually belong to a user, Mongoid will still query the database for users where the id is ```{"$in"=>[nil]}```. I see something like this in the logs: ```{"find"=>"users", "filter"=>{"_id"=>{"$in"=>[nil]}}...```

This change will skip the trip to the db if all the foreign keys are nil.

I applied the change to the v7.0.2 release and it worked for me. On v7.0.2 with this change, all the specs passed.

### DIFF
--- a/lib/mongoid/association/referenced/eager.rb
+++ b/lib/mongoid/association/referenced/eager.rb
@@ -65,7 +65,10 @@ module Mongoid
           #
           # @since 4.0.0
           def each_loaded_document
-            criteria = @association.klass.any_in(key => keys_from_docs)
+            doc_keys = keys_from_docs
+            return @association.klass.none if doc_keys.all?(&:nil?)
+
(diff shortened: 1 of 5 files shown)


p-mongo:
Thank you for the PR. This change does need a test case before we can merge it.

mattruzicka:
@p-mongo I added a couple specs
```
rspec ./spec/mongoid/association/referenced/belongs_to/eager_spec.rb:177
rspec ./spec/mongoid/association/referenced/has_and_belongs_to_many/eager_spec.rb:147
```

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4729

p-mongo:
Thank you @mattruzicka !

PR #4622, opened 2019-04-05
MONGOID-4728 Adds Support for $eq operator in Matchable

mmwillet:
Currently Matchable doesn't support `#_matches?` against selectors that use the mongodb `$eq` operator. This adds support for it to do so.


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4728

p-mongo:
Thank you for the patch!

PR #4624, opened 2019-04-09
Document how to use Rake tasks in non-Rails applications

p-mongo:


docs/tutorials/mongoid-indexes.txt
```diff
@@ -4,132 +4,173 @@ Indexes
 
 .. default-domain:: mongodb
 
+Specifying Indexes
+------------------
+
 You can define indexes on documents using the index macro. Provide the key for
-the index along with a direction. For additional options, supply them in a second
-options hash parameter.
+the index along with a direction. Aadditional options can be supplied in the
```
saghm: Looks like there's a typo here; it should be "Additional"

PR #4632, opened 2019-05-16
Fix MONGOID-4733 Document#as_json with :compact option returns nil if none of the fields are nil


p-mongo:


lib/mongoid/document.rb
```diff
@@ -200,11 +200,11 @@ def as_document
     #
     # @since 5.1.0
     def as_json(options = nil)
-      if options && (options[:compact] == true)
-        super(options).reject! { |k,v| v.nil? }
-      else
-        super(options)
+      rv = super
+      if options && options[:compact]
+        rv = rv.reject { |k,v| v.nil? }
```
johnnyshields: You can call `rv.compact` here. See: https://api.rubyonrails.org/classes/Hash.html#method-i-compact

But since `#compact` is already method on `Hash`, why make it an option for `as_json`? I don't see it in the ActiveRecord `as_json` API: https://api.rubyonrails.org/classes/ActiveModel/Serializers/JSON.html#method-i-as_json 

Should consider removing or deprecating `:compact` option.

p-mongo: Thanks, using Rails' compact is a good idea.

Unfortunately we cannot simply remove an option that is part of the public API.

johnnyshields: Why not deprecate the option and remove it in the next major Mongoid version? There's no practical reason to have this option.

p-mongo: OK that seems reasonable, do you mind creating a jira issue for this?

johnnyshields: https://jira.mongodb.org/browse/MONGOID-4737

johnnyshields: Also: https://jira.mongodb.org/browse/MONGOID-4738

PR #4633, opened 2019-05-16
Fix MONGOID-4645 belongs_to in namespaced children fails to detect a non-namespaced association


p-mongo:


johnnyshields:
@p-mongo please, please fix your CI checks here. It's very poor form to leave these failing on every PR.

p-mongo:
Thank you for the feedback, we have made a note to address the CI situation.

lib/mongoid/association/relatable.rb
```diff
@@ -426,6 +463,45 @@ def create_extension!(&block)
       def default_inverse
         @default_inverse ||= klass.relations[inverse_klass.name.underscore]
       end
+
+      def namespace_hierarchy(mod)
```
saghm: The semantics of this method seem non-trivial; I think it might be worth documenting what it does, maybe along with a simple example of what it would return.

saghm: Thanks!

PR #4638, opened 2019-05-29
Fix MONGOID-4742 String#numeric? can return values other than true & false

patrykk21:
Reference issue: https://jira.mongodb.org/browse/MONGOID-4742

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,10 @@ end
 group :development, :testing do
   gem 'yard'
   platforms :mri do
-    gem 'byebug'
+    if RUBY_VERSION < '2.3'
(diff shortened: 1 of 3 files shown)


saghm:
Thanks for the pull request! I've just started the CI run so we can make sure everything is working

p-mongo:
Thank you for the patch, the desired behavior is for `numeric?` to always return either true or false.

Looking at the diff, the `be_numeric?` expectations are redundant with the newly added tests for the actual return value. Would you mind removing all of the `be_numeric?` tests and just leaving one test per input that asserts that the return value is true or false?

patrykk21:
@p-mongo sure, here we go!

PR #4640, opened 2019-05-29
MONGOD-4749 Fix: association with class_name: '::SomeClass'

zarqman:
Following the rework of the `:class_name` option in v7.0.3, a value of '::SomeClass' causes a NameError exception. This fixes it so it works again (as it did in 7.0.2 and prior).

Example:
```
class Child
  include Mongoid::Document
  belongs_to :parent, class_name: '::OtherClass'
end
```


zarqman:
cf MONGOID-4749

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4749

PR #4643, opened 2019-06-18
Fix headings in all Ruby files

p-mongo:


johnnyshields:
As I understand it `# encoding: utf-8` is not needed in Ruby 2.0+. What's the reason?

p-mongo:
http://ruby-doc.org/core-2.6.3/Encoding.html states:

>  Ruby source files should declare its script encoding by a magic comment even when they only depend on US-ASCII strings or regular expressions.

PR #4646, opened 2019-06-19
Update Mongoid driver test matrix.

p-mongo:
Add stable (2.10) and oldstable (2.9) configurations,
as well as the minimum one (2.5).


johnnyshields:
Recommend to either:
1) add back testing for driver v2.7 (in addition to 2.8 and 2.9)   OR
2) update gemspec to drop support for driver < 2.8

egiurleo:
It looks like `.evergreen/.evg.yml` is a configuration file that lists out all the types of tests you want to run and sets environment variables. Then, in `.evergreen/run-tests.sh`, you use the variables you set to figure out which Gemfile is appropriate for the tests you're about to run. Am I understanding this correctly?

Also, is there a guide to writing Evergreen configuration files that I can read?

p-mongo:
That is correct. Evergreen documentation for configuration files (Evergreen calls them "project files") is here: https://github.com/evergreen-ci/evergreen/wiki/Project-Files

egiurleo:
@p-mongo Thank you! I'll make sure to read these over.

PR #4652, opened 2019-08-19
4730

p-mongo:


p-mongo:
This alters the gemspec to allow Rails 6. Prior to release the following additional changes need to be made:

- Support matrices updated in documentation
- Add server 4.2 support, this requires some test fixes

johnnyshields:
@p-mongo why is server 4.2 support dependent on Rails 6 support? The two are completely separate items, and we shouldn't delay Rails 6 arbitrarily.

p-mongo:
Rails 6 support does not depend on 4.2 server support.

PR #4658, opened 2019-08-29
Copy Evergreen configuration and gemfiles from 7.0 branch/master

p-mongo:


p-mongo:
This copies current evergreen configuration, a bunch of general test suite setup code and some test changes from master to 6.4 branch. There are still some test failures in the resulting tree but in general the test suite is now useful, in particular for backports of ongoing changes which generally rely on the modern test setup available on master.

PR #4660, opened 2019-08-29
RUBY-4685 use max_server_version instead of a separate method for geo near support


p-mongo:


p-mongo:
This was meant to reference https://jira.mongodb.org/browse/MONGOID-4685, not RUBY-4685.

PR #4666, opened 2019-09-19
MONGOID-4667 Test Mongoid on rails 5.1


p-mongo:


p-mongo:
This PR also changes default Rails version to be 6.0, and adds a 5.2 config.

PR #4668, opened 2019-10-07
Fix MONGOID-4718 Dynamic attribute readers/writers fail to define when attribute name has special chars


p-mongo:


docs/tutorials/mongoid-documents.txt
```diff
@@ -604,56 +604,158 @@ of the following methods:
...
+(this appies to models not using dynamic fields as well):
 
 .. code-block:: ruby
 
-   # Raise a NoMethodError if value isn't set.
-   person.gender
-   person.gender = "Male"
+  bob = Person.new(age: 42)
+  
+  # OK
```
egiurleo: Can you write `- string access` here as well? It makes these examples very easy to read.

docs/tutorials/mongoid-documents.txt
```diff
@@ -604,56 +604,158 @@ of the following methods:
...
+  bob['name']
+  # => nil
+  
+  # OK
```
egiurleo: Same with symbol access.

PR #4683, opened 2019-11-10
MONGOID-4798 Revise configuration documentation


p-mongo:


docs/tutorials/mongoid-configuration.txt
```diff
@@ -30,52 +30,86 @@ i.e. ``development``, ``test`` or ``production``. The third level key,
...
+manually. This can be done via the ``Mongoid.load!`` method, which takes
+the configuration file path as its argument, as follows:
 
 .. code-block:: ruby
 
+  # Use automatically detected environment name
+  Mongoid.load!("path/to/your/mongoid.yml")
+  
+  # Specify environment name manually
   Mongoid.load!("path/to/your/mongoid.yml", :production)
```
egiurleo: Where do you put this code block? Also in `application.rb`? Or does it depend?

p-mongo: This section applies to non-Rails applications. As such the bootstrap code can vary from one application to another.

Example for tenex: https://github.com/p-mongo/tenex/blob/master/lib/fe/boot.rb#L5

egiurleo: That makes sense. Thanks for clarifying.

PR #4687, opened 2019-11-19
Fix error in QueryCache when retrieving a result larger than batch_size from cache

1st8:
Hi,

after upgrading mongoid and mongo we encountered an error "Cannot restart iteration of a cursor which issued a getMore" when performing a query with a result set larger than the batch size for the second time.

It happens in `QueryCache::View#each`, when iterating a cached cursor with `@get_more_called` a second time, which triggers https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/cursor.rb#L144

- **The first commit provides a spec which reproduces the error.**
- The second commit provides a fix, by only reusing cursors with `@get_more_called == false`. I simplified the code while implementing the fix, because I found it difficult to follow the logic in the old implementation. 
  (Call to `cursor.to_a[0...limit.abs]` which is never returned. 
  Logic in `#process` was copypasta but `super` does just fine.)
- The third commit fixes the "Legacy read_with_retry invocation" which came from missing arguments.
- The fourth commit updates `#each` to be closer to the original `Iterable#each` implementation, just for completeness.

Feel free to cherry-pick what you need and to let me know what you think.

Cheers 🍻

### DIFF
--- a/lib/mongoid/query_cache.rb
+++ b/lib/mongoid/query_cache.rb
@@ -228,25 +228,31 @@ module Mongoid
       #
       # @since 5.0.0
       def each
-        return super if system_collection? || !QueryCache.enabled?
+        return super unless should_cache?
 
-        cursor = fetch_cached_cursor do
+        @cursor = nil
+        @cursor = fetch_cached_cursor do
           session = client.send(:get_session, @options)
+          # Expanded implementation of #read_with_retry_cursor
           read_with_retry(session, server_selector) do |server|
-            CachedCursor.new(view, send_initial_query(server, session), server, session: session)
+            result = send_initial_query(server, session)
+            CachedCursor.new(view, result, server, session: session)
           end
         end
-        cursor.each do |doc|
-          yield doc
-        end if block_given?
-        cursor
+        if block_given?
+          @cursor.each do |doc|
+            yield doc
+          end
+        else
+          @cursor.to_enum
+        end
       end
 
       private
 
       # Returns a currently cached iterable cursor or yields and caches result
-      def fetch_cached_cursor(&_block)
-        iterable_cached_cursor || cache_cursor(yield)
+      def fetch_cached_cursor(&block)
+        iterable_cached_cursor || cache_cursor(&block)
       end
 
       # Returns iterable cursor or nil
@@ -270,8 +276,13 @@ module Mongoid
         [ collection.namespace, selector, limit, skip, sort, projection, collation ]
       end
 
-      def cache_cursor(cursor)
-        QueryCache.cache_table[cache_key(limit: limit)] = cursor
+      # Caches the result of block
+      def cache_cursor(&block)
+        QueryCache.cache_table[cache_key(limit: limit)] = yield
+      end
+
+      def should_cache?
+        QueryCache.enabled? && !system_collection?
       end
 
       def system_collection?
--- a/lib/mongoid/query_cache.rb
+++ b/lib/mongoid/query_cache.rb
@@ -160,9 +160,9 @@ module Mongoid
         "#<Mongoid::QueryCache::CachedCursor:0x#{object_id} @view=#{@view.inspect}>"
       end
 
-      # The cache can be iterated again, if get more was never called
+      # The cache can be iterated again, if the result was completed in one batch
       def iterable_again?
-        !@get_more_called
+        @batch_count.to_i <= 1
       end
 
       private
@@ -170,7 +170,8 @@ module Mongoid
       def process(result)
         documents = super
 
-        if @cursor_id.zero? && !@get_more_called
+        @batch_count = @batch_count.to_i + 1
+        if @cursor_id.zero? && @batch_count == 1
           @cached_documents = documents
         end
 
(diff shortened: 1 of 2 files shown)


p-mongo:
Thank you for the patch. Unfortunately `@get_more_called` is an instance variable set by the driver, and as such it cannot be directly referenced by Mongoid. The way to go is probably to move the guts of query cache implementation into the driver, which is covered by https://jira.mongodb.org/browse/RUBY-1849.

1st8:
Hi Oleg,

thanks for the feedback.

> Unfortunately @get_more_called is an instance variable set by the driver, and as such it cannot be directly referenced by Mongoid.

I am not sure if I understand you correctly. 
Technically it is possible to access `@get_more_called` since `Mongoid::QueryCache::CachedCursor` inherits from `Mongo::Cursor`.
Specs are all green.

I saw https://jira.mongodb.org/browse/RUBY-1849, but didn't think it would be a good idea to submit an unwarranted PR with a change deep in the basic driver, you probably aggree.

This PR fixes a very real bug, QueryCache throwing when results are larger than batch_size, without, in my opinion, diverging too much from the old implementation.

What do you suggest to move forward?

p-mongo:
It is technically possible to access driver internals from Mongoid, but we don't want to do that because then the driver is not able to change its implementation without breaking existing applications that use published Mongoid versions.

If the bug can be fixed using public driver APIs, such a fix can go into Mongoid today. Otherwise I would focus the efforts on ruby-1849.

1st8:
Hi Oleg,

thanks for clarifying, that makes sense. I removed the usage of `@get_more_called` in the latest commit. 
What do you think?

Cheers!

1st8:
Hi Oleg, 
just a quick follow-up - is there anything else I can do here?
Cheers

1st8:
I just checked the code again and was able to remove the usage of `@cursor_id`, too. 
There isn't any usage of internal implementation details left, as far as I can see.

Let me know what you think.
Alternatively I could move the fixed QueryCache into its own gem, but I don't see a reason for mongoid to keep athebroken QueryCache implementation.

lib/mongoid/query_cache.rb
```diff
@@ -225,40 +229,63 @@ module View
...
-            read_with_retry do
-              server = server_selector.select_server(cluster)
-              cursor = CachedCursor.new(view, send_initial_query(server), server)
-              QueryCache.cache_table[cache_key] = cursor
-            end
+        return super unless should_cache?
+
+        @cursor = nil
+        @cursor = fetch_cached_cursor do
+          session = client.send(:get_session, @options)
```
p-mongo: get_session is a private driver method.

lib/mongoid/query_cache.rb
```diff
@@ -225,40 +229,63 @@ module View
...
+          # Expanded implementation of #read_with_retry_cursor
+          read_with_retry(session, server_selector) do |server|
+            result = send_initial_query(server, session)
```
p-mongo: The two lines above are also driver internals.

1st8:
Okay, I give up then.

We are looking forward to an official solution to the currently broken QueryCache and will keep our fix internally the meantime.

Cheers!

ryana:
@1st8 have you been using this in production? We're running into the same problem and are thinking about using your fork in our application.

1st8:
Yes absolutely!
I can tell you exactly which commit and what version of the driver we are
running tomorrow.

Too bad this isn't fixed in mongoid itself...
I thought about extracting the fixed QueryCache into its own Gem so we
could get back to an official mongoid version. Maybe I'll find some time
for this.

Ryan Angilly <notifications@github.com> schrieb am Di., 11. Feb. 2020,
19:42:

1st8:
Looks like a new QueryCache implementation is nearing in https://github.com/mongodb/mongo-ruby-driver/pull/1603

I think my fork might be good until then.

Cheers!

ryana:
Yeah we deployed your fork earlier today and it worked like a charm!

Very cool on the driver-level QC implementation. Didn't know about that. Thanks for the pointer.

joshcass:
@1st8 looks like work on https://github.com/mongodb/mongo-ruby-driver/pull/1603 has stalled since December. We're running into this issue ourselves. Have you given any more thought to releasing your fix as a gem?

p-mongo:
The query cache project (https://jira.mongodb.org/browse/RUBY-2130) is high on our priority list. We welcome contributions from the community - if interested please comment in the linked ticket.

joshcass:
I guess I have a hard time seeing my to contributing to software that a [publicly traded company](https://investors.mongodb.com/file/Index?KeyFile=401492336) maintains to support the services it charges money for. That's just me though. That said, as paying mongodb customers we sure would appreciate a fix for this soon.

p-mongo:
There is no expectation of contributions, but as I said they are welcome and this project is prioritized by the team.

1st8:
> Have you given any more thought to releasing your fix as a gem?

@joshcass This is currently not something I feel like pursuing, sorry. I share your disappointment about the handling of this issue and can't really understand why a treacherous bug like this is kept in the official ORM.
For the time being I will maintain my fork with the fix by rebasing it occasionally when we need it.

PR #4699, opened 2019-11-30
Add eager loading for polymorphic belongs_to associations

leoarnold:


lib/config/locales/en.yml
```diff
@@ -60,14 +60,6 @@ en:
           resolution: "Search for attributes that are in the database or set
             the Mongoid.raise_not_found_error configuration option to false,
             which will cause a nil to be returned instead of raising this error."
-        eager_load:
```
p-mongo: Everything having to do with this exception must remain in Mongoid for backwards compatibility reasons, even if Mongoid no longer uses this exception.

leoarnold:
@p-mongo :heavy_check_mark: done

p-mongo:
Thank you, I have this PR in my todo list.

p-mongo:
This will be merged as https://github.com/mongodb/mongoid/pull/4727.

PR #4704, opened 2019-12-05
MONGOID-4810 prohibit nil arguments to Criteria#all


p-mongo:
https://jira.mongodb.com/browse/MONGOID-4810


johnnyshields:
Why not return an empty (null) criteria instead?

p-mongo:
@johnnyshields Can you give an example illustrating what you are suggesting?

PR #4711, opened 2019-12-12
[wip] Unalias id/_id

egiurleo:


lib/mongoid/fields.rb
```diff
@@ -59,8 +60,10 @@ module Fields
         type: BSON::ObjectId
       )
 
-      alias :id :_id
-      alias :id= :_id=
+      alias_attribute(:id, :_id)
+
+      # alias :id :_id
+      # alias :id= :_id=
```
johnnyshields, line 66: Delete comment here

lib/mongoid/fields.rb
```diff
@@ -37,7 +37,8 @@ module Fields
     # Constant for all names of the id field in a document.
     #
     # @since 5.0.0
-    IDS = [ :_id, :id, '_id', 'id' ].freeze
+    # IDS = [ :_id, :id, '_id', 'id' ].freeze
```
johnnyshields, line 40: delete comment here

spec/mongoid/fields_spec.rb
```diff
@@ -208,6 +208,7 @@
     end
   end
 
+
```
johnnyshields, line 211: Whitespace not needed here

lib/mongoid/attributes.rb
```diff
@@ -335,8 +335,25 @@ def alias_attribute(name, original)
...
+        remove_method "#{name}_changed?"
+        remove_method "reset_#{name}!"
+        remove_method "reset_#{name}_to_default!"
+        remove_method "#{name}_was"
+        remove_method "#{name}_will_change!"
+        remove_method "#{name}_before_type_cast"
+      end
     end
 
+
```
johnnyshields, line 356: remove whitespace

lib/mongoid/attributes.rb
```diff
@@ -335,8 +335,25 @@ def alias_attribute(name, original)
           alias #{name}_before_type_cast #{original}_before_type_cast
         RUBY
       end
+
+      # TODO: documentation
+      def unalias_attribute(name)
```
p-mongo, line 340: What about adding a more generally usable  `remove_field` method that could remove both fields and aliases?

johnnyshields: I think that's more dangerous and has a different fundamental meaning. It's fine to remove the alias `id`. It would be risky to accidentally remove the `_id` field itself.

p-mongo:
Continuing in https://github.com/mongodb/mongoid/pull/4929.

PR #4713, opened 2019-12-17
override default single table inheritance type value

jamesroscoe:
fixes https://jira.mongodb.org/browse/MONGOID-4816

### DIFF
--- a/lib/mongoid/document.rb
+++ b/lib/mongoid/document.rb
@@ -353,7 +353,57 @@ module Mongoid
       #
       # @since 1.0.0
       def _types
-        @_type ||= (descendants + [ self ]).uniq.map(&:to_s)
+        @_type ||= (descendants + [ self ] + (@_type_mappings || {}).keys + descendants.map{|d| d._types }.flatten).map(&:to_s).uniq
+      end
+
+      # Returns the class name to instantiate after appying
+      # override_inheritance_type_value.
+      #
+      # @param [ String ] type The descendant type name (actual Ruby class name
+      # or arbitrary alias name configured using
+      # override_inheritance_type_value.
+      # @returns [ String ] Ruby document type name as camelcased string
+      def __type(type)
+        klass = (@_type_mappings || {})[type]
+        descendants.each {|d| break if klass ||= d.__type(type) } unless klass
+        klass
+      end
+
+      # Returns the class name to instantiate after appying
+      # override_inheritance_type_value.
+      #
+      # @param [ String ] type The descendant type name (actual Ruby class name
+      # or arbitrary alias name configured using
+      # override_inheritance_type_value.
+      # @eturns [ Class ] Ruby document type to instantiate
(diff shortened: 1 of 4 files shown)


p-mongo:
Hi @jamesroscoe , thank you for the patch.

Is this implementation based on Mongoose's one, and if so where can I read about the Mongoose's API? https://mongoosejs.com/docs/discriminators.html does not seem to have anything relevant.

lib/mongoid/document.rb
```diff
@@ -353,7 +353,57 @@ def instantiate(attrs = nil, selected_fields = nil)
...
+      # class names or source languages.
+      #
+      # @example Override the type name
+      #
+      # class HospitalWorker::Doctor < HospitalWorker::Base
+      #   override_inheritance_type_value 'MD'
+      #
+      # @param [ String ] name The name to store in the type field
+      def override_inheritance_type_value(name)
+        c_name = name.to_s.camelize
```
p-mongo: Why does this method alter the value provided to it?

p-mongo:
Does this patch also handle various cases of the discriminator type value being overridden and not overridden in a single inheritance chain? E.g. base model without override -> inherited model with override -> inherited model without override.

lib/mongoid/document.rb
jamesroscoe: I wanted to treat :alias, 'Alias' and 'alias' the same so I camelized them all before storing.

jamesroscoe:
I am looking into your Mongoose question. I just happened to start a new job with the JIRA case's reporter and was asked to look into it.

With respect to the second question, I believe it should work in the case you've described (it's adding additional lookups and searches for matches recursively down the inheritance chain) but I didn't add a test specifically for that.

lib/mongoid/document.rb
p-mongo: Converting to string is OK but transforming the values I would say should be left to the application.

jamesroscoe: I think that's a good point. Your way increases flexibility. I don't think it will affect anything. The only potential issue I see is that in factory.rb the _type value is camelized before searching for the type name. It might be necessary to skip that step.

jamesroscoe: Or maybe to compare the camelized version passed to ::__type against a camelized version from ::_types.

jamesroscoe:
So it does appear to be different from the Mongoose discriminator concept. But that includes a discriminator schema and I'm not sure this applies here.

p-mongo:
Have you been able to locate the relevant Mongoose documentation then?

p-mongo:
This PR has been superseded by the discriminator value override project, which has just completed and is now on master.

@jamesroscoe Would you mind reviewing the behavior and documentation on master to verify it meets your requirements?

https://docs.mongodb.com/mongoid/master/tutorials/mongoid-documents/#changing-the-discriminator-key

PR #4715, opened 2019-12-19
MONGOID-4826 Support hashed shard key declarations

p-mongo:
https://jira.mongodb.com/browse/MONGOID-4826


p-mongo:
I wanted to make CI green on jruby before this PR but let's get this PR in since I believe work here is complete.

docs/tutorials/sharding.txt
```diff
@@ -0,0 +1,111 @@
...
+  class Person
+    include Mongoid::Document
+    
+    belongs_to :country
+
+    # Shards by country_id.
+    shard_key country: 1
+    
+    # The collection must also have an index that starts with the shard key.
+    index ssn: 1
```
egiurleo: should this say `index country: 1`?

p-mongo: I think it should, thank you.

PR #4718, opened 2020-01-06
MONGOID-4832 Attribute validation should not allow false values for Hash and Array types

gPrado:
See: https://jira.mongodb.org/browse/MONGOID-4832


gPrado:
Closed in favor of https://github.com/mongodb/mongoid/pull/4719

PR #4719, opened 2020-01-06
Fix MONGOID-4832 Storing false values in fields of Array and Hash types fails but not in an expected way

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4832

PR #4723, opened 2020-01-22
Add translated message for restrict_with_error association

gersongroth:
When trying to destroy an object with a restrict_with_error dependent, the resulting error message is fixed.
Theses changes allow internationalize the error message, displaying which model is related to the object being destroyed.

### DIFF
--- a/lib/mongoid/association/depending.rb
+++ b/lib/mongoid/association/depending.rb
@@ -131,7 +131,9 @@ module Mongoid
 
       def _dependent_restrict_with_error!(association)
         if (relation = send(association.name)) && !relation.blank?
-          errors.add(association.name, RESTRICT_ERROR_MSG)
+          association_name_translated = ::I18n.translate("activerecord.models.#{association.name}", default: association.name.to_s)
+
+          errors.add(:base, :restrict_with_error_dependent_destroy, association: association_name_translated)
           throw(:abort, false)
         end
       end
--- a/spec/mongoid/association/depending_spec.rb
+++ b/spec/mongoid/association/depending_spec.rb
@@ -849,8 +849,11 @@ describe Mongoid::Association::Depending do
 
       it 'adds an error to the parent object' do
         expect(person.delete).to be(false)
-        expect(person.errors[:restrictable_posts].first).to be(
-          Mongoid::Association::Depending::RESTRICT_ERROR_MSG)
+
+        key_message = "#{Mongoid::Errors::MongoidError::BASE_KEY}.restrict_with_error_dependent_destroy"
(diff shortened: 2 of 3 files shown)


p-mongo:
Hi @gersongroth , thank you for the PR.

Can you please make a ticket at https://jira.mongodb.org/browse/MONGOID which includes a snippet of code that triggers the error message you are changing, the old message and the new message proposed by this PR?

lib/mongoid/association/depending.rb
```diff
@@ -131,7 +131,9 @@ def _dependent_restrict_with_exception!(association)
 
       def _dependent_restrict_with_error!(association)
         if (relation = send(association.name)) && !relation.blank?
-          errors.add(association.name, RESTRICT_ERROR_MSG)
+          association_name_translated = ::I18n.translate("activerecord.models.#{association.name}", default: association.name.to_s)
```
p-mongo, line 134: Why does this line reference `activerecord`?

I found https://stackoverflow.com/questions/18672861/rails-4-i18n-on-activerecord-attributes#23727347 which suggests that `human_attribute_name` should be used to translate attribute name, and it appears that Mongoid does support `human_attribute_name`.

lib/mongoid/association/depending.rb
```diff
@@ -131,7 +131,9 @@ def _dependent_restrict_with_exception!(association)
...
+
+          errors.add(:base, :restrict_with_error_dependent_destroy, association: association_name_translated)
```
p-mongo, line 136: Why is the error added to base instead of the association? As far as I can tell the error is on an object in the association and not on the base model.

spec/mongoid/association/depending_spec.rb
```diff
@@ -849,8 +849,11 @@ class DoubleAssocThree < DoubleAssoc; end
 
       it 'adds an error to the parent object' do
         expect(person.delete).to be(false)
-        expect(person.errors[:restrictable_posts].first).to be(
-          Mongoid::Association::Depending::RESTRICT_ERROR_MSG)
+
+        key_message = "#{Mongoid::Errors::MongoidError::BASE_KEY}.restrict_with_error_dependent_destroy"
+        expect(person.errors[:base].first).to eq(
+          ::I18n.translate(key_message, association: :restrictable_posts)
```
p-mongo, line 855: This assertion should use the actual expanded string, so that it is clear what the message produced looks like. Right now one has to run the code to determine what the message is.

p-mongo:
Superseded by https://github.com/mongodb/mongoid/pull/4798.

PR #4724, opened 2020-02-05
Fix evergreen configuration


p-mongo:


p-mongo:
Same fix as bson-ruby one.

PR #4727, opened 2020-02-10
MONGOID-4841 Add eager loading for polymorphic belongs_to associations


p-mongo:
https://jira.mongodb.com/browse/MONGOID-4841


leoarnold:
@p-mongo What was so bad about #4699?

p-mongo:
I made this PR purely for logistical reasons which have to do with the way Evergreen, our CI system, handles PRs from external contributors. It is still in my review queue, thank you for checking.

PR #4733, opened 2020-03-11
[MONGOID-4850] Consolidate delegation

leoarnold:
https://jira.mongodb.org/browse/MONGOID-4850

Since the gem already requires "active_support/core_ext" which includes

https://api.rubyonrails.org/classes/Module.html#method-i-delegate

we can remove all other variants of delegators and consolidate
delegation on this one way to do it.

### DIFF
--- a/lib/mongoid.rb
+++ b/lib/mongoid.rb
@@ -3,7 +3,6 @@
 
 require "support/ruby_version"
 
-require "delegate"
 require "time"
 require "set"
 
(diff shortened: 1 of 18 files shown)


p-mongo:
Thank you for the patch @leoarnold . Do you have a code snippet that reproduces the problem that you could provide for https://jira.mongodb.org/browse/MONGOID-4849 ?

leoarnold:
@p-mongo Let me check if I can add a unit test for that.

leoarnold:
@p-mongo Here you go: I added a test which fails with the error described in branch `7.1-stable` (more precisely https://github.com/mongodb/mongoid/commit/4911c90cb00c0908058b8d5a18a8afa155f28e16) and passes with this PR.

p-mongo:
Thank you, please see my comment in https://jira.mongodb.org/browse/MONGOID-4849?focusedCommentId=2968651&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2968651.

p-mongo:
We decided to fix this via https://github.com/mongodb/mongoid/pull/4739 but thank you for this PR.

PR #4734, opened 2020-03-16
DOCSP-9414: Replace Google Groups references with MongoDB Community F…

ccho-mongodb:
…orums

JIRA: https://jira.mongodb.org/browse/DOCSP-9414

Please advise on what should be done with the reference to the google group in the gemspec for the mailing list. Thanks!

@p-mongo


ccho-mongodb:
@p-mongo can you take a look when you get the chance? If everything looks good, can you merge this since I don't have write access? Thanks.

PR #4736, opened 2020-03-31
Fix the Mongoid Associations Extensions code example

cgriego:
For `target` begin changed to `_target`.


p-mongo:
@egiurleo Please adjust the merge commit message to use "associations" instead of "relations".

egiurleo:
@p-mongo done

PR #4749, opened 2020-04-24
Fix MONGOID-4874 Values seem to be lost when using add_to_set on a non-existent attribute on a loaded instance

jklina:
It seems when an instance is loaded from a database, its `attributes`
method returns an instance of `BSON::Document`. This causes some
surprising behavior when we use `#add_to_set` on that loaded instance.

When a value is assigned to a `BSON::Document` instance, it gets
modified:

https://github.com/mongodb/bson-ruby/blob/master/lib/bson/document.rb#L87-L89

During modification, the attribute loses its reference to the original
value, however the `add_to_set` method relies on this reference:

https://github.com/mongodb/mongoid/blob/136ccbc140b25719f48ff5185bb471f90c063148/lib/mongoid/persistable/pushable.rb#L27-L31

Since that reference is broken, it no longer update the instance's
attribute and the value is lost in memory, though still retained in the
DB.

This ensures that the reference remains intact.

### DIFF
--- a/lib/mongoid/persistable/pushable.rb
+++ b/lib/mongoid/persistable/pushable.rb
@@ -24,7 +24,13 @@ module Mongoid
       def add_to_set(adds)
         prepare_atomic_operation do |ops|
           process_atomic_operations(adds) do |field, value|
-            existing = send(field) || (attributes[field] ||= [])
+            existing = send(field) || attributes[field]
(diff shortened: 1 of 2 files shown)


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4874

PR #4751, opened 2020-04-26
MONGOID-4877 Document that mongoid.yml is preprocessed with ERb


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4877

PR #4757, opened 2020-04-28
Fix MONGOID-4861 Useless $and when calling Foo.and(Foo.or(...))


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4861

PR #4761, opened 2020-05-13
Fix MONGOID-4880 Criteria for a field clobbered instead of combined when using symbol operators


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4880

PR #4764, opened 2020-05-15
Fix MONGOID-4882 Embedded documents not setting parent association if declared with a class_name having "::" prefix

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4882

Haerezis:
I don't know if I'm allowed to comment on this MR but the documentation of `relation_class_name` say the following :
```
      # @note The return value of this method should not be used to determine
      #   whether two associations have the same target class, because the
      #   return value is not always a fully qualified class name. To compare
      #   classes, retrieve the class instance of the association target using
      #   the +relation_class+ method.

```

In that case, shouldn't we use `relation_class.name` instead of `relation_class_name` on this particular instance?

p-mongo:
@Haerezis You are most welcome to comment on this or any other PR in Mongoid.

The docstring does say what you quoted, but attempting to do that produces the issue described in https://jira.mongodb.org/browse/MONGOID-4750. We have https://jira.mongodb.org/browse/MONGOID-4751 filed to deal with the remainder of this situation and I made a note there to also update our docstrings.

Haerezis:
Ho, ok. Thanks for your work !

PR #4766, opened 2020-05-15
Fix MONGOID-4884 not creates incorrect queries when symbol operators are given


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4884

PR #4767, opened 2020-05-18
MONGOID-4885 Use belongs_to instead of embedded_in in model generator


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4885

PR #4768, opened 2020-05-19
MONGOID-4744 Audit tree for incorrect use of ^ and $ anchors, replace with \A \z as needed


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4744

PR #4769, opened 2020-05-19
MONGOID-4848 Add server 4.4 configurations to evergreen

p-mongo:
- Add 4.4 configurations
- Add documentation for map/reduce statistics, which are no longer available in 4.4
- Skip map/reduce statistics tests on 4.4
- Skip code with scope tests until 4886 is done
- Sort map/reduce results in tests because they arrive in different order in 4.4 after being executed through aggregation pipeline
- 4.4 allows a combination of hashed and non-hashed shard keys, use two hashed keys which 4.4 does not allow instead to produce failures

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -354,6 +354,10 @@ axes:
   - id: "mongodb-version"
     display_name: MongoDB Version
     values:
+      - id: "4.4"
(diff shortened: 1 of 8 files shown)


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4848

PR #4770, opened 2020-05-19
MONGOID-4875 Eliminate x = (attributes[:foo] ||= value) pattern


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4875

PR #4774, opened 2020-05-25
MONGOID-4700 Use limit 1 instead of -1


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4700

PR #4776, opened 2020-05-25
Fix MONGOID-4772 Model.client_name not usable due to private visibility unlike collection_name/database_name


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4772

PR #4780, opened 2020-05-26
MONGOID-4827 Change test suite to always define all examples


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4827

PR #4788, opened 2020-06-08
Fix MONGOID-4853 Querying Date fields using a TimeWithZone instance queries by the date in local time zone


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4853

PR #4791, opened 2020-06-20
MONGOID-4902 Include shard key when reloading documents for improved query performance with Atlas Global Write clusters

schuylr:
https://jira.mongodb.org/browse/MONGOID-4902

I included some extra documentation for sharding with Global Clusters, since I initially thought that the `shard_key` method didn't actually modify the atomic selector for the document. I thought the tutorial was only used for setting up sharding on the cluster itself via rake tasks.

### DIFF
--- a/docs/tutorials/mongoid-upgrade.txt
+++ b/docs/tutorials/mongoid-upgrade.txt
@@ -216,6 +216,33 @@ example, the following two approaches result in the same generated selector:
     #   class:    Band
     #   embedded: false>
 
+Use Shard Key On Reload
+-----------------------
+
+Minor change: Mongoid 7.1 modifies the ``reload`` method to perform a find
+command using the model's shard key if one exists. This improves the performance,
+of ``reload`` against sharded clusters, especially those that shard documents
+based on the physical location of the shards.
+
+Consider a class ``Band`` whose documents are sharded by the ``name`` key.
+
+Example Mongoid 7.1 behavior:
+
+.. code-block:: ruby
+
+  band = Band.create(name: "Destiny's Child")
+  band.reload
+  # Command logs: { "find"=>"bands", "filter"=>{ "_id"=>BSON::ObjectId('...') "name"=>"Destiny's Child" } }
+
+Example Mongoid 7.0 behavior:
+
+.. code-block:: ruby
+
+  band = Band.create(name: "Destiny's Child")
+  band.reload
+  # Command logs: { "find"=>"bands", "filter"=>{"_id"=>BSON::ObjectId('...') } }
+
+
 Ruby Version Support
 --------------------
 
--- a/docs/tutorials/mongoid-upgrade.txt
+++ b/docs/tutorials/mongoid-upgrade.txt
@@ -221,10 +221,18 @@ Use Shard Key On Reload
 
 Minor change: Mongoid 7.1 modifies the ``reload`` method to perform a find
 command using the model's shard key if one exists. This improves the performance,
-of ``reload`` against sharded clusters, especially those that shard documents
-based on the physical location of the shards.
+of ``reload`` against sharded clusters, especially on Atlas Global Clusters.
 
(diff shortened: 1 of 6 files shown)


p-mongo:
Hi @schuylr , thank you for the patch. Can you please:

- provide a prose description of the behavior change,
- add test case(s) and/or code snippet(s) illustrating the behavior change.

schuylr:
Hi @p-mongo - sure thing

Given a sharded Mongo Document model with a `location` field (or any other field as the shard key):

```
class Document
  include Mongoid::Document

  # other fields
  field :location, type: String, default: 'US'

  shard_key :location 
end
```

And an object of this model (i.e. `doc = Document.last` or `doc = Document.create`)

The `reload` command will now utilize the defined shard keys when re-fetching from the collection:

```
# before
doc.reload
MONGODB | [8] mongo:27017 #1 | jewlr_development.find | STARTED | {"find"=>"documents", "filter"=>{"_id"=>BSON::ObjectId('...')}, "$db"=>"...", "lsid"=>{ ... }}

# after 
doc.reload
MONGODB | [8] mongo:27017 #1 | jewlr_development.find | STARTED | {"find"=>"documents", "filter"=>{"_id"=>BSON::ObjectId('...'), "location"=>"US"}, "$db"=>"...", "lsid"=>{ ... }}
```

Without this change, Mongo Atlas Global Clusters have to [broadcast the read query to all zones across the entire global cluster](https://docs.atlas.mongodb.com/global-cluster-sharding/#global-write-cluster-read-operations), which have a performance hit of around 200ms per reload, by my measurements.

I'll commit a test case shortly

schuylr:
Sorry, I couldn't find an easy way to validate that `reload` was including a `shard_key` in the `find` query. The only way I think this could be done is if instrumentation was added to `Findable` which would include the query used to find the object into an attribute. Hopefully my prose and code snippets were enough.

p-mongo:
Thank you, I believe this is sufficient.

egiurleo:
@p-mongo Do the tests I added sufficiently cover this feature? Would it be valuable to add a testing config for a sharded cluster with multiple mongoses?

spec/mongoid/reloadable_spec.rb
```diff
@@ -353,5 +353,82 @@
...
+        Mongoid::Config.send(:clients=, CONFIG[:clients])
+        Mongoid::Clients.with_name(:other).subscribe(Mongo::Monitoring::COMMAND, EventSubscriber.new)
+      end
+
+      let(:subscriber) do
+        client = Mongoid::Clients.with_name(:other)
+        monitoring = if client.respond_to?(:monitoring, true)
+          client.send(:monitoring)
+        else
+          # driver 2.5
```
p-mongo, line 373: According to gemspec, Mongoid presently requires driver 2.7+, thus this branch should be removable (everywhere).

spec/mongoid/reloadable_spec.rb
```diff
@@ -353,5 +353,82 @@
...
+
+      context 'without embedded document' do
+        let(:profile) do
+          Profile.with(client: :other) do |klass|
+            klass.create
+          end
+        end
+
+        before do
+          Profile.with(client: :other) do |_|
```
p-mongo, line 397: I believe you can simply omit the block argument if you don't need it in the block.

docs/tutorials/mongoid-upgrade.txt
```diff
@@ -216,6 +216,41 @@ example, the following two approaches result in the same generated selector:
     #   class:    Band
     #   embedded: false>
 
+Use Shard Key On Reload
+-----------------------
+
+Minor change: Mongoid 7.1 modifies the ``reload`` method to perform a find
+command using the model's shard key if one exists. This improves the performance,
```
p-mongo, line 223: The comma at the end of the line here is  I believe incorrect.

PR #4792, opened 2020-06-24
MONGOID-4904 Use a generic client name in tutorial example rather than "secondary"


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4904

PR #4794, opened 2020-06-24
MONGOID-4905 Use length instead of count when evaluating length of Array


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4905

PR #4796, opened 2020-06-26
MONGOID-4703 deprecate geo_spacial


p-mongo:
https://jira.mongodb.com/browse/MONGOID-4703


p-mongo:
- add deprecation note
    - change docstrings to use geo_spatial
    - change tests to use geo_spatial
    - test both geo_spatial and geo_spacial with all input cases

PR #4798, opened 2020-06-30
MONGOID-4912 Use a translated message for when an association fails validation during deletion due to dependent documents existing

p-mongo:
https://jira.mongodb.com/browse/MONGOID-4912

### DIFF
--- a/lib/config/locales/en.yml
+++ b/lib/config/locales/en.yml
@@ -2,8 +2,6 @@ en:
   mongoid:
     errors:
       messages:
-        restrict_with_error_dependent_destroy:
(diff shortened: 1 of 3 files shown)


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4912

p-mongo:
Compared to https://github.com/mongodb/mongoid/pull/4723, this PR:

- puts the error message back on the association field, adjusting the wording accordingly
- removes the need to translate the field name since the message is now a simple postfix like "is not empty"
- specifies the expected error message as a literal in the test

PR #4799, opened 2020-06-30
MONGOID-4909 updated documentation to include correct dependency types

Neilshweky:
I tested all of the dependency types for both has_many and has_one and they all work as intended for each type.

### DIFF
--- a/docs/tutorials/mongoid-relations.txt
+++ b/docs/tutorials/mongoid-relations.txt
@@ -760,7 +760,7 @@ The default behavior of each association when no dependent option is provided is
 
    class Label
      include Mongoid::Document
-     has_many :bands, dependent: :restrict
+     has_many :bands, dependent: :restrict_with_exception
    end
 
    label = Label.first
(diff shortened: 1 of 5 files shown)


p-mongo:
The changes in the tutorial are looking good. Can you also review the source tree for incorrect values of the option. I see the following:

<pre>
lib/mongoid/association/referenced/has_and_belongs_to_many/proxy.rb:          # :delete or :destroy the appropriate removal will occur.
lib/mongoid/association/referenced/has_many/proxy.rb:          # :delete or :destroy the appropriate removal will occur.
docs/tutorials/mongoid-relations.txt:     has_many :bands, dependent: :restrict
lib/config/locales/en.yml:          summary: "When defining '%{relation}' with a :dependent => :restrict,
lib/config/locales/en.yml:          summary: "When defining '%{relation}' with a :dependent => :restrict,
spec/mongoid/errors/delete_restriction_spec.rb:        "When defining 'drugs' with a :dependent => :restrict,"
</pre>

p-mongo:
I believe this one is still remaining:

<pre>
docs/tutorials/mongoid-relations.txt:     has_many :bands, dependent: :restrict
</pre>

PR #4800, opened 2020-07-01
MONGOID-4890 change blank? to use exists and not count

Neilshweky:
This is a pretty weird fix... Since _target can be either an enumerable or an array I had to make a method in enumerable that was already defined on arrays. This method turned out to be any?, which returns a boolean based on whether there is an element in the array.

Also note that `in_memory` and `_added` are both hashes in the any? function.

### DIFF
--- a/lib/mongoid/association/referenced/has_many/enumerable.rb
+++ b/lib/mongoid/association/referenced/has_many/enumerable.rb
@@ -216,6 +216,21 @@ module Mongoid
               end
             end
 
+            # Does the enumerable have any? Will determine if there exists at least
+            # one element in the enumerable, whether or not it is _loaded.
+            #
+            # @example Doess the enumerable have any?
+            #   enumerable.any?
+            #
+            # @return [ true, false ] If the enumerable has any elements.
+            def any? 
+              if _loaded?
+                in_memory.count > 0
+              else
+                _unloaded.exists? || _added.count > 0
+              end
+            end
+
             # Get the first document in the enumerable. Will check the persisted
             # documents first. Does not load the entire enumerable.
             #
--- a/lib/mongoid/association/referenced/has_many/enumerable.rb
+++ b/lib/mongoid/association/referenced/has_many/enumerable.rb
@@ -223,7 +223,7 @@ module Mongoid
             #   enumerable.any?
             #
             # @return [ true, false ] If the enumerable has any elements.
-            def any? 
+            def any?
               if _loaded?
                 in_memory.count > 0
               else
--- a/spec/mongoid/association/depending_spec.rb
+++ b/spec/mongoid/association/depending_spec.rb
@@ -870,7 +870,7 @@ describe Mongoid::Association::Depending do
       end
 
       context 'when deleted inside a transaction' do
-        require_topology :replica_set
+        require_topology :replica_set, :sharded
         
         before do
           person.restrictable_posts << post
@@ -906,7 +906,7 @@ describe Mongoid::Association::Depending do
       end
 
       context "when inside a transaction" do
-        require_topology :replica_set
+        require_topology :replica_set, :sharded
 
         it 'doesn\'t raise an exception inside a transaction' do
           person.with_session do |session|
--- a/spec/mongoid/association/depending_spec.rb
+++ b/spec/mongoid/association/depending_spec.rb
@@ -868,6 +868,8 @@ describe Mongoid::Association::Depending do
     end
 
     context 'when deleted inside a transaction' do
+      require_topology :replica_set
+      
       before do
         person.restrictable_posts << post
       end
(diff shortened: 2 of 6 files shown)


lib/mongoid/association/referenced/has_many/enumerable.rb
```diff
@@ -216,6 +216,21 @@ def empty?
...
+            #   enumerable.any?
+            #
+            # @return [ true, false ] If the enumerable has any elements.
+            def any? 
+              if _loaded?
+                in_memory.count > 0
+              else
+                _unloaded.exists? || _added.count > 0
+              end
+            end
```
egiurleo, line 232: Could you make sure this method works with the `has_and_belongs_to_many` association as well?

spec/mongoid/association/depending_spec.rb
```diff
@@ -865,5 +865,21 @@ class DoubleAssocThree < DoubleAssoc; end
...
+      end
+
+      it 'doesn\'t raise an exception' do
+        person.with_session do |session|
+          session.start_transaction
+          person.destroy
+          session.commit_transaction
+        end
+      end
+    end
```
egiurleo, line 883: Could you add the following tests:
1. Test that this works with `has_and_belongs_to_many`
2. Add unit tests for the `any?` method
3. Add unit tests for the `any?` method on the `has_and_belongs_to_many` relation?

lib/mongoid/association/referenced/has_many/enumerable.rb
```diff
@@ -216,6 +216,21 @@ def empty?
...
+            #
+            # @example Doess the enumerable have any?
+            #   enumerable.any?
+            #
+            # @return [ true, false ] If the enumerable has any elements.
+            def any? 
+              if _loaded?
+                in_memory.count > 0
+              else
+                _unloaded.exists? || _added.count > 0
```
egiurleo, line 230: This is looking great so far! I have two remaining questions:
1. Under what circumstances does this `_added.count` case get executed?
2. If that happened in a transaction, would it cause an error?

Apologies if you're already testing that case, but I just want to make sure that's covered.

Neilshweky: Good catch! I just added a test like this...

egiurleo: Nice, thank you!

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
```diff
@@ -1757,6 +1757,75 @@
...
+        it "returns true" do
+          sandwich.with_session do |session|
+            session.start_transaction
+            expect { sandwich.meats << Meat.new }.to_not raise_error
+            expect(sandwich.meats.any?).to be true
+            session.commit_transaction
+          end
+        end
+      end
+    end
```
egiurleo, line 1827: Could you standardize the tests between habtm/proxy_spec and has_many/proxy_spec? It looks like there are tests in both files that do not exist in the other, which seems inconsistent to me.

spec/mongoid/association/depending_spec.rb
```diff
@@ -825,44 +825,95 @@ class DoubleAssocThree < DoubleAssoc; end
...
+      let(:person) do
+        Person.new houses: [House.new]
+      end
+
+      it "returns false" do
+        expect(person.destroy).to be false
+      end
+
+      context "when inside a transaction" do
+        require_topology :replica_set
```
egiurleo, line 908: Could you test these on sharded topologies as well?

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
Neilshweky: Agree, which is weird because I just copied #count's tests in each file

spec/mongoid/association/depending_spec.rb
Neilshweky: Done!

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
Neilshweky: Fixed this!

Neilshweky: Should I do the same for embeds_many? or is that one fine

p-mongo:
This PR needs to be rebased for conflicts.

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
```diff
@@ -1915,6 +1846,107 @@
...
+    context "when appending to a relation in a transaction" do
+      require_topology :replica_set, :sharded
+      
+      let!(:sandwich) do
+        Sandwich.create!
+      end
+
+      it "returns true" do
+        sandwich.with_session do |session|
+          session.start_transaction
```
p-mongo, line 1909: Please use with_transaction instead of start/commit pair.

Neilshweky:
> This PR needs to be rebased for conflicts.

@p I'm getting "current branch is up to date" when trying to rebase.... Any thoughts?

PR #4801, opened 2020-07-01
MONGOID-4708 update documentation to reflect default dependency option

Neilshweky:
From ActiveRecord's documentation: https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#module-ActiveRecord::Associations::ClassMethods-label-Dependent+associations
"When no option is given, the behavior is to do nothing with the associated records when destroying a record."

This is how we handle this situation as well... I updated the documentation to reflect this.


docs/tutorials/mongoid-relations.txt
```diff
@@ -742,7 +742,7 @@ to be deleted. The options are as follows:
 - ``:nullify``: Orphan the child document.
 - ``:restrict``: Raise an error if the child is not empty.
 
-The default behavior of each association when no dependent option is provided is to nullify.
+The default behavior of each association when no dependent option is provided is to do nothing with the associated record.
```
p-mongo, line 745: Could you clarify this some more using language similar to the following:

<pre>
If no ``:dependent`` option is provided,  deleting the parent record leaves the child record unmodified
(in other words, the child record continues to reference the now deleted parent record via the
foreign key field). The child may become orphaned if it is ordinarily only referenced via the parent.
</pre>

and adjust the description of nullification to be something like

<pre>
 - ``:nullify``: Set the foreign key field of the child document to nil. The child may become orphaned
  if it is ordinarily only referenced via the parent.
</pre>

PR #4802, opened 2020-07-04
MONGOID-4914 Unmask Enumerable#find

rgould:
See description in Jira: https://jira.mongodb.org/browse/MONGOID-4914

### DIFF
--- a/lib/mongoid/criteria.rb
+++ b/lib/mongoid/criteria.rb
@@ -59,7 +59,7 @@ module Mongoid
 
     # Call either Findable#find or Enumerable#find based on the given arguments
     #
-    # If the arguments match Enumerable#find signature, that will be invoked.
+    # If given a block, Enumerable#find will be invoked.
     # Otherwise Findable#find will be invoked
     #
     # The signature of Enumerable#find can be found in the stdlib documentation:
@@ -72,16 +72,13 @@ module Mongoid
     # @example Find given a block and a Proc, invokes Enumerable#find
     #   criteria.find(-> { "Default Band" }) { |item| item.name == "Milwaukee Mode" }
     #
-    # @example Find given just a Proc, invokes Enumerable#find, returns an Enumerator
-    #   enumerator = criteria.find(-> { "Default Band" })
-    #
     # @example Find given anything else, invokes Findable#find
     #   critera.find("1234")
     #
     # @return Same return value as either Findable#find or Enumerable#find, depending
     #   on which one was invoked.
     def find(*args, &block)
-      if block_given? || args.first&.is_a?(Proc)
+      if block_given?
         _enumerable_find(*args, &block)
       else
         _findable_find(*args)
--- a/lib/mongoid/criteria.rb
+++ b/lib/mongoid/criteria.rb
@@ -20,9 +20,11 @@ module Mongoid
   # against the database.
   class Criteria
     include Enumerable
+    alias :_enumerable_find :find
     include Contextual
     include Queryable
     include Findable
+    alias :_findable_find :find
     include Inspectable
     include Includable
     include Marshalable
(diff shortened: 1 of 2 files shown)


lib/mongoid/criteria.rb
```diff
@@ -55,6 +57,40 @@ def ==(other)
...
 
+    # Call either Findable#find or Enumerable#find based on the given arguments
+    #
+    # If the arguments match Enumerable#find signature, that will be invoked.
+    # Otherwise Findable#find will be invoked
+    #
+    # The signature of Enumerable#find:
+    #
+    #   find(ifnone = nil) { |obj| block } → obj or nil
+    #   find(ifnone = nil) → an_enumerator
```
p-mongo, line 68: We do not need to duplicate standard library documentation in our docstrings. Linking to stdlib docs I would say is sufficient and this way our documentation cannot get out of date.

lib/mongoid/criteria.rb
```diff
@@ -55,6 +57,40 @@ def ==(other)
...
+    #
+    # @example Find using a block, invokes Enumerable#find
+    #   criteria.find { |item| item.name == "Depeche Mode" }
+    #
+    # @example Find given a block and a Proc, invokes Enumerable#find
+    #   criteria.find(-> { "Default Band" }) { |item| item.name == "Milwaukee Mode" }
```
p-mongo, line 74: Is there actual usage of this pattern in the wild, considering one can just as easily say

<pre>
     #   criteria.find { |item| item.name == "Milwaukee Mode" } || 'Default Band'
</pre>

lib/mongoid/criteria.rb
```diff
@@ -55,6 +57,40 @@ def ==(other)
...
+    # @example Find given just a Proc, invokes Enumerable#find, returns an Enumerator
+    #   enumerator = criteria.find(-> { "Default Band" })
+    #
+    # @example Find given anything else, invokes Findable#find
+    #   critera.find("1234")
+    #
+    # @return Same return value as either Findable#find or Enumerable#find, depending
+    #   on which one was invoked.
+    #
+    # @since 7.1.1
```
p-mongo, line 85: We no longer use since tags, and this one isn't accurate since the change would go into the next major release.

lib/mongoid/criteria.rb
```diff
@@ -55,6 +57,40 @@ def ==(other)
...
+    def find(*args, &block)
+      if block_given? || args.first&.is_a?(Proc)
```
p-mongo, line 87: I propose changing this to `if block_given?` only.

lib/mongoid/criteria.rb
rgould: Good call - fixed at b58cf1de3

lib/mongoid/criteria.rb
rgould: Yeah, I've never seen it. I was just trying to cover all the cases for `Enumerable#find` :)

lib/mongoid/criteria.rb
rgould: removed at 397f4c2b9

lib/mongoid/criteria.rb
rgould: Fine by me :) Fixed at c08d43ffe

p-mongo:
Hi @rgould , thank you for making these changes.

I looked into this PR in https://github.com/mongodb/mongoid/pull/4817 and it seems the changes so far have to do with association finds and the top-level model finds seem to also be affected by the same issue:

```
irb(main):005:0> Band.find { |x| p x }
=> nil
```

Would you like to investigate the top-level find behavior? It looks like you added unit tests for Criteria and it seems we also need integration tests for finding on an association as well as finding on top level.

rgould:
Ah! Sure! I don't know how but I convinced myself it wasn't necessary. I'll try and take a look at it this weekend :) Thanks!

johnnyshields:
Should close in favor of #4817

p-mongo:
@rgould Are you able to make the changes mentioned? If not we can also do them.

rgould:
@p-mongo Sorry! Lots of surprise bureaucracy is getting in my way. I'll try and get it done this weekend, but if I don't, feel free to take it on if you have the capacity

rgould:
@p-mongo I started to look into this again today, and I'm not exactly sure where I should make this change. Also note that ActiveRecord doesn't document this feature on their end, though they do incidentally support it.

Do we want to make `Document` or maybe `Composable` also include `Enumerable`? If it's `Composable`, then the solution is consistent with how it was done for `Criteria`

PS. I don't have a lot of time in the next couple weeks, so if you want to pick this up and finish it I'm totally okay with that. If not I will come back to it when I find the time :)

p-mongo:
I filed  https://jira.mongodb.org/browse/MONGOID-5003 for the top-level find change, and this PR will be merged as https://github.com/mongodb/mongoid/pull/4817.

PR #4803, opened 2020-07-06
MONGOID-4788 Ensure $exists embedded matcher returns true for null values of defined fields


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4788

PR #4806, opened 2020-07-08
MONGOID-4801 Add tests for condition being a matching array in $eq embedded matcher


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4801

PR #4808, opened 2020-07-08
MONGOID-4898 changed count to use count_documents and added estimated_count

Neilshweky:
This PR has failing tests in the current driver version. The count_documents method does not accurately count object with collation, because it doesn't pass along the options to the count_documents method. The fix in driver was merged into master, I am just waiting for it to be released.


p-mongo:
@Neilshweky  Since we just released the driver, could you update the minimum required driver version to 2.10.5 in https://github.com/mongodb/mongoid/blob/master/mongoid.gemspec, bump https://github.com/mongodb/mongoid/blob/master/gemfiles/driver_min.gemfile and https://github.com/mongodb/mongoid/blob/master/gemfiles/driver_min_jruby.gemfile to 2.10-stable and resend this PR to evergreen?

lib/mongoid/contextual/mongo.rb
```diff
@@ -71,7 +71,31 @@ def cached?
...
+      # @example Get the count of documents with the provided options.
+      #   context.estimated_count(limit: 1)
+      #
+      # @example Get the count for where the provided block is true.
+      #   context.estimated_count do |doc|
+      #     doc.likes > 1
+      #   end
+      #
+      # @param [ Hash ] options The options, such as skip and limit to be factored
+      #   into the count.
```
p-mongo, line 91: Are these options really valid? https://docs.mongodb.com/manual/reference/method/db.collection.estimatedDocumentCount/ only mentions maxtimems.

spec/mongoid/findable_spec.rb
```diff
@@ -479,6 +479,32 @@
     end
   end
 
+  describe '.count' do 
+    context 'when the collection is not empty' do 
+      before do
+        Band.create(name: "Tool")
```
p-mongo, line 485: Please use `create!` whenever the return value is not checked.

lib/mongoid/findable.rb
```diff
@@ -60,6 +60,18 @@ def count
...
+    # Returns an estimated count of records in the database.
+    # If you want to specify conditions use where.
+    #
+    # @example Get the count of matching documents.
+    #   Person.estimated_count
+    #   Person.where(title: "Sir").estimated_count
+    #
+    # @return [ Integer ] The number of matching documents.
+    def estimated_count
+      with_default_scope.estimated_count
```
p-mongo, line 72: I believe calling estimated_count on a class with a default scope defined needs to be tested as I don't expect it to work (at all or correctly).

Neilshweky: There is a single test for estimated count that does pass... At least we know it doesn't crash... I'm not sure how to test it because I'm not sure what the outcome of the function is supposed to be (after all it is estimated)...

lib/mongoid/contextual/mongo.rb
```diff
@@ -71,7 +71,31 @@ def cached?
...
+      #   context.estimated_count do |doc|
+      #     doc.likes > 1
+      #   end
+      #
+      # @param [ Hash ] options The options, such as maxTimeMS to be factored
+      #   into the count.
+      #
+      # @return [ Integer ] The number of matches.
+      #
+      # @since 3.0.0
```
p-mongo, line 95: This since version is incorrect and should be removed.

lib/mongoid/findable.rb
p-mongo: Which test uses a default scope?

lib/mongoid/contextual/mongo.rb
```diff
@@ -71,7 +71,31 @@ def cached?
...
+      # @example Get the estimated number of matching documents.
+      #   context.estimated_count
+      #
+      # @example Get the count of documents with the provided options.
+      #   context.estimated_count(limit: 1)
+      #
+      # @example Get the count for where the provided block is true.
+      #   context.estimated_count do |doc|
+      #     doc.likes > 1
+      #   end
```
p-mongo, line 88: This usage with a block is also not tested as far as I can tell.

Neilshweky:
> Which test uses a default scope?

There is a test in findable spec that does `Band.estimated_count` which calls the `with_default_scope.estimated_count`.

lib/mongoid/contextual/mongo.rb
Neilshweky: Emily and I decided not to allow for the block notation with `estimated_count`. The reason `count` has a block is because it is a function often used on `Enumerable` and the user expects to be able to pass in a block, the however is not true with `estimated_count`.

If the user wants to use a filter for an `estimated_count` they can call the `estimated_count` method on a criteria that uses that filter.

lib/mongoid/findable.rb
Neilshweky: See my comment below

p-mongo:
You have some trailing whitespace in this PR, please re-save all files after waiting for https://github.com/mongodb/mongoid/pull/4850 to be merged and rebasing on master.

lib/mongoid/findable.rb
p-mongo: I believe if a default scope is defined, estimated_count should raise an error, unless you would like to make an argument to the contrary.

spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -195,6 +195,52 @@
...
+        described_class.new(criteria.cache)
+      end
+
+      before do
+        expect(context.view).to receive(:estimated_document_count).once.and_return(1)
+      end
+
+      it "returns the count cached value after first call" do
+        2.times do 
+          context.estimated_count
```
p-mongo, line 236: This test has no assertions?

p-mongo: Please move the expectation setup out of before into the example so that it is clear that there are assertions.

p-mongo:
The session problem is also fixed on master, you can get rid of my last commit here when you rebase.

spec/mongoid/contextual/mongo_spec.rb
Neilshweky: I found a similar pattern in other places, I fixed them as well.

lib/mongoid/errors/invalid_estimated_count_criteria.rb
```diff
@@ -0,0 +1,24 @@
...
+
+module Mongoid
+  module Errors
+
+    # Creates the exception raised when trying to call estimated_count
+    # on a filtered criteria.
+    #
+    # @param [ String ] class_name The class name.
+    #
+    # @api private
```
p-mongo, line 12: You can have api private on the constructor but the class overall is not private because it is used for exceptions reported to applications.

p-mongo: So maybe move this entire docstring to the constructor, since it seems to apply to the constructor and not the class.

lib/mongoid/errors/invalid_estimated_count_criteria.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+module Mongoid
+  module Errors
+
+    # Creates the exception raised when trying to call estimated_count
+    # on a filtered criteria.
+    #
+    # @param [ String ] class_name The class name.
```
p-mongo, line 10: What class is this the name of?

lib/mongoid/findable.rb
```diff
@@ -60,6 +60,18 @@ def count
       with_default_scope.count
     end
 
+    # Returns an estimated count of records in the database.
+    # If you want to specify conditions use where.
```
p-mongo, line 64: The conditions are not specifiable right?

lib/mongoid/findable.rb
```diff
@@ -60,6 +60,18 @@ def count
...
+    #
+    # @example Get the count of matching documents.
+    #   Person.estimated_count
+    #   Person.where(title: "Sir").estimated_count
```
p-mongo, line 68: And this example would raise an exception yes?

lib/mongoid/errors/invalid_estimated_count_criteria.rb
Neilshweky: The `klass` of the criteria used to call estimated count... I can put that in the docstring

PR #4812, opened 2020-07-10
MONGOID-4919 Use Mongo::Error::SessionsNotSupported for checking session support


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4919

lib/mongoid/clients/sessions.rb
```diff
@@ -41,12 +41,20 @@ module Sessions
...
+        end
         session = persistence_context.client.start_session(options)
         Threaded.set_session(session)
         yield(session)
       rescue Mongo::Error::InvalidSession => ex
-        if ex.message == Mongo::Session::SESSIONS_NOT_SUPPORTED
+        if
+          # Driver 2.13.0+
+          defined?(Mongo::Error::SessionsNotSupported) &&
+            Mongo::Error::SessionsNotSupported === ex or
```
egiurleo, line 54: Is there a reason to use `or` here vs `||`? All things being equal, I think I'd prefer consistent styles.

p-mongo: I think I did it for emphasis but I can redo with punctuation.

egiurleo: Thanks!

PR #4813, opened 2020-07-11
MONGOID-4920 Remove redundant _type field declarations in child classes


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4920

spec/mongoid/fields_spec.rb
```diff
@@ -1112,6 +1118,10 @@ def testing=(value)
       it "includes the child fields" do
         expect(circle.fields.keys).to include("radius")
       end
+
+      it 'includes _type field' do
+        expect(circle.fields.keys).to include("_type")
+      end
```
Neilshweky, line 1124: Is there a reason you're only testing this on one of `Shape`'s subclasses? I would love to see if the same is true in the Square class.... Especially because of how `inherited` is implemented....

p-mongo: What are you expecting  the difference between testing a square and testing a circle to be?

Neilshweky: The way inherited is implemented is `field` is only called once across all calls of `inherited` for one parent class, because of `unless fields.has_key?("_type")`. Now, the `field` function does add the field to all of the parent's subclasses, but at the time that `field` is called (i.e. in the first call to `inherited`) it's possible that Square is not part of Shape's descendants yet.

PR #4814, opened 2020-07-13
MONGOID-4921 add discriminator_key as a class variable

Neilshweky:
Added discriminator key as an attr_accessor on the class. Pre-req to doing MONGOID-4817.


lib/mongoid.rb
```diff
@@ -111,4 +111,14 @@ def client(name)
...
+
+  # The default discriminator key for Mongoid
+  #
+  # @example get the base discriminator key
+  #   Mongoid.discriminator_key
+  #
+  # @example set the base discriminator key
+  #   Mongoid.discriminator_key = "test"
+  attr_accessor :discriminator_key
+  self.discriminator_key = '_type'
```
p-mongo, line 123: Let's change this to a Mongoid config option such that it is settable in the configuration file. Please add the tutorial documentation as well.

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,87 @@
...
+
+  describe "#discriminator_key" do
+
+    context "when the discriminator key is not changed" do 
+      it "equals _type" do
+        expect(Instrument.discriminator_key).to eq("_type")
+      end
+
+      it "children's discriminator keys equal _type" do
+        expect(Guitar.discriminator_key).to eq("dkey")
```
p-mongo, line 263: This assertion does not appear to go with the test label.

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,87 @@
...
+        expect(Piano.discriminator_key).to eq("_type")
+      end
+
+      it "the base discriminator key is _type" do
+        expect(Mongoid.discriminator_key).to eq("_type")
+      end
+    end
+
+    context "when the discriminator key is changed at the base level" do
```
p-mongo, line 272: Suggest "changed globally"

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,87 @@
...
+      end
+    end
+
+    context "when the discriminator key is changed at the base level" do 
+      before do
+        Mongoid.discriminator_key = "hello"
+      end
+
+      after do
+        Mongoid.discriminator_key = "_type"
```
p-mongo, line 278: Can this be assigned `nil` such that subsequent code would use the global default?

spec/support/models/instrument.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+class Instrument
+    include Mongoid::Document
+    field :owner, type: String
```
p-mongo, line 6: If this field isn't used please take it out.

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,87 @@
...
+    context "when the discriminator key is changed in the parent" do 
+      before do
+        Instrument.discriminator_key = "hello2"
+      end
+
+      after do 
+        Instrument.discriminator_key = "_type"
+      end
+
+      it "changes in the child classes" do
```
p-mongo, line 301: Let's make it so that discriminator_key cannot be changed midway in the inheritance hierarchy, like store_in works.

lib/mongoid/traversable.rb
```diff
@@ -205,10 +222,11 @@ def inherited(subclass)
         # We only need the _type field if inheritance is in play, but need to
         # add to the root class as well for backwards compatibility.
         unless fields.has_key?("_type")
-          field(:_type, default: self.name, type: String)
+          field(self.discriminator_key, default: self.name, type: String)
         end
+
         subclass_default = subclass.name || ->{ self.class.name }
-        subclass.field(:_type, default: subclass_default, type: String, overwrite: true)
+        subclass.field(self.discriminator_key, default: subclass_default, type: String, overwrite: true)
```
p-mongo, line 229: The above two hunks don't belong in this PR I think.

lib/mongoid/traversable.rb
```diff
@@ -17,6 +17,23 @@ def _parent=(p)
...
+      end
+
+      def self.discriminator_key=(value) 
+        @discriminator_key = value
+        # unless fields.has_key?(@discriminator_key)
+        #   field(@discriminator_key, default: self.name, type: String)
+        # end 
+
+        descendants.each do |child| 
+          child.discriminator_key = value
```
p-mongo, line 32: Let's try replacing this with `class_attribute`.

p-mongo:
Let's not allow reads of discriminator_key on instance level and add a test for that.

spec/mongoid/traversable_spec.rb
Neilshweky: did you mean to put this comment on Instrument.discriminator_key? 

Also, setting the `Instrument.discriminator_key` to nil changes the global default because we delegate the `discriminator_key` method to `::Mongoid`

lib/mongoid/traversable.rb
Neilshweky: The problem with doing this is I can't override the methods, so I can't disallow users from setting in a child class

spec/mongoid/traversable_spec.rb
p-mongo: Yes I meant this comment to be on a class.

lib/mongoid/errors/setting_discriminator_key_on_child.rb
```diff
@@ -0,0 +1,23 @@
...
+# encoding: utf-8
+
+module Mongoid
+  module Errors
+
+    # This error is raised when trying to create a scope with an name already
+    # taken by another scope or method
+    #
+    # @example Create the error.
+    #   ScopeOverwrite.new(Person,'teenies')
```
p-mongo, line 11: This example is not for this class.

lib/mongoid/errors/setting_discriminator_key_on_child.rb
```diff
@@ -0,0 +1,23 @@
...
+    class SettingDiscriminatorKeyOnChild < MongoidError
```
p-mongo, line 12: Suggest `InvalidDiscriminatorKeyClass` or `InvalidDiscriminatorKeyTarget` or `InvalidDiscriminatorKeyChange`.

lib/mongoid/traversable.rb
```diff
@@ -17,6 +17,25 @@ def _parent=(p)
       @__parent = p
     end
 
+    included do 
+      module DiscriminatorKeyAssignment
```
p-mongo, line 21: Please move the module out of the `included` block and mark it `@api private`.

lib/mongoid/traversable.rb
```diff
@@ -17,6 +17,25 @@ def _parent=(p)
...
+        def discriminator_key=(value)
+          if hereditary?
+            raise Errors::SettingDiscriminatorKeyOnChild.new(self.to_s, self.superclass.to_s)
```
p-mongo, line 24: `to_s` calls here should not be needed.

lib/mongoid/traversable.rb
```diff
@@ -17,6 +17,25 @@ def _parent=(p)
...
+          end
+
+          value = Mongoid.discriminator_key unless value
```
p-mongo, line 27: This should remove the override instead of writing the global value in.

spec/mongoid/config_spec.rb
```diff
@@ -197,6 +197,34 @@
     end
   end
 
+  context 'when discriminator_key option' do
```
p-mongo, line 200: This sentence fragment is missing a verb, alternatively you can say something like  `context 'discriminator_key override'`.

spec/mongoid/config_spec.rb
```diff
@@ -197,6 +197,34 @@
...
+    context 'is not set in the config' do
+      it 'sets the value to _type by default' do
```
p-mongo, line 202: s/sets/is/. The test is reading the value, hence the heading should talk about reading and not writing.

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,108 @@
       end
     end
   end
+
+  describe "#discriminator_key" do
+
+    context "when the discriminator key is not changed" do
```
p-mongo, line 257: s/changed/set on a class/, for example.

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,108 @@
...
+
+      after do
+        Mongoid.discriminator_key = "_type"
+      end
+
+      it "the value changes" do 
+        expect(Mongoid.discriminator_key).to eq("hello")
+      end
+
+      it "all discriminator keys change" do
```
p-mongo, line 287: This should probably be 3 separate tests with each test having its own descriptive label as to what the test is testing.

spec/support/models/instrument.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+class Instrument
+    include Mongoid::Document
```
p-mongo, line 5: 2 space indent please.

lib/mongoid/traversable.rb
Neilshweky: Would you know how to achieve this? I've been googling but I can't figure out how to remove the override (or redelegate to Mongoid)...

Things I've tried: 
`delegate :discriminator_key, to: ::Mongoid unless value`

Neilshweky: I think my implementation, while isn't exactly what you would expect, is a better implementation... Here's the use case I'm thinking about. Imagine the following code was executed:
```
Instrument.discriminator_key = nil
Instrument.discriminator_key = nil
```
If we remove the override of the discriminator_key on the first line and redelegate `discriminator_key` to `Mongoid`, then the second line will set the global discriminator_key to nil, which is presumably not what we want. In my implementation, `Instrument.discriminator_key` will return the global discriminator key after each line.

lib/mongoid/errors/invalid_discriminator_key_target.rb
```diff
@@ -0,0 +1,20 @@
...
+  module Errors
+
+    # This error is raised when trying to set or get the 
+    # discriminator key on a child class.
+    class InvalidDiscriminatorKeyTarget < MongoidError
+      def initialize(class_name, superclass)
+        super(
+          compose_message(
+            "invalid_discriminator_key_target",
+            { class_name: class_name,  superclass: superclass}
```
egiurleo, line 14: Nitpick - missing a space before the closing curly brace.

lib/mongoid/errors/invalid_discriminator_key_target.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+module Mongoid
+  module Errors
+
+    # This error is raised when trying to set or get the 
+    # discriminator key on a child class.
+    class InvalidDiscriminatorKeyTarget < MongoidError
+      def initialize(class_name, superclass)
```
egiurleo, line 10: Could you write an api doc comment for this method?

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,164 @@
       end
     end
   end
+
+  describe "#discriminator_key" do
+
+    context "when the discriminator key is not set on a class" do
+      it "equals _type" do
```
egiurleo, line 258: Could you make this test description more descriptive?

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,164 @@
...
+      end
+
+      it "child discriminator keys equal _type: Piano" do
+        expect(Piano.discriminator_key).to eq("_type")
+      end
+
+      it "child discriminator keys equal _type: Guitar" do
+        expect(Guitar.discriminator_key).to eq("_type")
+      end
+    end
```
egiurleo, line 273: I don't mean to be a stickler for rspec conventions (lol) but could you change these test descriptions so the line reads like a sentence? (e.g. `it "sets the global discriminator key to _type"`)

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,164 @@
...
+        end
+  
+        it "changes in the child class: Guitar" do
+          expect(Guitar.discriminator_key).to eq("_type")
+        end
+      end
+
+      context "when resetting the discriminator key after nil" do
+        before do
+          Instrument.discriminator_key = "hello4"
```
egiurleo, line 350: I think you need to put a line before this one that sets the discriminator key to nil? Correct me if I'm wrong here.

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,164 @@
...
+      end
+
+      it "doesn't change in the sibling" do
+        expect(Piano.discriminator_key).to eq("_type")
+      end
+
+      it "doesn't change in the parent" do
+        expect(Instrument.discriminator_key).to eq("_type")
+      end
+    end
```
egiurleo, line 393: I think you need to add some sort of `before` block that tries to set the discriminator key on `Guitar`.

spec/mongoid/traversable_spec.rb
Neilshweky: Yeah, that's a good point... The `after` in the previous test sets it to `nil` so that's what I didn't originally put it there... But you're right, I should.

spec/mongoid/traversable_spec.rb
Neilshweky: I changed it to a let block... does that work?

spec/mongoid/traversable_spec.rb
Neilshweky: sure!

spec/mongoid/traversable_spec.rb
Neilshweky: done!

lib/mongoid/errors/invalid_discriminator_key_target.rb
Neilshweky: yup!

lib/mongoid/errors/invalid_discriminator_key_target.rb
Neilshweky: good one! can't believe I missed that... It's bothering me like crazy just looking at it.

lib/mongoid/traversable.rb
Neilshweky: Resolved!

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,169 @@
...
+        expect(Mongoid.discriminator_key).to eq("_type")
+      end
+
+      it "doesn't change in the sibling" do
+        expect(Piano.discriminator_key).to eq("_type")
+      end
+
+      it "doesn't change in the parent" do
+        expect(Instrument.discriminator_key).to eq("_type")
+      end
```
egiurleo, line 397: For these tests, I would add a line before the expectation that invokes guitar and throws away the exception

guitar rescue nil
While the first test within this context invokes guitar, I think you'd want to be able to run any of these tests individually and have it still do everything you want it to do.

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,169 @@
...
+        end
+  
+        it "changes in the child class: Guitar" do
+          expect(Guitar.discriminator_key).to eq("hello4")
+        end
+      end
+    end
+
+    context "when the discriminator key is changed in the child" do
+      let(:guitar) do
```
egiurleo, line 373: Could you give this helper a slightly more descriptive name like set_discriminator_key?

PR #4820, opened 2020-07-15
MONGOID-4928 implemented custom discriminator key fields

Neilshweky:
If the discriminator key changes before the class is created then it is created with the correct field, otherwise a second field is added in addition to type.

### DIFF
--- a/lib/mongoid/traversable.rb
+++ b/lib/mongoid/traversable.rb
@@ -239,7 +239,7 @@ module Mongoid
 
         # We only need the _type field if inheritance is in play, but need to
         # add to the root class as well for backwards compatibility.
-        unless fields.has_key?("_type")
+        unless fields.has_key?(self.discriminator_key)
           default_proc = lambda { self.class.name }
           field(self.discriminator_key, default: default_proc, type: String)
         end
--- a/lib/mongoid/traversable.rb
+++ b/lib/mongoid/traversable.rb
@@ -37,7 +37,11 @@ module Mongoid
           end
         end 
 
-        unless fields.has_key?(self.discriminator_key) || descendants.length == 0
+        # This condition checks if the new discriminator key would overwrite
+        # an existing field.
+        # This condition also checks if the class has any descendants, because
+        # if it doesn't then it doesn't need a discriminator key. 
+        if !fields.has_key?(self.discriminator_key) && !descendants.empty?
           default_proc = lambda { self.class.name }
           field(self.discriminator_key, default: default_proc, type: String)
         end
--- a/lib/mongoid/traversable.rb
+++ b/lib/mongoid/traversable.rb
@@ -26,7 +26,6 @@ module Mongoid
           raise Errors::InvalidDiscriminatorKeyTarget.new(self, self.superclass)
         end
 
-        
         if value
           super
         else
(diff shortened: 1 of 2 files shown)


spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,328 @@
...
+        expect(Mongoid.discriminator_key).to eq("_type")
+      end
+
+      it "doesn't change in the sibling" do
+        expect(Piano.discriminator_key).to eq("_type")
+      end
+
+      it "doesn't change in the parent" do
+        expect(Instrument.discriminator_key).to eq("_type")
+      end
```
egiurleo, line 397: For these tests, I would add a line before the expectation that invokes guitar and throws away the exception
```
guitar rescue nil
```
While the first test within this context invokes guitar, I think you'd want to be able to run any of these tests individually and have it still do everything you want it to do.

spec/mongoid/traversable_spec.rb
```diff
@@ -251,4 +251,328 @@
...
+        end
+  
+        it "changes in the child class: Guitar" do
+          expect(Guitar.discriminator_key).to eq("hello4")
+        end
+      end
+    end
+
+    context "when the discriminator key is changed in the child" do
+      let(:guitar) do
```
egiurleo, line 373: Could you give this helper a slightly more descriptive name like `set_discriminator_key`?

lib/mongoid/traversable.rb
```diff
@@ -35,6 +35,11 @@ def discriminator_key=(value)
           class << self
             delegate :discriminator_key, to: ::Mongoid
           end
+        end 
+
+        unless fields.has_key?(self.discriminator_key) || descendants.length == 0
```
p-mongo, line 40: Can this condition be reworded using `if` please?

p-mongo: And you can use `descendants.empty?` instead of `descendants.length == 0`.

p-mongo: Also can this conditional have a comment above it explaining what it is for (i.e. adding the new field for the new discriminator_key to already instantiated classes that have descendants, and thus that have had the previous discriminator key field added to them already).

lib/mongoid/traversable.rb
```diff
@@ -234,9 +239,9 @@ def inherited(subclass)
 
         # We only need the _type field if inheritance is in play, but need to
         # add to the root class as well for backwards compatibility.
-        unless fields.has_key?("_type")
+        unless fields.has_key?(self.discriminator_key)
           default_proc = lambda { self.class.name }
-          field(:_type, default: default_proc, type: String)
+          field(self.discriminator_key, default: default_proc, type: String)
```
p-mongo, line 244: :+1:

spec/mongoid/traversable_spec.rb
```diff
@@ -422,5 +422,164 @@
...
+
+        it "creates a _type field in the child: Piano" do
+          expect(Piano.fields.keys).to include("_type")
+        end
+      end
+  
+      context "when the discriminator key is changed at the base level" do
+        context "after class creation" do
+          before do
+            class GlobalDiscriminiatorParent
```
p-mongo, line 444: Typo in "Discriminator".

lib/mongoid/traversable.rb
Neilshweky: done!

PR #4823, opened 2020-07-17
MONGOID-4817 use #discriminator_key function instead of _type

Neilshweky:
So far I have changed criteria.rb to no longer hard-code _type.

### DIFF
--- a/lib/mongoid/criteria.rb
+++ b/lib/mongoid/criteria.rb
@@ -292,7 +292,7 @@ module Mongoid
         args.unshift(:_id)
       end
       if klass.hereditary?
-        super(*args.push(:_type))
+        super(*args.push(klass.discriminator_key))
       else
         super(*args)
       end
@@ -555,8 +555,7 @@ module Mongoid
     # @since 3.0.3
     def type_selectable?
       klass.hereditary? &&
-        !selector.keys.include?("_type") &&
-        !selector.keys.include?(:_type)
+        !selector.keys.include?(self.discriminator_key)
     end
 
     # Get the selector for type selection.
--- a/lib/mongoid/criteria.rb
+++ b/lib/mongoid/criteria.rb
@@ -169,7 +169,7 @@ module Mongoid
     # @since 2.0.0
     def field_list
       if options[:fields]
-        options[:fields].keys.reject{ |key| key == "_type" }
+        options[:fields].keys.reject{ |key| key == klass.discriminator_key }
       else
         []
       end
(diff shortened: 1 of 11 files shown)


Neilshweky:
It seems like the `selector_with_type_selection` function is never used... Not sure why it's still there, but it's the only place that `type_selectable?` is used... I changed it anyway to use `#discriminator_key` instead of `_type` and I guess it doesn't hurt to leave in extra functions, I just find it weird that they're there (and `api private`).

lib/mongoid/criteria.rb
```diff
@@ -555,8 +555,7 @@ def merge_type_selection
     # @since 3.0.3
     def type_selectable?
       klass.hereditary? &&
-        !selector.keys.include?("_type") &&
-        !selector.keys.include?(:_type)
+        !selector.keys.include?(self.discriminator_key)
```
p-mongo, line 558: This still needs to account for string and symbol keys.

p-mongo:
`selector_with_type_selection` is currently not used by anything. However, `type_selectable?` *is* used by Mongoid.

lib/mongoid/criteria.rb
```diff
@@ -292,7 +292,7 @@ def only(*args)
         args.unshift(:_id)
       end
       if klass.hereditary?
-        super(*args.push(:_type))
+        super(*args.push(klass.discriminator_key))
```
p-mongo, line 295: This was changed from symbol to string.

lib/mongoid/factory.rb
```diff
@@ -7,8 +7,6 @@ module Mongoid
   module Factory
     extend self
 
-    TYPE = "_type".freeze
```
p-mongo, line 10: This constant needs to stay here with `@deprecated` tag.

spec/mongoid/document_spec.rb
```diff
@@ -881,27 +881,60 @@ class Manager < Person
...
+          it "sets the class type" do
+            expect(person._type).to eq("Person")
+          end
 
-        it "sets the class type" do
-          expect(person._type).to eq("Person")
+          it "raises an error when inappropriate class is provided" do
+            expect {
+              manager.becomes(String)
+            }.to raise_error(ArgumentError)
```
p-mongo, line 904: Can you add a regexp match on the exception message here please?

spec/mongoid/inspectable_spec.rb
```diff
@@ -32,6 +32,33 @@
       it "displays field aliases" do
         expect(inspected).to include("t(test):")
       end
+
+      it "displays the default discriminator key" do 
+        expect(inspected).to include("_type: \"Person\"")
```
p-mongo, line 37: Can you replace the string with e.g. `%q,_type: "Person",` please.

lib/mongoid/criteria.rb
Neilshweky: do you want me to put a `.to_sym` here?

lib/mongoid/criteria.rb
Neilshweky: I fixed this, but I don't understand why it was needed. `self.discriminator_key` is going to equal whatever the user set the discriminator key, whether it's a symbol or a string, no?

PR #4824, opened 2020-07-18
MONGOID-4907 Add test cases for traversing arrays as hashes and hashes as arrays when embedded matching


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4907

PR #4825, opened 2020-07-18
MONGOID-4929 Test passing $options to $regex when embedded matching


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4929

PR #4826, opened 2020-07-19
MONGOID-4931 use #discriminator_key function instead of _type

Neilshweky:
This ticket is for updating the use of the _type field to use the #discriminator_key function instead.


lib/mongoid/criteria.rb
```diff
@@ -292,7 +292,7 @@ def only(*args)
         args.unshift(:_id)
       end
       if klass.hereditary?
-        super(*args.push(:_type))
+        super(*args.push(klass.discriminator_key.to_sym))
```
Neilshweky, line 295: It seems like everything in this array is a symbol...

spec/integration/discriminator_key_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+
+      let(:child) do 
+        PreGlobalIntDiscriminatorChild.new
+      end
+
+      it "creates a field with new discriminator key" do 
+        expect(child.test).to eq("PreGlobalIntDiscriminatorChild")
+      end
+      
+      it "does not have the default value as a field" do
```
p-mongo, line 106: :+1:

PR #4827, opened 2020-07-20
MONGOID-4932 convert symbol to string

Neilshweky:


lib/mongoid.rb
```diff
@@ -111,4 +111,18 @@ def client(name)
   #
   # @since 1.0.0
   def_delegators Config, *(Config.public_instance_methods(false) - [ :logger=, :logger ])
+
+
+  # Module used to prepend the discriminator key assignment function to change 
+  # the value assigned to the discriminator key to a string.
+  module GlobalDiscriminatorKeyAssignment
+    def discriminator_key=(value)
+      value = value.to_s
```
Neilshweky, line 120: do I need validation here?

p-mongo: Let's validate discriminator key is not one of reserved  field names, I suggest doing that in a separate PR.

PR #4828, opened 2020-07-20
MONGOID-4933 add validation on assignment of discriminator key for reserved keywords

Neilshweky:
I added a new error type because the `InvalidField` error requires a class that includes Mongoid::Document, so I needed an error type to raise globally, i.e. in Mongoid.


p-mongo:
`InvalidField` also appears to include diagnostics which I am not seeing in the exception you added here, what happened to that?

spec/mongoid/traversable_spec.rb
```diff
@@ -770,5 +770,85 @@ class PreLocalSymDiscriminatorChild < PreLocalSymDiscriminatorParent
         end
       end
     end
+
+    context "when setting an invalid discriminator key on the class level", :focus do
```
p-mongo, line 774: :focus should not be here right?

spec/mongoid/traversable_spec.rb
```diff
@@ -770,5 +770,85 @@ class PreLocalSymDiscriminatorChild < PreLocalSymDiscriminatorParent
         end
       end
     end
+
+    context "when setting an invalid discriminator key on the class level", :focus do 
+      context "when duplicate_fields_exception is true" do
+        before do
+          Mongoid.duplicate_fields_exception = true
```
p-mongo, line 777: Where is this reset back to default?

Neilshweky:
> `InvalidField` also appears to include diagnostics which I am not seeing in the exception you added here, what happened to that?

The location method: 
```
def location(klass, name)
  @location ||=
          (klass.instance_method(name).source_location || [ "Unknown", 0 ])
end
```
needs to know where the instance method is defined on that klass... However, `Mongoid` in and of itself is not a class, and thus does not implement those instance methods...

spec/mongoid/traversable_spec.rb
```diff
@@ -770,5 +770,85 @@ class PreLocalSymDiscriminatorChild < PreLocalSymDiscriminatorParent
...
+          end
+
+          class GlobalDuplicateDiscriminatorKeyChild < GlobalDuplicateDiscriminatorKeyParent
+          end
+        end
+
+        it "raises an error" do
+          expect do 
+            GlobalDuplicateDiscriminatorKeyParent.field("dkey")
+          end.to raise_error(Mongoid::Errors::InvalidField)
```
p-mongo, line 831: Why is this an error?

spec/mongoid/traversable_spec.rb
Neilshweky: Good call, my bad

spec/mongoid/traversable_spec.rb
Neilshweky: Because on creating a child class that inherits from the parent, a field is created called `dkey` because it is the discriminator key. Then when creating another field called `dkey` it triggers the duplicate field error.

spec/mongoid/traversable_spec.rb
Neilshweky: yeah my bad.

lib/mongoid/traversable.rb
```diff
@@ -27,8 +29,7 @@ def discriminator_key=(value)
         end
 
         if value
-          # I don't need to do any sort of validation here because it will be done
-          # on field creation.
+          Mongoid::Fields::Validators::Macro.validate(self, value, {})
```
Neilshweky, line 32: Validating here gives an undesired outcome, in my opinion... If a user sets the discriminator key to a field that already exists, it shouldn't fail, it should do nothing... That's why we have the `if` statement online the 47 of this file... Do you agree @p ?

Neilshweky: Especially because this is going to be validated when the field is added on line 49... I guess the only edge case is when creating an invalid discriminator key before the child class is instantiated, and then the validation doesn't happen until it does get instantiated...

Alternatively, I can do what I did in `Mongoid` and throw an `InvalidDiscriminatorKey` error on the field name being in `Mongoid.destructive_fields`.

spec/mongoid/traversable_spec.rb
p-mongo: :+1:

lib/mongoid.rb
```diff
@@ -116,7 +116,14 @@ def client(name)
   # Module used to prepend the discriminator key assignment function to change 
   # the value assigned to the discriminator key to a string.
   module GlobalDiscriminatorKeyAssignment
+    # This class is used for obtaining the method definition location for
+    # Mongoid methods.
+    class InvalidFieldHost
```
p-mongo, line 121: Can you also `@api private` this please?

lib/mongoid.rb
```diff
@@ -116,7 +116,14 @@ def client(name)
   # Module used to prepend the discriminator key assignment function to change 
   # the value assigned to the discriminator key to a string.
   module GlobalDiscriminatorKeyAssignment
```
p-mongo, line 118: And this module.

lib/mongoid/fields/validators/macro.rb
```diff
@@ -61,28 +62,44 @@ def validate_relation(klass, name, options = {})
...
         # @example Check the field name.
-        #   Macro.validate_name(Model, :name)
+        #   Macro.validate_field_name(Model, :name)
         #
         # @param [ Class ] klass The model class.
         # @param [ Symbol ] name The field name.
         #
         # @raise [ Errors::InvalidField ] If the name is not allowed.
         #
         # @since 3.0.0
```
p-mongo, line 77: `@api private`

spec/mongoid/traversable_spec.rb
```diff
@@ -770,5 +770,94 @@ class PreLocalSymDiscriminatorChild < PreLocalSymDiscriminatorParent
...
+          class GlobalDuplicateDiscriminatorKeyParent
+            include Mongoid::Document
+          end
+
+          class GlobalDuplicateDiscriminatorKeyChild < GlobalDuplicateDiscriminatorKeyParent
+          end
+        end
+
+        after do 
+          Mongoid.duplicate_fields_exception = true
```
p-mongo, line 833: false?

spec/mongoid/traversable_spec.rb
```diff
@@ -770,5 +770,94 @@ class PreLocalSymDiscriminatorChild < PreLocalSymDiscriminatorParent
...
+          Mongoid.discriminator_key = "_type"
+        end
+
+        it "raises an error" do
+          expect do 
+            GlobalDuplicateDiscriminatorKeyParent.field("dkey")
```
p-mongo, line 839: Can you spell this with `class_eval`?

p-mongo:
```
GlobalDuplicateDiscriminatorKeyParent.class_eval do
  field("dkey") 
end
```

PR #4830, opened 2020-07-21
MONGOID-4934 discriminator key documentation

Neilshweky:


docs/tutorials/mongoid-documents.txt
```diff
@@ -1147,6 +1147,109 @@ collection. An additional attribute ``_type`` is stored in order to make sure wh
 from the database the correct document is returned. This also holds true for the embedded
 documents ``Circle``, ``Rectangle``, and ``Shape``.
 
+Changing the Discriminator Key
+------------------------------
+Mongoid now supports changing the discriminator key from the default ``_type``. There are a few
```
p-mongo, line 1152: You can remove "now". We use the "Mongoid now supports ..." style in upgrade guide/release notes but not in the main body of documentation. Here we can just say "Mongoid supports ...".

docs/tutorials/mongoid-documents.txt
```diff
@@ -1147,6 +1147,109 @@ collection. An additional attribute ``_type`` is stored in order to make sure wh
...
+   end
+
+Here a call to the ``discriminator_key=`` setter was added to the parent class. Now, on 
+creation of a Rectangle or Circle, a ``shape_type`` field will be added. 
+
+Note that the discriminator key can only be modified in the parent class, and an error
+will be raised if trying to set it on the child class.
+
+If the discriminator key is changed after the child class is created, a new field is
+added with the new discriminator key value, and the old field will remain unchanged.
```
p-mongo, line 1192: :+1:

docs/tutorials/mongoid-documents.txt
```diff
@@ -1147,6 +1147,109 @@ collection. An additional attribute ``_type`` is stored in order to make sure wh
...
+     field :width, type: Float
+     field :height, type: Float
+   end
+
+After setting the global discriminator key, all classes will use ``_the_type`` as
+the discriminator key and will not contain a ``_type`` field.
+
+Note that similarly to defining the discriminator key on the class level, when defining
+the discriminator key on the global level, it must be defined before the child class is defined
+for the new discriminator key to register in that class.
```
egiurleo, line 1247: Could you be a bit more clear on what it means for the discriminator key to "register"? Clearer phrasing might be "...for child class to use the global discriminator key" or something along those lines.

This note is more about personal preference, but I would remove the part that says "similarly to defining the discriminator key on the class level" just because takes longer to read the sentence, but I leave the final decision up to you.

PR #4831, opened 2020-07-22
MONGOID-4937 Test double and triple negation with $not when embedded matching


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4937

PR #4832, opened 2020-07-22
MONGOID-4791 Support logical operators under $elemMatch when embedded matching


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4791

PR #4836, opened 2020-07-24
MONGOID-4941 Keep track of discriminator_mapping in Parent

Neilshweky:


lib/mongoid/traversable.rb
```diff
@@ -54,18 +53,38 @@ class << self
...
       class_attribute :discriminator_key, instance_accessor: false
       class_attribute :discriminator_value, instance_accessor: false
-      self.discriminator_value = self.name
-
+            
       class << self
         delegate :discriminator_key, to: ::Mongoid
         prepend DiscriminatorAssignment
+        
+        # @api private
```
p-mongo, line 69: Can the type be added here please  for our own internal consumption.

lib/mongoid/traversable.rb
```diff
@@ -54,18 +53,38 @@ class << self
...
+        attr_accessor :discriminator_mapping
+      end
+
+      # Add a discriminator mapping to the parent class. This mapping is used when
+      # receiving a document to identify its class.
+      #
+      # @api private
```
p-mongo, line 76: Add parameter types here also please.

lib/mongoid/traversable.rb
```diff
@@ -54,18 +53,38 @@ class << self
...
+      def self.add_discriminator_mapping(value, klass=self)
+        self.discriminator_mapping ||= {}
+        self.discriminator_mapping[value] = klass
+        superclass.add_discriminator_mapping(value, klass) if hereditary?
+      end
+      
+      # Get the discriminator mapping from the parent class
```
p-mongo, line 83: Parameter and return type doc please.

lib/mongoid/traversable.rb
```diff
@@ -54,18 +53,38 @@ class << self
...
+      #
+      # @api private
+      def self.get_discriminator_mapping(value)
+        self.discriminator_mapping[value]
```
p-mongo, line 87: Is this method expected to do more in the future?

Neilshweky: nope, this is it, I believe... I did in the past so that's why I had it there... But I honestly kind of like it instead of dealing with the discriminator_mapping hash directly

Neilshweky: Scratch that, I did have to add a check that discriminator_mapping exists (i.e. when it's not hereditary)

lib/mongoid/traversable.rb
```diff
@@ -54,18 +53,47 @@ class << self
...
-      self.discriminator_value = self.name
-
+            
       class << self
         delegate :discriminator_key, to: ::Mongoid
         prepend DiscriminatorAssignment
+        
+        # @api private
+        #
+        # @return [ Hash ] The current mapping of discriminator_values to classes
```
p-mongo, line 71: `Hash<String, Class>` ?

PR #4837, opened 2020-07-24
MONGOID-4942 Use #discriminator_mapping attribute instead of class.name

Neilshweky:


lib/mongoid/document.rb
```diff
@@ -353,7 +353,7 @@ def instantiate(attrs = nil, selected_fields = nil)
       #
       # @since 1.0.0
       def _types
-        @_type ||= (descendants + [ self ]).uniq.map(&:to_s)
+        @_type = (descendants + [ self ]).uniq.map { |type| type.discriminator_value || type.to_s }
```
p-mongo, line 356: What happened to `||=` here?

p-mongo: And shouldn't discriminator_value always be defined?

Neilshweky: The `||=` was removed on purpose... Emily and I worked through this together... There was a problem with the test beginning on line 132 of `discriminator_value_spec.rb`... Basically when `count` is called a `View` is created which contains a specific filter i.e. `{ "_type" => "Circle` }`. The problem is when changing the discriminator key/value and calling count again, this filter is not changed because the `@type` was cached, so the wrong count would be returned. Removing the `||=` allows the filter to be calculated when it needs to be.

Neilshweky: The `discriminator_value` is not always defined, I've recently discovered. Specifically in the test in `document_spec.rb` line 108... In this test a descendant of `Person` is created using `Class.new(Person)`, and in this new class the discriminator value is null and the test expects something of the form "<Class:.....>".

p-mongo: The `@_type` is a cache, so if you need the value to be recalculated every call you should remove the assignment to `@_type` and the clear in traversable.rb.

Neilshweky: How do I clear the `@_type` cache?

Neilshweky: I added a _clear_types method that clears the `@_type` cache.

lib/mongoid/traversable.rb
```diff
@@ -28,6 +28,8 @@ def discriminator_key=(value)
           raise Errors::InvalidDiscriminatorKeyTarget.new(self, self.superclass)
         end
 
+        self._clear_types if self.respond_to?(:_clear_types)
```
p-mongo, line 31: Why is the respond_to check here?

lib/mongoid/traversable.rb
```diff
@@ -53,19 +55,37 @@ class << self
...
+
     included do
       class_attribute :discriminator_key, instance_accessor: false
       class_attribute :discriminator_value, instance_accessor: false
             
       class << self
         delegate :discriminator_key, to: ::Mongoid
         prepend DiscriminatorAssignment
-        
+        prepend DiscriminatorRetrieval
```
p-mongo, line 87: I believe this does not need a prepend, so `include` should suffice?

lib/mongoid/traversable.rb
```diff
@@ -93,8 +113,10 @@ def self.add_discriminator_mapping(value, klass=self)
       #
       # @api private
       def self.get_discriminator_mapping(value)
-        self.discriminator_mapping[value]
+        self.discriminator_mapping[value] if self.discriminator_mapping
```
p-mongo, line 116: The docstring needs to be updated to indicate that nil is a possible return value, and when this would happen.

lib/mongoid/traversable.rb
Neilshweky: I don't understand what this means?

lib/mongoid/traversable.rb
Neilshweky: 👍

lib/mongoid/traversable.rb
Neilshweky: This function is often called before _clear_types (or even _types) is declared resulting in failure... I figured since _types doesn't even exist yet then calling _clear_types doesn't do anything at all.

lib/mongoid/traversable.rb
Neilshweky: How do I use `include` to achieve this functionality?

lib/mongoid/document.rb
```diff
@@ -353,7 +353,19 @@ def instantiate(attrs = nil, selected_fields = nil)
...
+      end
+
+      # Clear the @_type cache. This is generally called when changing the discriminator
+      # key/value on a class.
+      #
+      # @example Get the types.
+      #   document._clear_types
+      #
+      # @api private
+      def _clear_types
```
p-mongo, line 366: Can this be `_mongoid_clear_types` please?

p-mongo:
Please use unique class names so that the following does not happen:

```
/home/w/apps/mongoid/spec/mongoid/traversable_spec.rb:1215: warning: already initialized constant NewClassPerson
/home/w/apps/mongoid/spec/mongoid/traversable_spec.rb:1215: warning: previous definition of NewClassPerson was here
/home/w/apps/mongoid/spec/mongoid/traversable_spec.rb:1216: warning: already initialized constant NewClassPerson2
/home/w/apps/mongoid/spec/mongoid/traversable_spec.rb:1216: warning: previous definition of NewClassPerson2 was here
```

p-mongo:
evergreen retry

PR #4840, opened 2020-07-27
MONGOID-4893 Add Range tests for embedded matchers


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4893

PR #4843, opened 2020-07-29
Fix MONGOID-4952 Unable to access embedded fields in projected query results

wolfoo2931:
In case __selected_fields contains a key which is the same as assoc_key the filtered variable will be set to {"" => 1}.

I could not exactly find out where the value of __selected_fields comes from (where it is computed). So I do not exactly know why the assoc_key is contained in the __selected_fields.

---

From @egiurleo: Mongoid 7.1 raises an exception when a user attempts to access an embedded field on a projected query result. This is because of a bug in the `_mongoid_filter_selected_fields` method. From my understanding, this method is meant to filter which fields from an embedded document can be accessed on a projected query result in the format `{ 'field.name' => 1 }. The problem was that the method would add a blank key to the filter results when the user tried to access the embedded field.


egiurleo:
Hi, @wolfoo2931! Thank you for the patch. I'd love if you could provide us with a bit more information:
* Could you explain the exact problem this patch fixes?
* Could you provide a code snippet that demonstrates the problem? This could be in the form of a failing test.

johnnyshields:
@egiurleo this is a critical issue for us as well, it is a regression from 7.0 --> 7.1

For us `Reservation.only(:orders)` (where orders is embeds_many) broke.

wolfoo2931:
@egiurleo Thanks to @johnnyshields I was able to add a test that reproduces the bug.
I hope it helps.

johnnyshields:
@wolfoo2931 tthank you. @egiurleo @p-mongo please make this a priority to review & merge.

egiurleo:
evergreen retry

egiurleo:
@wolfoo2931 I slightly amended your fix and added some more tests. Do my changes suit your use case?

p-mongo:
I changed save to save! and relation to association, in the entire file while I was on it.

I also adjusted the fix to be more in line with what the method is supposed to do and added a note that hopefully makes it more understandable.

wolfoo2931:
@p-mongo I can confirm the updated path still solves our problem.

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4952

p-mongo:
4.2 and lower servers allow projecting an association and its fields in the same query, revised the fix to support that behavior.

johnnyshields:
@p-mongo thank you for reviewing behavior of past versions on this issue.

PR #4845, opened 2020-07-30
MONGOID-4842 use drivers query cache implementation

gjchong25:


lib/mongoid/query_cache.rb
```diff
@@ -18,7 +18,11 @@ class << self
       #
       # @since 4.0.0
       def cache_table
-        Thread.current["[mongoid]:query_cache"] ||= {}
+        if defined?(Mongo::QueryCache)
```
johnnyshields, line 21: Mongo is a hard dependency for Mongoid. Why do we need an if-statement to if it's defined? We should always use Mongo::QueryCache

gjchong25: Mongo's QueryCache was only recently implemented and available on certain versions, so this check would make sure Mongoid can fall back on its own implementation of QueryCache if the driver's doesn't exist.

johnnyshields: How about we just bump the minimum dependency version for `mongo` gem then keep the Mongoid code clean? It should not be a problem for 99.9% of users.

johnnyshields: Ah I see you have a discussion with Emily in the JIRA ticket. It's ok. Not how I would do it but ok.

johnnyshields: Should raise a ticket as a reminder to remove this code in the Mongoid minor version (7.2) and bump the `mongo` gem dependency at that time.

PR #4846, opened 2020-07-30
MONGOID-4947 include block and pattern params in any?

Neilshweky:
Apologies for all the whitespace changes in the diff.

### DIFF
--- a/lib/mongoid/association/referenced/has_many/enumerable.rb
+++ b/lib/mongoid/association/referenced/has_many/enumerable.rb
@@ -224,6 +224,8 @@ module Mongoid
             #
             # @return [ true, false ] If the enumerable has any elements.
             def any?
+              return super if block_given?
+
               if _loaded?
                 in_memory.count > 0
               else
--- a/lib/mongoid/association/referenced/has_many/enumerable.rb
+++ b/lib/mongoid/association/referenced/has_many/enumerable.rb
@@ -219,12 +219,17 @@ module Mongoid
             # Does the enumerable have any? Will determine if there exists at least
             # one element in the enumerable, whether or not it is _loaded.
             #
+            # Note that in Enumerable's any? method if both a pattern and a block are
+            # passed in, only the pattern is used. I implemented the same functionality
+            # here.
+            #
             # @example Doess the enumerable have any?
             #   enumerable.any?
             #
             # @return [ true, false ] If the enumerable has any elements.
-            def any?
-              return super if block_given?
+            def any?(pattern=nil, &block)
+              return super(pattern) if pattern
+              return super(&block) if block_given?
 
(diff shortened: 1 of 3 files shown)


johnnyshields:
This is a regression introduced in 7.1. Please merge soon, thanks!

lib/mongoid/association/referenced/has_many/enumerable.rb
```diff
@@ -219,11 +219,18 @@ def empty?
...
+            # here.
+            #
             # @example Doess the enumerable have any?
             #   enumerable.any?
             #
             # @return [ true, false ] If the enumerable has any elements.
-            def any?
+            def any?(pattern=nil, &block)
+              return super(pattern) if pattern
+              return super(&block) if block_given?
```
p-mongo, line 232: Both of these calls should forward both arguments to super, i.e. I believe they can be simplified to 

```
return super if pattern || block_given?
```

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
```diff
@@ -256,6 +256,111 @@
...
+        end
+
+        it "returns false when the predicate is false" do
+          expect(
+            enumerable.any? { |doc| false }
+          ).to be false
+        end
+      end
+
+      context "when an option is given" do
```
p-mongo, line 291: s/option/argument/ and in the below tests

p-mongo:
@johnnyshields Which Mongoid version is behaving correctly and which version is behaving incorrectly?

johnnyshields:
@p-mongo version 7.0.x is fine. Also version 3.x is fine. I haven't tested 4/5/6

lib/mongoid/association/referenced/has_many/enumerable.rb
Neilshweky: This doesn't seem to work, actually.

egiurleo: When you say it doesn't work, what behavior are you observing?

Neilshweky: The block is not being passed along in the super call

egiurleo: Thanks for explaining!

p-mongo: You still need to pass both arguments in both cases, you could try `super(pattern, &block)` if needed.

Neilshweky:
@p-mongo for some reason this still causes failures... Also I don't really need to pass both arguments in both cases, because in Enumerble's implementation of any? if there is both a pattern and a block, the block is ignored.

PR #4848, opened 2020-07-31
MONGOID-4949 update inheritance documentation to include note about bogus discriminator values

Neilshweky:


p-mongo:
While the change in this PR is good and should be retained, the ticket is actually about having values in the database that do not correspond to any of the types. This PR only says what happens when Circle is queried, it doesn't say how (or whether) one could retrieve documents with bogus type values.

So in addition to the changes here, there should be some language added that says, for example, that when querying using the parent class, all documents in the collection where the discriminator value doesn't map to a known subclass of parent will be returned as instances of the parent class.

Neilshweky:
When querying the parent class of a collection all documents are returned anyway... It's when the child is queried that this matters.

docs/tutorials/mongoid-documents.txt
```diff
@@ -1154,6 +1154,9 @@ documents ``Circle``, ``Rectangle``, and ``Shape``.
     whatever the discriminator value was set to), all other discriminator values will be considered an object
     of the Shape class.
 
+    Similarly, when querying any Parent class, even if a document does not have a discriminator value that
+    belongs to any of its descendants, this document is still returned.
```
p-mongo: I would say this as:

>     Similarly, when querying by parent classes (``Canvas`` in this example), any documents in the collection that  does not have a discriminator value, or whose discriminator value does not map to either the parent or any of its descendants, will be returned as an instance of the parent class.

PR #4852, opened 2020-08-06
MONGOID-4957 Allows nested conflicts on update

vsevolod:
Sometimes we have conflicts inside conflicts modifiers. It returns this error
```bash
  Failure/Error: expect { a.save }.not_to raise_error
       expected no Exception, got #<Mongo::Error::OperationFailure: Updating the path 'bs' would create a conflict at 'bs' (40) (on 127.0.0.1:27017, legacy retry, attempt 1)> with backtrace: ...
```

This patch replaces conflicts to an instance of Modifiers class, so we can use nested conflicts.


johnnyshields:
This spec is a regression going from 3.1 to 7.0. I am not sure about 4.x, 5.x, and 6.x.

spec/mongoid/atomic/modifiers_spec.rb
```diff
@@ -342,6 +342,36 @@
...
+          it "adds conflicts to the conflicts hash" do
+            expect(modifiers).to eq(
+              { "$push" => {
+                "addresses.0.locations" => { '$each' => [{ "street" => "Bond St" }] } },
+                conflicts: { "$push" =>
+                  { "addresses" => { '$each' => [
+                      { "street" => "Oxford St" }
+                    ] }
+                  },
+                  conflicts: { "$push" =>
```
p-mongo, line 364: I don't think I understand this. Why are conflicts nested?

vsevolod: We have three changes for one association `addresses`:
1. `{ "addresses" => { "street" => "Oxford St" } }`
2. `{ "addresses.0.locations" => { "street" => "Bond St" } }`
3. `{ "addresses.0.locations" => { "number" => "164" } }`
It can't be completed into one query, so in modifiers, we have `:conflict` key, which make the different request. But 2 and 3 changes can't complete into one query too. That's why we use nested conflicts and make the third request. I think this is a very rare case, but anyway.

p-mongo: The argument to $each is an array, so I think I expect that that array should include all required attribute sets in it.

johnnyshields: Why couldn't we do instead:

```
                conflicts: { "$push" =>
                  { "addresses" => { '$each' => [
                      { "street" => "Oxford St" }
                    ] },
                    "addresses.0.locations" => { "$each" => [
                        { "number" => "164" }
                      ] }
                    }
                  }
```

Wouldn't this be equivalent?

vsevolod: ok. I relied on this case:
```yaml
{
  "$set"=>
    {"u_at"=>2020-08-16 11:33:09.161392083 UTC,
    "bs.0.u_at"=>2020-08-16 11:33:09.16102952 UTC},
  :conflicts=>
    {"$push"=>
      {"bs.0.cs"=>
        {"$each"=>
          [{"_id"=>BSON::ObjectId('5f39192143c23478de24298e'),
            "c_name"=>"C Name",
            "u_at"=>2020-08-16 11:33:09.160779392 UTC,
            "c_at"=>2020-08-16 11:33:09.160779392 UTC}]},
      "bs"=>
        {"$each"=>
          [{"_id"=>BSON::ObjectId('5f39191043c23478de24298d'),
            "b_name"=>"New B Name",
            "u_at"=>2020-08-16 11:33:09.161229624 UTC,
            "c_at"=>2020-08-16 11:33:09.161229624 UTC}]}}}}
```
There is one `set` and two different `push`. They can't combine into one query. That's why we need to use three requests to complete w/o any errors.
The current test case is forced. I can change `modifiers.push(nested2)` to `modifiers.set(nested2)`. It will be more natural

johnnyshields: @p see test case here: https://github.com/mongodb/mongoid/pull/4866. This should more clearly illustrate the problem.

johnnyshields:
This can be closed in favor of #4866

vsevolod:
This PR will fix the issue https://github.com/mongodb/mongoid/pull/4866

PR #4857, opened 2020-08-13
MONGOID-4969 Remove unnecesary checks

vsevolod:


p-mongo:
@vsevolod Can you please describe the change that this PR is proposing?

johnnyshields:
@p-mongo this is a 4x performance improvement on large documents: https://gist.github.com/vsevolod/28055a93f3ec2539f2582b2c91420990

The dynamic attributes logic being removed here appears to serve no meaningful purpose. It looks like possibly a holdover from earlier Mongoid versions. The `clone_document` method will still include all dynamic attributes so long as they have a setter defined in the document class.

p-mongo:
The
```
Foo.new(attrs)
```

pattern is well established. Changing that to `Foo.new` and individual attribute setting seems like a clear BC break to me. Plus, setting attributes individually may run into practical problems in case several of them need to be processed together for whatever reason by the application.

p-mongo:
I would also like to see a more detailed breakdown of where the time is going as well as actual benchmarks (including source code). For example, the `update!` call may be replaced with `[]=` which potentially would eliminate one hash construction per key.

p-mongo:
I see that the ticket references the benchmark, which is https://gist.github.com/vsevolod/28055a93f3ec2539f2582b2c91420990.

johnnyshields:
@p-mongo what work remains before this can be merged?

p-mongo:
The compatibility concern I mentioned a few comments up needs to be addressed.

johnnyshields:
@p-mongo `Foo.new(attrs)` comes from `ActiveModel::AttributeAssignment`. If you read the code, you see that it uses public method setters under-the-hood. Because our code uses both public method setters and attribute hash assignment, it covers 100% of what `ActiveModel` does as far as setting attributes is concerned.

p-mongo:
I tested the behavior in question in https://github.com/p-mongo/tests/blob/master/mongoid-4969-bench/construct.rb and this PR changes the behavior.

p-mongo:
I made two changes that improve clone efficiency without changing behavior in https://github.com/mongodb/mongoid/pull/4901, also adding tests for current constructor invocation behavior. These changes improve performance by 33%. This PR improves performance by 100% per the referenced benchmark (not the claimed 4x as far as I can see) but changes behavior.

The original motivation for this PR, as I understand it, was support for protected attributes. I commented on how this may be addressed in the future in https://jira.mongodb.org/browse/MONGOID-4969?focusedCommentId=3456981&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3456981.

p-mongo:
Closing per previous comment.

PR #4858, opened 2020-08-13
MONGOID-4522 add tests for #any? not retrieving entire result set, like #exists?

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4522

PR #4865, opened 2020-08-21
MONGOID-4977 Use recent driver APIs in Mongoid QueryCache for compatibility with sessions & modern retryable reads

mdehoog:
https://jira.mongodb.org/browse/MONGOID-4977

The current `Mongoid::QueryCache::View.each` method is in sync with `Mongo::Collection::View::Iterable` from [5 years ago](https://github.com/mongodb/mongo-ruby-driver/commit/2df954f609d0d29b517a27784e0408894e39466b).

Since then, `Iterable` has been updated with a number of new features:
 - `close_query` [support](https://github.com/mongodb/mongo-ruby-driver/commit/e7dce35c2e66a24a954659de433db560abf23e6b)
 - `sessions` [support](https://github.com/mongodb/mongo-ruby-driver/commit/9769cbeea128c5d77e243486306f255cfa543aa7)
 - [support](https://github.com/mongodb/mongo-ruby-driver/commit/fc8d2e9af4907d0a6786d096bf9df3bcf7a7da38) for `write?` or `out?` queries

The primary issue with this is that any queries through the query cache don't utilize sessions. This PR is designed to update the query cache implementation to match the current implementation of `Mongo::Collection::View::Iterable` (see https://github.com/mongodb/mongo-ruby-driver/blob/v2.12.1/lib/mongo/collection/view/iterable.rb#L37-L56).

It adds support for:
 - the `write?` check (skips the query cache)
 - storing the cursor in `@cursor` for `close_query` support
 - calling `client.get_session` and passing the session to `read_with_retry`, `send_initial_query`, and the `CachedCursor` object

Secondly, `Mongoid::QueryCache::CachedCursor.process` has been changed to delegate to the super method in `Mongo::Cursor.process`, which picks up a bunch of changes, most importantly ending the implicit session.

### DIFF
--- a/spec/mongoid/query_cache_spec.rb
+++ b/spec/mongoid/query_cache_spec.rb
@@ -2,6 +2,7 @@
 # encoding: utf-8
 
 require "spec_helper"
+require "support/session_registry" 
 
 describe Mongoid::QueryCache do
 
--- a/spec/mongoid/query_cache_spec.rb
+++ b/spec/mongoid/query_cache_spec.rb
@@ -10,6 +10,16 @@ describe Mongoid::QueryCache do
     Mongoid::QueryCache.cache { spec.run }
   end
 
+  before(:all) do
+    # It is likely that there are other session leaks in the driver
+    # and/or Mongoid that are unrelated to the query cache. Clear the
+    # SessionRegistry at the start of these tests in order to detect leaks that
+    # occur only within the scope of these tests.
+    #
+    # Other session leaks will be detected and addressed as part of RUBY-2391.
+    SessionRegistry.instance.clear_registry
+  end
+
   before do
     SessionRegistry.instance.verify_sessions_ended!
   end
--- a/spec/mongoid/query_cache_spec.rb
+++ b/spec/mongoid/query_cache_spec.rb
@@ -2,7 +2,6 @@
 # encoding: utf-8
 
 require "spec_helper"
-require "support/session_registry" 
 
 describe Mongoid::QueryCache do
 
@@ -21,7 +20,7 @@ describe Mongoid::QueryCache do
     SessionRegistry.instance.clear_registry
   end
 
-  before do
+  after do
     SessionRegistry.instance.verify_sessions_ended!
   end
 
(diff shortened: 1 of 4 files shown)


mdehoog:
This fix still has a bug with cursor `GetMore` operations:
```
vendor/bundle/ruby/2.6.0/gems/mongo-2.12.1/lib/mongo/cursor.rb:152:in `each': Cannot restart iteration of a cursor which issued a getMore
```
Will look a bit deeper when I get a chance

egiurleo:
@mdehoog This is a known issue -- I will make that fix in a follow-up PR.

egiurleo:
@mdehoog @p-mongo I added session leak detection to the query cache tests and fixed another leak as a result. Let me know what you think.

I will be addressing the "Cannot restart iteration of a cursor which issued a getMore" as a separate PR (see MONGOID-4978).

johnnyshields:
Thank you all for work on this 👍

spec/mongoid/query_cache_spec.rb
```diff
@@ -2,6 +2,7 @@
 # encoding: utf-8
 
 require "spec_helper"
+require "support/session_registry"
```
p-mongo, line 5: This shouldn't be needed if spec helper is required?

spec/mongoid/query_cache_spec.rb
```diff
@@ -10,6 +11,20 @@
...
+    # It is likely that there are other session leaks in the driver
+    # and/or Mongoid that are unrelated to the query cache. Clear the
+    # SessionRegistry at the start of these tests in order to detect leaks that
+    # occur only within the scope of these tests.
+    #
+    # Other session leaks will be detected and addressed as part of RUBY-2391.
+    SessionRegistry.instance.clear_registry
+  end
+
+  before do
```
p-mongo, line 24: Shouldn't this be `after` ?

spec/mongoid/query_cache_spec.rb
egiurleo: Yup, good catch!

spec/mongoid/query_cache_spec.rb
egiurleo: Also a good catch! I think I was having trouble running it in an after block at some point, so I switched it and then never switched it back. It works now though.

p-mongo (approved):

PR #4866, opened 2020-08-26
Fix for MONGOID-4957

johnnyshields:
After further investigation, the problem with MONGOID-4957 is how conflicts are handled.

Currently, **conflicts can occur within conflicts**. The test case demonstrates this failure.

In order to fix this, we need to add de-confliction logic within the conflict resolution logic. The logic I've added does this in a reasonably minimal number of queries.

### DIFF
--- a/lib/mongoid/persistable/updatable.rb
+++ b/lib/mongoid/persistable/updatable.rb
@@ -137,11 +137,13 @@ module Mongoid
             coll = collection(_root)
             selector = atomic_selector
             coll.find(selector).update_one(positionally(selector, updates), session: _session)
+
+            # This causes the issue
             conflicts.each_pair do |key, value|
               coll.find(selector).update_one(positionally(selector, { key => value }), session: _session)
             end
 
-            # This code will fix the issue
+            # This code will fix the issue (replace above)
             # conflicts.each_pair do |modifier, values|
             #   inner_conflicts = values.group_by {|key, _| key.split(".", 2)[0] }.values
             #   while batch = inner_conflicts.map(&:pop).compact.to_h.presence
--- a/lib/mongoid/persistable/updatable.rb
+++ b/lib/mongoid/persistable/updatable.rb
@@ -140,6 +140,14 @@ module Mongoid
             conflicts.each_pair do |key, value|
               coll.find(selector).update_one(positionally(selector, { key => value }), session: _session)
             end
+
+            # This code will fix the issue
+            # conflicts.each_pair do |modifier, values|
+            #   inner_conflicts = values.group_by {|key, _| key.split(".", 2)[0] }.values
+            #   while batch = inner_conflicts.map(&:pop).compact.to_h.presence
+            #     coll.find(selector).update_one(positionally(selector, { modifier => batch }), session: _session)
+            #   end
+            # end
           end
         end
       end
--- a/spec/mongoid/persistable/savable_spec.rb
+++ b/spec/mongoid/persistable/savable_spec.rb
@@ -263,37 +263,8 @@ describe Mongoid::Persistable::Savable do
         let!(:crate) { truck.crates.create!(volume: 0.4) }
 
         it 'persists the new documents' do
-          expect(truck.crates.size).to eq 1
-          expect(truck.crates[0].volume).to eq 0.4
-          expect(truck.crates[0].toys.size).to eq 0
-
           truck.crates.first.toys.build(name: "Teddy bear")
           truck.crates.build(volume: 0.8)
-
-          # The following is equivalent to the two lines above:
-          #
-          # truck.crates_attributes = {
-          #   '0' => {
-          #     "toys_attributes" => {
-          #       "0" => {
-          #         "name" => "Teddy bear"
-          #       }
-          #     },
-          #     "id" => crate.id.to_s
-          #   },
-          #   "1" => {
-          #     "volume" => 0.8
-          #   }
-          # }
-
-          expect(truck.crates.size).to eq 2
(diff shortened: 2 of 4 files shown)


johnnyshields:
I've tested this same case on Mongoid 3.1.7. Interestingly, it registers conflicts as well, however `#save` method returns true and the *results are not actually saved*. This indicates that perhaps this was failing silently before and no one noticed it.

johnnyshields:
OK, I think I'm close to figuring this out.

If I have 2+ levels of nesting, i.e. A > B > C, it is impossible to do combined `$push` and/or `$set` operations which create a new B and a new C at the same time. They need to be done as a series of separate operations. More to come.

johnnyshields:
@p please look at this fix

p-mongo:
I added several tests covering deep nested attribute assignment in https://github.com/mongodb/mongoid/pull/4868 after simplifying the models you've proposed to remove pieces not relevant to attribute assignment (timestamps, association polymorphism and callback cascading). No changes to Mongoid code were needed for these tests to pass.

In particular, the test proposed in this PR is now in Mongoid and passes.

If your issue is not resolved on master please rebase this PR on master and add new tests as appropriate.

johnnyshields:
I've rebased this PR. @p-mongo in #4868 you removed timestamps and `cascade_callbacks`  which gave you a false positive. Please check again.

p-mongo:
I added the tests in https://github.com/mongodb/mongoid/pull/4873.

For the fix, I believe the correct fix is to move some of the operations from conflicts to the first update (which is possible in the examples added in https://github.com/mongodb/mongoid/pull/4873) rather than issuing additional writes as this PR proposes.

johnnyshields:
@p-mongo I think we should do **both** approaches:
1. move some of the operations from conflicts to the first update
2. add my conflict management logic.

The failure is fundamentally coming from the DB. Even if you can fix this particular test case, I will make another even more complex test case that will fail without #2.

Let me explain a bit better what this PR does. Supposing I have the following conflicts:

```
1. crates.0.updated_at
2. crates.0.toys.0.name
3. crates.0.toys.1.name
4. dogs.0.name
5. dogs.0.bones.0.color
6. foo
7. bar
```

Currently, all conflicts are being attempted to be sent at once, however, 1+2+3 conflict and 4+5 conflict. My logic groups them into the smallest number of batches possible to send:

```
Batch 1: [
crates.0.updated_at
dogs.0.name
foo
bar
]

Batch 2: [
crates.0.toys.0.name
dogs.0.bones.0.color
]

Batch 3: [
crates.0.toys.1.name
]
```

johnnyshields:
Here is another way to visualize this:
image

p-mongo:
I don't see what is conflicting about those updates at all, and I expect them all to be sent in the same query.

johnnyshields:
The problem is that the DB rejects them. Please experiment with it, you'll see the same thing.

p-mongo:
You haven't provided complete queries, what you expect me to test?

p-mongo:
```
client = Mongo::Client.new(['localhost:14440'])
client['foo'].delete_many
client['foo'].insert_one({})
client['foo'].update_one({},
'$set' => %w(
crates.0.updated_at
crates.0.toys.0.name
crates.0.toys.1.name
dogs.0.name
dogs.0.bones.0.color
foo
bar
).map do |s|
[s, 1]
end.to_h
)
client['foo'].find.first

# => {"_id"=>BSON::ObjectId('5f4c767a2c97a6228afed3e3'), "bar"=>1, "crates"=>{"0"=>{"toys"=>{"0"=>{"name"=>1}, "1"=>{"name"=>1}}, "updated_at"=>1}}, "dogs"=>{"0"=>{"bones"=>{"0"=>{"color"=>1}}, "name"=>1}}, "foo"=>1}
```

johnnyshields:
I'm sorry, I meant test on the **test case in the PR.** The above comment was just a conceptual example, not an actual test case.

johnnyshields:
By the way, I think the issue is with `push`, not `set`. It could be that the logic I'm proposing here is only needed for `push`, I'm not sure.

p-mongo:
The test cases in this PR were merged  in https://github.com/mongodb/mongoid/pull/4873. If you believe there are additional cases that should be tested, please rebase this PR on master and add them.

johnnyshields:
@p-mongo I already did. This PR is rebased. I force-pushed to it.

Please see this diff:
https://github.com/mongodb/mongoid/pull/4866/files

p-mongo:
As I mentioned in the previous comment I already incorporated these tests into Mongoid.

johnnyshields:
No, you did not. Read the [diff](https://github.com/mongodb/mongoid/pull/4866/files) carefully please.

```
# spec/support/models/crate.rb 
  include Mongoid::Timestamps::Short
```

```
# spec/support/models/vehicle.rb 
  embeds_many :crates	  embeds_many :crates, cascade_callbacks: true
```

These lines are what causes the issue.

johnnyshields:
Ah ok, you made a second PR. I see. Well, I'm not sure about that one. Can you please refer to the specs I have here?

johnnyshields:
OK, from your most recently-merged PR it appears you are reproducing the issue correctly.

Again, I think the issue is with how conflicts are handled, and I suggest the fix in this PR. Please try applying my fix and see if it resolves your pending issues (of course, it won't prevent conflicts, but it allows models to be saved when they are conflicted).

p-mongo:
I referenced this PR from https://jira.mongodb.org/browse/MONGOID-4982 so that we can refer to it when we are working on this issue.

PR #4867, opened 2020-08-26
Improve Mongoid::Atomic::Modifiers#push_conflict? regex

johnnyshields:
Currently, Mongoid::Atomic::Modifiers#push_conflict? has a regex which is too loose. The regex should match exactly the first segment of a key.

addresses <--> addresses.0.street should match
addresses <--> dresses should not match


johnnyshields:
Closing in favor of PR #4869

PR #4872, opened 2020-08-29
rectified email link under credit section

nitish-awasthi:
Under credits section-
Email link-**durran at gmail dot com** _changed to_ **<durran@gmail.com>**


egiurleo:
Hi, @nitish-awasthi! The email is actually formatted intentionally to prevent scraping (as much as possible haha), so I'm going to leave it as is. Thank you for the contribution, though!

PR #4873, opened 2020-08-30
Failing tests for MONGOID-4982 Cannot update a field on embedded association member, add another mem...

p-mongo:
ber and add a nested association to the first member in the same save call


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4982

PR #4876, opened 2020-09-14
MONGOID-4978 Stop raising error when performing multi-batch query with QueryCache enabled

egiurleo:
This is a backport of RUBY-2338 into Mongoid ([see corresponding driver PR](https://github.com/mongodb/mongo-ruby-driver/pull/2072)).

There was a bug in the query cache implementation where it would cache _all_ queries, not just single batch queries. Thus, when the user tried to perform the same multi-batch query twice, it would attempt to re-iterate the cached query, which isn't allowed in the Ruby driver, and would raise an error.

### DIFF
--- a/spec/mongoid/query_cache_spec.rb
+++ b/spec/mongoid/query_cache_spec.rb
@@ -460,13 +460,19 @@ describe Mongoid::QueryCache do
 
    context 'when querying colleciton larger than the batch size' do
      before do
-       101.times { Band.create! }
+       Band.destroy_all
+       101.times { |i| Band.create!(_id: i) }
      end
 
      it 'does not raise an exception when querying multiple times' do
        expect do
-         Band.all.to_a
-         Band.all.to_a
+         results1 = Band.all.to_a
+         expect(results1.length).to eq(101)
+         expect(results1.map { |band| band["_id"] }).to eq([*0..100])
+
+         results2 = Band.all.to_a
+         expect(results2.length).to eq(101)
+         expect(results2.map { |band| band["_id"] }).to eq([*0..100])
        end.not_to raise_error
      end
    end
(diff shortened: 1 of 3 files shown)


spec/mongoid/query_cache_spec.rb
```diff
@@ -611,7 +624,7 @@
       Mongoid::QueryCache.enabled = true
       10.times { Band.create! }
 
-      Band.batch_size(4).all.any?
+      Band.batch_size(4).to_a
```
egiurleo, line 627: I found out that calling `any?` with no block was actually causing a session leak. I documented it in [RUBY-2391](https://jira.mongodb.org/browse/RUBY-2391).

p-mongo: Does this leak apply to Mongoid's QC code also?

spec/mongoid/query_cache_spec.rb
```diff
@@ -458,6 +458,19 @@
...
     end
 
+   context 'when querying colleciton larger than the batch size' do
+     before do
+       101.times { Band.create! }
+     end
+
+     it 'does not raise an exception when querying multiple times' do
+       expect do
+         Band.all.to_a
```
p-mongo, line 468: Can these also assert at least the length of arrays returned and ideally contents?

p-mongo (approved):

spec/mongoid/query_cache_spec.rb
egiurleo: Good call! Done.

spec/mongoid/query_cache_spec.rb
egiurleo: Good question. I'll make a separate PR after this one.

p-mongo (approved):

PR #4884, opened 2020-09-21
Reorganize test splits

p-mongo:


p-mongo:
This PR significantly speeds up the test run. I had to adjust some tests that broke due to other tests mutating global state (e.g. model fields and which models are loaded).

egiurleo (approved):
Nice, thank you!

PR #4885, opened 2020-09-21
MONGOID-4756: modify query cache tests to check that query cache is used

egiurleo:
I took all tests that didn't have an `expect_query` call and added one. I also discovered that calling `Xyz.first` queries with a sort by id, so I added a note to MONGOID-4972 about it.

### DIFF
--- a/spec/mongoid/query_cache_spec.rb
+++ b/spec/mongoid/query_cache_spec.rb
@@ -255,24 +255,50 @@ describe Mongoid::QueryCache do
     end
 
     it 'returns all documents' do
-      expect(Person.all.to_a.count).to eq(3)
-      Person.first
-      expect(Person.all.to_a.count).to eq(3)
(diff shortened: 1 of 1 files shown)


spec/mongoid/query_cache_spec.rb
```diff
@@ -255,24 +255,50 @@
     end
 
     it 'returns all documents' do
-      expect(Person.all.to_a.count).to eq(3)
-      Person.first
-      expect(Person.all.to_a.count).to eq(3)
+      # Mongoid adds a sort by _id to the Person.first call, which is why
+      # these commands issue two queries instead of one.
+      expect_query(2) do
```
egiurleo, line 260: I added a note to MONGOID-4972 to document this.

p-mongo: :+1:

p-mongo (approved):

PR #4888, opened 2020-09-24
MONGOID-4972 Document query cache limitations + new driver query cache

egiurleo:


docs/tutorials/mongoid-queries.txt
```diff
@@ -967,14 +978,69 @@ recommended to use the block form described above:
...
+
+The following is a list of limitations of the Mongoid query cache:
+
+* The query cache does not cache query results that exceed the batch size.
+  The default batch size is 1000 documents.
+
+* The query cache does not take into account read preference or read concern
+  when deciding whether to return cached results. When performing the same
+  query with a different read preference or read concern with the query cache
+  enabled, the driver may return incorrect results.
```
p-mongo, line 1018: s/driver/Mongoid/ ? Or I would phrase it as "incorrect results may be returned" to avoid explicitly mentioning either one.

docs/tutorials/mongoid-queries.txt
```diff
@@ -967,14 +978,69 @@ recommended to use the block form described above:
...
+  do not invalidate the query cache. Cached results may be stale after performing
+  any of these operations.
+
+* Aggregation results are not cached.
+
+A Note on using ``#first``
+``````````````````````````
+
+Calling the ``first`` method on a model class imposes an ascending sort by
+the ``_id`` field on the underlying query. This may produce unexpected behavior.
```
p-mongo, line 1030: Suggest "This may produce unexpected behavior with respect to query caching."

docs/tutorials/mongoid-queries.txt
```diff
@@ -967,14 +978,69 @@ recommended to use the block form described above:
...
+However, because of the sort imposed on the second query, both methods
+will query the database and separately cache their results.
+
+.. code-block:: ruby
+
+  Band.all.to_a
+  #=> Queries the database and caches the results
+
+  Band.first
+  #=> Queries the database again because of the sort
```
p-mongo, line 1043: Is `Band.all.limit(1).first` going to produce the first cached document? If yes perhaps add this as a solution to this situation.

p-mongo (approved):

PR #4889, opened 2020-09-24
MONGOID-4956 add query cache information to release notes

egiurleo:


docs/tutorials/mongoid-upgrade.txt
```diff
@@ -362,6 +362,25 @@ Example Mongoid 7.1 behavior:
 Mongoid provides :ref:`sharding management Rake tasks <sharding-management>`
 to shard collections according to shard keys declared in models.
 
+Query Cache Moved to Driver
+------------------------------------
+
+Minor change: The Mongoid query cache has been deprecated in favor of a new
```
p-mongo, line 368: Similarly to the other PR, I suggest starting with the fact that Mongoid will use the driver's QC in 2.14, then listing the benefits, then stating the legacy QC is deprecated, and using the term  "the legacy Mongoid query cache" to refer to it.

Also can this link to the QC section in the docs (will potentially need an anchor added to the section so that it can be linked to).

docs/tutorials/mongoid-upgrade.txt
```diff
@@ -362,6 +362,25 @@ Example Mongoid 7.1 behavior:
...
+The interface for enabling and disabling the query cache in Mongoid has not
+changed. When using driver versions 2.14.0 and newer, this interface will
+enable or disable the query cache in the driver.
+
+The driver query cache is more correct and more effective than the legacy
+Mongoid query cache. If you plan to use the query cache, it is recommended
+that you upgrade to driver version 2.14.
+
+To read more about the query cache improvements made in the driver, see
+`the Ruby driver documentation <https://docs.mongodb.com/ruby-driver/tutorials/query-cache/>`_.
```
p-mongo, line 383: I believe this is a 404, I think https://docs.mongodb.com/ruby-driver/current/tutorials/query-cache/ would work.

egiurleo: Yup, thank you!

p-mongo (approved):
:+1:

PR #4890, opened 2020-09-25
Fix MONGOID-4993 Cleanup code in QC.cached/uncached resets cache enabled to nil when new driver is used


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4993

egiurleo (approved):

PR #4892, opened 2020-09-30
Fix query cache tests after API changes in driver

egiurleo:


spec/mongoid/query_cache_spec.rb
```diff
@@ -726,9 +741,24 @@
       Band.batch_size(4).to_a
     end
 
-    it 'does not cache the result' do
-      expect_query(1) do
-        expect(Band.all.map(&:id).size).to eq(10)
+    context 'with drivre query cache' do
```
p-mongo, line 744: driver is misspelled

p-mongo (approved):

egiurleo:
@p-mongo Sure thing.

spec/mongoid/query_cache_spec.rb
egiurleo: Whoops thanks!

PR #4893, opened 2020-10-02
MONGOID-5000 Add a test for bypassing the query cache on system collections

egiurleo:


spec/support/models/system_role.rb
```diff
@@ -0,0 +1,7 @@
+class SystemRole
+ include Mongoid::Document
+
+  # NOTE: this model is for test purposes only. It is not recommended that you
+  # store Mongoid documents in system collections.
```
egiurleo, line 5: Adding a note here in case someone stumbles upon this code in the wild and thinks this is a reasonable pattern haha

p-mongo: :+1:

p-mongo (approved):

PR #4894, opened 2020-10-03
MONGOID-4948 Use driver version constraints instead of looking for query cache


p-mongo:
https://jira.mongodb.com/browse/MONGOID-4948


egiurleo (approved):
This is great! Thank you.

PR #4896, opened 2020-10-06
Fix MONGOID-5001 Cannot clone a model instance that embeds another model which uses inheritance and is inheritance root


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5001

egiurleo (approved):

PR #4898, opened 2020-10-09
Fix MONGOID-4873 Regexp::Raw usage breaks when Mongoid is loaded


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4873

egiurleo (approved):
This is clever! Just to confirm, the user is never expected to interact directly with the `Raw_` module, right?

p-mongo:
That's right, the module in question is included into BSON::Regexp::Raw.

PR #4900, opened 2020-10-19
Fix method visibility via `Mongoid::Association::Proxy`

kitop:
See https://jira.mongodb.org/projects/MONGOID/issues/MONGOID-5009

When defining belongs_to/has_one associations, the methods in the associated object change visibilty and ruby's `defined?` doesn't detect them.
When doing `parent.child = child`, the associated object is cloned (object_id changes) and while the class, ancestors, and everything looks and feels the same, something happens where the private methods are made public, and public methods are there but not recognized by ruby's `defined?`.
Same happens when doing `child.parent = parent`

After looking for a bit, found the root cause in `Mongoid::Association::Proxy`. It undefines all methods to then send them to `_target` via `method_missing`
https://github.com/mongodb/mongoid/blob/v7.1.4/lib/mongoid/association/proxy.rb#L16-L20
https://github.com/mongodb/mongoid/blob/v7.1.4/lib/mongoid/association/proxy.rb#L130-L138

In there, it's using `send` to call the method in target, that makes it delegate _any_ method to it, including private ones. It should be using `public_send`. That solves the "private_methods_stay_private" case.

In addition to that, when overriding `method_missing` it should override `respond_to_missing?` as well, so ruby knows to find those methods. See https://thoughtbot.com/blog/always-define-respond-to-missing-when-overriding.

Opening this against `7.1-stable` since I was having trouble running tests in master.

### DIFF
--- a/lib/mongoid/association/proxy.rb
+++ b/lib/mongoid/association/proxy.rb
@@ -137,6 +137,10 @@ module Mongoid
         _target.public_send(name, *args, &block)
       end
 
+      def respond_to_missing?(name, *args)
+        _target.respond_to?(name)
+      end
+
       # When the base document illegally references an embedded document this
       # error will get raised.
       #
--- a/lib/mongoid/association/proxy.rb
+++ b/lib/mongoid/association/proxy.rb
@@ -134,7 +134,7 @@ module Mongoid
       # @param [ Array ] args The arguments passed to the method.
       #
       def method_missing(name, *args, &block)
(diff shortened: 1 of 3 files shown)


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5009

p-mongo:
The tests for belongs_to were merged in https://github.com/mongodb/mongoid/pull/4912, the fix will be in https://github.com/mongodb/mongoid/pull/4932.

PR #4903, opened 2020-10-21
Drop collections instead of deleting their contents to fix CI on 2.6 servers

p-mongo:


p-mongo:
I believe https://jira.mongodb.org/browse/RUBY-2432 is the actual problem, hence this PR shouldn't be necessary once 2432 is fixed but the drop/create in tx tests could be useful to keep.

johnnyshields:
Would propose to drop support for MongoDB < 3.0 in the next major Mongoid version (8). Realistically no one should be using < 3.6

p-mongo:
I don't think that is necessary. Most of the work is done for older version support is by the driver.

p-mongo:
The issue was fixed in the driver.

PR #4906, opened 2020-10-26
MONGOID-4913 Implement StringifiedSymbol type

ksadoff:
[This ticket](https://jira.mongodb.org/browse/MONGOID-4913) creates a StringifiedSymbol type in Mongoid, to allow users to work with the Symbol class in Ruby in a way that is interoperable with other drivers. This extension exposes data as the Symbol type to the Ruby developer, while storing the content as a string on MongoDB's end, maintaining compatibility with drivers that do not support Symbol types.


ksadoff:
evergreen restart

ksadoff:
evergreen restart

lib/mongoid/extensions/stringified_symbol.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+module Mongoid
+  module Extensions
+    module StringifiedSymbol
```
p-mongo, line 6: I suggest moving this class directly into `Mongoid` module, "extensions" generally refers to modifying behavior of core/standard library classes.

lib/mongoid/fields.rb
```diff
@@ -573,6 +574,8 @@ def field_for(name, options)
       def unmapped_type(options)
         if "Boolean" == options[:type].to_s
           Mongoid::Boolean
+        elsif "StringifiedSymbol" == options[:type].to_s
```
p-mongo, line 577: Am I correct that all 3 of the following work:

```
field :foo, type: :stringified_symbol
field :foo, type: 'StringifiedSymbol'
field :foo, type: Mongoid::Extensions::StringifiedSymbol
```

If yes, I think the second use isn't really needed given the first and the third ones.

spec/support/models/stringified_symbol.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+class StringifiedSymbol
```
p-mongo, line 4: Can you please change the name of this class to be something else?

spec/support/models/stringified_symbol.rb
```diff
@@ -0,0 +1,8 @@
...
+  include Mongoid::Document
+  store_in collection: "stringified_symbols", client: :other
```
p-mongo, line 6: `store_in` shouldn't be necessary since the default should work just fine.

spec/support/models/stringified_symbol.rb
```diff
@@ -0,0 +1,8 @@
...
+  field :stringified_symbol, type: StringifiedSymbol
```
p-mongo, line 7: I suggest using a field name that isn't the same as the field type also, for example something like `:status` might be a good choice.

spec/integration/stringified_symbol_field_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    let(:string_query) do
+      {'stringified_symbol' => {'$eq' => 'test'}}
+    end
+
+    let(:symbol_query) do
+      {'stringified_symbol' => {'$eq' => :test}}
+    end
+
+    it "finds a symbol" do
```
p-mongo, line 23: Suggest `it "can be queried with a string"` to match the other test case.

p-mongo:
Per our conversation:

- StringifiedSymbol I think should be a class, not a module, per the custom field documentation (https://docs.mongodb.com/mongoid/master/tutorials/mongoid-documents/#custom-fields)
- Remove the string StringifiedSymbol alias
- Rename the test model so that it is a domain entity name, e.g. Order
- Rename the field on the test model to a domain entity field, e.g. status
- Use https://github.com/p-mongo/mongoid/commit/3b68bffd8cbac67267927e7800ba0175743da0dd instead of the current event subscriber code
- Add  tests where the value stored in the database is a number and an array
- Add tests that assign a number and an array to the field
- The value should be stringified in Ruby when it is assigned
- The value read from the database should be stringified then symbolized

lib/mongoid/extensions/stringified_symbol.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      # Convert the object from its mongo friendly ruby type to this type.
+      #
+      # @example Demongoize the object.
+      #   Symbol.demongoize(object)
+      #
+      # @param [ Object ] object The object to demongoize.
+      #
+      # @return [ Symbol ] The object.
+      def demongoize(object)
```
p-mongo, line 17: All of the methods in this class need to have `@api private` on them.

lib/mongoid/extensions/stringified_symbol.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+module Mongoid
+  class StringifiedSymbol
```
p-mongo, line 5: Can you please add a brief description of this class here as  a docstring?

lib/mongoid/fields.rb
p-mongo: I believe we discussed removing this change?

spec/integration/stringified_symbol_field_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+      s.save
+      event = update_events.first
+      expect(event.command["updates"].first["u"]["$set"]["status"]).to eq("other")
+    end
+  end
+
+  context "when reading an integer" do
+
+    before do
+      Order.create(status: 42, _id: 13)
```
p-mongo, line 130: The insert here needs to be done through the driver (i.e. via `insert_one`) as otherwise Mongoid can stringify upon save.

spec/integration/stringified_symbol_field_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    it "receives the value as a symbol" do
+      expect(Order.find(13).status).to eq(:"42")
+    end
+
+    it "saves the value as a string" do
+      s = Order.find(13)
+      s.status = 24
+      s.save
```
p-mongo, line 140: `s.save!`

spec/integration/stringified_symbol_field_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+      event = update_events.first
+      expect(event.command["updates"].first["u"]["$set"]["status"]).to eq("24")
+    end
+  end
+
+  context "when reading an array" do
+    before do
+      Order.create(status: [0, 1, 2], _id: 14)
```
p-mongo, line 148: This also needs to be written through the driver.

p-mongo (commented):
There are two more changes that will need to be made:

- Tutorial documentation added under https://docs.mongodb.com/mongoid/master/tutorials/mongoid-documents/#fields
- Also https://docs.mongodb.com/mongoid/master/tutorials/mongoid-documents/#bson-symbol-type needs to be updated to reference the StringifiedSymbol as the recommended migration path
- The class needs to be aliased from Mongoid::Document

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,31 @@ Types that are not supported as dynamic attributes since they cannot be cast are
 - ``DateTime``
 - ``Range``
 
+The StringifiedSymbol Type
+--------------------------
+
```
egiurleo, line 182: I would add a bit of background explaining that using the `Symbol` field type will cause Mongoid to serialize a field into `BSON::Symbol`s, but that `BSON::Symbol` is deprecated (basically give a bit more background about why someone would want to use this type).

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,31 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+This type allows users to write Ruby code that makes use of the built-in Ruby Symbol type,
```
egiurleo, line 183: I would just say "This type allows users to make use of the built-in..." (since it's assumed that anybody using Mongoid is writing Ruby code heh)

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,31 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+while ensuring interoperability with other drivers. The StringifiedSymbol type stores all data
+on the MongoDB backend as a String, while returning values to the user as a Symbol.
```
egiurleo, line 185: I don't think we use the term "MongoDB backend" -- I would use "the database" instead.

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,31 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+
+For example, when setting a field to a string value:
```
egiurleo, line 187: I'm a bit confused by this sentence -- you're setting the field to a StringifiedSymbol value in the example below, right?

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,31 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+
+    field :status, type: StringifiedSymbol
+  end
+
+  post = Post.new(status: "hello")
+  post.status # => :hello
+
+All values will be stringified upon being sent to the database, and all values will be converted
+to symbols when returned to the user. So, in the above example, setting ``:status`` to ``42``
+would result in ``"42"`` being stored in the database, and ``:"42"``
```
egiurleo, line 202: This is really clear! Can you change the symbol to just be `:42` rather than `:"42"`?

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,31 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+being returned to the user.
+
```
egiurleo, line 204: Two more things that I'd love to see in this documentation:
1. Could you add a sentence or two about the fact that this type will lazily migrate any BSON::Symbols already stored in the database to BSON::Strings when the user updates the document being stored? (Let me know if you have questions about how this works.)
2. Could you add a link to the [BSON Symbol Type](https://docs.mongodb.com/mongoid/current/tutorials/mongoid-documents/#bson-symbol-type) section of the docs for anybody who wants to learn more about using BSON::Symbols with Mongoid?

egiurleo: There are examples of how to link things throughout the docs but it's a bit finicky so let me know if you need any help with that.

egiurleo (changes requested):
Great start to the docs! I left a few comments (happy to chat about those if they aren't clear).

lib/mongoid/extensions.rb
```diff
@@ -38,6 +38,7 @@ def transform_keys
 require "mongoid/extensions/regexp"
 require "mongoid/extensions/set"
 require "mongoid/extensions/string"
+require "mongoid/extensions/stringified_symbol"
```
p-mongo, line 41: Can you move this file up one level please?

spec/support/models/order.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+# encoding: utf-8
+
+class Order
+  include Mongoid::Document
+  field :status, type: Mongoid::StringifiedSymbol
+end
```
p-mongo, line 7: Please add a newline a the end of file.

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,38 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+.. code-block:: ruby
+
+  class Post
+    include Mongoid::Document
+
+    field :status, type: StringifiedSymbol
+  end
+
+  post = Post.new(status: "hello")
```
egiurleo, line 201: In this example, could you pass "hello" as a symbol, since that's probably how the user would want to use it?

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,38 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+The StringifiedSymbol Type
+--------------------------
+
+When using the built-in Symbol field type, Mongoid defaults to serializing the field into the BSON::Symbol
+type. For more information on the BSON::Symbol type, see
+`here <https://docs.mongodb.com/mongoid/master/tutorials/mongoid-documents/#bson-symbol-type>`_.
+BSON::Symbol is deprecated, so the StringifiedSymbol type allows users to write code
+that makes use of the built-in Symbol type while ensuring interoperability with other drivers.
+The StringifiedSymbol type stores all data on the database as a String, while returning values to the
+user as a Symbol.
```
egiurleo, line 189: I would say "to the application" instead of the "user" here.

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,38 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+    include Mongoid::Document
+
+    field :status, type: StringifiedSymbol
+  end
+
+  post = Post.new(status: "hello")
+  post.status # => :hello
+
+All values will be stringified upon being sent to the database, and all values will be converted
+to symbols when returned to the user. This includes values that cannot be converted directly to
```
egiurleo, line 205: user --> application

egiurleo: And could you be more specific about what would happen to values that cannot be converted to symbols?

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,38 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+symbols, such as integers and arrays.
+
+This type will lazily migrate any BSON::Symbols already stored in the database to BSON::Strings. If
```
egiurleo, line 208: I think this is up to personal preference but I would say "perform a lazy migration" instead of "lazily migrate" 😆

egiurleo: Also, could you stylize any class names with code? I.e. ``BSON::Symbol`` instead of BSON::Symbol. (It's two ticks `` in rst instead of just one like github).

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,38 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
+the user changes their model to use StringifiedSymbol instead of Symbol, the documents will be saved
```
egiurleo, line 209: I would try to use indirect language as much as possible, so instead of saying "if the user does xyz," say "If the ``StringifiedSymbol`` type is applied to a field that contains ``BSON::Symbol`` data..."

docs/tutorials/mongoid-documents.txt
p-mongo: :+1:

docs/tutorials/mongoid-documents.txt
p-mongo: I agree that the first example would make more sense with `:hello`, but this example could be copied/moved under the next paragraph that talks about values being converted to symbols when returned to application.

docs/tutorials/mongoid-documents.txt
p-mongo: I think adding an example where `42` is assigned to the status and the returned value would be `:42` would do the trick.

docs/tutorials/mongoid-documents.txt
p-mongo: :+1:

docs/tutorials/mongoid-documents.txt
ksadoff: The way we have mongoize and demongoize implemented right now (`object.to_s` and `object.to_s.to_sym`, respectively) results in `:"42"` being returned, so that's what I've shown in the new example.

docs/tutorials/mongoid-documents.txt
```diff
@@ -709,6 +755,8 @@ For cases when you do not want to have ``BSON::ObjectId`` ids, you can override
 BSON Symbol Type
 ----------------
 
+TODO: ADD IN INFO ABOUT LAZY MIGRATION USING STRINGIFIEDSYMBOL
```
egiurleo, line 758: I'm assuming you want to remove this?

docs/tutorials/mongoid-documents.txt
```diff
@@ -176,6 +177,51 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
 - ``Range``
 
+The StringifiedSymbol Type
+--------------------------
+
+When using the built-in ``Symbol`` field type, Mongoid defaults to serializing the field into the ``BSON::Symbol``
+type. For more information on the ``BSON::Symbol`` type, see
+`here <https://docs.mongodb.com/mongoid/master/tutorials/mongoid-documents/#bson-symbol-type>`_.
+``BSON::Symbol`` is deprecated, so the ``StringifiedSymbol`` type allows the use of the built-in ``Symbol``
+type while ensuring interoperability with other drivers. The ``StringifiedSymbol`` type stores all data
```
egiurleo, line 187: I would make this sentence a bit more direct by saying something like, "Use the ``StringifiedSymbol`` field type in order to use the built-in ``Symbol`` type in an application while ensuring..." If there is a best practice, we should tell the user this explicitly.

egiurleo (approved):
This is looking great! I left two more comments on the docs but I'm happy with it once those are addressed.

p-mongo (approved):

docs/tutorials/mongoid-documents.txt
```diff
@@ -215,23 +212,23 @@ An example usage is shown below:
...
-
-Values which are stored in the database as symbols will be returned to the
-application as symbols, but will be stored as strings at the next save.
-This permits transparent lazy migration from fields that currently store either
-strings or BSON symbols in MongoDB to the ``StringifiedSymbol`` field type.
+  post.status
+  # => :"42"
 
+If the ``StringifiedSymbol`` type is applied to a field that contains ``BSON::Symbol``
+data, the documents will be stored with ``String`` values instead of the deprecated ``BSON::Symbol``
```
p-mongo, line 229: It is not technically correct to speak about data stored in the database in terms of Ruby types. The correct language to use would refer to BSON types. Thus, "BSON symbol" and not `BSON::Symbol` for example.

docs/tutorials/mongoid-documents.txt
```diff
@@ -177,22 +177,19 @@ Types that are not supported as dynamic attributes since they cannot be cast are
...
-BSON symbols is difficult in programming languages that do not have a native
-symbol type. The ``StringifiedSymbol`` field type stores values as strings
-in MongoDB, thus enabling easy interoperability with software written in other
-programming languages, but exposes the values as instances of ``Symbol`` in
-Ruby programs.
+values that should be exposed as symbols to Ruby applications. When using the ``Symbol`` field type,
+Mongoid defaults to storing values as BSON symbols. For more information on the
+``BSON::Symbol`` type, see :ref:`here <bson-symbol>`.
+However, the ``BSON::Symbol`` type is deprecated and is difficult to work with in programming languages
+without native symbol types, so the ``StringifiedSymbol`` type allows the use of symbols
```
p-mongo, line 190: Same here, other languages do not have a `BSON::Symbol` type since this type is Ruby driver-specific.

egiurleo (approved):

p-mongo (approved):

PR #4911, opened 2020-11-11
Fix MONGOID-5002 parent.child = Child.new raises TypeError when there's no inverse relation for belongs_to with has_one parent


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5002

ksadoff (approved):

PR #4913, opened 2020-11-12
Fix MONGOID-5013 Re-associating identical object destroys itself for dependent: :destroy


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5013

lib/mongoid/association/referenced/has_one/proxy.rb
```diff
@@ -54,9 +54,14 @@ def nullify
...
           # @since 2.0.0.rc.1
           def substitute(replacement)
+            # If the same object currently associated is being assigned,
+            # rebind the association and save the target but do not destroy
+            # the target.
+
             unbind_one
             if persistable?
-              if _association.destructive?
+              # TODO can this entire method be skipped if self == replacement?
```
ksadoff, line 63: Is this for a future ticket or were you planning on doing it for this one?

ksadoff (approved):

lib/mongoid/association/referenced/has_one/proxy.rb
p-mongo: It's a note to whoever looks at/changes this method next but I have no plans to investigate this in the foreseeable future.

p-mongo: embeds_one I believe skips the entire method but I am not sure which behavior is more correct since I don't yet have an in-depth understanding of binding logic in Mongoid.

PR #4914, opened 2020-11-12
MONGOID-5018 test cases


p-mongo:
https://jira.mongodb.com/browse/MONGOID-5018

### DIFF
--- a/spec/integration/associations/has_one_spec.rb
+++ b/spec/integration/associations/has_one_spec.rb
@@ -67,6 +67,54 @@ describe 'has_one associations' do
     end
   end
 
+  context 'when calling methods on target' do
+    let(:parent) do
+      HomCollege.create!.tap do |college|
+        HomAccreditation.create!(college: college)
+      end
+    end
+
+    shared_examples 'delegates to the field' do |reloaded: false|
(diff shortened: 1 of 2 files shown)


durran:
@p-mongo My assumption was that the tests would fail when on the ticket it stated was reproducible. Was this in a different environment other that the test quite that is was reproduced?

p-mongo:
@durran  If I understand your question correctly, the failing test cases are marked pending.

spec/integration/associations/has_one_spec.rb
```diff
@@ -67,6 +67,54 @@
...
+
+      context 'field name that conflicts with Kernel' do
+        it 'delegates to the field' do
+          parent.accreditation.format.should == 'fmt'
+        end
+
+        context 'using send' do
+          it 'delegates to the field' do
+            if reloaded
+              pending 'MONGOID-4018'
```
p-mongo, line 98: This one is pending.

durran: Ahhh. I missed that. Thanks.

durran (approved):
LGTM

PR #4917, opened 2020-11-13
MONGOID-4906 Implement bitwise operators

ksadoff:


lib/mongoid/matcher/bits_all_clear.rb
```diff
@@ -0,0 +1,24 @@
...
+        # p condition
+        case condition
+        when Array
+        #  array of bits
+        # TODO
+        when BSON::Binary
+          #condition is a binary string
+          # TODO: add logic to convert to integer
+        when Integer
+          value & condition
```
p-mongo, line 17: I think this would need to be `value & condition > 0` or `value & condition == 0` depending on the desired condition as 0 is truthy in Ruby.

lib/mongoid/matcher/type.rb
```diff
@@ -0,0 +1,86 @@
...
+          !!value === value
+        when 9
+          # Date
+          Date === value
+        when 10
+          # Null
+          NilClass === value
+        when 11
+          # Regex
+          Regexp::Raw === value
```
p-mongo, line 47: This *probably* should also accept value being of class `::Regexp` (the Ruby stdlib one) but if you are not sure let's discuss.

lib/mongoid/matcher/type.rb
```diff
@@ -0,0 +1,86 @@
...
+          Binary === value
+        when 6
+          # Undefined
+          BSON::Undefined === value
+        when 7
+          # ObjectId
+          BSON::ObjectId === value
+        when 8
+          # Boolean
+          !!value === value
```
p-mongo, line 38: If the value is supposed to be of a boolean type I believe the correct expression would be `TrueClass === value || FalseClass === value`.

lib/mongoid/matcher/type.rb
```diff
@@ -0,0 +1,86 @@
...
+        when 9
+          # Date
+          Date === value
```
p-mongo, line 41: This is most likely supposed to be a timestamp type, therefore  I expect `Time === value` to be more correct and potentially `Time === value || DateTime === value || Date === value` could be appropriate also.

lib/mongoid/matcher/type.rb
```diff
@@ -0,0 +1,86 @@
...
+          String === value
+        when 3
+          # Object
+          Object === value
+        when 4
+          # Array
+          Array === value
+        when 5
+          # Binary data
+          Binary === value
```
p-mongo, line 29: `BSON::Binary` ?

lib/mongoid/matcher/type.rb
```diff
@@ -0,0 +1,86 @@
...
+          Regexp::Raw === value
+        when 12
+          # DBPointer deprecated
+          DBPointer === value
+        when 13
+          # JavaScript
+          Code === value
+        when 14
+          # Symbol deprecated
+          Symbol === value
```
p-mongo, line 56: We also have `BSON::Symbol::Raw` that I think should be allowed here.

lib/mongoid/matcher/type.rb
ksadoff: There is a Timestamp check later on, so I'll add that logic there.

p-mongo (commented):
It looks like all four matchers are essentially the same, such that they can be expressed by one class that does almost all of the work and four wrapper classes that specify the operator (== or != 0) and the any/all flag.

p-mongo (approved):

p-mongo:
I added test cases for negative integers, negative floats and fixed bits any clear.

PR #4923, opened 2020-11-18
MONGOID-4906 Implement $type

ksadoff:


lib/mongoid/matcher/type.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    # @see https://docs.mongodb.com/manual/reference/operator/query/type/
+    #
+    # @api private
+    module Type
+      module_function def matches?(exists, value, condition)
+        case condition
+        when Array
+          condition.any? do |v|
+            condition = v
```
p-mongo, line 12: This doesn't look right.

lib/mongoid/matcher/type.rb
```diff
@@ -0,0 +1,84 @@
...
+          end
+        when 1
+          # Double
+          Float === value
+        when 2
+          # String
+          String === value
+        when 3
+          # Object
+          Object === value
```
p-mongo, line 22: This should probably be `Hash` and not `Object`.

lib/mongoid/matcher/type.rb
```diff
@@ -0,0 +1,84 @@
...
+          BSON::ObjectId === value
+        when 8
+          # Boolean
+          TrueClass === value || FalseClass === value
+        when 9
+          # Date
+          Date === value || Time === value || DateTime === value
+        when 10
+          # Null
+          BSON::NilClass === value
```
p-mongo, line 43: Just `NilClass`

p-mongo (approved):
I added a few test cases and fixed the following:

- arrays with multiple elements as $type arguments
- null type when the field being checked is not present
- accept Ruby integers in int32/int64 matches if they are in the right range

spec/integration/matcher_operator_data/type_object.yml
```diff
@@ -0,0 +1,25 @@
+- name: existing field - matches ObjectId
```
ksadoff, line 1: I think the filename for this and `type_object_id.yml` were swapped accidentally.

p-mongo (approved):

PR #4929, opened 2020-11-28
MONGOID-4812 Allow id to be a regular field in models (not an alias of _id)

p-mongo:
https://jira.mongodb.com/browse/MONGOID-4812


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4812

johnnyshields:
@p-mongo we might want to reconsider the API here in light of the request not to have an `_id` field at all (I think that was mainly for embedded docs)

p-mongo:
What are you suggesting exactly?

johnnyshields:
I guess the other one would be a separate PR. Nevermind.

durran (approved):
LGTM

PR #4931, opened 2020-12-01
MONGOID-5019 Document omitting _id field from embedded documents


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5019

johnnyshields:
I am skeptical that removing `_id` from embedded documents in this manner magically works. Have you tested updating and deleting such documents? How does Mongoid / MongoDB know which doc to delete (`pull`) if there's no `_id`?

p-mongo:
@johnnyshields  I don't see where in this PR I am claiming that updating and deleting work.

If you need this functionality and it is not working, please feel free to create a Jira ticket and we will look into it.

johnnyshields:
If it doesn't work, don't you think you should put a caveat in the documentation? The wording you have in this PR doesn't make any mention or warning about possible side-effects.

p-mongo:
The acceptance criteria for this PR were given in the referenced ticket. 

I am unaware of any open tickets that would be related to the functionality described to this PR that can be referenced as caveats.

Again, if you have a specific issue with Mongoid, please create a Jira ticket and we will look into it. If you are curious about whether a MongoDB-supported product is appropriate for your use case, and you wish someone to perform this research for you, MongoDB provides [professional services](https://www.mongodb.com/products/consulting) that do this.

johnnyshields:
This is a comment directly about the documentation changes **in this PR**. There is no reason to raise a separate ticket.

**In this PR**, you are advising users on how they can remove the ID field, without mentioning potential side effects (broken functionality, etc.) in doing so. All I am asking is to improve this documentation by adding the appropriate warnings.

durran:
I agree with both sides here as there are side effects of doing what is documented here, however just showing how to do it was the scope of the ticket. I think it would probably be best to open a new ticket that asks to update the documentation to provide caveats of actually doing it. (Ie, updating arrays of embedded documents would not work)

durran (approved):

johnnyshields:
@durran documentation which advises users to do dangerous things should clearly explain to users what the risks are. I'm stunned that the Mongoid team only considers the "scope of the ticket" and not the broader context/risks. Shaking my head as to why you approved this PR in its current form.

Nevertheless I have raised a new ticket to rectify this. https://jira.mongodb.org/browse/MONGOID-5025

PR #4933, opened 2020-12-06
MONGOID-5019 note that _id is only assigned for top-level documents


p-mongo:
https://jira.mongodb.com/browse/MONGOID-5019


durran (approved):
LGTM

PR #4934, opened 2020-12-07
Fix MONGOID-4982 Cannot update a field on embedded association member, add another member and add a nested association to the first member in the same save call

johnnyshields:
Fixes MONGOID-4982

### DIFF
--- a/spec/mongoid/persistable/savable_spec.rb
+++ b/spec/mongoid/persistable/savable_spec.rb
@@ -270,30 +270,12 @@ describe Mongoid::Persistable::Savable do
           truck.crates.first.toys.build(name: "Teddy bear")
           truck.crates.build(volume: 0.8)
 
-          # The following is equivalent to the two lines above:
-          #
-          # truck.crates_attributes = {
-          #   '0' => {
-          #     "toys_attributes" => {
-          #       "0" => {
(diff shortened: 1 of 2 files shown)


spec/mongoid/persistable/savable_spec.rb
```diff
@@ -270,31 +270,13 @@
...
-          #         "name" => "Teddy bear"
-          #       }
-          #     },
-          #     "id" => crate.id.to_s
-          #   },
-          #   "1" => {
-          #     "volume" => 0.8
-          #   }
-          # }
-
```
p-mongo, line 288: Why is this removed?

johnnyshields: It's just a dead-weight comment. Why do we need it?

p-mongo:
https://jira.mongodb.org/browse/MONGOID-4982

durran (approved):
One thing we should look into for the future is to clean up all the conflicting updates logic once Mongoid has a minimum mongodb version of 3.6. With the array updates feature we can possible get rid of all this. But for now LGTM.

p-mongo:
We have a follow-up ticket to this in https://jira.mongodb.org/browse/MONGOID-5026, I transferred your comment there. thanks for pointing this out.

PR #4935, opened 2020-12-16
MONGOID-5027 Rails 6.1 support

francelwebdev:
Hi,
I use rails 6.1 and mongoid and i get error when i run bundle. So i update the dependency to activemodel <= 6.1 .

### DIFF
--- a/mongoid.gemspec
+++ b/mongoid.gemspec
@@ -32,7 +32,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = ">= 2.3"
   s.required_rubygems_version = ">= 1.3.6"
 
-  s.add_dependency("activemodel", [">= 5.1", "<=6.1"])
+  s.add_dependency("activemodel", [">= 5.1", "<6.2"])
   s.add_dependency("mongo", ['>=2.10.5', '<3.0.0'])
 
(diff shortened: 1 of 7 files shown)


mongoid.gemspec
```diff
@@ -32,7 +32,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = ">= 2.3"
   s.required_rubygems_version = ">= 1.3.6"
 
-  s.add_dependency("activemodel", [">= 5.1", "<6.1"])
+  s.add_dependency("activemodel", [">= 5.1", "<=6.1"])
```
khustochka, line 35: I would suggest making it "< 6.2", because if it is compatible with 6.1, it will most likely be compatible with 6.1.1 and so on.

p-mongo:
https://jira.mongodb.org/browse/MONGOID-5027

alcaeus (approved):

mongoid.gemspec
francelwebdev: OK. thanks.

alcaeus:
@francelwebdev was the pull request closed in error? Changes looked good and would ensure support for Rails 6.1 in upcoming Mongoid versions.

PR #4937, opened 2020-12-20
Test cases for MONGOID-4476 Writes in callbacks of nested embedded documents are not persisted when top-level document is modified via update_attributes

p-mongo:
https://jira.mongodb.com/browse/MONGOID-4476


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4476

PR #4941, opened 2021-01-05
Replace parent with module_parent_name

PanosCodes:
When trying to run `bin/rails g mongoid:config` on a rails 6.1+ app results in the following error

```/usr/local/bundle/gems/railties-6.1.0/lib/rails/railtie.rb:209:in `method_missing': undefined method `parent' for App::Application:Class (NoMethodError)```

Since Rails 6.0 the parent method is replaced with module_parent
https://github.com/rails/rails/pull/34051

### DIFF
--- a/lib/rails/generators/mongoid/config/config_generator.rb
+++ b/lib/rails/generators/mongoid/config/config_generator.rb
@@ -15,7 +15,7 @@ module Mongoid
       end
 
       def app_name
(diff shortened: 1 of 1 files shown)


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5042

p-mongo:
This was merged as https://github.com/mongodb/mongoid/pull/4960, thank you.

PR #4942, opened 2021-01-06
Fix MONGOID-5035 Query embedded documents by date field with $gt/gte/lt/lte does not find existing documents


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5035

alcaeus (approved):

PR #4944, opened 2021-01-07
::I18n.translate takes keyword arguments

reidmorrison:
Ruby v3 is failing with the following exception since :I18n.translate takes the options as keyword arguments and not a hash:
~~~
ArgumentError: wrong number of arguments (given 2, expected 0..1)
            /Users/rmorrison/.rvm/gems/ruby-3.0.0/gems/i18n-1.8.7/lib/i18n.rb:195:in `translate'
            /Users/rmorrison/.rvm/gems/ruby-3.0.0/gems/mongoid-7.2.0/lib/mongoid/errors/mongoid_error.rb:51:in `translate'
            /Users/rmorrison/.rvm/gems/ruby-3.0.0/gems/mongoid-7.2.0/lib/mongoid/errors/mongoid_error.rb:66:in `translate_problem'
            /Users/rmorrison/.rvm/gems/ruby-3.0.0/gems/mongoid-7.2.0/lib/mongoid/errors/mongoid_error.rb:25:in `compose_message'
            /Users/rmorrison/.rvm/gems/ruby-3.0.0/gems/mongoid-7.2.0/lib/mongoid/errors/document_not_found.rb:32:in `initialize'
            /Users/rmorrison/.rvm/gems/ruby-3.0.0/gems/mongoid-7.2.0/lib/mongoid/reloadable.rb:26:in `new'
            /Users/rmorrison/.rvm/gems/ruby-3.0.0/gems/mongoid-7.2.0/lib/mongoid/reloadable.rb:26:in `reload'
~~~

### DIFF
--- a/lib/mongoid/errors/mongoid_error.rb
+++ b/lib/mongoid/errors/mongoid_error.rb
@@ -48,7 +48,7 @@ module Mongoid
       #
       # @return [ String ] A localized error message string.
       def translate(key, options)
-        ::I18n.translate("#{BASE_KEY}.#{key}", options)
+        ::I18n.translate("#{BASE_KEY}.#{key}", **options)
       end
 
       # Create the problem.
--- a/lib/mongoid/validatable/associated.rb
+++ b/lib/mongoid/validatable/associated.rb
@@ -43,7 +43,7 @@ module Mongoid
         ensure
           document.exit_validate
         end
-        document.errors.add(attribute, :invalid, options) unless valid
+        document.errors.add(attribute, :invalid, **options) unless valid
       end
     end
   end
--- a/lib/mongoid/validatable/presence.rb
+++ b/lib/mongoid/validatable/presence.rb
@@ -34,15 +34,15 @@ module Mongoid
             document.errors.add(
               attribute,
               :blank_in_locale,
-              options.merge(location: _locale)
+              **options.merge(location: _locale)
             ) if not_present?(_value)
           end
         elsif document.relations.has_key?(attribute.to_s)
           if relation_or_fk_missing?(document, attribute, value)
-            document.errors.add(attribute, :blank, options)
+            document.errors.add(attribute, :blank, **options)
           end
         else
-          document.errors.add(attribute, :blank, options) if not_present?(value)
+          document.errors.add(attribute, :blank, **options) if not_present?(value)
         end
       end
 
--- a/lib/mongoid/validatable/uniqueness.rb
+++ b/lib/mongoid/validatable/uniqueness.rb
@@ -68,7 +68,7 @@ module Mongoid
       # @since 2.4.10
       def add_error(document, attribute, value)
         document.errors.add(
-          attribute, :taken, options.except(:case_sensitive, :scope).merge(value: value)
+          attribute, :taken, **options.except(:case_sensitive, :scope).merge(value: value)
         )
       end
 


mateusnava:
UP!

p-mongo:
Mongoid is not currently tested against Ruby 3, and given that the driver doesn't support Ruby 3 officially Mongoid support for Ruby 3 is still some ways away. This PR can remain open but I expect to address it as part of Mongoid Ruby 3 work.

johnnyshields:
If specs are passing, what is the harm merging this as-is?

p-mongo:
Two reasons:

1. This change can break existing applications. Most changes have some risk of breaking existing applications, and the changes are made because they improve behavior and the improvement outweighs the potential risk. But, since this change doesn't fix any failing tests, what improvement this change provides isn't clear.
2. Since this change doesn't have any tests, if it gets broken at a future time the breakage won't be detected.

johnnyshields:
@reidmorrison can you add tests here, and make sure they pass on all supported Ruby / Rails versions?

toddnestor:
@p-mongo this also fixes deprecation warnings in Ruby 2.7 which is a supported version of Ruby according to the Readme, would that justify getting this merged sooner than the Ruby 3 work?

CamiMeneses:
Please merge it. 
This is how it looks now. It is just a 'find' and it's not working, it happens when the id is not correct and the second one it's when you try to create with validations but you are not passing the data. 
I have tried with his changes and it fixes these terrible problems.
Screen Shot 2021-04-16 at 5 31 54 PM

johnnyshields:
Let's get a few tests first to demonstrate nothing is breaking on older versions.

hanishjadala:
As Ruby 3 is supported by Mongoid, what is the next previous version of mongoid which has not issues and which rails version it supports?

p-mongo:
If there is a clear demonstration of an issue that occurs on Ruby 2.x and this PR fixes it, I'll consider merging it sooner.

midnight-wonderer:
For those waiting for the fixes, I wanted to point out that you can monkey patch the methods yourself as a workaround.

I don't use mongoid for validations, so the only patch I need is:
~~~ruby
require 'mongoid'

module Mongoid
  module Errors
    class MongoidError < StandardError
      def translate(key, options)
        ::I18n.translate("#{BASE_KEY}.#{key}", **options)
      end
    end
  end
end
~~~
The patch looks relatively safe since the `translate` method is dedicated to this very purpose.

Apart from this, mongoid seems to work well on my Ruby 3 application.
I don't use many features of mongoid, though. (no relation, no validation, etc.)

johnnyshields:
Hi all, lets please **ADD TESTS** to this PR. I hope that MongoDB team will be reasonable and merge the PR if it has tests.

midnight-wonderer:
I read it differently.
If what p-mongo said represents MongoDB team decision. They have practically declined the pull request, just keeping the public relations. And I am sure they already have someone on their team looking into this issue already.

With the former comment, maybe more tests are needed, and they can merge the PR.  
The latter one is quite clear what the intention was.
Merging this PR might require someone on their team to rebase. Adding tests might require that same one to merge the test case.

Given what he knows, this is the best action he suggested.  
I would recommend just adding some workaround for development for now and wait for the announcement.
Someone can try, but the likelihood of merging this PR isn't there.

johnnyshields:
@midnight-wonderer Mongo team will accept this PR when they choose to support Rails 3. Hopefully they come to their senses and either make that decision soon, or decide to accept this PR in advance of that. Anyway basic tests don't take much effort, so lets add them.

techbelle:
Hi all, I thought I might weigh in here and lend some clarity on the Ruby team position on the PR. Currently we try to not accept PRs without adequate tests in them as a policy, which seems reasonable enough of a position (see also the current UMinnesota insanity with the Linux kernel)... but otherwise the quarter before a major server release is always incredibly difficult for the team - even the things that we _really_ want to work on, or investigate, we do not have the time to do that because of the influx of new feature support for MongoDB server.

I can absolutely assure this group that Ruby 3.0 support is top of mind and on our schedule, and following that will be the mongoid related items for Ruby 3.0. Currently that work is slated to begin in July 2021, which I realize is a very very long time from now for folks impacted by this problem - but if there is any way to start it before then, we will.

Rachelle

reidmorrison:
Without the above changes the existing Mongoid tests should fail on Ruby v3. If not then there is clearly a test coverage problem that is outside the scope of this fix.
With these fixes the tests pass when running against both older and newer versions of Ruby.
Since these fixes do not change any behavior then the existing test coverage should be sufficient across all supported versions of Ruby, including Ruby v3.

For anyone that needs to run Ruby v3 with Mongoid 7.2 you can point your Gemfile to the source repo:
```ruby
gem "mongoid", git: "https://github.com/reidmorrison/mongoid", branch:"7.2-ruby_3"
```

p-mongo:
@reidmorrison Am I understanding your last comment correctly that with this change the entire Mongoid test suite is green on Ruby 3?

johnnyshields:
@p-mongo if this change is passing on the supported Mongoid versions, let's please merge it.

p-mongo:
There is no demonstration of the issue provided in this PR. Such a demonstration can be added as a failing test or as a code snippet demonstrating the problem using, for example, https://github.com/p-mongo/tests/blob/master/mongoid-5081/test.rb as a template.

As already mentioned earlier, we will add Ruby 3 support at a future time. If you are requesting expedited review of this particular PR please help us do so by providing a clear demonstration of the issue that this PR addresses.

johnnyshields:
OK, here's your Ruby 3.0 PR which includes the changes from this one:
https://github.com/mongodb/mongoid/pull/5001

p-mongo:
Superseded by #5009.

PR #4954, opened 2021-01-22
use https protocol for spec/shared submodule

awesome:
fixes inability of users to update submodules when they do not have an SSH key configured for GitHub


p-mongo:
Thank you for the fix.

p-mongo:
I did not intend for the description to be clobbered here, it was a side effect of retitling via my tooling.

PR #4959, opened 2021-03-01
MONGOID-5047 Take Rails versions from Evergreen configuration in app tests


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5047

PR #4961, opened 2021-03-11
fix: Compatibility with keyword arguments in Ruby 3.0.0

mateusnava:
In this Pull Request I fixed some incompatibilities with Ruby 3.0.

Reference: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/


mateusnava:
Closed, because there is other pull request with the same changes - https://github.com/mongodb/mongoid/pull/4944/files

PR #4962, opened 2021-03-15
Fix MONGOID-5060 #delete causes dependent associations to be destroyed


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5060

docs/tutorials/mongoid-upgrade.txt
```diff
@@ -23,6 +23,48 @@ Upgrading to Mongoid 7.3
...
+Previous versions of Mongoid invoked dependent behavior when deleting parents.
+
+To invoke dependent behavior, use the ``destroy`` method:
+
+.. code-block:: ruby
+  
+  # Deletes the album from the database
+  band.destroy
+
+The behavior of Mongoid 7.3 is consistent with how ActiveRecord behaves.
```
alexbevi, line 65: > The behavior of Mongoid 7.3 is consistent with how ActiveRecord behaves.

The behavior of Mongoid 7.3 is consistent with how [ActiveRecord associations](https://guides.rubyonrails.org/association_basics.html) behave.

alexbevi (approved):

docs/tutorials/mongoid-upgrade.txt
p-mongo: I prefer to not link to AR docs when not necessary to avoid link rot as well as sending users away from our documentation.

PR #4967, opened 2021-03-23
Fix MONGOID-5068 ElemMatch matcher operates on pre-stringified keys in condition verification path against impossible values

dalton-braze:
`Mongoid::Matcher::ElemMatch.matches?()` didn't consistently convert input keys within its condition to Strings. This PR doesn't include a new unit test verifying this behavior because `ElemMatch` is clearly out of sync with the `k = k.to_s` pattern used in other `lib/Mongoid/matcher/*.rb` files. Existing examples of correctly using `k = k.to_s` in other of these Matcher sub-modules are:
* [expression.rb](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/matcher/expression.rb)
* [field_expression.rb](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/matcher/field_expression.rb)
* [not.rb](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/matcher/not.rb)

### DIFF
--- a/lib/mongoid/matcher/elem_match.rb
+++ b/lib/mongoid/matcher/elem_match.rb
@@ -15,7 +15,8 @@ module Mongoid
           # Validate the condition is valid, even though we will never attempt
           # matching it.
           condition.each do |k, v|
-            if k.to_s.start_with?('$')
+            k = k.to_s
+            if k.start_with?('$')
               begin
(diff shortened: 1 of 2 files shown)


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5068

PR #4970, opened 2021-03-23
MONGOID-5032 fix the rest of issues


p-mongo:
https://jira.mongodb.com/browse/MONGOID-5032

### DIFF
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -253,11 +253,22 @@ module Mongoid
     private
 
     def selection_excluded?(name, selection, field)
-      selection[name] == 0
+      path = name.split('.')
+
+      selection.find do |k, included|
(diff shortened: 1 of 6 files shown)


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5032

p-mongo:
This PR extracts the logic of determining whether an attribute is allowed to be accessed due to configured projections into a new Projector class.

Subsequently, this logic is expanded to permit sub-paths in both directions for inclusionary projections (a projection of foo.bar allows foo to be accessed in Mongoid), and prohibit same and deeper paths for exclusionary projections (an exclusionary projection of foo.bar permits accessing foo and prohibits accessing foo.bar.*). 

Additionally, the Projector handles all integer values, not just 0 and 1, and true and false as projection values, per the server behavior.

alexbevi (approved):
LGTM

PR #4974, opened 2021-03-25
Fix nil preventing fallbacks on localized fields

sentience:
In #3490, a fix to prevent falling back to another locale when the current locale's value was `false` introduced a regression where the value `nil`, which was previously treated as a missing translation, became accepted as a translated value too.

This PR assumes that was an unwanted change, fixes it, and introduces a regression test.

### DIFF
--- a/lib/mongoid/fields/localized.rb
+++ b/lib/mongoid/fields/localized.rb
@@ -87,7 +87,7 @@ module Mongoid
         end
         return value unless value.nil?
         if fallbacks? && ::I18n.respond_to?(:fallbacks)
-          object[::I18n.fallbacks[locale].map(&:to_s).find{ |loc| object.has_key?(loc) }]
+          object[::I18n.fallbacks[locale].map(&:to_s).find{ |loc| !object[loc].nil? }]
         end
       end
     end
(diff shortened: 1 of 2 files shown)


p-mongo:
Hi @sentience , thank you for the PR.

I reviewed https://github.com/ruby-i18n/i18n/wiki/Fallbacks and I don't see any mention there that a nil translation should be treated as a missing one. If you believe this is the correct behavior, can you please reference documentation where this is stated?

johnnyshields:
@p-mongo even if undocumented, we should preserve the existing behavior prior to #3490. Like @sentience, my app is assuming that `nil` should get a fallback.

p-mongo:
I understand that the behavior of Mongoid appears to have changed in the referenced PR, but, this change was made in 2014 which is 7 years ago. There can be many users relying on the current behavior.

As such, to change the behavior now, I believe the new behavior needs to be demonstrated as being superior to the current behavior, even if the new behavior happens to be the same as what the behavior was 8 years ago.

johnnyshields:
@p-mongo my apologies, you are correct.

p-mongo:
@sentience I will close this PR for the time being, happy to reopen if you can provide the additional information requested.

PR #4978, opened 2021-03-31
Fix MONGOID-5059 `set` nilifies attributes excluded via `only`


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5059

alexbevi (approved):
LGTM

PR #4982, opened 2021-04-09
Fix MONGOID-5076 reload fails to find document when the document's sharded field has been updated

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5076

alexbevi (approved):
LGTM

PR #4983, opened 2021-04-16
Use double splat (**) when sending options toActiveModel::Errors#add method

toddnestor:
Gets rid of a few of these deprecation notices:

```
warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```


toddnestor:
This is actually handled in #4944 I see, closing this PR.

PR #4984, opened 2021-04-23
MONGOID-4668 Add test for regexp field

comandeo:
If a document has a field with type Regexp, it is still possible to assign strings to this field. In this case the field still should be persisted as regexp.

### DIFF
--- a/spec/mongoid/document_fields_spec.rb
+++ b/spec/mongoid/document_fields_spec.rb
@@ -87,13 +87,6 @@ describe Mongoid::Document do
   end
 
   describe 'Regexp field' do
-    class Template
-      include Mongoid::Document
-
-      field :name, type: String
-      field :pattern, type: Regexp
-    end
-
     it 'persists strings as regexp' do
       mop = Mop.create!(regexp_field: 'foo')
       expect(mop.regexp_field).to be_a Regexp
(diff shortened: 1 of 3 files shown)


spec/mongoid/document_fields_spec.rb
```diff
@@ -85,4 +85,19 @@
       end
     end
   end
+
+  describe 'Regexp field' do
+    class Template
```
p-mongo, line 90: It is my understanding that class definitions in Ruby are global (even though this one appears to be scoped to the rspec group). Therefore I suggest one of the following instead:

- Renaming the class to have the scope in its name, e.g. DocumentFieldsSpecTemplate.
- Moving it to document_fields_models.rb file in the current directory, with a namespaced name like DocumentFieldsModels::Template or DfmTemplate.
- Putting the class into our (huge) models directory.
- Reusing an existing class from the models directory without modifications to that class (additions can be OK depending on what they are).

comandeo: Good catch, thanks! I used one of existing classes from models.

p-mongo:
Could you also use either https://docs.mongodb.com/manual/reference/operator/query/type/ or https://docs.mongodb.com/manual/reference/operator/aggregation/type/ to assert the type of the value stored in the database?

p-mongo (approved):

PR #4986, opened 2021-04-27
Fix MONGOID-5054 $eq or $regex must be used in place of implicit equality based on value type


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5054

comandeo (approved):

PR #4987, opened 2021-04-27
MONGOID-4757 Raise warning for unknown config params

comandeo:


lib/mongoid/clients/factory.rb
```diff
@@ -59,12 +60,20 @@ def default
...
         raise Errors::NoClientsConfig.new unless configuration
-        if configuration[:uri]
-          Mongo::Client.new(configuration[:uri], options(configuration))
+        config = configuration.dup
+        uri = config.delete(:uri)
+        database = config.delete(:database)
+        hosts = config.delete(:hosts)
+        opts = config.delete(:options) || {}
+        unless config.empty?
+          default_logger.warn("Unknown config options detected: #{config}. Please check your config files.")
```
p-mongo, line 69: The options can come from a hash, not necessarily a file, right? If so I suggest removing the second sentence and just leaving the first one.

lib/mongoid/clients/factory.rb
```diff
@@ -78,9 +87,8 @@ def driver_version
         Mongo::VERSION.split('.')[0...2].map(&:to_i)
       end
 
-      def options(configuration)
-        config = configuration.dup
-        options = config.delete(:options) || {}
+      def options(opts)
```
p-mongo, line 90: Did the API of this method change? I think a brief docstring would be helpful to understand what this method is supposed to receive in the opts.

lib/mongoid/clients/factory.rb
comandeo: Right, fixed.

lib/mongoid/clients/factory.rb
comandeo: Added some docstring.

spec/mongoid/clients/factory_spec.rb
```diff
@@ -389,16 +397,19 @@
...
+        bad_two: { uri: "mongodb://127.0.0.1:1234,127.0.0.1:5678/#{database_id}" }.merge(unknown_opts)
       }
     end
 
-    before do
+    around(:each) do |example|
+      old_config = Mongoid::Config.clients
       Mongoid::Config.send(:clients=, config)
+      example.run
+      Mongoid::Config.send(:clients=, old_config)
```
p-mongo, line 412: I think this line needs to be in an `ensure` block otherwise it won't run if the test fails.

lib/mongoid/clients/factory.rb
```diff
@@ -87,6 +87,12 @@ def driver_version
         Mongo::VERSION.split('.')[0...2].map(&:to_i)
       end
 
+      # Prepare options for Mongo::Client based on Mongoid client configuration.
+      #
+      # @param [ Hash ] opts Parameters from options section of Mongoid client configuration.
+      # @return [ Hash ] Options that should be passed to Mongo::Client driver.
```
p-mongo, line 93: Client is not really a "driver", can you s/driver/constructor/ please?

spec/mongoid/clients/factory_spec.rb
comandeo: This is how I did it initially, but it turned out that `around` hook always executes. Even in case of an exception raised inside an example. It is mentioned in rspec docs - https://relishapp.com/rspec/rspec-core/v/3-8/docs/hooks/around-hooks#an-around-hook-continues-to-run-even-if-the-example-throws-an-exception

We still can add an `ensure` block here, but it might be confusing.

lib/mongoid/clients/factory.rb
comandeo: Sorry 😊 
Fixed.

spec/mongoid/clients/factory_spec.rb
p-mongo: I see, this is correct then.

p-mongo (approved):

PR #4988, opened 2021-04-27
MONGOID-5073 Make purge and truncate use overrides

comandeo:


spec/mongoid/config_spec.rb
```diff
@@ -545,4 +545,34 @@ def self.logger
...
+
+    after do
+      Mongoid.override_database(nil)
+    end
+
+    describe "#purge!" do
+      it 'respects persistence context overrides' do
+        House.create!(name: '1', model: 'Big')
+        Mongoid.purge!
+        expect(House.count).to be == 0
```
p-mongo, line 566: For symmetry can this test please assert the count is 1 before purging/truncation, and I believe the equality can be expressed as

```
        expect(House.count).to eq 0
```

p-mongo (approved):

p-mongo (approved):
Thanks!

PR #4999, opened 2021-05-24
Fix MONGOID-5098，ActiveSupport::TimeWithZone in range will ignore timezone

xuxiangyang:
[MONGOID-5098](https://jira.mongodb.org/browse/MONGOID-5098)


johnnyshields:
@xuxiangyang can you add tests for this?

johnnyshields:
@agolin95 @comandeo this one looks good to merge

johnnyshields:
@xuxiangyang can you please merge latest master for this PR. I think master contains CI fixes.

comandeo (approved):

p-mongo:
As mentioned in https://jira.mongodb.org/browse/MONGOID-5098, this PR and/or the ticket needs to provide a code snippet that demonstrates the problem. Currently this PR only includes low-level unit tests, the code that is added/changed isn't meant to be invoked by applications directly.

If the problem is the lack of UTC conversion in this example:


```
irb(main):002:0> Band.where(foo: Time.now..Time.now)
=> 
#<Mongoid::Criteria
  selector: {"foo"=>{"$gte"=>2021-08-03 02:52:36.681803167 -0400, "$lte"=>2021-08-03 02:52:36.681803724 -0400}}
  options:  {}
  class:    Band
  embedded: false>

irb(main):003:0> Band.where(foo: Time.now)
=> 
#<Mongoid::Criteria
  selector: {"foo"=>2021-08-03 06:53:23.714128115 UTC}
  options:  {}
  class:    Band
  embedded: false>
```

... then this test should be added (perhaps in addition to the low-level tests that are already part of this PR).

johnnyshields:
@xuxiangyang p-mongo's request is reasonable here. Please add an integration spec something like this:

```
b1 = Band.create!(foo: time1)
b2 = Band.create!(foo: time2)
b3 = Band.create!(foo: time3)

expect(Band.where(foo: Time.zone.parse(time1 + 1.minute)..Time.zone.parse(time3 - 1.minute).to_a).to eq([b2])
```

xuxiangyang:
I have no Idea where to place the spec code.... Could you give me some help?

johnnyshields:
Sure am glad to help. Will raise a PR to your branch.

johnnyshields:
@xuxiangyang please merge this PR: https://github.com/xuxiangyang/mongoid/pull/1

johnnyshields:
@xuxiangyang ah wait my PR doesn't include dynamic columns. I will add that.

johnnyshields:
@xuxiangyang it appears all the specs I added are also passing on master branch. Can you investigate why that is? I've included specs for both static and dynamic field queries.

johnnyshields:
@xuxiangyang I've added more specs related to Range persistence here https://github.com/mongodb/mongoid/pull/5024. It seems like time is always persisted in UTC

lib/mongoid/extensions/range.rb
```diff
@@ -73,7 +73,7 @@ def mongoize(object)
           return nil if object.nil?
           return object if object.is_a?(::Hash)
```
johnnyshields, line 74: If `Hash`, the keys [:min and :max] should be mongoized.

johnnyshields:
@p-mongo based on my investigation of this PR (see #5023, #5024) in terms of integration with Mongoid <--> Mongo Driver <--> DB it appears there is no change, and the driver is probably converting TimeWithZone to UTC. I haven't looked in detail.

**However**, it is still a good idea to merge this, because the `#mongoize` / `#evolve` methods for Time do convert to UTC, so we should be consistent and do the same with Range, even if the driver also does such a conversion. (Otherwise, we should remove UTC conversion everywhere in Mongoid--probably not a good idea.)

lib/mongoid/criteria/queryable/extensions/range.rb
```diff
@@ -59,7 +59,7 @@ module ClassMethods
             # @since 1.0.0
             def evolve(object)
               return object unless object.is_a?(::Range)
-              { "$gte" => object.min, "$lte" => object.max }
+              { "$gte" => object.min.mongoize, "$lte" => object.max.mongoize }
```
johnnyshields, line 62: This needs to consider `exclude_end`. Should be a separate PR.

johnnyshields:
@xuxiangyang I've raised https://github.com/mongodb/mongoid/pull/5025 for a much better implementation of the `Range.evolve` method. I'll probably look at the `mongoize` method soon.

lib/mongoid/extensions/range.rb
johnnyshields: Above is handled in #5025

xuxiangyang:
@johnnyshields `evolve` is a much better implementation~

johnnyshields:
I've raised 3 PRs which together solve this issue completely:

#5026- MONGOID-5098 Standardize Range.mongoize
#5025 - Standardize/improve Range queries -- .where(field: 1..3)
#5023 - Improve specs for timezone handling

PR #5004, opened 2021-06-10
Add ruby 3.0 support

comandeo:


lib/mongoid/validatable/presence.rb
```diff
@@ -34,15 +34,15 @@ def validate_each(document, attribute, value)
...
             ) if not_present?(_value)
           end
         elsif document.relations.has_key?(attribute.to_s)
           if relation_or_fk_missing?(document, attribute, value)
-            document.errors.add(attribute, :blank, options)
+            document.errors.add(attribute, :blank, **options)
           end
         else
-          document.errors.add(attribute, :blank, options) if not_present?(value)
+          document.errors.add(attribute, :blank, **options) if not_present?(value)
```
hgraham594, line 45: Thanks, much appreciated.

gemfiles/standard.rb
```diff
@@ -21,8 +21,6 @@ def standard_dependencies
     gem 'timecop'
     gem 'rspec-retry'
     gem 'benchmark-ips'
-    gem 'rspec-expectations', '~> 3.7', '>= 3.8.4'
-    gem 'rspec-mocks-diag', '~> 3.0'
```
p-mongo, line 25: We need this right?

mongoid.gemspec
```diff
@@ -29,10 +29,14 @@ Gem::Specification.new do |s|
     warn "[#{s.name}] Warning: No private key present, creating unsigned gem."
   end
 
-  s.required_ruby_version     = ">= 2.3"
+  s.required_ruby_version     = ">= 2.5"
```
p-mongo, line 32: I think this needs to be reverted as well, since we are going to release this diff on all stable branches.

mongoid.gemspec
```diff
@@ -29,10 +29,14 @@ Gem::Specification.new do |s|
...
   s.required_rubygems_version = ">= 1.3.6"
 
-  s.add_dependency("activemodel", [">= 5.1", "<6.2"])
+  if RUBY_VERSION.start_with?('2.')
+    s.add_dependency("activemodel", [">=5.2", "<6.3"])
```
p-mongo, line 36: 6.2 isn't out yet, so unless I'm missing something the upper bound should remain as < 6.2.

Does 5.1 no longer work on Ruby 2 with the current Mongoid tree?

p-mongo (commented):
I think this PR should really be split up into 3:

Server 5.0
Ruby 2.3/2.4 removal
Ruby 3 support

Since there are changes needed for server 5.0 support, that should be a separate PR.

Ruby 2.3/2.4 removal can't be done on stable branches, thus needs to be in a separate PR.

comandeo:
Converting this to draft to avoid an accidental merge. This one should be split.

p-mongo:
Superseded by https://github.com/mongodb/mongoid/pull/5009.

PR #5007, opened 2021-07-01
allows ActionController::Parameters to be passed as nested attributes

braindeaf:
Whilst upgrading to Rails 5 we found that ActionController::Parameters doesn't respond to .with_indifferent_access. So this caused problems here in Nested Many because this assumes that the attributes are either passed as an array or a hash and ActionController::Parameters is assumed to be an Array.

### DIFF
--- a/lib/mongoid/association/nested/many.rb
+++ b/lib/mongoid/association/nested/many.rb
@@ -1,4 +1,3 @@
-require 'pry'
 # frozen_string_literal: true
 # encoding: utf-8
 
@@ -28,7 +27,7 @@ module Mongoid
             raise Errors::TooManyNestedAttributeRecords.new(existing, options[:limit])
           end
           attributes.each do |attrs|
-            if attrs.is_a?(Hash)
+            if attrs.is_a?(::Hash)
               process_attributes(parent, attrs.with_indifferent_access)
             else
               process_attributes(parent, attrs[1].with_indifferent_access)
--- a/lib/mongoid/association/nested/many.rb
+++ b/lib/mongoid/association/nested/many.rb
@@ -1,3 +1,4 @@
+require 'pry'
 # frozen_string_literal: true
 # encoding: utf-8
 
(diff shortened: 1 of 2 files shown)


p-mongo:
Hi @braindeaf , thank you for the PR.

Can you please add a code snippet demonstrating the problem which includes the stack trace demonstrating how an instance of Parameters is getting to the initialize method? This isn't obvious from the diff in this PR. Thanks!

lib/mongoid/association/nested/many.rb
```diff
@@ -45,6 +45,7 @@ def build(parent, options = {})
         # @param [ Hash ] attributes The attributes hash to attempt to set.
         # @param [ Hash ] options The options defined.
         def initialize(association, attributes, options = {})
+          attributes = attributes&.to_h if attributes.respond_to?(:keys)
```
p-mongo, line 48: What is the purpose of  `&` on this line?

p-mongo:
I am also confused why you say

> attributes are either passed as an array or a hash

The docstring on the method states that the parameter is an attributes hash without mentioning arrays. Are you saying the docstring is incorrect?

lib/mongoid/association/nested/many.rb
braindeaf: I had initially assumed this could be nil, so .to_h is only called if attributes isn't nil. It's not need after I added the :keys test .

braindeaf:
> I am also confused why you say
> 
> > attributes are either passed as an array or a hash
> 
> The docstring on the method states that the parameter is an attributes hash without mentioning arrays. Are you saying the docstring is incorrect?

Honestly, I find this code a little confusing because we're expecting *initialize* to take a Hash and it seems that the attributes are sorted by id if it responds to .with_indifferent_access. But what else would it be. I suppose it could be nil and *initialize* assigns it and it that's a problem it will just fail later. I think I assumed it could also be an Array because in the *build* method it makes a distinction between whether the Hash#each's block argument is a Hash or not. I'm not actually sure that's possible or that's a pre-Ruby 2.0 thing. So my bad. Fixing.

braindeaf:
Sorry to waste your time. I've just found this is a horrible side-effect of our setup. I couldn't create a failing spec and can only conclude that the problem is with us.

mongoid 7.3.0
strong_parameters_continued 

and then permitting all attributes with 

model.update(params.fetch(:something, {}).permit!)

As soon as I nuked attr_accessible definitions it works :) Thank you for your time.

PR #5008, opened 2021-07-01
Fix MONGOID-5116 Using 'without' breaks access to attributes of embedded documents

simonc:
As reported in https://jira.mongodb.org/browse/MONGOID-5116, there's an issue when a record was loaded by a query that was built using `without` and we then try to access a field on one of its embeds.

```ruby
duck = Pet.new(name: 'Duck')
Person.create!(username: 'Dev', title: 'CEO', pet: duck)

person = Person.where(username: 'Dev').without(:title).first

person.pet.name
#! ActiveModel::MissingAttributeError:
#!  Missing attribute: 'name'
```

This is due to the way fields filtering happens as it only accounts for cases where `only` was used.

This PR is a fix that will ignore fields selection if it's only rejecting fields that don't involve the association.

### DIFF
--- a/lib/mongoid/association/accessors.rb
+++ b/lib/mongoid/association/accessors.rb
@@ -121,6 +121,9 @@ module Mongoid
       # (embedded) association with the given key, or nil if no projection
       # is to be performed.
       #
+      # Also returns nil if exclusionary projection was requested but it does
+      # not exclude the field of the association.
+      #
       # For example, if __selected_fields is {'a' => 1, 'b.c' => 2, 'b.c.f' => 3},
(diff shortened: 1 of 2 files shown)


p-mongo (approved):

comandeo (approved):
👏

p-mongo:
I believe the app tests are red due to prep work for https://github.com/mongodb/mongoid/pull/5004, but since that PR isn't quite ready to go in yet I'll merge this one as is.

simonc:
Thanks for merging this! ❤️

PR #5009, opened 2021-07-01
RUBY-5044 Add Ruby 3 support

comandeo:


comandeo:
https://jira.mongodb.org/browse/MONGOID-4833
https://jira.mongodb.org/browse/MONGOID-5044

gemfiles/standard.rb
```diff
@@ -21,8 +27,12 @@ def standard_dependencies
     gem 'timecop'
     gem 'rspec-retry'
     gem 'benchmark-ips'
-    gem 'rspec-expectations', '~> 3.7', '>= 3.8.4'
-    gem 'rspec-mocks-diag', '~> 3.0'
+    if RUBY_VERSION.start_with?('2.')
+      # Brind the dependencies back the following is fixed:
```
p-mongo, line 31: Bring

p-mongo (commented):
I am still wondering - does Rails 5.1 continue to work on Ruby 2?

comandeo:
> I am still wondering - does Rails 5.1 continue to work on Ruby 2?

Yes, it does. It is not supported anymore, though. That's why we may want to remove it from our list of supported versions.

p-mongo:
I support doing that  for 7.4 but I don't think it should be done in this PR.

gemfiles/standard.rb
p-mongo: Also 

> Bring the dependencies back *after* the following is fixed:

comandeo:
Sure, you're right. This one does not remove support for 5.1 - https://github.com/mongodb/mongoid/pull/5009/files#diff-f776cb607ce9083ed350e5839c8034744e53b0a1da6541533a0e9871138942d5R36

.evergreen/config.yml
```diff
@@ -773,15 +799,29 @@ buildvariants:
...
+- matrix_name: app-tests-ruby-2.7
   matrix_spec:
     driver: current
     ruby: ruby-2.7
     mongodb-version: '4.4'
     topology: standalone
     app-tests: yes
     # Not currently testing rails master
-    rails: ['5.1', '5.2', '6.0', '6.1']
+    rails: ['5.2', '6.0', '6.1']
```
p-mongo, line 824: Does this not remove test coverage for 5.1?

.evergreen/config.yml
```diff
@@ -697,45 +723,45 @@ buildvariants:
...
-    rails: '5.2'
-  display_name: "${rails}, ${driver}, ${mongodb-version}"
+    rails: '6.0'
+  display_name: "${ruby}, ${driver}, ${mongodb-version}, ${topology}, ${rails}"
   run_on:
     - rhel70-small
   tasks:
      - name: "test"
-- matrix_name: "rails-6.0"
+- matrix_name: "rails-6.1"
```
p-mongo, line 758: This shouldn't be needed either. The default rails should be the most recent one allowed by the gemspec, which currently is 6.1. We have gemfiles for older versions, meaning 6.0 through 5.1.

.evergreen/config.yml
comandeo: Oh, good catch, thanks!

.evergreen/config.yml
comandeo: I'll bring back gemfile and configuration for 5.1

gemfiles/standard.rb
comandeo: Fixed

.evergreen/config.yml
comandeo: Fixed

p-mongo (approved):

PR #5012, opened 2021-07-12
MONGOID-5125 find: does not return the correct results for has_many relations

bradrees:
See: https://jira.mongodb.org/projects/MONGOID/issues/MONGOID-5125

Since #4817 `find` can take a block, however this is not used when a has many relation proxy is used. This is a simple change to allow the block to be passed through to the criteria.

Example:

```
class Widget 
  include Mongoid::Document

  has_many :screws

  def find_flat_screw 
    screws.find { |screw| screw.head == 'flat' }
  end
end
```

The find method above will currently return an empty set incorrectly, instead of the first item in the has_many collection

### DIFF
--- a/lib/mongoid/association/embedded/embeds_many/proxy.rb
+++ b/lib/mongoid/association/embedded/embeds_many/proxy.rb
@@ -219,8 +219,8 @@ module Mongoid
           # @param [ Array<Object> ] args Various arguments.
           #
           # @return [ Array<Document>, Document ] A single or multiple documents.
-          def find(*args)
-            criteria.find(*args)
+          def find(*args, &block)
+            criteria.find(*args, &block)
           end
 
           # Instantiate a new embeds_many association.
--- a/lib/mongoid/association/embedded/embeds_many/proxy.rb
+++ b/lib/mongoid/association/embedded/embeds_many/proxy.rb
@@ -210,16 +210,27 @@ module Mongoid
           # Finds a document in this association through several different
           # methods.
           #
+          # This method delegates to +Mongoid::Criteria#find+. If this method is
+          # not given a block, it returns one or many documents for the provided
+          # _id values.
+          #
+          # If this method is given a block, it returns the first document
+          # of those found by the current Criteria object for which the block
+          # returns a truthy value.
+          #
           # @example Find a document by its id.
           #   person.addresses.find(BSON::ObjectId.new)
           #
           # @example Find documents for multiple ids.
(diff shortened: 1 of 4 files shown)


johnnyshields:
@bradrees can you please add tests for this?

spec/mongoid/association/embedded/embeds_many/proxy_spec.rb
```diff
@@ -2148,6 +2148,44 @@ class TrackingIdValidationHistory
...
+
+      it "finds one" do
+        expect(
+          author.videos.find do |video|
+            video.title == 'video one'
+          end
+        ).to eq(video_one)
+      end
+
+      it "finds multiple" do
```
p-mongo, line 2173: This actually returns a single document, right? Is the test meant to exercise the sitution when multiple documents match the block condition?

lib/mongoid/association/embedded/embeds_many/proxy.rb
```diff
@@ -217,10 +217,11 @@ def exists?
           #   person.addresses.find([ BSON::ObjectId.new, BSON::ObjectId.new ])
           #
           # @param [ Array<Object> ] args Various arguments.
+          # @param [ Proc ] block Optional block to pass.
```
p-mongo, line 220: Can these docstrings be augmented to 1) either include a copy of  the text added in https://github.com/mongodb/mongoid/pull/4817/files that explains the block usage, or link to the documentation of the methods documented in https://github.com/mongodb/mongoid/pull/4817/files, so that in either case the block usage is documented, and 2) provide an example with the block form since currently the other two invocations are documented with examples (could be only the single line)?

spec/mongoid/association/embedded/embeds_many/proxy_spec.rb
comandeo: Oh, my bad. We should expect multiple documents to be found here. Will fix and check again.

comandeo: Corrected.

lib/mongoid/association/embedded/embeds_many/proxy.rb
comandeo: Added

p-mongo (approved):

comandeo:
Thank you very much for your contribution @bradrees !

bradrees:
@comandeo Thanks for sorting out the tests! It's been a busy week for me, so I really appreciate it!

PR #5013, opened 2021-07-13
MONGOID-5126 Bring Mongoid::Document's === implementation in alignment with Ruby behavior

mdehoog:
This PR modifies the `===` method override on `Mongoid::Document` classes to just call `is_a?`.

Currently, for a document such as:
```ruby
class User
  include Mongoid::Document
  field :first_name, type: String
end
```
this returns `true`:
```ruby
User === Object
=> true
```
However, the `Object` class is not an instance of `User`, which means it's overriding the expected semantics of `===` for classes (`is_a?`).

This breaks things like `RSpec::Mock`, which uses `===` to determine which class to mock: https://github.com/rspec/rspec-mocks/blob/b1b32716114727a9257e834f0b72015208ca217f/lib/rspec/mocks/space.rb#L108. This causes errors like:
```ruby
allow_any_instance_of(User).to receive(:first_name)
=> RSpec::Mocks::MockExpectationError: Object does not implement: first_name
```

Note this is a breaking change, and may need to be released as a part of a major release. E.g. this will break:
```ruby
case document.class
when User
  # do something
end
```
and needs to be switched to:
```ruby
case document
when User
  # do something
end
```

### DIFF
--- a/docs/release-notes.txt
+++ b/docs/release-notes.txt
@@ -9,6 +9,7 @@ Release Notes
 .. toctree::
   :titlesonly:
 
+  release-notes/mongoid-7.4
   release-notes/mongoid-7.3
   release-notes/mongoid-7.2
   release-notes/mongoid-7.1
--- a/docs/release-notes/mongoid-7.4.txt
+++ b/docs/release-notes/mongoid-7.4.txt
@@ -22,12 +22,12 @@ the complete list of issues fixed in each release, including bug fixes.
 ---------------------------------------
 
 **Breaking change:** In Mongoid 7.4, the ``===`` operator works the same way
-as it does in Ruby, and is equivalent to calling ``is_a?`` on the left hand
-side with the right hand side as the argument:
+as it does in Ruby, and is equivalent to calling ``is_a?`` on the right hand
+side with the left hand side as the argument:
 
 .. code-block:: ruby
 
-  instance === ModelClass
+  ModelClass === instance
   
   # equivalent to:
   instance.is_a?(ModelClass)
--- a/docs/release-notes/mongoid-7.4.txt
+++ b/docs/release-notes/mongoid-7.4.txt
@@ -1,0 +1,90 @@
+***********
+Mongoid 7.4
+***********
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+This page describes significant changes and improvements in Mongoid 7.4.
+The complete list of releases is available `on GitHub
+<https://github.com/mongodb/mongoid/releases>`_ and `in JIRA
+<https://jira.mongodb.org/projects/MONGOID?selectedItem=com.atlassian.jira.jira-projects-plugin:release-page>`_;
+please consult GitHub releases for detailed release notes and JIRA for
+the complete list of issues fixed in each release, including bug fixes.
+
+
+``===`` Operator Matches Ruby Semantics
+---------------------------------------
+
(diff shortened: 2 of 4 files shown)


agolin95:
Tracking in: https://jira.mongodb.org/browse/MONGOID-5126

johnnyshields:
@comandeo CI looks like it needs help here.

comandeo:
For some reason, state was not reported back from Evergreen. Unfortunately, all tests (except app tests) failed with the following:

```
1) Mongoid::Equality.=== when comparable is a relation of this document returns true
   Failure/Error: expect(klass === relation).to be true
   
     expected true
          got false
   # ./spec/mongoid/equality_spec.rb:103:in `block (4 levels) in <top (required)>'
   # ./spec/lite_spec_helper.rb:68:in `block (3 levels) in <top (required)>'
   # ./spec/lite_spec_helper.rb:67:in `block (2 levels) in <top (required)>'

2) Mongoid::Equality.=== when comparable is the same class returns true
   Failure/Error: expect(klass === Person).to be true
   
     expected true
          got false
   # ./spec/mongoid/equality_spec.rb:110:in `block (4 levels) in <top (required)>'
   # ./spec/lite_spec_helper.rb:68:in `block (3 levels) in <top (required)>'
   # ./spec/lite_spec_helper.rb:67:in `block (2 levels) in <top (required)>'

3) Mongoid::Equality#=== when comparing to a class when the class is the same returns true
   Failure/Error: expect(person === Person).to be true
   
     expected true
          got false
   # ./spec/mongoid/equality_spec.rb:160:in `block (5 levels) in <top (required)>'
   # ./spec/lite_spec_helper.rb:68:in `block (3 levels) in <top (required)>'
   # ./spec/lite_spec_helper.rb:67:in `block (2 levels) in <top (required)>'

4) Mongoid::Equality#=== when comparing to a class when the class is a superclass returns true
   Failure/Error: expect(Doctor.new === Person).to be true
   
     expected true
          got false
   # ./spec/mongoid/equality_spec.rb:174:in `block (5 levels) in <top (required)>'
   # ./spec/lite_spec_helper.rb:68:in `block (3 levels) in <top (required)>'
   # ./spec/lite_spec_helper.rb:67:in `block (2 levels) in <top (required)>'
```

comandeo:
I think the first failure is a deal breaker here. Consider this example:

```ruby
class Person
  include Mongoid::Document

  field :name, :type => String

  has_many :posts, :validate => false
end

class Post
  include Mongoid::Document

  field :title, :type => String

  belongs_to :person
end

person = Person.create(name: 'Author')
post = person.posts.create(title: 'Some post')

Person === post.person # => now false, but should ne true
```

mdehoog:
@comandeo running that snippet you linked works with the patch for me...
```ruby
Person === post.person # => returns `true`
```

mdehoog:
@comandeo my fault, I should not have removed the entire implementatio. This was based on an incorrect assumption that `===` would inherit the default behavior for Ruby classes, but it didn't work for relations (due to the delegation?):
```ruby
Person === Person.new # true
Person === person # true
Person === post.person # false
```

I've updated the implementation to explicitly call `is_a?`, which fixes the issue.

johnnyshields:
FYI the latest one from @mdehoog matches ActiveRecord
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/core.rb#L421

johnnyshields:
@comandeo please try running the CI again on this, or report what the errors are?

p-mongo (approved):

mdehoog:
Thanks for the docs and specs @p-mongo 🙌

comandeo (approved):

PR #5014, opened 2021-07-15
Do not reference bson master from driver stable/oldstable config


p-mongo:


comandeo (approved):

p-mongo:
Backported through 7.0.

PR #5015, opened 2021-07-19
MONGOID-5123 Fix methods incompatible with Ruby 3

comandeo:
https://jira.mongodb.org/browse/MONGOID-5123

This PR fixes issues with argument delegation in various versions of Ruby. It also adds some tests to cover cases that are handled differently in different Ruby versions.


alexbevi (approved):
LGTM

PR #5016, opened 2021-07-23
MONGOID-5127: Add `distance_field` to `Mongoid::Contextual::GeoNear`

karlbaker02:
This PR adds a method for `distance_field` to `Mongoid::Contextual::GeoNear`. This method is [specified in the documentation](https://docs.mongodb.com/v2.4/reference/operator/aggregation/geoNear/#pipe._S_geoNear) for MongoDB from the current version `5.0` all the way back to at least version `2.4.0`, however it's currently not settable via this class, returning the error `NoMethodError: undefined method `distance_field' for #<Mongoid::Contextual::GeoNear:0x0000563a2fc32d18>. Did you mean?  distance_multiplier`.

By adding the `distance_field` method the calculated distances as part of _geoNear_ will be projected in the document results.

### DIFF
--- a/lib/mongoid/contextual/geo_near.rb
+++ b/lib/mongoid/contextual/geo_near.rb
@@ -46,6 +46,22 @@ module Mongoid
         end
       end
 
+      # Projects an output field containing the distance from each document
+      # to the point in the command.
+      #
+      # @example Specify the distance field.
+      #   geo_near.distance_field("dist")
+      #
+      # @param [ String ] value The distance field.
+      #
+      # @return [ GeoNear ] The GeoNear wrapper.
+      #
+      # @since 2.4.0
+      def distance_field(value)
+        command[:distanceField] = value
+        self
+      end
+
       # Provide a distance multiplier to be used for each returned distance.
       #
       # @example Provide the distance multiplier.
--- a/spec/mongoid/contextual/geo_near_spec.rb
+++ b/spec/mongoid/contextual/geo_near_spec.rb
@@ -158,6 +158,27 @@ describe Mongoid::Contextual::GeoNear do
       end
     end
 
+    context "when providing a distance field" do
+
+      let(:criteria) do
+        Bar.all
+      end
+
+      let(:geo_near) do
+        described_class.new(collection, criteria, [ 52, 13 ])
+      end
+
+      let(:results) do
+        geo_near.distance_field("distance").entries
+      end
+
+      it "returns calculated distances" do
+        expect(results.first.distance.calculated).to eq(0.390512483795333)
+        expect(results.last.distance.calculated).to eq(0.460977222864644)
+      end
+
+    end
+
     context "when providing a distance multiplier" do
 
       let(:criteria) do


agolin95:
Tracked in https://jira.mongodb.org/browse/MONGOID-5127

lib/mongoid/contextual/geo_near.rb
```diff
@@ -46,6 +46,22 @@ def each
...
+      # to the point in the command.
+      #
+      # @example Specify the distance field.
+      #   geo_near.distance_field("dist")
+      #
+      # @param [ String ] value The distance field.
+      #
+      # @return [ GeoNear ] The GeoNear wrapper.
+      #
+      # @since 2.4.0
```
alcaeus, line 59: I'm not sure this is correct. I'd expect the version provided by `@since` to be the Mongoid version that introduced the API. Looking at the other methods, I don't think those versions refer to server versions as this one does.

spec/mongoid/contextual/geo_near_spec.rb
```diff
@@ -158,6 +158,27 @@
...
+        described_class.new(collection, criteria, [ 52, 13 ])
+      end
+
+      let(:results) do
+        geo_near.distance_field("distance").entries
+      end
+
+      it "returns calculated distances" do
+        expect(results.first.distance.calculated).to eq(0.390512483795333)
+        expect(results.last.distance.calculated).to eq(0.460977222864644)
```
alcaeus, line 177: I don't feel comfortable adding this kind of check as it's one of those "tests that just start failing at some point" thing.

Shouldn't we instead check that a distance is provided, that it is a float, and that the values are not equal as that seems to be the only thing we can be sure of from the setup?

karlbaker02:
Hi @alcaeus. Thanks for your review, I will reply inline.

lib/mongoid/contextual/geo_near.rb
karlbaker02: Fair point. I did my best to look through the previous docs but I found that I could only go as far back as `2.4.0`, which is why I called it out in the PR description just in case this wasn't correct. Any help to track this down then would be appreciated.

spec/mongoid/contextual/geo_near_spec.rb
karlbaker02: If we don't want to add in the exact distances here, the approach you describe sounds like a good alternative. Equally, there could also be a check to ensure that the first result's distance < last result's distance in addition/to replace a check that the values aren't equal.

comandeo:
Hi @karlbaker02, thank you for the PR!

Unfortunately, there is a test failing:

```
 1) Mongoid::Contextual::GeoNear#each when providing a distance field returns calculated distances
    Failure/Error: return super unless attributes.has_key?(attr.reader)
    
    NoMethodError:
      undefined method `distance' for #<Bar _id: 60fff8941f62c27f2c469ec6, name: nil, location: [52.3, 13.25], lat_lng: nil, geo_near_distance: 0.39051248379533055>
    # ./lib/mongoid/attributes/dynamic.rb:140:in `method_missing'
    # ./spec/mongoid/contextual/geo_near_spec.rb:176:in `block (4 levels) in <top (required)>'
    # ./spec/lite_spec_helper.rb:68:in `block (3 levels) in <top (required)>'
    # ./spec/lite_spec_helper.rb:67:in `block (2 levels) in <top (required)>'
```

It seems that is fails on all configurations it is supposed to be executed - mongodb version <= 4.0

Could you please check?

spec/mongoid/contextual/geo_near_spec.rb
alcaeus: That sounds like a good option, what do you think @comandeo?

karlbaker02:
> geo_near_distance

Hi @comandeo ,

This is where my knowledge of MongoDB tapers off somewhat. I _believe_ the reason that the tests are failing for MongoDB <= version 4.0 is because `geoNear` is referring to the command, not the aggregation `$geoNear`. [Noting in the docs for MongoDB 4.2](https://docs.mongodb.com/v4.2/release-notes/4.2-compatibility/#remove-support-for-the-geonear-command), `geoNear` has been removed and replaced by `$geoNear` the aggregation.

The point here is that I may have been mistaken in what I thought I was calling in those older MongoDB versions, as `distanceField` is _not_ actually one of the [options for the command](https://docs.mongodb.com/v4.0/reference/command/geoNear/#definition), whereas [it is an option for the aggregation in MongoDB 4.2](https://docs.mongodb.com/v4.2/reference/operator/aggregation/geoNear/#definition) (and above).

The output from one of the failing tests is also very useful (I haven't been able to get tests running locally), particularly that the [distance _is_ being included via the property `geo_near_distance`](https://github.com/mongoid/mongoid/blob/63ece9973709a837a1817824d202b367db3e3688/lib/mongoid/contextual/geo_near.rb#L233), which takes its value from the `dis` attribute.

Would certainly appreciate your thoughts on this.

spec/mongoid/contextual/geo_near_spec.rb
johnnyshields:
```
        expect(results.first.distance.calculated).to be_within(0.0001).of(0.39051)
        expect(results.last.distance.calculated).to be_within(0.0001).of(0.46097)
```

comandeo:
Hi @karlbaker02 ,

I did some investigation, and I'm afraid you are right here. `Mongoid::Contextual::GeoNear` is a wrapper around `geoNear` command, which was removed in 4.2. That's why there is `max_server_version '4.0'` constraint in the corresponding spec. As you correctly mentioned, the command does not support `distanceField` attribute, so it is effectively ignored.  If we execute `db.runCommand({geoNear: 'bars', near: [52, 13], distanceField: "distance"})` in mongo shell, it returns same results as `db.runCommand({geoNear: 'bars', near: [52, 13]})`.

The calculated distance is available via `geo_near_distance` attribute. So, specs will pass if we adjust them to something like:

```ruby
it "returns calculated distances" do
  expect(results.first.geo_near_distance).to be_within(0.0001).of(0.39051)
  expect(results.last.geo_near_distance).to be_within(0.0001).of(0.46097)
end
```

However, it will also work without your changes.

Therefore, I do not see what we can do here, and suggest closing this PR and the corresponding issue in JIRA. Please let us know if you have any objections.

p-mongo:
I agree with @comandeo , the command in question is https://docs.mongodb.com/v4.0/reference/command/geoNear/index.html and it doesn't take a `distanceField` option per the docs. Mongoid does include a minimal aggregation pipeline DSL which is documented in https://docs.mongodb.com/mongoid/master/reference/associations/#builder-dsl, it doesn't currently implement a `geo_near` stage but I think that would be the place to add it.

karlbaker02:
> Hi @karlbaker02 ,
> 
> I did some investigation, and I'm afraid you are right here. `Mongoid::Contextual::GeoNear` is a wrapper around `geoNear` command, which was removed in 4.2. That's why there is `max_server_version '4.0'` constraint in the corresponding spec. As you correctly mentioned, the command does not support `distanceField` attribute, so it is effectively ignored. If we execute `db.runCommand({geoNear: 'bars', near: [52, 13], distanceField: "distance"})` in mongo shell, it returns same results as `db.runCommand({geoNear: 'bars', near: [52, 13]})`.
> 
> The calculated distance is available via `geo_near_distance` attribute. So, specs will pass if we adjust them to something like:
> 
> ```ruby
> it "returns calculated distances" do
>   expect(results.first.geo_near_distance).to be_within(0.0001).of(0.39051)
>   expect(results.last.geo_near_distance).to be_within(0.0001).of(0.46097)
> end
> ```
> 
> However, it will also work without your changes.
> 
> Therefore, I do not see what we can do here, and suggest closing this PR and the corresponding issue in JIRA. Please let us know if you have any objections.

Hi @comandeo ,

Thanks for confirming the situation. It would be good to have the Ruby driver provide an interface consistent with the available options for MongoDB >= 4.2, but I appreciate that the info is available via `geo_near_distance` in any case.

If everyone is in agreement then I'll go ahead and close this PR, thanks to all for contributing to it 🙇

agolin95:
Closing out according to above conversation, thanks all!

PR #5021, opened 2021-08-06
MONGOID-5118 Add 5.0 configurations to evergreen

comandeo:
This PR proposes new structure to our evergreen configurations. We extensively test actual versions using stable version of the driver:
- ruby 2.6 - 3.0
- jruby 9.2
- mongodb 3.6 - 5.0
- rails 5.2 - 6.1

We test all possible combinations of these on all mongodb topologies.

For other versions of components that may be not that actual, but still supported by mongoid (ruby <= 2.5, mongodb <= 3.4, rails 5.1, oldstable and min driver, etc.) we limit number of configurations. 

However, this setup starts 161 tasks on Evergreen for every PR opened. Maybe this is too many, but I am not sure. I am looking forward to your feedback.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -327,6 +327,10 @@ axes:
   - id: "mongodb-version"
     display_name: MongoDB Version
     values:
+      - id: "5.0"
+        display_name: "MongoDB 5.0"
+        variables:
+           VERSION: "5.0"
       - id: "4.4"
         display_name: "MongoDB 4.4"
         variables:
@@ -438,10 +442,6 @@ axes:
   - id: "os"
     display_name: OS
     values:
-      # https://jira.mongodb.org/browse/MONGOID-5133
-      # - id: ubuntu-14.04
-      #   display_name: "Ubuntu 14.04"
-      #   run_on: ubuntu1404-small
       - id: ubuntu-16.04
         display_name: "Ubuntu 16.04"
         run_on: ubuntu1604-small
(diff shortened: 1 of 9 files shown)


comandeo:
https://jira.mongodb.org/browse/MONGOID-5118
https://jira.mongodb.org/browse/MONGOID-5134

alcaeus:
> However, this setup starts 161 tasks on Evergreen for every PR opened. Maybe this is too many, but I am not sure. I am looking forward to your feedback.

Yeah, that's a lot of tasks to run on every pull request. The DBX build stability policy defines a set of builds that should be kept in a passing state in all supported branches. IMO, it is sufficient to only run those builds on pull requests, while still allowing to start specific builds (e.g. to verify a bug fix in the exact environment it was discovered in).

For pull requests, I'd suggest a different approach than starting all builds. Instead of creating a build matrix from all axes (Ruby language version, driver version, MongoDB server version, Rails version, ...) we should test each axis with only the latest item in all other axes. As an example, we wouldn't test all Ruby versions against all MongoDB and Rails versions, but only against the latest MongoDB and Rails versions. We would also test all MongoDB versions with the latest Ruby and Rails versions, and all Rails versions with the latest Ruby and Mongoid versions, and so on.

This would limit the pipeline size that's run on PR builds, while still covering all axes sufficiently, as well as allowing to manually add skipped build variants to the patch build.

p-mongo:
Mongoid does not have much code conditioned on server version. If I remember correctly, the existing matrices were constructed from the following principles:

- At least one build per each server version
- At least one build per Ruby version

The server x Ruby version combinations were then distributed to keep the overall number of builds down. For example, I believe I assigned older Ruby versions to the older servers in general.

This coverage can be attained by grouping, for example:

- 5.0 server -> Ruby 3.0
- 4.4 -> 2.7
- 4.2 -> 2.7
- 4.0 -> 2.6
- 3.6 -> 2.5
- 3.4 -> 2.3
- 3.2 -> 2.4
- 3.0 -> 2.5
- 2.6 -> 2.3

The topology coverage is somewhat important, I suggest ensuring each topology of each server version has at least one build.

Then there are the various special cases like i18n and app tests.

comandeo:
@alcaeus @p-mongo Reduced the number of tasks to 41. Looks like a good compromise to me.

p-mongo:
Am I correct in seeing that app tests for Rails 5.1 are gone? If yes I think those need to be reinstated.

comandeo:
> Am I correct in seeing that app tests for Rails 5.1 are gone? If yes I think those need to be reinstated.

Also for rails 6.0. Added both.

p-mongo (approved):

p-mongo:
I think the failing test is a result of the driver closing push monitor connection while it's being read from, and has nothing to do with Mongoid. I am thinking a stress test which creates a client, waits a random amount of time (say 0..3 seconds), then closes the client, and runs 5-10 threads performing these operations concurrently, should reproduce the warning that is produced.

PR #5023, opened 2021-08-10
MONGOID-5098 Improve specs for timezone handling

johnnyshields:
Cover more cases of Time vs. ActiveSupport::TimeWithZone in integration tests.

Only specs; no logic change.

Related to MONGOID-5098 but is not a fix for the issue.


spec/integration/criteria/time_with_zone_spec.rb
```diff
@@ -4,29 +4,140 @@
...
+  let(:now_utc) { Time.now.utc }
+  let(:now_in_zone) { now_utc.in_time_zone(time_zone) }
 
-    before do
-      time = Time.now
-      expect(time.utc_offset).not_to eq(time.in_time_zone(time_zone_name).utc_offset)
+  shared_examples_for 'time zone queries' do
+    let!(:agency1) { Agency.create!(updated_at: now_utc - 20.minutes, dynamic_time: now_utc - 20.minutes).reload }
+    let!(:agency2) { Agency.create!(dynamic_time: now_in_zone).reload }
+    let!(:agency3) { Agency.create!(updated_at: now_in_zone + 20.minutes, dynamic_time: now_in_zone + 20.minutes).reload }
```
p-mongo, line 13: Can these variables please be renamed to reflect their contents, e.g. `agency_utc` instead of `agency1`.

spec/support/models/agency.rb
```diff
@@ -4,5 +4,6 @@
 class Agency
   include Mongoid::Document
   include Mongoid::Timestamps::Updated
+  include Mongoid::Attributes::Dynamic
```
p-mongo, line 7: Instead of doing this which is rather invasive can you please either 1) find an existing model that already includes Dynamic and, if needed, add a datetime field there or 2) create a new model that contains the fields you need for these tests.

spec/integration/criteria/time_with_zone_spec.rb
```diff
@@ -4,29 +4,140 @@
...
+        expect(view_gt.pluck(:_id).sort).to eq([agency2, agency3].pluck(:_id).sort)
+        expect(view_range.pluck(:_id).sort).to eq([agency2].pluck(:_id).sort)
+      end
+    end
+  end
+
+  context 'query with Time' do
+    let(:query_time) { now_utc }
+
+    context 'in a UTC time zone' do
```
p-mongo, line 103: This context and the outer context on line 100 need to have their descriptions expanded to clarify which times they refer to, because the outer context is already defining a UTC time hence it's unclear why the outer context isn't "in UTC time zone".

If your intention is to have the query times in various time zones, then line 101 needs to be clarified as to why that time is already in UTC (and not simply `Time.now` for example) and this context and its siblings need to say something like "when query argument is a time in UTC time zone".

p-mongo (commented):
Overall this looks to be very thorough, thank you for this PR.

AR defines its association thusly:

```
activerecord/lib/active_record/associations.rb:      def has_and_belongs_to_many(name, scope = nil, options = {}, &extension)
```
The scope argument was added in 4.0. Previously the definition looked like this:
```
activerecord/lib/active_record/associations.rb:      def has_and_belongs_to_many(name, options = {}, &extension)
```

I imagine such a change would be a breaking one, hence I see the value in using a keyword argument for scope. At the same time it should be relatively straightforward to support both invocations by examining the type of `scope` and if it's a Hash, treating it as options. It is my impression that this compatibility mechanism would continue to work even with the double splat syntax and Ruby 3 which is what Rails currently has in its codebase:
```
activerecord/lib/active_record/associations.rb:        def has_and_belongs_to_many(name, scope = nil, **options, &extension)
```

spec/support/models/agency.rb
johnnyshields: Done > have changed to Book without any model modifications

spec/integration/criteria/time_with_zone_spec.rb
johnnyshields: Done --> book_earlier, book_now, and book_later

spec/integration/criteria/time_with_zone_spec.rb
johnnyshields: Done > "when zone of queried time is UTC"

johnnyshields:
@p-mongo all comments addressed.

johnnyshields:
@p-mongo shall we merge this?

johnnyshields:
Let's merge?

comandeo (approved):

p-mongo (approved):

PR #5027, opened 2021-08-11
MONGOID-5149 Skip inverse detection when inverse_of: nil is set

simonc:
When an association is defined with the `inverse_of: nil` option, I believe its inverse should not be looked up in any case.

This PR fixes that by skipping the lookup when the option is explicitly defined and set to `nil` but will still let the lookup happen if the option was not specified and is thus `nil`.

I didn't push as to create a `inverse_of_set?` method in the `Options` module as it would not be defined for any other option but I suppose it could make sense.

## Why this PR

We actually got a bug while migrating to mongoid 7 where we have the kind of relationship I defined in the spec and due to the way inverse are looked up, trying to get the inverse of `previous_academy` here would return the inverse of `current_academy` and that would in the end cause this error:

```
ActiveModel::MissingAttributeError:
  Missing attribute: 'current_academy_id'
```

Which doesn't really make sense.

### DIFF
--- a/spec/mongoid/association/referenced/has_many_models.rb
+++ b/spec/mongoid/association/referenced/has_many_models.rb
@@ -24,6 +24,14 @@ class HmmAddress
   belongs_to :company, class_name: 'HmmCompany'
 end
 
+class HmmAcademy
+  include Mongoid::Document
+
+  has_many :students, class_name: 'HmmStudent', inverse_of: :current_academy
+
+  field :name, type: String
+end
+
 class HmmSchool
   include Mongoid::Document
 
@@ -35,6 +43,9 @@ end
 
 class HmmStudent
   include Mongoid::Document
+
+  belongs_to :current_academy, class_name: 'HmmAcademy', inverse_of: :students, optional: true
+  belongs_to :previous_academy, class_name: 'HmmAcademy', inverse_of: nil, optional: true
 
   belongs_to :school, class_name: 'HmmSchool'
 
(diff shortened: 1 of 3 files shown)


spec/mongoid/association/referenced/has_many_models.rb
```diff
@@ -36,6 +44,9 @@ class HmmSchool
 class HmmStudent
   include Mongoid::Document
 
+  belongs_to :current_academy, class_name: 'HmmAcademy', inverse_of: :students, optional: true
+  belongs_to :previous_academy, class_name: 'HmmAcademy', inverse_of: nil, optional: true
```
p-mongo, line 48: I think these should also go into a new model, unless there is something else from HmmStudent that your test would need.

p-mongo (commented):
Hi @simonc , thank you for the PR.

Can you please create your own test case instead of adding your test code to an existing test case. Your issue is with the inverse being explicitly not set, therefore I would expect the test to reflect that in its description (the test you modified has to do with projections).

spec/mongoid/association/referenced/has_many_models.rb
simonc: I can create a dedicated model for this yes.

simonc:
Hi @p-mongo. Actually it's somewhat related and that's why I've added it to the existing test as it's the very reason I faced the issue in the first place.

If the projection lists fields for an association that can conflict with another one due to the bug, you get the `MissingAttributeError`. Without projection you don't get an error as the inverse will be wrongly defined but not cause an issue.

That said, I can still create a separate spec for the specific case.

simonc:
@p-mongo I updated the PR to have its own models and spec.

johnnyshields:
👍 looks good to me

p-mongo:
Thank you very much @simonc , this looks very clear now.

p-mongo (approved):

comandeo (approved):

PR #5029, opened 2021-08-17
MONGOID-5029 #empty method should use an #exists? rather than a #count query

johnnyshields:
Fixes MONGOID-5029
- `#empty?` method should use an `exists?` rather than a `count` query, which is more efficient (single document existence check vs. counting total docs)
- `#empty?` (and `#any?`) should short-circuit avoid making a query if `_added` is set
- `#any?` should call `!empty?` since empty is its exact inverse.

Specs DONE -- Ready for merge
* [x] verify that an exists query is used instead of count (both empty/any)
* [x] verify that if `_added` is set no DB query is performed (both empty/any)


lib/mongoid/association/referenced/has_many/enumerable.rb
```diff
@@ -209,9 +209,9 @@ def each
           # @since 2.1.0
           def empty?
             if _loaded?
-              in_memory.count == 0
+              in_memory.length == 0
```
p-mongo, line 212: Can this simply call `in_memory.empty?` ?

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
```diff
@@ -201,6 +201,122 @@
     end
   end
 
+  describe "#empty?" do
+
+    let(:person) do
+      Person.create
```
p-mongo, line 207: `create!` please, and elsewhere.

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
```diff
@@ -201,6 +201,122 @@
...
+
+      let(:criteria) do
+        Post.where(person_id: person.id)
+      end
+
+      let!(:enumerable) do
+        described_class.new(criteria)
+      end
+
+      let!(:empty) do
```
p-mongo, line 228: Why does this use `let!` instead of `let`?

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
```diff
@@ -201,6 +201,122 @@
...
+        enumerable.empty?
+      end
+
+      it "returns false" do
+        expect(empty).to be false
+      end
+
+      it "retains the correct length" do
+        expect(enumerable.length).to eq(2)
```
p-mongo, line 237: If you intended to call `enumerable.empty?` and then perform additional assertions, these examples should be moved under their own context which has a `before` block which calls `.empty?`, and the context should be labeled something like "after .empty? is called".

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
```diff
@@ -201,6 +201,122 @@
...
+      end
+
+      it "retains the correct length when calling to_a" do
+        expect(enumerable.to_a.length).to eq(2)
+      end
+
+      context "when iterating over the relation a second time" do
```
p-mongo, line 244: We use "association" because that is the correct term. "Relation" (per the original relational DB model) refers to values in the columns being related into a single table.  For BC reasons the method names retain "relation" in a number of places, but all comments, local variables, prose etc. should use "association".

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
johnnyshields: I'll do it here, but we really should just create a PR which does this across the entire repo. The reason I didn't use `create!` is because I copy-pasted from existing specs.

I can raise the PR once I have CI visibility. It will break many tests.

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
johnnyshields: Again, I've just copy-pasted from existing specs. I will do the change here, but I recommend you raise a PR which does this change uniformly across the entire repo.

lib/mongoid/association/referenced/has_many/enumerable.rb
johnnyshields: Yes, fixed.

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
johnnyshields: Fixed in this PR.

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
johnnyshields: Because I copied the spec from `#any` and it was done that way there. I've changed it for both.

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
johnnyshields: OK, I've changed it. Again I copied an existing spec.

spec/mongoid/association/referenced/has_many/enumerable_spec.rb
johnnyshields: Changed in this PR.

johnnyshields:
@p-mongo all requested changes done, let's merge.

p-mongo:
Hi Johnny,

> It would be great if you can raise PRs to fix some of the issues you highlighted uniformly across the codebase.

I'm not sure this is realistic. Changing existing tests would require us to review all of them to ensure the tests continue to work as intended. When the tests are otherwise being changed, or new tests are being added, those get reviewed, and at this point it is relatively clear whether `create!` can be used in place of `create` for example. So, to take this PR as an example, I reviewed about 350 lines of changes, but our entire codebase is almost 3 million lines of spec tests.

> As a general guideline, if something is done a certain way 1000 times in the codebase, we shouldn't make it a requirement to change for specific PRs in order to merge them--the fix should be done uniformly in a separate PR.

There was never a requirement to change existing code (unless it's being changed already for some other reason). The request was for new code to follow the current requirements.

> Contributors shouldn't be penalized for following apparently established conventions of the repo.

If by "penalized" you refer to changing existing code, again, you don't have to do that. With any established project there is some code that isn't written according to current practices, if this code happens to get touched by a change it may be necessary to bring it into compliance with the current requirements. Same thing for new code, it is helpful to copy-paste existing specs for example instead of coming up with new specs from scratch but if the existing specs don't meet current requirements they may need to be revised. You still obtain the benefit of referercing/copying existing specs.

johnnyshields:
> I'm not sure this is realistic.

It's completely realistic. I just did it here: https://github.com/mongodb/mongoid/pull/5061. If the specs are passing, then it's "working as intended".

p-mongo:
> If the specs pass, then it's "working as intended".

Unfortunately this isn't true. For example if you replace 10 specs with the same spec copied 10 times, the specs would pass but they wouldn't work "as intended".

p-mongo (approved):

comandeo (approved):

johnnyshields:
Thanks!

PR #5032, opened 2021-08-18
MONGOID-5099 Improve error when loading an empty or invalid config file

johnnyshields:
Fixes MONGOID-5099

Improves error when loading an empty or invalid mongoid.yml config file and adds specs to #load_yaml method. Note that certain errors (file missing, ERB, YAML parsing) bubble up as per existing behavior.

This change is only intended to avoid an internal cryptic error happening within Mongoid's code (NoMethodError (undefined method `[]' for false:FalseClass). Please also note that this does not change the conditions under which an exception occurs, it just changes the class of the exception.


p-mongo (approved):

comandeo (approved):

johnnyshields:
Thanks for the improvements!

PR #5033, opened 2021-08-18
MONGOID-5111 [Documentation Only] Add docs for limit/skip/batch_size

johnnyshields:
Fixes MONGOID-5111

- Documents limit, skip, and batch_size operators
- Add warning to Queries + Persistence section about limit, etc being ignored.

**QUESTION:** Is there a way I can preview these docs--what format are they? I've made this PR without actually seeing a rendered preview, so please check the docs are formatted properly (especially links, etc.)

### DIFF
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -987,6 +987,69 @@ the default scope being evaluated first:
     embedded: false>
 
 
+Pagination
+==========
+
+Mongoid provides the pagination operators: ``limit``, ``skip``, and ``batch_size``.
+
+.. _limit:
+
+``limit``
+---------
+
+``limit`` sets the total number of documents to be returned by a query:
+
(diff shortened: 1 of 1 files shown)


p-mongo:
Hi Johnny,

>  Is there a way I can preview these docs--what format are they?

The full build process for these is rather complicated. We have a "developer" build process documented in 
 https://github.com/mongodb/mongoid/blob/master/docs/README.md which should produce a close enough result - the content should be fully rendered but headers/navigation would be absent.

As far as I know you are the first contributor to inquire about building the docs, please let us know if the instructions in the referenced readme don't work for you.

p-mongo (approved):

johnnyshields:
Made one more minor change. Checked with sphinx and it looks good.

comandeo (approved):

PR #5036, opened 2021-08-18
MONGOID-4755 None criteria class should support #sum, #avg, #min, #max, #aggregates

johnnyshields:
Fixes MONGOID-4755

None criteria class should support #sum, #avg, #min, #max in a way that exactly matches the interfaces of normal Criteria class.

In addition:
- Adds missing #aggregates method to Aggregable::Memory (for interface consistency)
- Bug fix: Aggregable::Mongo#aggregates "sum" should be zero, not nil when no docs found.
- None#length is hardcoded to zero now; it doesn't need to call entries.length.
- Improve method documentation
- DRY up specs


p-mongo:
I created an integration test in https://github.com/mongodb/mongoid/pull/5044 which, after that PR is merged, should be used in this PR to ascertain that all contexts behave the same for the methods being modified.

I also see that the Aggregable implementations exist in their own modules (Aggregable::Mongo, Aggregable::Memory) and it seems to me that there should be Aggregable::None added for consistency with those rather than the aggregation methods being added to the top-level None.

johnnyshields:
@p-mongo good call on `Aggregable::None`. I've updated the PR and I found two additional things:
- Aggregable::Memory and Aggregable::None was missing #aggregates method (for interface consistency q/ Aggregable::Mongo)
- Bug fix: Aggregable::Mongo#aggregates "sum" should be zero, not nil when no docs found.

johnnyshields:
Let's merge?

@comandeo looks like one test has a transient failure here: https://github.com/mongodb/mongoid/pull/5036/checks?check_run_id=3555802141

p-mongo (approved):

comandeo (approved):

PR #5037, opened 2021-08-18
MONGOID-5141 [Code Comments Only] Remove all @since comments

johnnyshields:
Fixes MONGOID-5141

`@since` comments are generally not useful:

1. The project is now quite mature. `@since` was perhaps useful when there were radical changes between 1.0, 2.0, and 3.0, but those days are long gone.
2. They do not reflect significant interface changes to methods
3. They are not accurate, for example they have been kept in-place when code was moved to other classes
4. We're not adding them on new PRs, and it confuses first-time contributors
5. It adds clutter and makes code harder to read.

Instead of tagging each method with `@since`, we should simply be following Semver, and be mindful when we are change public method behavior.

### DIFF
--- a/lib/mongoid.rb
+++ b/lib/mongoid.rb
@@ -42,8 +42,6 @@ module Mongoid
   extend self
 
   # A string added to the platform details of Ruby driver client handshake documents.
-  #
-  # @since 6.1.0
   PLATFORM_DETAILS = "mongoid-#{VERSION}".freeze
 
   # The minimum MongoDB version supported.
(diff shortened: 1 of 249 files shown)


p-mongo:
I have no objections. @comandeo is currently on PTO, if he agrees when he's back we'll get this into master.

p-mongo:
Note that, since the commit doesn't include the ticket number, whoever is merging this PR should ensure the ticket number makes it into the final commit message.

johnnyshields:
> Note that, since the commit doesn't include the ticket number, whoever is merging this PR should ensure the ticket number makes it into the final commit message.

I'll try to do that from now on, was not aware that was the policy.

comandeo (approved):
:+1:

PR #5038, opened 2021-08-18
MONGOID-4500 + MONGOID-5421: Improvement to Mongoid::Contextual::Aggregable::Memory handling of nil / non-numeric values / type conversion.

johnnyshields:
Fixes MONGOID-4500

`Mongoid::Contextual::Aggregable::Memory` provides `#min, #max, #sum, #avg` for `Criteria` whose documents have currently been loaded to memory (most commonly embedded documents.) These functions are particularly useful when working with embedded documents.

This PR improves those methods so that the in-memory behavior of is closer (*but intentionally not exactly the same as*) MongoDB's server behavior. The primary difference with the server is that this PR allows summing/averaging mixed strings and integer values, e.g. for the use case that fields have `type: BigDecimal` (which is stored in the DB as a string.)

The following table summarizes the change for `sum, avg, min, max`:

| Case | Current In-Memory Behavior | New In-Memory Behavior | Database Behavior |
| --- | --- | --- | --- |
| All results are Integer/Float | Works correctly | Works correctly | Works correctly |
| nil is present in the results | Throws error | Ignores nil | Ignores nil |
| Non-numeric values (e.g. Date) are present in the results | Throws error | Ignores non-numeric | `sum, avg` = ignores non-numeric; `min, max` = behaves strangely (e.g. returns a seemingly random non-numeric value) |
| Numeric-like strings are present in the result (e.g. stored by BigDecimal) | Throws error (even when BigDecimal is used) | Converts numeric-like Strings to numbers | `sum, avg` = ignores numeric-like Strings (undesirable for user); `min, max` = behaves strangely |

### DIFF
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -382,6 +382,54 @@ Mongoid 7 behavior:
   end
 
 
+``max``, ``min``, ``avg``, ``sum`` Aggregations On Embedded Documents
+---------------------------------------------------------------------
+
+Mongoid 8.0 converts fields of embedded documents that are being numerically
+aggregated (that is, aggregated using the ``max``, ``min``, ``avg`` and ``sum``
+methods) to numbers, if possible, prior to performing the aggregation.
+Any values which are not numeric are ignored.
+
+This conversion permits numeric aggregations over ``BigDecimal`` values stored
+in the database as strings.
+
+Mongoid 8.0 behavior:
+
+.. code-block:: ruby
+
+  class Cat
+    include Mongoid::Document
+
+    field :age, type: Integer
+
+    embeds_many :kittens
+  end
+
+  class Kitten
+    include Mongoid::Document
+
+    embedded_in :cat
+
+    field :age, type: Integer
+  end
+
+  Cat.create!(age: 12)
+  Cat.create!(age: '11')
+  Cat.create!(age: 'hello')
+
+  Cat.sum(:age)
+  # => 23
+
+  cat = Cat.create!
+
+  cat.kittens.create!(age: 12)
+  cat.kittens.create!(age: '11')
+  cat.kittens.create!(age: 'hello')
+
+  cat.kittens.sum(:age)
+  # => 23
+
+
 ``#pluck`` on Embedded Criteria Returns ``nil`` Values
 ------------------------------------------------------
 
--- a/lib/mongoid/contextual/aggregable/memory.rb
+++ b/lib/mongoid/contextual/aggregable/memory.rb
@@ -27,9 +27,13 @@ module Mongoid
         #
         # @param [ Symbol ] field The field to average.
         #
-        # @return [ Float ] The average.
+        # @return [ Numeric ] The average.
         def avg(field)
-          any? ? sum(field).to_f / count.to_f : nil
+          total = count { |doc| doc.send(field).numeric? }
+          return nil unless total > 0
+
+          total = total.to_f if total.is_a?(Integer)
+          sum(field) / total
         end
 
         # Get the max value of the provided field. If provided a block, will
@@ -46,10 +50,12 @@ module Mongoid
         #
         # @param [ Symbol ] field The field to max.
         #
-        # @return [ Float, Document ] The max value or document with the max
+        # @return [ Numeric | Document ] The max value or document with the max
         #   value.
         def max(field = nil)
-          block_given? ? super() : aggregate_by(field, :max_by)
+          return super() if block_given?
+
+          aggregate_by(field, :max)
         end
 
         # Get the min value of the provided field. If provided a block, will
@@ -66,10 +72,12 @@ module Mongoid
         #
         # @param [ Symbol ] field The field to min.
         #
-        # @return [ Float, Document ] The min value or document with the min
+        # @return [ Numeric | Document ] The min value or document with the min
         #   value.
         def min(field = nil)
-          block_given? ? super() : aggregate_by(field, :min_by)
+          return super() if block_given?
+
+          aggregate_by(field, :min)
         end
 
         # Get the sum value of the provided field. If provided a block, will
@@ -83,13 +91,12 @@ module Mongoid
         #
         # @param [ Symbol ] field The field to sum.
         #
-        # @return [ Float ] The sum value.
+        # @return [ Numeric ] The sum value.
         def sum(field = nil)
-          if block_given?
-            super()
-          else
-            any? ? super(0) { |doc| doc.public_send(field) } : 0
-          end
+          return super() if block_given?
+          return 0 unless any?
+
+          aggregate_by(field, :sum)
         end
 
         private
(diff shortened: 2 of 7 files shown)


p-mongo:
Hi @johnnyshields , 

Have you seen https://docs.mongodb.com/manual/reference/bson-type-comparison-order/? This might explain the ordering of different data types you encountered.

johnnyshields:
Sure, that explains the min/max behavior from the DB when mixing types.

johnnyshields:
@p-mongo do you have any further comment here? Do you agree with my approach to do what is most sensible "in Ruby memory" rather than attempting to exactly replicate MongoDB (as is done in other `::Memory::` modules in Mongoid)

p-mongo:
I may need to do some more testing but my first thought was that having different sort order in different circumstances would be problematic. For example, wouldn't this mean that if a user changed an embedded association to a referenced one, their objects would change order?

johnnyshields:
@p-mongo can you please clarify your point here?

This PR only affects the `sum, avg, min, max` methods, which will yield the same result irrespective of sort order. It does not change the behavior of the `sort` method.

(Moreover, the existing behavior is that in-memory sorting vs. in-database sorting can be different; by design the `::Memory::` module does **not** attempt to emulate the exact behavior of the DB.)

p-mongo:
Hi @johnnyshields , I reviewed the server documentation (https://docs.mongodb.com/manual/reference/operator/aggregation/max/, https://docs.mongodb.com/manual/reference/operator/aggregation/min/, https://docs.mongodb.com/manual/reference/operator/aggregation/sum/, https://docs.mongodb.com/manual/reference/operator/aggregation/avg/) and the handling of nil values seems to be straightforward. If you can create a new PR with just the nil value change I think we will be able to review that quickly, and after that we can return to the rest of this PR.

johnnyshields:
@p-mongo this PR is not very complex. Please kindly review it as-is.

johnnyshields:
@p-mongo please mark this as `oleg-todo`. I have responded to Oleg's comment re: break into smaller PR's that I don't think such approach is appropriate here.

johnnyshields:
@p-mongo can we please merge this. This is safe to merge.

johnnyshields:
@p-mongo I've updated PR description to better explain the use-case related to BigDecimal (string vs integer summing)

johnnyshields:
@p-mongo please merge or mark this as "oleg-todo"

johnnyshields:
Can we please merge this?

johnnyshields:
Changes look good 👍

johnnyshields:
Let's merge this?

p-mongo:
Hi @johnnyshields , I was reviewing this PR again and when I was typing up the docs I saw no difference between 7.4 and 7.5 behavior. Given that the tests were all for Mongoid internals there isn't a test case to illustrate this either.

johnnyshields:
@p-mongo I'm not sure what you mean? In the description of the PR I've documented the behavior difference, and my original specs also covered this behavioral change: https://github.com/mongodb/mongoid/pull/5038/commits/a89c598a82d7f1bda701e9dadc00ab252b534881

It looks like you've now refactored it into YAML. Are you sure you've covered 100% of the original specs and cases in my table above?

johnnyshields:
Hmm... I'm looking at your YAML specs now. In it, you are only calling sum/aggregate on single documents in your arrays. This is poor test coverage and it doesn't capture the original intent of this PR, which is to handle summing **mixed type values**, such as `[nil, 100, 200.5, BigDecimal(4000)]`

johnnyshields:
@p-mongo can you please comment on this? (See comments above)

johnnyshields:
@p-mongo any chance to complete this ticket on your end?

johnnyshields:
@p-mongo I have made the specs more robust here. This is ready to merge, however, I have one question.

In my spec, I am seeing this behavior:

```
class Band
   ...
   field :mojo, type: Object
end

band = Band.create!(mojo: BigDecimal('123')

band.mojo #=> returns BSON::Decimal128 object !!
```

This is problematic to get a BSON::Decimal128, because:

* `BSON::Decimal128#numeric?` returns `false`
   * I've changed it to return `true`
* `BSON::Decimal128` does not implement Ruby stdlib `Numeric` interface, i.e. methods like `#to_i, #to_f, #coerce` etc.
  * I've implemented logic to call `BSON::Decimal128#to_big_decimal`

This feels a bit hacky. I think Mongo team should decide whether:
* A) Make BSON::Decimal128 should be a proper `Numeric`-like object, i.e. implement `#to_i, #to_f, #coerce` and `.is_a?(Numeric), numeric?` etc. These could delegate to `to_big_decimal` for now.
* B) Always convert BSON::Decimal128 to BigDecimal in Mongoid--never expose it to the user in a way that it could accidentally get mixed up in numeric calculations.
* C) Do both A and B -- have (B) "always use BigDecimal" be the golden path, but implement (A) for safety to avoid app errors.

Approach C would be my preference.

johnnyshields:
@Neilshweky I see you are improving support for BigDecimal in BsonRuby library. Please comment on the above?

p-mongo:
The fields need to be declared appropriately. If you want to be performing numeric operations on a field, the field needs to be of a sensible numeric type. Object is not appropriate in this case.

You can of course have fields of mixed types, but if you declare a field of type Object, then your application should, before any operation, inspect the type of value in a particular document and perform any needed conversions.

You may find  https://jira.mongodb.org/browse/MONGOID-5221 related to this discussion.

johnnyshields:
I agree that real-world applications **should** declare fields which they intend to use for math operations with a numeric type (Integer, Float, etc.)

**However**, it is also valid to use Object as a type, rely on default demongoization, and perform `sum`, etc. on such fields. The reason for doing this is to **mirror what MongoDB itself does** when calling `sum` on mixed data, which is the core goal of this PR. Though its not a standard use case, it can and should be supported.

johnnyshields:
@p-mongo this is ready for your final review.

johnnyshields:
@p-mongo let's merge this.

johnnyshields:
@p-mongo any reason why this PR hasn't been merged?

p-mongo:
The following are valid use cases:

1. Performing numeric aggregation over a value stored in the database as a number, with no declared application type

2. Performing numeric aggregation over a value stored in the database as a string, with declared application type of BigDecimal.

The case where the value is stored as a string and there is no declared application type is not one that is produced by Mongoid and is outside the scope of https://jira.mongodb.org/browse/MONGOID-4500.

p-mongo:
Going back to the ticket that this PR is for, https://jira.mongodb.org/browse/MONGOID-4500:

The change is for aggregation on memory context to ignore nil and non-numeric values.

Conversions between types aren't in scope of this ticket at all, if this is what you want to do please open a new ticket explaining your use case.

Please review this PR and ensure it only makes aggregation on memory context ignore nil and non-numeric values as per the ticket description.

johnnyshields:
@p-mongo I have raised MONGOID-5421 to cover the type conversion aspects. As the changes are closely related (analogous to Neil's PR for MONGOID-5222, in which he is changing nil / type conversion behavior together), I would like to ask to do both MONGOID-5421 and MONGOID-4500 together in the same PR.

johnnyshields:
I've raised a version of this PR here which only changes the nil handling behavior: https://github.com/mongodb/mongoid/pull/5382

johnnyshields:
Replaced by https://github.com/mongodb/mongoid/pull/5389

PR #5040, opened 2021-08-18
MONGOID-4737 Deprecate the :compact option on Document#as_json

johnnyshields:
Fixes MONGOID-4737

### DIFF
--- a/spec/mongoid/document_spec.rb
+++ b/spec/mongoid/document_spec.rb
@@ -501,7 +501,7 @@ describe Mongoid::Document do
       end
     end
 
-    context ':compact option' do
+    context 'deprecated :compact option' do
       # Since rails 6 differs in how it treats id fields,
       # run this test on one version of rails. Currently rails 6 is in beta,
       # when it is released this version should be changed to 6.
@@ -511,6 +511,22 @@ describe Mongoid::Document do
         # These tests require a specific set of defined attributes
         # on the model
         expect(church.as_json.keys.sort).to eq(%w(_id location name))
+      end
+
+      context 'deprecation' do
+        let(:church) do
+          Church.create!(name: 'St. Basil')
+        end
+
+        it 'logs a deprecation warning when :compact is given' do
+          expect_any_instance_of(Logger).to receive(:warn).with('#as_json :compact option is deprecated. Please call #compact on the returned Hash object instead.')
+          church.as_json(compact: true)
+        end
+
+        it 'does not log a deprecation warning when :compact is not given' do
+          expect_any_instance_of(Logger).to_not receive(:warn)
+          church.as_json
+        end
       end
 
       context 'there is a nil valued attribute' do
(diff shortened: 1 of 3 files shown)


p-mongo:
Hi Johnny, thank you for this PR.

I see that AS offers https://everydayrails.com/2021/07/31/rails-custom-deprecation-warnings.html, which Mongoid should be able to utilize. What do you think of using this deprecation machinery instead of simply writing to Mongoid's log?

An important question here is whether the deprecation log is going to the same logger as the one Mongoid uses in Rails apps, which I haven't investigated yet.

If the loggers are the same I think going through AS deprecation machinery would make sense.

johnnyshields:
By default `ActiveSupport::Deprecation.warn` is going to use Rails logger with a fallback to $stderr. However, it looks that there is a way to add a Mongoid-specific behavior that would use Mongoid's logger. I would vote to merge this first, and then do a separate PR which adds ActiveSupport::Deprecation consistently for all deprecations in the codebase. I'd be glad to handle that as a follow-up.

spec/mongoid/document_spec.rb
```diff
@@ -541,7 +557,7 @@ class << self; attr_accessor :name; end
         end
 
         it 'returns all fields and _id' do
-          actual = church.as_json(compact: true)
+          actual = church.as_json(compact: true).compact
```
p-mongo, line 560: Did you mean to remove the `compact: true` argument here?

p-mongo:
I filed https://jira.mongodb.org/browse/MONGOID-5146 to look into the AS deprecation machinery.

spec/mongoid/document_spec.rb
johnnyshields: Oops yes. Fixed.

johnnyshields:
Ready for merge.

p-mongo (approved):

spec/mongoid/document_spec.rb
```diff
@@ -513,6 +513,22 @@ class << self; attr_accessor :name; end
...
+          Church.create!(name: 'St. Basil')
+        end
+
+        it 'logs a deprecation warning when :compact is given' do
+          expect_any_instance_of(Logger).to receive(:warn).with('#as_json :compact option is deprecated. Please call #compact on the returned Hash object instead.')
+          church.as_json(compact: true)
+        end
+
+        it 'does not log a deprecation warning when :compact is not given' do
+          expect_any_instance_of(Logger).to_not receive(:warn)
```
comandeo, line 527: MINOR: I think this may be a potentially flaky test. Any warning logged anywhere in Mongoid will make this test fail. I believe it will be more robust (and, maybe, more readable) if we specify more explicitly what warning we should not receive.

comandeo (approved):

spec/mongoid/document_spec.rb
johnnyshields: Fixed.

PR #5041, opened 2021-08-18
MONGOID-5143 [Code Comments Only] Remove magic comment "# encoding: utf-8"

johnnyshields:
Fixes MONGOID-5143

It has not had any affect since Ruby 2.1, as that version made UTF-8 the default encoding. Min ruby we support is 2.3, so this is fine to remove.


p-mongo:
No objections.

comandeo (approved):

PR #5042, opened 2021-08-18
MONGOID-5144 Add a deprecation warning when using the legacy query cache

johnnyshields:


spec/mongoid/query_cache_spec.rb
```diff
@@ -29,6 +29,8 @@
       min_driver_version '2.14'
 
       it 'does not log a deprecation warning' do
+        Mongoid::QueryCache.remove_instance_variable('@legacy_query_cache_warned')
```
p-mongo, line 32: This happens to fail when the instance variable does not exist.

PR #5043, opened 2021-08-19
MONGOID-5131: Set up build pipeline for outside contributors

johnnyshields:
Fixes MONGOID-5131

Uses Github Actions based on an Evergreen auto-converter script.

You can regenerate the config with:

```
ruby .evergreen/make-github-actions
```

The script is pretty good/robust as long as there are no major changes to the Evergreen config format and variable nomenclature. Most configurations defined in Evergeen are supported. The following are skipped for now:
- `topology=sharded-cluster`. This is possible to support if we can get a better Github Actions config for running MongoDB itself (I opened a DRIVERS ticket for this but I can't find it--maybe it's private after I created it?)
- Two ad-hoc tests. Perhaps these can be cleaned up and/or integrated into the main suite in a separate ticket?
   - `i18n-with-fallbacks`
   - `app-test`

You can see the pipeline running on my Github repo: https://github.com/johnnyshields/mongoid/actions/runs/1148246726 (or click the colored dot below next to commit hash.)

Note that some of the builds are passing! The others look like failures due to flakey tests (it would be great if team could fix these--the first step to having reliable tests is to have a reliable and visible CI)

image


johnnyshields:
@alcaeus

johnnyshields:
@p-mongo what does the "deferred" label mean here? Can you check if there's a way to enable this build pipeline as "allowed to fail" so it runs in parallel with Evergreen, but is not required to merge a PR? It would be really useful for me.

p-mongo:
@alcaeus said he wanted to set up the CI pipeline so I'm going to leave this PR for him to comment on when he is back from PTO.

Unless we set up a requirement for green CI to merge PRs I do not see why red CI would prevent merging.

johnnyshields:
@alcaeus looks like we're waiting for your comment on this one.

.github/workflows/test.yml
```diff
@@ -0,0 +1,453 @@
+# This file was auto-generated by .evergreen/make-github-actions
+# at 2021-08-21T15:14:55Z
+---
+name: Run Mongoid Tests
+'on':
+- push
+- pull_request
```
alcaeus, line 7: Note: this will test every push and every pull request to all branches. For the PHP driver, I opted for the following configuration:
https://github.com/mongodb/mongo-php-library/blob/master/.github/workflows/tests.yml#L4..L11

This tests supported branches (the latest development branch and all stable release branches), both when changes are pushed and when a pull request targets one of those branches.

I don't think we necessarily need to limit this now, just pointing this out as we may find ourselves running more builds than we should, which I'd like to avoid in order to reduce load on open source build architecture.

.github/workflows/test.yml
```diff
@@ -0,0 +1,453 @@
...
+jobs:
+  build:
+    name: "${{matrix.ruby}} driver-${{matrix.driver}} mongodb-${{matrix.mongodb}}
+      ${{matrix.topology}}"
+    env:
+      CI: true
```
alcaeus, line 13: Thank you for adding this, I wanted to check if GH actions sets this automatically but haven't gotten to it. This ensures we definitely expose this as a CI job. 👍

.github/workflows/test.yml
```diff
@@ -0,0 +1,453 @@
...
+      TESTOPTS: "-v"
+    runs-on: ubuntu-latest
+    continue-on-error: "${{matrix.experimental}}"
```
alcaeus, line 16: Oh, I totally forgot about this one - I was wondering how to solve experimental builds (e.g. testing against new language versions). This may ensure that we're always testing builds we expect to be stable and could add them as required checks for merging.

.github/workflows/test.yml
```diff
@@ -0,0 +1,453 @@
...
+        include:
+        - mongodb: '4.4'
+          ruby: ruby-3.0
+          topology:
+          task: test
+          driver: current
+          rails:
+          i18n:
+          gemfile: Gemfile
+          experimental: false
```
alcaeus, line 29: I suppose this defines all builds in the matrix individually (instead of using `matrix.ruby` and `matrix.topology` arrays to have GH actions generate the matrix automatically) since it's auto-generated. No need to change this since we don't expect people to make manual changes in this file.

I'll note that it makes it a little more difficult to figure out what's being tested by looking at the file, but that's a problem that can be solved another day.

.github/workflows/test.yml
```diff
@@ -0,0 +1,453 @@
...
+          gemfile: gemfiles/rails-5.2.gemfile
+          experimental: false
+    steps:
+    - name: repo checkout
+      uses: actions/checkout@v2
+    - name: start mongodb
+      uses: supercharge/mongodb-github-action@1.6.0
+      with:
+        mongodb-version: "${{matrix.mongodb}}"
+        mongodb-replica-set: rs0
```
alcaeus, line 433: Depending on the configurations being used, we can leverage drivers-evergreen-tools for this:
```suggestion
    - name: start mongodb
      id: "setup-mongodb"
      uses: mongodb-labs/drivers-evergreen-tools@master
      with:
        version: "${{matrix.mongodb}}"
        topology: "${{matrix.topology}}"
```

I'll note that this may require changing the values of `matrix.topology`, so I'll let @p-mongo and @comandeo decide how to proceed here. I'm fine using something other than drivers-evergreen-tools if it makes the job at hand easier.

.github/workflows/test.yml
```diff
@@ -0,0 +1,453 @@
...
+        git submodule init
+        git submodule update
+      env:
+        BUNDLE_GEMFILE: "${{matrix.gemfile}}"
+    - name: test
+      timeout-minutes: 60
+      continue-on-error: "${{matrix.experimental}}"
+      run: bundle exec rake spec
+      env:
+        BUNDLE_GEMFILE: "${{matrix.gemfile}}"
```
alcaeus, line 453: Not sure how the connection string is passed to the driver, but in PHP we use the `MONGODB_URI` env variable that needs to be set to the output of the `start mongodb` step earlier:
```suggestion
        BUNDLE_GEMFILE: "${{matrix.gemfile}}"
        MONGODB_URI: ${{ steps.setup-mongodb.outputs.cluster-uri }}
```

alcaeus (commented):
@johnnyshields thanks for reminding me, this got lost in a heap of notifications.

I think the minimum that should be added by this PR is a single, stable build. This can test the latest stable versions of everything on a standalone topology, thus giving outside contributors a first indication if their changes work. The way this is set up, outside contributors would be able to push changes to their fork and see CI results, which is what we're after.

Since some builds are marked as experimental and allowed to fail (while also ensuring that non-experimental builds continue to run), I'm fine not having all builds stable if @p-mongo and @comandeo agree with me. If we do so, I'd appreciate if we could track the remaining work in Jira so as to not leave a bunch of failing CI around for longer than necessary.

As to the requirement of green CI, we currently have not yet enabled branch protection in this repository, so failing builds don't matter. Even with branch protection, we can choose which builds we require to be green for the PR to be mergeable, so we would be able to enable this functionality even if only a subset of the builds added here pass.

Once again, thanks @johnnyshields for working on this. I'll leave it to the Ruby team to make a decision as to what builds we require, but I do believe that we should focus on getting a few stable builds merged and taking care of more complex builds later. I would also like to point out that we should not test all possible builds on GH actions AND evergreen: a subset of builds required to be stable can help as a good indication whether the PR is ready for a review, or if there are build issues to figure out that a contributor could solve by themselves. More complex builds that require more computing time should remain in Evergreen only, so that we opt in to more stringent testing in the interest of saving resources. I'm aware that this means that it is the team's responsibility to fix failing builds in Evergreen, as outside contributors can't see those logs. I'm working on a process that allows us to spend more time working with outside contributors and get pull requests like this one handled in a more streamlined way.

.github/workflows/test.yml
johnnyshields: Let's run it all the time, so that contributors can see the results on their PRs. It's free--why not run it everywhere?

.github/workflows/test.yml
johnnyshields: Yep you got it.

johnnyshields:
@alcaeus I've merged your suggested changes (GH marks the comment as resolved when I do that).

It would be great if the Ruby team could take this over at this point as it's going to be hard for me as an outsider to follow all the idiosyncrasies of how you want to set this up. I think I've adequately demonstrated what's possible here--please take it to the finish line.

johnnyshields:
Also to reiterate my core point here: Github Actions can do everything that Evergreen can do, and it does it better, cleaner, more transparently, and at zero running cost to MongoDB, with zero concern about security etc.

GH Actions didn't exist at the time your company wrote Evergreen, but now it does. The MongoDB drivers team should push for an exemption from your corporate silliness and just go with Github.

comandeo (approved):

comandeo:
Specs fail on evergreen because I marked one test as `pending`. Weird. One more reason to move to GH actions :-)

johnnyshields:
image

p-mongo:
A pending test will fail the suite when the test succeeds, hence is not suitable for tests that e.g. fail on some configurations but succeed on others. To always skip a test, use `skip`.

comandeo:
@p-mongo Ready

spec/support/spec_config.rb
```diff
@@ -19,6 +19,9 @@ def addresses
     if @uri
       @uri.servers
     else
+      STDERR.puts "Environment variable 'MONGODB_URI' is not set, so the default url will be used."
+      STDERR.puts "This may lead to unexpected test failures because service discovery will raise unexpected warnings."
+      STDERR.puts "Please consider providing the correct uri via MONGODB_URI environment variable."
```
p-mongo, line 24: :+1:

p-mongo (approved):

johnnyshields:
Thank you MongoDB team for handling this 🚀

PR #5039, opened 2021-08-18
MONGOID-5136 5136


p-mongo:
https://jira.mongodb.com/browse/MONGOID-5136

PR #5045, opened 2021-08-20
MONGOID-5274 / MONGOID-5142 - Rework #touch with embedded documents

johnnyshields:
Fixes MONGOID-5142
Fixes MONGOID-5274
Correct fix for MONGOID-5136 (the previous one was non-optimal)
Replaces PR https://github.com/mongodb/mongoid/pull/5039 (includes specs from it, which are passing here)

## Changes

- Perform all touch operations on the document in a single write operation (MONGOID-5142)
- Remove touches from the atomic set operations, so that they are no longer there when saving the next time.
- Cleanup/refactor touch method to be much more readable. The code is also now symmetrical for parents and child docs, everything is nicely recursive.

### DIFF
--- a/docs/release-notes/mongoid-9.0.txt
+++ b/docs/release-notes/mongoid-9.0.txt
@@ -73,3 +73,37 @@ the type conversion.
 Note that in prior Mongoid versions, typecasting Date to Time during
 persistence operations was already correctly using the
 ``Mongoid.use_activesupport_time_zone`` setting.
+
+
+```#touch`` method on embedded documents correctly handles ``touch: false`` option
+----------------------------------------------------------------------------------
+
+When the ``touch: false`` option is set on an ``embedded_in`` relation,
+calling the ``#touch`` method on an embedded child document will not
+invoke ``#touch`` on its parent document.
+
+.. code-block:: ruby
+
+  class Address
+    include Mongoid::Document
+    include Mongoid::Timestamps
+
+    embedded_in :mall, touch: false
+  end
+
+  class Mall
+    include Mongoid::Document
+    include Mongoid::Timestamps
+
+    embeds_many :addresses
+  end
+
+  mall = Mall.create!
+  address = mall.addresses.create!
+
+  address.touch
+  #=> updates address.updated_at but not mall.updated_at
+
+In addition, the ``#touch`` method has been optimized perform one
+persistence operation per parent document, even when using multiple
+levels of nested embedded documents.
--- a/docs/release-notes/mongoid-9.0.txt
+++ b/docs/release-notes/mongoid-9.0.txt
@@ -104,7 +104,7 @@ invoke ``#touch`` on its parent document.
   address.touch
   #=> updates address.updated_at but not mall.updated_at
 
-In addition, the ``#touch`` method has been optimized perform one
+In addition, the ``#touch`` method has been optimized to perform one
 persistence operation per parent document, even when using multiple
 levels of nested embedded documents.
 
--- a/lib/mongoid/association/options.rb
+++ b/lib/mongoid/association/options.rb
@@ -117,6 +117,8 @@ module Mongoid
       #
       # @return [ true | false ] returns true if this association is
       #   automatically touched, false otherwise. The default is false.
+      #
+      # @api private
       def touchable?
         !!@options[:touch]
       end
--- a/lib/mongoid/association/options.rb
+++ b/lib/mongoid/association/options.rb
@@ -112,8 +112,11 @@ module Mongoid
         @touch_field ||= options[:touch] if (options[:touch].is_a?(String) || options[:touch].is_a?(Symbol))
       end
 
-      private
-
+      # Whether the association object should be automatically touched
+      # when its inverse object is updated.
+      #
+      # @return [ true | false ] returns true if this association is
+      #   automatically touched, false otherwise. The default is false.
       def touchable?
         !!@options[:touch]
       end
--- a/lib/mongoid/touchable.rb
+++ b/lib/mongoid/touchable.rb
@@ -81,8 +81,8 @@ module Mongoid
         updates = atomic_updates['$set']
         return {} unless updates
 
-        touchable_keys = Set.new(%w(updated_at u_at))
-        touchable_keys << field.to_s
+        touchable_keys = Set['updated_at', 'u_at']
+        touchable_keys << field.to_s if field.present?
 
         updates.keys.each_with_object({}) do |key, touches|
           if touchable_keys.include?(key.split('.').last)
--- a/lib/mongoid/touchable.rb
+++ b/lib/mongoid/touchable.rb
@@ -81,8 +81,8 @@ module Mongoid
         updates = atomic_updates['$set']
         return {} unless updates
 
-        touchable_keys = %w(updated_at u_at)
-        touchable_keys << field.to_s if field.present?
+        touchable_keys = Set.new(%w(updated_at u_at))
+        touchable_keys << field.to_s
 
         updates.keys.each_with_object({}) do |key, touches|
           if touchable_keys.include?(key.split('.').last)
--- a/lib/mongoid/touchable.rb
+++ b/lib/mongoid/touchable.rb
@@ -145,7 +145,7 @@ module Mongoid
           without_autobuild do
             if relation = __send__(name)
               # This looks up touch_field at runtime, rather than at method definition time.
-              # If touch_field is nil, it will only touch the default field (updated_at)
+              # If touch_field is nil, it will only touch the default field (updated_at).
               relation.touch(association.touch_field)
             end
           end
--- a/lib/mongoid/touchable.rb
+++ b/lib/mongoid/touchable.rb
@@ -144,13 +144,9 @@ module Mongoid
         define_method(method_name) do
           without_autobuild do
             if relation = __send__(name)
-              if association.touch_field
-                # Note that this looks up touch_field at runtime, rather than
-                # at method definition time.
-                relation.touch(association.touch_field)
-              else
-                relation.touch
-              end
+              # This looks up touch_field at runtime, rather than at method definition time.
+              # If touch_field is nil, it will only touch the default field (updated_at)
+              relation.touch(association.touch_field)
             end
           end
         end
--- a/lib/mongoid/touchable.rb
+++ b/lib/mongoid/touchable.rb
@@ -60,6 +60,9 @@ module Mongoid
         run_callbacks(:touch)
       end
 
+      # Indicates whether the parent exists and is touchable.
+      #
+      # @api private
       def __touchable_parent?
         _parent && _association&.inverse_association&.touchable?
       end
--- a/lib/mongoid/touchable.rb
+++ b/lib/mongoid/touchable.rb
@@ -22,46 +22,68 @@ module Mongoid
       # @return [ true/false ] false if document is new_record otherwise true.
       def touch(field = nil)
         return false if _root.new_record?
-        current = Time.configured.now
+
+        touches = __gather_touch_updates(Time.configured.now, field)
+        _root.send(:persist_atomic_operations, '$set' => touches) if touches.present?
+
+        __run_touch_callbacks_from_root
+        true
+      end
+
+      # Recursively sets touchable fields on the current document and each of its
+      # parents, including the root node. Returns the combined atomic $set
+      # operations to be performed on the root document.
+      #
+      # @param [ Time ] now The timestamp used for synchronizing the touched time.
+      # @param [ Symbol ] field The name of an additional field to update.
+      #
+      # @return [ Hash<String, Time> ] The touch operations to perform as an atomic $set.
+      #
+      # @api private
+      def __gather_touch_updates(now, field = nil)
         field = database_field_name(field)
-        write_attribute(:updated_at, current) if respond_to?("updated_at=")
-        write_attribute(field, current) if field
+        write_attribute(:updated_at, now) if respond_to?("updated_at=")
+        write_attribute(field, now) if field
 
-        # If the document being touched is embedded, touch its parents
-        # all the way through the composition hierarchy to the root object,
-        # because when an embedded document is changed the write is actually
-        # performed by the composition root. See MONGOID-3468.
-        if _parent
-          # This will persist updated_at on this document as well as parents.
-          # TODO support passing the field name to the parent's touch method;
-          # I believe it should be read out of
-          # _association.inverse_association.options but inverse_association
-          # seems to not always/ever be set here. See MONGOID-5014.
-          _parent.touch
+        touch_parent = _parent && _association&.inverse_association&.touchable?
 
-          if field
-            # If we are told to also touch a field, perform a separate write
-            # for that field. See MONGOID-5136.
-            # In theory we should combine the writes, which would require
-            # passing the fields to be updated to the parents - MONGOID-5142.
-            sets = set_field_atomic_updates(field)
-            selector = atomic_selector
-            _root.collection.find(selector).update_one(positionally(selector, sets), session: _session)
-          end
-        else
-          # If the current document is not embedded, it is composition root
-          # and we need to persist the write here.
-          touches = touch_atomic_updates(field)
-          unless touches["$set"].blank?
-            selector = atomic_selector
-            _root.collection.find(selector).update_one(positionally(selector, touches), session: _session)
+        touches = __extract_touches_from_atomic_sets(field) || {}
+        touches.merge!(_parent.__gather_touch_updates(now) || {}) if touch_parent
+        touches
+      end
+
+      # Recursively runs :touch callbacks for the document and its parents,
+      # beginning with the root document and cascading through each successive
+      # child document.
+      #
+      # @api private
+      def __run_touch_callbacks_from_root
+        _parent.__run_touch_callbacks_from_root if _parent
+        run_callbacks(:touch)
+      end
+
+      private
+
+      # Extract and remove the atomic updates for the touch operation(s)
+      # from the currently enqueued atomic $set operations.
+      #
+      # @param [ Symbol ] field The optional field.
+      #
+      # @return [ Hash ] The field-value pairs to update atomically.
+      #
+      # @api private
+      def __extract_touches_from_atomic_sets(field = nil)
+        updates = atomic_updates['$set']
+        return {} unless updates
+
+        touchable_keys = %w(updated_at u_at)
+        touchable_keys << field.to_s if field.present?
+
+        updates.keys.each_with_object({}) do |key, touches|
+          if touchable_keys.include?(key.split('.').last)
+            touches[key] = updates.delete(key)
           end
         end
-
-        # Callbacks are invoked on the composition root first and on the
-        # leaf-most embedded document last.
-        run_callbacks(:touch)
-        true
       end
     end
 
@@ -82,7 +104,10 @@ module Mongoid
       association.inverse_class.tap do |klass|
         klass.after_save method_name
         klass.after_destroy method_name
-        klass.after_touch method_name
+
+        # Embedded docs handle touch updates recursively within
+        # the #touch method itself
+        klass.after_touch method_name unless association.embedded?
       end
     end
 
--- a/lib/mongoid/touchable.rb
+++ b/lib/mongoid/touchable.rb
@@ -45,10 +45,8 @@ module Mongoid
         write_attribute(:updated_at, now) if respond_to?("updated_at=")
         write_attribute(field, now) if field
 
-        touch_parent = _parent && _association&.inverse_association&.touchable?
-
         touches = __extract_touches_from_atomic_sets(field) || {}
-        touches.merge!(_parent.__gather_touch_updates(now) || {}) if touch_parent
+        touches.merge!(_parent.__gather_touch_updates(now) || {}) if __touchable_parent?
         touches
       end
 
@@ -58,8 +56,12 @@ module Mongoid
       #
       # @api private
       def __run_touch_callbacks_from_root
-        _parent.__run_touch_callbacks_from_root if _parent
+        _parent.__run_touch_callbacks_from_root if __touchable_parent?
         run_callbacks(:touch)
+      end
+
+      def __touchable_parent?
+        _parent && _association&.inverse_association&.touchable?
       end
 
       private
@@ -142,7 +144,7 @@ module Mongoid
               if association.touch_field
                 # Note that this looks up touch_field at runtime, rather than
                 # at method definition time.
-                relation.touch association.touch_field
+                relation.touch(association.touch_field)
               else
                 relation.touch
               end
--- a/lib/mongoid/touchable.rb
+++ b/lib/mongoid/touchable.rb
@@ -23,10 +23,10 @@ module Mongoid
       def touch(field = nil)
         return false if _root.new_record?
 
-        touches = __gather_touch_updates(Time.configured.now, field)
+        touches = _gather_touch_updates(Time.configured.now, field)
         _root.send(:persist_atomic_operations, '$set' => touches) if touches.present?
 
-        __run_touch_callbacks_from_root
+        _run_touch_callbacks_from_root
         true
       end
 
@@ -40,13 +40,13 @@ module Mongoid
       # @return [ Hash<String, Time> ] The touch operations to perform as an atomic $set.
       #
       # @api private
-      def __gather_touch_updates(now, field = nil)
+      def _gather_touch_updates(now, field = nil)
         field = database_field_name(field)
         write_attribute(:updated_at, now) if respond_to?("updated_at=")
         write_attribute(field, now) if field
 
-        touches = __extract_touches_from_atomic_sets(field) || {}
-        touches.merge!(_parent.__gather_touch_updates(now) || {}) if __touchable_parent?
+        touches = _extract_touches_from_atomic_sets(field) || {}
+        touches.merge!(_parent._gather_touch_updates(now) || {}) if _touchable_parent?
         touches
       end
 
@@ -55,15 +55,15 @@ module Mongoid
       # child document.
       #
       # @api private
-      def __run_touch_callbacks_from_root
-        _parent.__run_touch_callbacks_from_root if __touchable_parent?
+      def _run_touch_callbacks_from_root
+        _parent._run_touch_callbacks_from_root if _touchable_parent?
         run_callbacks(:touch)
       end
 
       # Indicates whether the parent exists and is touchable.
       #
       # @api private
-      def __touchable_parent?
+      def _touchable_parent?
         _parent && _association&.inverse_association&.touchable?
       end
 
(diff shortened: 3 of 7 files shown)


johnnyshields:
@p-mongo if you can help with the specs here would be greatly appreciated.

p-mongo:
Hi @johnnyshields , can you please update the title and description of this PR to reflect what it contains? MONGOID-5136 is already fixed. 

Can you also clarify whether this PR is ready for review? If it isn't it can be marked [wip] or draft.

johnnyshields:
@p-mongo I've marked this as WIP. The ticket is MONGOID-5142. I was hoping someone from the Mongo team could contribute the specs listed as TODOs.

johnnyshields:
I've raised https://github.com/mongodb/mongoid/pull/5202 to point out some strange/incorrect behavior re: `#touch` method. This PR is on hold until the issues are resolved.

johnnyshields:
@Neilshweky please review this.

lib/mongoid/touchable.rb
```diff
@@ -22,46 +22,68 @@ module InstanceMethods
...
+      #
+      # @api private
+      def __extract_touches_from_atomic_sets(field = nil)
+        updates = atomic_updates['$set']
+        return {} unless updates
+
+        touchable_keys = %w(updated_at u_at)
+        touchable_keys << field.to_s if field.present?
+
+        updates.keys.each_with_object({}) do |key, touches|
```
Neilshweky, line 82: TIL this is `reduce` without having to explicitly return the object. Cool.

Neilshweky:
If I understand this correctly, your PR makes the following changes: 
- touches are now persisted immediately (backwards breaking).
  - which causes touch callbacks to be called earlier.
- touches are no longer included in the atomic_updates calculation on `save`.
  - can you put a test for this?
- touch operations are done in one operation.
- embedded touches no longer use after_touch callbacks to update associated documents.

Would you say this is a good summary of the effective changes in this PR?

I think the only backwards breaking change here would be the fact that touches are persisted immediately. Personally this doesn't bother me as a change, although release notes would have to be written about it. @p-mongo is that change okay in your opinion?

Neilshweky (commented):
Please review my comments above.

lib/mongoid/association/options.rb
```diff
@@ -112,8 +112,10 @@ def touch_field
         @touch_field ||= options[:touch] if (options[:touch].is_a?(String) || options[:touch].is_a?(Symbol))
       end
 
-      private
-
+      # Whether the association object should be automatically touched
+      # when its inverse object is updated.
+      #
+      # @return [ true, false ] Default is false.
```
Neilshweky, line 118: # @return [ true | false ] returns true if this association is
#   automatically touched, false otherwise. The default is false.

johnnyshields:
@Neilshweky thanks for the review. Let me first respond to your summary of changes so we're on same page:

> 1. touches are now persisted immediately (backwards breaking). which causes touch callbacks to be called earlier.
I think the old code was the same. The old code has `_root.collection.find(selector).update_one(...)`. Please double check this one?

> 2. touches are no longer included in the atomic_updates calculation on save. can you put a test for this?
No. `touch_atomic_updates` was never actually part of `atomic_updates`. Its code just happened to be in the same file, where it doesn't belong.

> 3. touch operations are done in one operation.
Yes. This is the primary intention of the PR.

> 4. embedded touches no longer use after_touch callbacks to update associated documents.
In the new code, calling the `#touch` method on an embedded now recursively touches its parent(s) within the `#touch` method itself, so there's no need for to use `after_touch` callbacks. So this is not actually a change to the end result behavior.

lib/mongoid/touchable.rb
johnnyshields: I learned this and many other lessons from Rubocop 😅

lib/mongoid/touchable.rb
```diff
@@ -22,46 +22,68 @@ module InstanceMethods
...
+      # Extract and remove the atomic updates for the touch operation(s)
+      # from the currently enqueued atomic $set operations.
+      #
+      # @param [ Symbol ] field The optional field.
+      #
+      # @return [ Hash ] The field-value pairs to update atomically.
+      #
+      # @api private
+      def __extract_touches_from_atomic_sets(field = nil)
+        updates = atomic_updates['$set']
```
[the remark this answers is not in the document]
johnnyshields: Hmm... I'm keeping this consistent with the old `set_field_atomic_updates` method. Calling `atomic_updates` will build the updates, and I think `$set` will then include both `atomic_sets` and `delayed_atomic_sets`. If you're 100% confident there won't be knock-on effects, I'm fine to change, otherwise I'd say do it in a follow-up PR.

lib/mongoid/association/options.rb
johnnyshields: Done

spec/mongoid/touchable_spec.rb
```diff
@@ -77,51 +77,33 @@
...
           entrance.touch
 
-          building.updated_at.should == update_time
+          expect(building.updated_at).to eq start_time
         end
 
         it 'does not persist updated updated_at on parent' do
           entrance
           update_time
           entrance.touch
```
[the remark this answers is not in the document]
johnnyshields: I've modified it. Erm... its a bit strange here, ":touch is not set" is for embedded while ":touch is false" is for referenced; the specs themselves are duplicated code. I've improved it now with this:

```ruby
      shared_examples 'does not update the parent when :touch is not set' do
        it_behaves_like 'does not update the parent when :touch is false'
     end
```

Problem solved.

p-mongo:
Neither https://jira.mongodb.org/browse/MONGOID-5142 nor https://jira.mongodb.org/browse/MONGOID-5142 specify making touch perform an immediate write, so I would say this shouldn't be done.

If you believe it is a desired behavior, @johnnyshields please feel free to create a ticket for it describing your use case for it. My thinking is that perhaps `touch!` could be added which would immediately persist the write (and that would have to deal with validations, for example what if there is a validation requiring updated_at to be older than a certain time, touching such a document could make it invalid).

p-mongo:
https://jira.mongodb.org/browse/MONGOID-5274 was the other ticket.

johnnyshields:
> perform an immediate write, so I would say this shouldn't be done.

@p-mongo correct me if I'm wrong, but as mentioned in comment above to Neil it seems the existing code also performs an immediate write. See `_root.collection.find(selector).update_one(...)` in legacy code.

johnnyshields:
@Neilshweky I've responded to all your comments here. Please kindly resolve the comments or give further response.

Neilshweky:
> > perform an immediate write, so I would say this shouldn't be done.
> 
> @p-mongo as mentioned in comment above to Neil, I believe the existing code also performs an immediate write. See `_root.collection.find(selector).update_one(...)` in existing code.
> 
> I agree that an immediate write is **not** desirable, however, here I intentionally tried to keep the behavior consistent. I think we should defer the write in a follow-up ticket.

I can confirm that the current functionality on master is to persist touch operations immediately.

Neilshweky:
> Can you add a test for touching with grandchildren? I would like to see how this works with multiple generations of descendants.

@johnnyshields can you take care of this and then I'll give this another look?

spec/mongoid/touchable_spec.rb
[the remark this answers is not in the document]
Neilshweky: Do you think you can make the models consistent in touchable_spec? 

Right now we have:

- TouchableSpec
  - Embedded
    - Entrance (touch not set on embedded_in)
    - Floor (touch true on embedded_in)
  - Referenced
    - Entrance (touch false on embedded_in)
    - Floor (touch true on embedded_in)

Can you change entrance on embedded to have touch: false for Embedded, and create a third model (maybe exits?) that has touch not set, then explicitly create tests for touch not being set?

Neilshweky (commented):
Please make the testing changes mentioned above:
- multi-generational test
- third model with touch not set

johnnyshields:
@Neilshweky thanks for the fast response.

> Please make the testing changes mentioned above

Sure I'll do in next 1-2 days

> I can confirm that the current functionality on master is to persist touch operations immediately.

I've raised https://jira.mongodb.org/browse/MONGOID-5392 to track the fix for this. It should be done separately from this PR.

johnnyshields:
@Neilshweky so firstly great idea on doing multi-level specs. I've committed some WIP, however, I found a bug which is a blocker: https://jira.mongodb.org/browse/MONGOID-5398 - Classnames with modules dont work for `embedded_model._association&.inverse_association`. Refer to ticket for example.

(Given this bug, I'm sort of confused on how any specs are working at all with embeds here. Strange.)

johnnyshields:
@Neilshweky polite reminder -- I'm still blocked here due to MONGOID-5398. I'm glad to resume work and complete this PR whenever MONGOID-5398 is resolved.

Neilshweky:
Hey @johnnyshields, taking another look at this rn.

A few things:
1. I checked out this PR and ran the tests in touchable_spec and they all seem to be failing. Why are they failing?
2. Do you have a test that shows the failure related to MONGOID-5398? Can you put one in touchable_spec and mark it pending?
3. MONGOID-5398 seems to be a little far out, and I think it might require a major version (just conjecture but that's what I would guess), so I'm wondering if there's a way we can get around this. I have some ideas so I would love for you to get those tests fixed and put in that test for the module problems.

johnnyshields:
@Neilshweky

1. I encountered MONGOID-5398 while writing the tests, so I just committed messy WIP here; I'm aware the tests aren't passing.

2. MONGOID-5398 contains a clear demonstration of the failure. The specs are the ones at the end of `touchable_spec.rb` with the parent/child/grandchild touches. If you read the ticket body of MONGOID-5398 it should be obvious how it correlates to what I have in `touchable_spec.rb`.

3. / 4. The workaround would be to change the tests classes here to get rid of namespaces, and that would sidestep MONGOID-5398 and make the PR tests **appear to be green**. **However** the touch functionality would still be **broken in the real-world** for embedded classes with namespaces, which is a very common use case (e.g. 100% of embedded classes in my app are), so I think this is not a particularly meaningful approach.

Neilshweky:
@johnnyshields I personally think a lot of what's in this PR is good and I would like for some of these changes to make it into master...

Would you mind cleaning up this PR and getting it into a form that would be releasable other than the MONGOID-5398 problems? 

I am generally okay with putting this into master and just saying that classes inside modules need to include a class_name option on their associations in the context of this ticket since that doesn't really have anything to do with this ticket, it just is other buggy behavior.

I also wonder if there's a way to avoid using the `inverse_association` method altogether so I would like to play with it if you don't mind.

johnnyshields:
OK, I was unaware that adding `class_name` fixes [MONGOID-5398](https://jira.mongodb.org/browse/MONGOID-5398). I've just tested and it indeed works 🎉. That's enough for me to get around the blocker here and complete this PR. Please give me 2-3 weeks or so b/c I'm a bit busy at [work](https://www.tablecheck.com/en/join/) recently.

Neilshweky:
> OK, I was unaware that adding `class_name` fixes [MONGOID-5398](https://jira.mongodb.org/browse/MONGOID-5398). I've just tested and it indeed works 🎉. That's enough for me to get around the blocker here and complete this PR. Please give me 2-3 weeks or so b/c I'm a bit busy at [work](https://www.tablecheck.com/en/join/) recently.

Hey @johnnyshields, just checking on this!

Neilshweky:
Hey @johnnyshields, two things: 
- I have started reviewing some of those PRs:  
  - I reviewed #5267 and I'm waiting on your answer
  - Those two doc PRs I plan on reading those before the end of the week
- I believe that this PR could add value that we would be ready to use in the current state of Mongoid. Some of the other PRs that you have open are part of larger projects (i.e. attribute storage refactoring), that we simply have not started. 

Now, in order for us to accept this PR we need a few things:
- Can you type up an example with two levels of embedded documents and type up the queries before and after these changes?
  - This would demonstrate the value of this PR, as well as the correctness of the implementation. 
- Can you fix all of the tests that are still broken?

johnnyshields:
@Neilshweky thank you for your help to get the other PRs reviewed. I will try to do a pass on this PR over the weekend.

johnnyshields:
@Neilshweky need a bit more time on this but definitely still in my queue.

Neilshweky:
> @Neilshweky need a bit more time on this but definitely still in my queue.

Hey what's the status on this?

johnnyshields:
@Neilshweky I'll look at it now.

johnnyshields:
@Neilshweky all multi-level specs done; this is ready for final review. Good idea by the way, I did catch one edge case while writing the multi-level specs.

Neilshweky:
@johnnyshields looks like there's still some failures

johnnyshields:
@Neilshweky please check my [most recent commit](https://github.com/mongodb/mongoid/pull/5045/commits/e9b1d38c3bdae806db24970f03208ebef741f71f). I believe the failing spec was incorrect; multi-level touch should not work unless `touch: true` is explicitly set on the relations.

johnnyshields:
Tests are green 🎉

spec/mongoid/touchable_spec.rb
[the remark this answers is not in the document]
johnnyshields: Adding a third model can be done as part of [MONGOID-5016](https://jira.mongodb.org/browse/MONGOID-5016) where we plan to change the default behavior.

spec/mongoid/touchable_spec.rb
```diff
@@ -731,11 +731,309 @@
...
+
+      context 'parent > referenced child > referenced grandchild' do
+
+        let(:parent_cls) { TouchableSpec::Referenced::Building }
+
+        let(:child_cls) do
+          TouchableSpec::Referenced::Floor
+        end
+
+        context 'grandchild touch: true' do
```
Neilshweky, line 1005: is there a reason you didn't expand these tests into `child touch: true/false?` like you did the other ones?

lib/mongoid/touchable.rb
```diff
@@ -22,46 +22,73 @@ module InstanceMethods
...
+        return {} unless updates
+
+        touchable_keys = %w(updated_at u_at)
+        touchable_keys << field.to_s if field.present?
```
Neilshweky, line 85: maybe make this as a set being that we're checking membership in it? Marginal performance improvement but no reason not to...

lib/mongoid/touchable.rb
```diff
@@ -22,46 +22,73 @@ module InstanceMethods
...
-            _root.collection.find(selector).update_one(positionally(selector, touches), session: _session)
-          end
-        end
+        write_attribute(:updated_at, now) if respond_to?("updated_at=")
+        write_attribute(field, now) if field
 
-        # Callbacks are invoked on the composition root first and on the
-        # leaf-most embedded document last.
+        touches = __extract_touches_from_atomic_sets(field) || {}
+        touches.merge!(_parent.__gather_touch_updates(now) || {}) if __touchable_parent?
```
Neilshweky, line 49: So I'm tracing through this code and this feels a little repetitive, can you explain why this is necessary? I'll explain... 

At the deepest child (take the building > floor > sofa example), the `__gather_touch_updates` call returns:
```
{ "floors.0.sofas.0.updated_at" => 2022-10-03 15:12:53 UTC }
```
Now one level up from that (the floor level): 
```
{ "floors.0.updated_at" => 2022-10-03 15:12:53 UTC, "floors.0.sofas.0.updated_at" => 2022-10-03 15:12:53 UTC }
```
And at the top (building) level: 
```
{ "updated_at" => 2022-10-03 15:12:53 UTC, "floors.0.updated_at" => 2022-10-03 15:12:53 UTC, "floors.0.sofas.0.updated_at" => 2022-10-03 15:12:53 UTC }
```
This, I think, results from the fact that the atomic updates of a child are included in its parent (does this hold always?) 

Would a more efficient solution be to find the greatest touchable ancestor and ``__extract_touches_from_atomic_sets`` once from there?

lib/mongoid/touchable.rb
```diff
@@ -117,7 +147,7 @@ def define_relation_touch_method(name, association)
               if association.touch_field
                 # Note that this looks up touch_field at runtime, rather than
                 # at method definition time.
-                relation.touch association.touch_field
+                relation.touch(association.touch_field)
```
Neilshweky, line 150: I have a feeling this conditional is unnecessary... If the touch_field doesn't exist we'll just pass `nil` anyway which is the default parameter value... Might not be worth changing but just a note.

Neilshweky (commented):
@johnnyshields great work here! I have a few comments below if you don't mind addressing them...

lib/mongoid/touchable.rb
```diff
@@ -22,46 +22,73 @@ module InstanceMethods
...
+      # parents, including the root node. Returns the combined atomic $set
+      # operations to be performed on the root document.
+      #
+      # @param [ Time ] now The timestamp used for synchronizing the touched time.
+      # @param [ Symbol ] field The name of an additional field to update.
+      #
+      # @return [ Hash<String, Time> ] The touch operations to perform as an atomic $set.
+      #
+      # @api private
+      def __gather_touch_updates(now, field = nil)
```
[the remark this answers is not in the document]
johnnyshields: I think I got it from somewhere else in the code, I forget. I can make it single _ if you prefer.

The reason for _ prefix is similar to `@api private`, it denotes a method that really shouldn't be called externally.

lib/mongoid/touchable.rb
johnnyshields: If I remember correctly, the reason for doing this is that, not only do we want to get the document updates (which are applied to root), we also want to mutate the objects in-memory to reflect the touch. Hence all the `write_attribute` calls. Does this make sense?

lib/mongoid/touchable.rb
johnnyshields: Done.

lib/mongoid/touchable.rb
johnnyshields: It is unnecessary, great catch. I've removed it and added a comment instead.

lib/mongoid/touchable.rb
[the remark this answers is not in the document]
johnnyshields: I've gone ahead and used single _

lib/mongoid/touchable.rb
Neilshweky: you can do all that without calling `__extract_touches_from_atomic_sets` on each "generation", no?

lib/mongoid/touchable.rb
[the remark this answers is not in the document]
Neilshweky: okay cool

lib/mongoid/touchable.rb
```diff
@@ -22,46 +22,73 @@ module InstanceMethods
       # @return [ true/false ] false if document is new_record otherwise true.
       def touch(field = nil)
         return false if _root.new_record?
-        current = Time.configured.now
+
+        touches = __gather_touch_updates(Time.configured.now, field)
+        _root.send(:persist_atomic_operations, '$set' => touches) if touches.present?
```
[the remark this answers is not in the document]
Neilshweky: yeah for sure... That one should be a lot more easily rebased than this one will... I wouldn't wish such a fate on you LMAO

spec/mongoid/touchable_spec.rb
johnnyshields: Originally I thought it wasn't needed, but why not? I've just added it.

lib/mongoid/touchable.rb
```diff
@@ -144,13 +144,9 @@ def define_relation_touch_method(name, association)
...
             if relation = __send__(name)
-              if association.touch_field
-                # Note that this looks up touch_field at runtime, rather than
-                # at method definition time.
-                relation.touch(association.touch_field)
-              else
-                relation.touch
-              end
+              # This looks up touch_field at runtime, rather than at method definition time.
+              # If touch_field is nil, it will only touch the default field (updated_at)
```
Neilshweky, line 148:
```suggestion
              # If touch_field is nil, it will only touch the default field (updated_at).
```

johnnyshields:
@Neilshweky all comments addressed.

lib/mongoid/touchable.rb
Neilshweky: nitpick... will do a more full review when you address the rest of my comments

spec/mongoid/touchable_spec.rb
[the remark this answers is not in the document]
Neilshweky: yeah adding the default is not necessary on second thought.

lib/mongoid/touchable.rb
johnnyshields: My brain is out of mana on this issue. I suggest we merge as-is and do the refactor as a follow-up.

You're probably right that it can be made more efficient in Ruby, however, the first thing this PR is trying to achieve is to make the database calls more efficient (which it achieves).

spec/mongoid/touchable_spec.rb
Neilshweky: Thanks!

johnnyshields:
@Neilshweky I appreciate if you can take over this PR from here, if there are further changes needed. I have a busy few weeks ahead of me and won't be about to devote further attention for awhile.

lib/mongoid/touchable.rb
Neilshweky: I'm not quite ready to approve this, if you don't mind just thinking out this last point with me.... I decided to see how difficult this would be to fix, and I discovered an interesting decision you made (I'm not sure if this was accidentally or purposefully), but I would like to bring it up.... 

I noticed that in the recursive call to `_gather_touch_updates` the field is not passed to the parent as well. Now, is this done on purpose, since the given field should only be touched on the deepest level, whereas on its ancestors the `updated_at` field should be touched? 

If this is true, then it makes sense to call `_extract_touches_from_atomic_sets` on each level.

Neilshweky: However I noticed one other thing which is _not_ taken into account, and I would be okay with leaving that as a followup, but MONGOID-5508 is not taken into account here. The recursive call here would have to check the association for a touch field and pass it in.

Neilshweky: Verify the above comments for me and then I should be okay to approve.

Neilshweky:
@comandeo I am pretty much done with my review, if you wouldn't mind taking a look here.

lib/mongoid/touchable.rb
johnnyshields: > Is this done on purpose, since the given field should only be touched on the deepest level, whereas on its ancestors the updated_at field should be touched.

The intent is that `child.touch(:field_a)` would touch `:field_a` and `:updated` on the child, and then *only* touch `:updated_at` (not `:field_a`) on the parent. Supposing I also have:

```
class Child
  embedded_in :parent, touch: :field_b
```

Then I need to touch `:field_b` and `:updated_at` on the parent, regardless of which field I touch on the child.

Does that make sense?

johnnyshields: As we've noted elsewhere, there are more refactors needed especially for persistence-related auto-touching, so I expect we'll be refactoring all this again anyway.

Neilshweky: yes it does make sense.... My latter point was that this: 
>  Then I need to touch :field_b and :updated_at on the parent, regardless of which field I touch on the child.

is not being accomplished in this PR. I'm okay with that, and I'm willing for that to be done in MONGOID-5508.

Neilshweky:
Good job @johnnyshields, this really cleans up this part of the code.

comandeo (approved):
Great job! :clap:

Neilshweky:
Open questions/todos:

- Should this go into 8.1/9.0?

Let's do 9.0 since the repairing of `touch: false` on embedded associations can be a BC break.

- What is default behavior of touch option for embedded associations?

Does this PR change the default touch behavior for embedded associations?

- Since we have decided to put this into 9.0 there needs to be release notes stating that `touch: false` for embedded associations has been repaired... @johnnyshields can you handle that? You can even open a separate PR if you like...

johnnyshields:
@Neilshweky agreed it should be 9.0.

Are we also going to change the default value to touch: true for embedded? That will have some bearing on the release notes.

Neilshweky:
I would love to change the default of touch to true for embedded associations, and I will write up a proposal for it in a ticket (MONGOID-5016?)... Assume that we will be changing the default to true, so that (2) will not change... Are there any backwards breaking changes in this PR? Anything that we would need to put in the release notes?

johnnyshields:
There are no backwards breaking changes. The release notes for the touch: true could have a line like:

`In addition, there have been some refactor of the #touch method to generate fewer queries, and edge cases where touch: false was not working previously have been standardized.`

Important note `touch: true` would only be the new default for `embedded_in` relations (i.e. touch child causes touch on parent). Referenced relations need to stay as `touch: false`.

There's also some ticket to remove `touch: false` support on `embedded_in`, but I think that's a bridge too far for Mongoid 9.0 in terms of preserving behavior. Should revisit that later.

Neilshweky:
@johnnyshields take a look at the last commit I made here. Turns out in one of the tests that used a `meth`, the `shared_examples` wasn't using it. I changed that and fixed the test (there's a problem in MONGOID-5504 that I will have to fix)... I also explicitly specified the `touch` option on all models so that when we flip the default, the tests won't fail... I'll add new tests when I do MONGOID-5016

johnnyshields:
Ah yep, great catch! Last commit is good.

Neilshweky:
In the interest of keeping my head on straight, here are the TODOs and followup work: 
- write 9.0 release notes @johnnyshields is handling this (if necessary)
- Followup: flip the default of the `touch` option on embedded associations to true MONGOID-5016
- Followup: verify that the `touch: :field` functionality works correctly MONGOID-5508

Neilshweky:
> There are no backwards breaking changes. The release notes for the touch: true could have a line like:

I think this might be a good idea but we can do this as part of MONGOID-5016
> `In addition, there have been some refactor of the #touch method to generate fewer queries, and edge cases where touch: false was not working previously have been corrected.`
> 
> Important note - `touch: true` would only be the new default for `embedded_in` relations (i.e. touch child causes touch on parent). Referenced relations need to stay as `touch: false`.
 
This is MONGOID-5016 which I repurposed for switching the default
> There's also some ticket to remove `touch: false` support on `embedded_in`, but I think that's a bridge too far for Mongoid 9.0 in terms of preserving behavior. Should revisit that later.

Neilshweky:
@johnnyshields even though there are no backwards breaking changes, the `touch: false` option has been fixed and I think that warrants a release note.

docs/release-notes/mongoid-9.0.txt
```diff
@@ -75,6 +75,40 @@ persistence operations was already correctly using the
...
+    embeds_many :addresses
+  end
+
+  mall = Mall.create!
+  address = mall.addresses.create!
+
+  address.touch
+  #=> updates address.updated_at but not mall.updated_at
+
+In addition, the ``#touch`` method has been optimized perform one
```
Neilshweky, line 107:
```suggestion
In addition, the ``#touch`` method has been optimized to perform one
```

Neilshweky (approved):

johnnyshields:
@Neilshweky your spec change adding `touch: false` seems to have broken a few specs. Suggest to revert it and fix it in the next PR.

Neilshweky:
I will fix it and merge when its done

PR #5048, opened 2021-08-20
Fix MONGOID-5006 Link default auth source documentation to driver instead of incorrectly claiming "admin" is always the default

johnnyshields:
Fixes MONGOID-5006

Documentation only.

Also fixes some broken links.


docs/reference/callbacks.txt
```diff
@@ -13,7 +13,7 @@ Callbacks
 Document Callbacks
 ==================
 
-Mongoid supports the following callbacks for :doc:`documents </tutorials/documents>`:
+Mongoid supports the following callbacks for :doc:`documents </tutorials/mongoid-documents>`:
```
p-mongo, line 16: I don't think this is right: https://github.com/mongodb/mongoid/blob/master/docs/tutorials/documents.txt

docs/tutorials.txt
```diff
@@ -11,4 +11,4 @@ Tutorials
 
   tutorials/getting-started-sinatra
   tutorials/getting-started-rails
-  tutorials/documents
+  tutorials/mongoid-documents
```
p-mongo, line 14: I do not understand this change either, can you explain please?

p-mongo (commented):
Other than the documents -> mongoid-documents change which I do not understand, the rest of this PR has my approval.

docs/reference/callbacks.txt
johnnyshields: Try for yourself:
* https://docs.mongodb.com/mongoid/current/tutorials/documents/
* https://docs.mongodb.com/mongoid/current/tutorials/mongoid-documents/

johnnyshields: Oops, you are correct. The docs which are online at docs.mongodb.com are stale. Fixed this so resolving.

docs/tutorials.txt
johnnyshields: You were correct. Resolved

johnnyshields:
Documents -> mongoid-documents change is reverted. Please merge, thanks!

comandeo (approved):

johnnyshields:
Let's merge?

docs/reference/configuration.txt
```diff
@@ -168,7 +168,10 @@ can be configured.
           # on 2.4 and 2.6 is :plain)
           auth_mech: :scram
 
-          # The database or source to authenticate the user against. (default: admin)
+          # The database or source to authenticate the user against.
+          # If no auth source is specified, then a default will be set by the client
```
p-mongo, line 172: I am going to change "client" to "driver" because I think this is helpful to clarify that the source is determined by a whole other project from Mongoid.

johnnyshields: I agree 👍

johnnyshields: Just changed it now

docs/reference/configuration.txt
```diff
@@ -168,7 +168,12 @@ can be configured.
           # on 2.4 and 2.6 is :plain)
           auth_mech: :scram
 
-          # The database or source to authenticate the user against. (default: admin)
+          # When authenticating, look up the user in this database.
+          # This parameter is also called the authentication database or the
+          # auth source.
```
johnnyshields, line 173: How about:

```
Specify the auth source, i.e. the database or other source which contains the user's login credentials.
```

p-mongo (approved):

docs/reference/callbacks.txt
p-mongo: Yes, the master documentation on `docs.` isn't regularly updated (in fact in the last year+ it's only really been updated when we release, thus in practice it is the same as `current` most of the time).

docs/reference/configuration.txt
p-mongo: My apologies, I approve/request reviews from my tooling which doesn't show comments, so I didn't see this comment prior to approving & requesting review from @comandeo .

The phrase "or other source" wasn't clear to me, therefore I reviewed the definition of auth source. In https://docs.mongodb.com/ruby-driver/master/reference/authentication/#auth-source (which matched my existing understanding) we say:

> A user’s auth source is the database where that user’s authentication credentials are stored.

This claims that auth source is always the database.

However, I then checked server LDAP auth docs at https://docs.mongodb.com/manual/core/security-ldap-external/#user-management, which say:

> With LDAP authorization, user creation and management occurs on the LDAP server. MongoDB requires creation of roles on the admin database, with the name of each role exactly matching a LDAP group Distinguished Name (DN). This is in contrast to MongoDB managed authorization, which requires creating users on the $external database.

Meaning, for LDAP authentication, users aren't in fact stored in any MongoDB database, and our driver documentation in this regard isn't quite correct.

I replaced my text with your suggestion and I added a note to refer to server documentation for figuring out what auth sources are allowed, what do you think?

johnnyshields: I've made a refinement to your addition which includes a link to the server docs page. Please check it.

p-mongo: :+1: Thank you for that.

p-mongo (approved):

comandeo (approved):

p-mongo:
https://jira.mongodb.org/browse/MONGOID-5006

PR #5049, opened 2021-08-20
MONGOID-5005 - .sum, .count, and similar aggregables should ignore sort if not limiting/skipping

johnnyshields:
In the count/sum/min/max/avg pipeline, the `$sort` stage is unnecessary if `limit` and `skip` aren't present--even worse, having `$sort` can cause performance issues if the sort isn't indexed. (I learned that the hard way 😄 )

Fixes MONGOID-5005


johnnyshields:
@p-mongo would appreciate your review on this. This has caused some production bugs for us and I'd like to make sure other Mongoid users aren't affected.

johnnyshields:
Let's merge?

comandeo (approved):

p-mongo (approved):

PR #5052, opened 2021-08-21
MONGOID-???? Replace field[0] == ?$ with field.start_with?('$')

johnnyshields:
Easier to read and 30%+ faster.

```ruby
require 'benchmark'
Benchmark.bm do |b|
  b.report do
    10_000_000.times { '$foobar'[0] == ?$ }
  end
  b.report do
    10_000_000.times { '$foobar'.start_with?('$') }
  end
end

#       user     system      total        real
#   1.656000   0.000000   1.656000 (  1.661555)
#   1.094000   0.000000   1.094000 (  1.105380)
```

### DIFF
--- a/lib/mongoid/criteria/queryable/selectable.rb
+++ b/lib/mongoid/criteria/queryable/selectable.rb
@@ -597,7 +597,7 @@ module Mongoid
               end
               _mongoid_expand_keys(new_s).each do |k, v|
                 k = k.to_s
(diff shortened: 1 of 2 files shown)


p-mongo:
I am working on retrieving the tickets you made that are currently inaccessible so that we don't need to retype their contents. Other than that this change LGTM.

johnnyshields:
I may not have created a ticket for this specific issue... I forget.

PR #5053, opened 2021-08-21
MONGOID-???? - Aggregable::Mongo should only pipeline the operators each method needs.

johnnyshields:
To avoid conflicts, this PR includes the changes from https://github.com/mongodb/mongoid/pull/5036. Please review and merge that one first.

---

This PR improves `count`, `sum`, `avg`, `min`, and `max` performance.

Currently when any of the aforementioned methods are called, Mongoid creates an aggregation pipeline which calculates all 5 of them, even though a single value is actually returned.

In the vast majority of use cases, there is no point to calculating all 5 values in the aggregation pipeline. I believe result is can be cached if QueryCache is enabled, but in practice I have very few cases where I'm calling count then sum on a given field.

After the changes in this PR, users who wish to retrieve multiple values can still do so, with greater selectivity than before:

```
Band.aggregates(:likes) => { sum: 1082, count: 100, avg: 10.2, min: 0, max: 23 }

Band.aggregates(:likes, :sum, :count) => { sum: 1082, count: 100 }
```

## TODO

- Specs

### DIFF
--- a/lib/mongoid/contextual/aggregable/memory.rb
+++ b/lib/mongoid/contextual/aggregable/memory.rb
@@ -16,11 +16,9 @@ module Mongoid
         #   If no documents are present, then returned Hash will have
         #   count, sum of 0 and max, min, avg of nil.
         def aggregates(field)
-          { "count" => count(field),
-            "sum" => sum(field),
-            "avg" => avg(field),
-            "min" => min(field),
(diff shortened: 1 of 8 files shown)


johnnyshields:
After testing in production, this doesn't seem to make a significant performance difference. Closing.

PR #5061, opened 2021-08-29
MONGOID-5173 Specs should use bang (!) methods as much as possible

johnnyshields:
Bang methods .create!, .save!, and .update_attributes! ensure that the methods actually persisted to the DB and do not have silent failures. They will raise an error (which will make test red) if they fail.

Follow-up from conversation in https://github.com/mongodb/mongoid/pull/5029


spec/mongoid/persistable/creatable_spec.rb
```diff
@@ -4,14 +4,14 @@
 
 describe Mongoid::Persistable::Creatable do
 
-  describe ".create" do
+  describe ".create!" do
```
p-mongo, line 7: This removes all test coverage for `create` method.

p-mongo (commented):
Hi Johnny,

Did you read (i.e. review) this diff? Green CI doesn't guarantee the tests are *correct* in that they are testing, for example, the desired functionality.

spec/mongoid/persistable/creatable_spec.rb
johnnyshields: I'm still reading through the diffs I haven't done a final pass on this yet.

johnnyshields: OK, diff reviewed.

johnnyshields:
@p-mongo in some tests, only one of `#create` or `#create!` is tested. For these, I have intentionally changed to `#create!`, because firstly, `#create!` calls `#create` internally (i.e. greater code coverage) and second, it is more likely to cause a failure (i.e. an error) if the model didn't save as the test writer intended.

In the process of making this PR I did find several places where models were not being saved as intended. The most common was cases of `Account` missing `name` field.

johnnyshields:
@p-mongo this PR is going to continually generate conflicts as new PRs are merged so it would be ideal if we could merge it first.

johnnyshields:
@p-mongo @comandeo if the specs are passing can we please merge this?

johnnyshields:
@comandeo let's merge this?

spec/mongoid/persistable/savable_spec.rb
```diff
@@ -296,7 +296,7 @@
           # no conflicts.
           #expect(truck.atomic_updates[:conflicts]).to eq nil
 
-          expect { truck.save! }.not_to raise_error
+          expect(truck.save).to eq true
```
p-mongo, line 299: @johnnyshields I am having difficulty understanding the changes in this file, could you explain them please?

p-mongo (commented):
The rest of this PR seems to only affect `update_attributes`. It looks like we currently only have tests for `update_attributes` and not `update_attributes!`, so perhaps both should be minimally covered?

spec/mongoid/persistable/savable_spec.rb
johnnyshields:
```
expect { truck.save! }.not_to raise_error
```

is equivalent to:

```
expect(truck.save).to eq true
```

Both statements verify the document saved successfully. I prefer the latter in this case; we are specifically testing that the document saved.

johnnyshields:
@p-mongo is my comment here sufficient to merge? Tests are green...

p-mongo (approved):

comandeo (approved):

PR #5064, opened 2021-08-31
MONGOID-5177 Fix flaky contextual spec

comandeo:


spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -640,8 +640,25 @@
           end
 
           it "does not load all documents" do
-            expect(Mongo::Logger.logger).to receive(:debug?).exactly(2).times.and_call_original
+            subscriber = Class.new do
```
p-mongo, line 643: Can either this be moved into a full class in `support` or, if this is substantially similar to the driver's `EventSubscriber`, that class be moved to MRSS and then used from here?

p-mongo (approved):

PR #5066, opened 2021-09-03
Add a more descriptive error when spec load fails

johnnyshields:
Raise a friendlier load error when running specs so that outside contributors know what to do.


spec/spec_helper.rb
```diff
@@ -28,10 +28,17 @@ def database_id_alt
...
+begin
+  require 'mrss/cluster_config'
+  require 'support/client_registry'
+  require 'mrss/constraints'
+  require 'mrss/event_subscriber'
+rescue LoadError
+  raise LoadError.new <<~MSG.strip
+    The test suite requires shared tooling to be installed.
+    Please refer to spec/README.md for instructions.
+  MSG
```
p-mongo, line 40: Thank you for this change, I think it is helpful.

Can you please include the original exception message in the new exception message using a pattern like

```
rescue LoadError => exc

raise LoadError, "...: #{exc.class}: #{exc}"
```

johnnyshields: In this case, since this is purely for development users in the console, I don't think it's useful to include the original error, it just makes it harder to read.

(If the original error was intelligible, I wouldn't have to make this PR :)

comandeo (approved):

PR #5077, opened 2021-09-11
MONGOID-5183: adds queries instead of overriding it

sunny4381:


p-mongo (approved):

comandeo:
@sunny4381 Thank you very much for your contribution!

PR #5081, opened 2021-09-13
MONGOId-5185 Remove tests for _id serialization

comandeo:
I suggest removing these tests from Mongoid. There is no code that does something with `_id`. And we agreed that we rely on `bson-ruby` implementation of json serialisation.


p-mongo (approved):
I think these tests are indeed not required.

PR #5082, opened 2021-09-13
MONGOID-4592 Add examples of using read_attribute and write_attribute to implement custom field behavior

p-mongo:
https://jira.mongodb.com/browse/MONGOID-4592

### DIFF
--- a/docs/reference/fields.txt
+++ b/docs/reference/fields.txt
@@ -643,6 +643,70 @@ with systems that use the ``id`` field to store value different from ``_id``.
 Customizing Field Behavior
 ==========================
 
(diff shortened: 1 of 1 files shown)


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4592

comandeo (approved):

comandeo:
Regarding spec failure: this one is very flaky - https://github.com/mongodb/mongoid/blob/master/spec/mongoid/clients/options_spec.rb

I think we may want to either completely rewrite it or get rid of it, since counting connection on server side is not reliable.

PR #5083, opened 2021-09-27
MONGOID-5189 Proper order of *_save callbacks

comandeo:


lib/mongoid/interceptable.rb
```diff
@@ -37,6 +37,9 @@ module Interceptable
       define_model_callbacks :build, :find, :initialize, :touch, only: :after
       define_model_callbacks :create, :destroy, :save, :update, :upsert
 
+      # @api private
+      define_model_callbacks :save_relations
```
p-mongo, line 41: Can this be `save_associations` please and when the PR is ready for review, can this declaration also have a note added to it explaining its purpose (e.g. "This callback is used internally by Mongoid to save association targets for referenced associations after the parent model is saved", if I understood its purpose correctly).

comandeo: I believe there can be even a better name. Now it is confusing we set it up, it is now like this:
```ruby
klass.after_save_relations save_method, unless: :autosaved?
# or
klass.after_save_associations save_method, unless: :autosaved?
```
In reality the `save_method` will be executed not *after* associations are saved. It will be called after the parent is saved, and actually saves associations. That's why I started with `internal_save`.

p-mongo: Would `save_parent` be a better choice? Then
```
klass.after_save_parent save_method, unless: :autosaved?
```

p-mongo (approved):
LGTM if this is ready for review.

PR #5087, opened 2021-10-01
MONGOID-5191 Proper order for after_create callback

comandeo:


docs/release-notes/mongoid-7.5.txt
```diff
@@ -0,0 +1,46 @@
...
+* --> Child persisted in MongoDB <--
+* Child, :around_create_open
+* Child, :around_create_close
+* Child, :after_create
+* Child, :around_save_close
+* Child, :after_save
+* Parent, :around_create_close
+* Parent, :after_create
+* Parent, :around_save_close
+* Parent, :after_save
```
p-mongo, line 46: Can you also add the 7.4 order please like we are doing in the release notes elsewhere (i.e. after showing what happens in the new version, we also show what was happening in the previous version so tht users can easily see the change/compare behavior).

p-mongo (approved):

docs/release-notes/mongoid-7.5.txt
comandeo: Done. I used a table to make it easier to see differences

p-mongo (approved):
:+1:  Nice

PR #5090, opened 2021-10-06
MONGOID-5192 Rails compatible callbacks order for embedded relations

comandeo:


lib/mongoid/interceptable.rb
```diff
@@ -115,22 +115,38 @@ def run_before_callbacks(*kinds)
...
+        super(kind, &block)
       else
         true
       end
     end
 
+    # Run the callbacks for embedded documents.
+    #
+    # @param [ Symbol ] kind The type of callback to execute.
+    def run_children_callbacks(kind, &block)
```
p-mongo, line 137: I believe this method gets included in application models, therefore can it be 1) marked `@api private` and 2) be renamed to e.g. `_mongoid_run_children_callbacks` please?

p-mongo: I think `_mongoid_run_child_callbacks` would be more grammatically correct, since "child" is an adjective here I think.

spec/mongoid/interceptable_spec.rb
```diff
@@ -1835,12 +1841,63 @@ class TestClass
       end
     end
 
+    context "embeds_one" do
+      let(:registry) { InterceptableSpec::CallbackRegistry.new }
+
+      let(:parent) do
+        InterceptableSpec::CbEmbedsOneParent.new(registry).tap do |p|
+          p.child = InterceptableSpec::CbEmbedsOneChild.new(registry)
```
p-mongo, line 1849: If you don't mind, `s/p/parent/` please, to save us debugging grief later. I see that I probably started with `p` as the variable name here which was a poor choice on my part.

p-mongo (approved):

lib/mongoid/contextual/map_reduce.rb
```diff
@@ -147,7 +147,7 @@ def output
       def raw
         validate_out!
         cmd = command
-        opts = { read: cmd.delete(:read) } if cmd[:read]
+        opts = { read: criteria.options.fetch(:read) } if criteria.options[:read]
```
p-mongo, line 150: Thank you for this, I gather this repairs failures that I've seen in another PR.

lib/mongoid/interceptable.rb
```diff
@@ -115,22 +115,43 @@ def run_before_callbacks(*kinds)
...
+    # @api private
+    def _mongoid_run_child_callbacks(kind, children: nil, &block)
+      child, *tail = (children || cascadable_children(kind))
+      if child.nil?
+        return block&.call
+      elsif tail.empty?
+        return child.run_callbacks(child_callback_type(kind, child), &block)
+      else
+        return child.run_callbacks(child_callback_type(kind, child)) do
+          _mongoid_run_child_callbacks(kind, children: tail, &block)
```
p-mongo, line 150: :+1: Cool!

PR #5092, opened 2021-10-13
MONGOID-5104 Add AR-style dirty attributes

comandeo:
Dirty data are now cleared and `changes` moved `previous_changes` immediately after a document is persisted in MongoDB. It means that old attribute values are available in `before_` callbacks, and updated attribute values are available in `after_` callbacks.


p-mongo (approved):
I actually don't understand the changes here, can you please add a one sentence summary to the description and the resulting merge commit message should anyone happen to be debugging something in the future and be looking at this commit as a possible cause?

johnnyshields:
@p-mongo @comandeo please see this follow-up ticket: https://jira.mongodb.org/browse/MONGOID-5217

PR #5104, opened 2021-12-17
Fix MONGOID-5198 Calling children_changed? on a deep cyclical data structure will cause semi-infinite looping


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5198

comandeo (approved):

PR #5108, opened 2021-12-18
MONGOID-5098 Bug fix: Range.mongoize should mongoize its members

johnnyshields:
Fixes MONGOID-5098 (along with #5023 and #5025 which were already merged.)

MongoDB supports `Range` as a storable field type. Currently, when storing a `Range` of `Time` objects, the `Range#mongoize` method does not call `Time#mongoize` on the inner objects.

In addition, this PR improves `Range.mongoize` handling of `Hash` as an input, by removing keys which are unrelated to `Range` and would be junk data in DB. This case is *highly unlikely* to be used in any real-world app, but the existing Mongoid code supports it so we should make it work properly.

(* This PR does NOT add support for endless/beginless ranges; that will be addressed in a separate PR.)

### DIFF
--- a/lib/mongoid/extensions/range.rb
+++ b/lib/mongoid/extensions/range.rb
@@ -59,12 +59,30 @@ module Mongoid
         #
         # @return [ Hash ] The object mongoized.
         def mongoize(object)
-          return nil if object.nil?
-          return object.mongoize if object.is_a?(::Hash)
-          return object.mongoize if object.is_a?(String)
-          hash = { "min" => object.first&.mongoize, "max" => object.last&.mongoize }
+          case object
+          when NilClass then nil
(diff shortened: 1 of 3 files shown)


johnnyshields:
@p-mongo this is ready for review/merge.

johnnyshields:
Let's run Evergreen and merge?

johnnyshields:
@p-mongo may we get someone to review this? This is a pretty straightforward PR.

p-mongo (approved):

johnnyshields:
Evergreen seems stuck again?

comandeo (approved):

comandeo:
Evergreen tests are all green.

PR #5109, opened 2021-12-18
MONGOID-5173 Specs should use bang (!) methods (without describe/context change)

johnnyshields:
Fixes MONGOID-5173

This is similar to PR https://github.com/mongodb/mongoid/pull/5061 except this PR does NOT change any bang methods in describe/context blocks.

(As an aside: I still do think the context blocks should also be changed--3 files total--but we can do that in a separate PR.)

### DIFF
--- a/spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
+++ b/spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
@@ -3653,11 +3653,15 @@ describe Mongoid::Association::Referenced::HasAndBelongsToMany::Proxy do
   context "when using a different primary key" do
 
     let(:dog) do
-      Dog.create(name: 'Doggie')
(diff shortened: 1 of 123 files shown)


johnnyshields:
@p-mongo this is ready for review/merge.

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
```diff
@@ -1992,15 +1992,15 @@
         context "when using string keys" do
 
           let(:agent) do
-            Agent.create(number: "007")
+            Agent.create!(number: "007")
           end
 
           before do
-            agent.accounts.create(name: "testing again")
+            agent.accounts.send(method, name: "test again")
```
p-mongo, line 1999: I think this is correct, the code previously had a bug :+1:

p-mongo (approved):

PR #5111, opened 2021-12-28
GH Actions tests for Rails 7

johnnyshields:
This PR adds GH Actions tests supports for Rails 7.

Currently, this PR is blocked for Ruby 3.1 + Rails 7.0.0 support due to a last-minute change (hence I've disallowed 7.0.0 as a version.) This will be fixed in Rails 7.0.1. See https://github.com/rails/rails/pull/43951.

### DIFF
--- a/mongoid.gemspec
+++ b/mongoid.gemspec
@@ -33,17 +33,15 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = ">= 2.5"
   s.required_rubygems_version = ">= 1.3.6"
 
-  if RUBY_VERSION.start_with?('2.')
-    s.add_dependency("activemodel", [">=5.1", "<6.2"])
-  elsif RUBY_VERSION.start_with?('3.')
-    s.add_dependency("activemodel", [">=6.0", "<6.2"])
-  end
+  # activemodel 7.0.0 cannot be used due to Class#descendants issue (see: https://github.com/rails/rails/pull/43951)
+  s.add_dependency("activemodel", ['>=5.1', '<7.1', '!= 7.0.0'])
   s.add_dependency("mongo", ['>=2.10.5', '<3.0.0'])
-  # Using this gem is recommended for handling argument delegation issues,
+
+  # The ruby2_keywords gem is recommended for handling argument delegation issues,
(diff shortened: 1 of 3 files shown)


p-mongo:

As I mentioned in the bson-ruby PR, we will need to do a few things in order for ruby 3.1 to be testable in evergreen. If you extract the changes to yaml safe load etc. that don't deal with ruby or rails version into their own PR I think those can be merged right away.

johnnyshields:
YAML.safe_load changes here: https://github.com/mongodb/mongoid/pull/5112

mongoid.gemspec
```diff
@@ -33,17 +33,15 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = ">= 2.5"
   s.required_rubygems_version = ">= 1.3.6"
 
-  if RUBY_VERSION.start_with?('2.')
-    s.add_dependency("activemodel", [">=5.1", "<6.2"])
-  elsif RUBY_VERSION.start_with?('3.')
-    s.add_dependency("activemodel", [">=6.0", "<6.2"])
-  end
+  # activemodel 7.0.0 cannot be used due to Class#descendants issue (see: https://github.com/rails/rails/pull/43951)
+  s.add_dependency("activemodel", ['>=5.1', '<7.1', '!= 7.0.0'])
```
p-mongo, line 37: I think we need a comment here saying that on Ruby 3 activemodel >= 6.0 is required to retain that note but the previous conditional that is being removed in this PR probably isn't valid in the gemspec, @comandeo do you agree?

johnnyshields: Rails 5.x does not support Ruby 3. This is not something specific to Mongoid, so I don't see why the Mongoid gem would need to add any comment or conditional logic here.

AlexKovynev:
it seems it can be merged now cause 7.0.1 is out https://twitter.com/rails/status/1479214221506322433

johnnyshields:
Specs are green! Let's merge @p-mongo

p-mongo:
I folded this PR into https://github.com/mongodb/mongoid/pull/5122.

PR #5112, opened 2021-12-28
MONGOID-5207 Use YAML.safe_load in specs

johnnyshields:
Change YAML.load to YAML.safe_load as it is the default in Ruby 3.1+. This change only affects specs.


johnnyshields:
@p-mongo ready to merge

comandeo (dismissed):

simi (approved):

comandeo (approved):

p-mongo (approved):
Thank you for repairing these issues, these changes should ideally be in the shared tooling (MRSS) but that can be done later.

PR #5115, opened 2021-12-29
MONGOID-5206 fix bug where embedded document is not re-embedded

Neilshweky:


lib/mongoid/association/embedded/embeds_one/proxy.rb
```diff
@@ -53,6 +53,7 @@ def substitute(replacement)
                 # The associated object will be replaced by the below update if non-nil, so only
                 # run the callbacks and state-changing code by passing persist: false in that case.
                 _target.destroy(persist: !replacement) if persistable?
+                _target.new_record = true
```
Neilshweky, line 56: maybe this should be inside the destroy?

Neilshweky: A little explanation on why this fixes the bug... We have assignments:
```
canvas.palette = palette
canvas.palette = nil 
canvas.palette = palette
```
Where canvas embeds_one palette.

Previously, what was happening was, on the first assignment, palette was considered a "new record" (`new_record?=true`) and thus palette was being inserted into the database. However, on the third assignment, where trying to reassign the palette, palette is no longer considered a new record, because it had been inserted previously. This is not exactly accurate, because the second assignment ultimately removed the palette from the database, so it needs to be reinserted. Since the palette's new_record is false, Mongoid ends up "updating" the document, which doesn't reinsert it into the database. 

The change I introduce here, respecifies palette as a "new record" when it gets removed from the database, so if it is reassigned, it will be reinserted into the database.

p-mongo: This is a great explanation, can you please copy it into a comment in the source so that we can refer to it in the future when someone looks at this code?

Neilshweky: sure!

p-mongo (approved):
:+1:  Thank you!

PR #5116, opened 2021-12-29
MONGOID-5208 fix error on reloading nil embedded document

Neilshweky:


lib/mongoid/reloadable.rb
```diff
@@ -82,7 +82,7 @@ def reload_embedded_document
     def extract_embedded_attributes(attributes)
       atomic_position.split(".").inject(attributes) do |attrs, part|
         attrs = attrs[part =~ /\d/ ? part.to_i : part]
-        attrs
+        attrs || {}
```
p-mongo, line 85: I don't think this is correct. If there are not attributes at the specified path (`atomic_position`), this method should probably return nil and the docstring should be updated to reflect that.

lib/mongoid/reloadable.rb
```diff
@@ -22,7 +22,7 @@ def reload
       end
 
       reloaded = _reload
-      if Mongoid.raise_not_found_error && (!reloaded || reloaded.empty?)
+      if Mongoid.raise_not_found_error && reloaded.empty?
```
p-mongo, line 25: The docstring on `_reload` should also then be updated to say that if the embedded document no longer exists in the dabatase, `_reload` would return nil.

Subsequently, here I think the correct fix is to check `reloaded.nil?` either instead of or in addition to `reloaded.empty?`.

I am also curious what it means for `reloaded` to be empty here - the document exists but has no attributes? How does a user attain such a beast?

lib/mongoid/reloadable.rb
Neilshweky: okay that was my original fix, I'll go back to that

lib/mongoid/reloadable.rb
Neilshweky: Yeah I'm not exactly sure how this case is achieved, but in general I balk at removing safety checks.

p-mongo:
This PR also doesn't have tests or am I missing how it's tested? Is the reproduce code in the ticket already in the test suite elsewhere?

Neilshweky:
> This PR also doesn't have tests or am I missing how it's tested? Is the reproduce code in the ticket already in the test suite elsewhere?

Is the test I added in `reloadable_spec.rb` not showing up for you? You might have an outdated version of the PR because one of the comments you made was on code that no longer exists in the most recent commit.

p-mongo:
This PR needs to be rebased on master.

lib/mongoid/reloadable.rb
p-mongo: I checked this situation I am not a fan of how that is implemented either, let's leave it alone for now since it's not in scope of your work.

Neilshweky:
Yup! I just did that

spec/mongoid/reloadable_spec.rb
```diff
@@ -311,6 +311,31 @@
...
+      end
+
+      let(:reload) do
+        palette.reload
+      end
+
+      it "doesn't raise an error" do
+        expect do
+          reload
+        end.to raise_error(Mongoid::Errors::DocumentNotFound, /Document\(s\) not found for class Palette with id\(s\)/)
```
p-mongo, line 335: The example description is the opposite of its behavior.

p-mongo (approved):
You can merge after fixing up the description and waiting for CI.

spec/mongoid/reloadable_spec.rb
Neilshweky: yes I just fixed that

PR #5119, opened 2022-01-01
Prefer field :type as Symbol + custom field type symbols

johnnyshields:
**Prior to this PR**, Mongoid already supports defining field `:type` as a `Symbol` rather than a `Class`:

```ruby
  field :first_name, type: :string  # works today, even without this PR!
```

Generally speaking, it is much better to to use Symbol than Class, because Class uses autoloading etc.

This PR adds the following:
- Main documentation is updated to **prefer** using Symbol rather than Class. I haven't done this change in 100% of doc files; I plan to do that after this PR is merged in a separate PR.
- Usage of Class is **deprecated** and now logs a warning. Usage of Class is intended to be removed in Mongoid 8.0.
- It is now possible to define custom symbols using the following:

```ruby
  Mongoid::Fields.configure do
    type :point, Point
  end
```

### DIFF
--- a/docs/reference/fields.txt
+++ b/docs/reference/fields.txt
@@ -13,15 +13,21 @@ Field Definition
    :class: singlecol
 
 
+.. _field-types:
+
 Field Types
 ===========
 
-Even though MongoDB is a schemaless database and allows data to be stored
-as strings, Mongoid permits the application to declare the type of data
-stored in the various fields of a document. Field type declarations affect
-the following:
+MongoDB stores underlying document data using
+`BSON types <https://docs.mongodb.com/manual/reference/bson-types/>`_, and
+Mongoid converts BSON types to Ruby types at runtime in your application.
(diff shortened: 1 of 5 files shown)


johnnyshields:
@p-mongo tests are now written. This is ready for final review and merge.

p-mongo (changes requested):
Hi @johnnyshields ,

The portion of this PR that we can merge is the one described in https://jira.mongodb.org/browse/MONGOID-5336 which is permitting applications to define their own aliases. Can you please remove from the diff the changes that have to do with replacing class use with symbol use?

johnnyshields:
@p-mongo I will be happy to split out that code soon. However, please see my reply in [MONGOID-5336](https://jira.mongodb.org/browse/MONGOID-5336), in the future, I still believe there is strong justification to switch to symbols.

johnnyshields:
@p-mongo PR split out, please see: https://github.com/mongodb/mongoid/pull/5269

PR #5123, opened 2022-01-05
MONGOID-5207 Test Ruby 3.1 on Evergreen

comandeo:


.evergreen/config.yml
```diff
@@ -531,15 +540,15 @@ axes:
           APP_TESTS: yes
 
 buildvariants:
-- matrix_name: "ruby-3.0"
+- matrix_name: "ruby-3"
   matrix_spec:
-    ruby: ["ruby-3.0"]
+    ruby: ["ruby-3.0", "ruby-3.1"]
     driver: ["current"]
     topology: '*'
```
p-mongo, line 547: Maybe just test one topology on 3.0?

p-mongo (approved):

.evergreen/config.yml
comandeo: Done, good idea.

PR #5125, opened 2022-01-06
MONGOID-5212 Support for Decimal128 type

Neilshweky:


lib/mongoid/errors/unmapped_bson_type.rb
```diff
@@ -0,0 +1,22 @@
+# encoding: utf-8
+module Mongoid
+  module Errors
+
+    # This error is raised when attempting to demongoize a value from the database
+    # that doesn't map to the field type.
+    class UnmappedBSONType < MongoidError
```
Neilshweky, line 7: Are we keeping this error around? See Emily's comment [here](https://github.com/mongodb/mongoid/pull/4363#discussion_r93001937)

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -58,7 +68,31 @@ def demongoize(object)
...
+          end
+        end
+
+        private
+
+        def demongoize_from_decimal128(object)
+          if Mongoid.map_big_decimal_to_decimal128
+            object.to_big_decimal
+          else
+            raise Mongoid::Errors::UnmappedBSONType.new(object)
```
Neilshweky, line 86: I don't think I agree with this. When you opt in to using the BigDecimal type, you expect to lose some precision. If you really cared about trailing zeroes then you would have used Decimal128.

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -58,7 +68,31 @@ def demongoize(object)
...
+          end
+        end
+
+        def mongoize_to_decimal128(object)
+          if object.is_a?(BigDecimal)
+            BSON::Decimal128.new(object)
+          elsif object.numeric?
+            BSON::Decimal128.new(object.to_s)
+          end
```
Neilshweky, line 95: I think we need an else condition here?

lib/mongoid/errors/unmapped_bson_type.rb
p-mongo: From what I am understanding, I don't think so. If user specifies BigDecimal as field type + Decimal128 serialization (which would have to be explicitly enabled at least in the first Mongoid release incorporating the integration) they are accepting the common functionality of the two types, including the fact that trailing zeroes will be lost in a roundtrip.

lib/mongoid/extensions/big_decimal.rb
Neilshweky: From the tests, it actually seems like returning `nil` is intended here. Does that make sense?

p-mongo: What is the purpose of this method? If it is only called from mongoize above why not put all logic there?

lib/mongoid/extensions/big_decimal.rb
p-mongo: If a user specified BigDecimal field type and database data is a Decimal128, I think they should get back a BigDecimal. This combination won't be currently produced by Mongoid from what I understand.

lib/mongoid/extensions/big_decimal.rb
Neilshweky: yup it seems like they abstracted some of the logic into its own function, even though it seems like it's only called once

lib/mongoid/extensions/big_decimal.rb
Neilshweky: done

lib/mongoid/extensions/big_decimal.rb
Neilshweky: I removed those functions and put the logic back into the bigger functions

lib/mongoid/errors/unmapped_bson_type.rb
Neilshweky: done

lib/mongoid/config.rb
```diff
@@ -73,6 +73,7 @@ module Config
 
     # Return stored times as UTC.
     option :use_utc, default: false
+    option :map_big_decimal_to_decimal128, default: false
```
p-mongo, line 76: This line needs an empty line above it to separate this option from the previous option.

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -46,7 +50,13 @@ module ClassMethods
         #
         # @return [ BigDecimal, nil ] A BigDecimal derived from the object or nil.
         def demongoize(object)
-          object && object.numeric? ? BigDecimal(object.to_s) : nil
+          if object
+            if object.is_a?(BSON::Decimal128)
+              object.to_big_decimal
+            elsif object.numeric?
```
p-mongo, line 56: I would like the `numeric?` condition to be dropped and this path be taken for non-numeric input also. I see that non-numeric input was already trashed previously but I think this is not good behavior.

Test with a time field:

```
  class Foo
    include Mongoid::Document

    field :a
  end

  Foo.delete_all

  Foo.create!(a: 'zz')

  class Foo
    field :a, type: BigDecimal
  end

  p Foo.first, Foo.first.a
```

produces:

```
NoMethodError: undefined method `getlocal' for "zz":String
```

Test with the code as proposed with BigDecimal field:

```
  class Foo
    include Mongoid::Document

    field :a
  end

  Foo.delete_all

  Foo.create!(a: 'zz')

  class Foo
    field :a, type: Time
  end

  p Foo.first, Foo.first.a
```

produces:

```
#<Foo _id: 61d76a95a15d5d4e9c038186, a: "zz">
nil
```

I think it would be better to always attempt to construct a BigDecimal, allowing *that* call to raise whatever exceptions on problems.

@comandeo  thoughts?

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -46,7 +50,13 @@ module ClassMethods
         #
         # @return [ BigDecimal, nil ] A BigDecimal derived from the object or nil.
         def demongoize(object)
-          object && object.numeric? ? BigDecimal(object.to_s) : nil
+          if object
```
p-mongo, line 53: This condition also causes this method to map `false` to `nil`, I suggest `unless object.nil?` instead.

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -58,7 +68,17 @@ def demongoize(object)
         #
         # @return [ String, nil ] A String representing the object or nil.
         def mongoize(object)
-          object && object.numeric? ? object.to_s : nil
+          if object
```
p-mongo, line 71: I suggest `unless object.nil?` as per above comment.

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -58,7 +68,17 @@ def demongoize(object)
...
+            if Mongoid.map_big_decimal_to_decimal128
+              if object.is_a?(BigDecimal)
+                BSON::Decimal128.new(object)
+              elsif object.numeric?
```
p-mongo, line 75: If this method (attempts) to handle a variety of input types, it seems to me that the second condition of this if statement (ahead of `object.numeric?` branch) should be `elsif object.is_a?(BSON::Decimal128)` which would return `object` verbatim.

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -58,7 +68,17 @@ def demongoize(object)
...
+                BSON::Decimal128.new(object.to_s)
+              end
```
p-mongo, line 77: ... and I would like to have an `else` here returning `object`, same as with the long comment above.

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -58,7 +68,17 @@ def demongoize(object)
...
+            elsif object.numeric?
+              object.to_s
+            end
```
p-mongo, line 80: The outer `if` should also check for BSON::Decimal128 and have the `else` branch, same as the inner `if`.

spec/mongoid/criteria/queryable/extensions/big_decimal_spec.rb
```diff
@@ -6,55 +6,121 @@
 
   describe ".evolve" do
 
-    context "when provided a big decimal" do
+    context 'when map_big_decimal_to_decimal128 is false' do
```
p-mongo, line 9: This test needs to either set the option to false or assert that its value is false, to guard against the option value being different due to e.g. a previously executed test.

spec/mongoid/extensions/big_decimal_spec.rb
```diff
@@ -8,401 +8,853 @@
...
+  end
+
+  context 'when map_big_decimal_to_decimal128 is true' do
+
+    before do
+      Mongoid.map_big_decimal_to_decimal128 = true
+    end
+
+    after do
+      Mongoid.map_big_decimal_to_decimal128 = false
```
p-mongo, line 440: I would prefer an `around` block instead which would read the current setting value and restore the setting to the current value, instead of hardcoding our present default like this.

spec/mongoid/fields_spec.rb
```diff
@@ -1234,71 +1234,153 @@ class DiscriminatorChild2 < DiscriminatorParent
...
+      end
+
+      context "when setting to a string" do
+
+        before do
+          band.sales = decimal.to_s
+        end
+
+        it "persists as a BSON::Decimal128" do
+          expect(band.attributes["sales"]).to eq(BSON::Decimal128.new(decimal.to_s))
```
p-mongo, line 1348: :+1:

lib/mongoid/extensions/big_decimal.rb
johnnyshields: False can be ignored as it isn't handled by below statements (it is neither a Decimal128 nor numeric?). Recommend to as a guard method `return unless object`

lib/mongoid/extensions/big_decimal.rb
johnnyshields: I disagree. I think demongoize should return nil unless the object can be represented as a BigDecimal. Should check the behavior of demongoize on other classes.

lib/mongoid/extensions/big_decimal.rb
johnnyshields: At a minimum the `object` must be mongoized, so return `object.mongoize`

lib/mongoid/extensions/big_decimal.rb
Neilshweky: @p-mongo I kind of agree with @johnnyshields here... Here are some examples of invalid values to demongoize:
```
irb(main):004:0> Float.demongoize("1.2asda")
=> 0.0
irb(main):005:0> Integer.demongoize("1234rqwdd")
=> 0
```
Now, I'm not sure why `0` is the right return value here but the important thing is that it doesn't raise an error. Let me know what you think...

johnnyshields: @Neilshweky is correct should make this return `0`. Important thing is that it doesn't raise error.

lib/mongoid/extensions/big_decimal.rb
Neilshweky: I changed this @p-mongo but I don't think it will make a difference since false is not numeric. I'll leave it because it's more expressive but there's no functionality change. 
```
irb(main):001:0> BigDecimal.demongoize(false)
=> nil
irb(main):002:0> false.numeric?
=> false
```

lib/mongoid/extensions/big_decimal.rb
Neilshweky: Won't that cause an infinite loop @johnnyshields ?

johnnyshields: No.

lib/mongoid/config.rb
Neilshweky: done

lib/mongoid/extensions/big_decimal.rb
Neilshweky: Actually after rereading your comment, I agree @p-mongo. Let's let BigDecimal throw an error

lib/mongoid/extensions/big_decimal.rb
Neilshweky: done

lib/mongoid/extensions/big_decimal.rb
Neilshweky: done

lib/mongoid/extensions/big_decimal.rb
johnnyshields: Wait--`demongoize` should NOT throw an error. Can you find examples of other classes where demongoize throws an error?

Imagine the user experience for this this... DB object can't be read... nightmare.

spec/mongoid/criteria/queryable/extensions/big_decimal_spec.rb
Neilshweky: done

spec/mongoid/extensions/big_decimal_spec.rb
Neilshweky: done

lib/mongoid/extensions/big_decimal.rb
Neilshweky: @p-mongo @comandeo what do you think?

lib/mongoid/extensions/big_decimal.rb
Neilshweky: Ah I see. For my curiosity, can you point out an example where this would cause a change in functionality?

spec/mongoid/criteria_spec.rb
```diff
@@ -3422,20 +3422,111 @@ def self.ages; self; end
...
+          let(:fans) do
+            BigDecimal('139432.0002')
+          end
+
+          let!(:band) do
+            Band.create!(name: "Boards of Canada", fans: fans)
+          end
+
+          let(:from_db) do
+            Band.where(fans: fans.to_s).first
```
Neilshweky, line 3498: This is an interesting case. A `BigDecimal` was inserted with `map_big_decimal_to_decimal128=false` so it was stored as a string. Now, if we try to query it using a `BigDecimal`, it won't find it. Note this field is _dynamic_ so it can't be expected to know that strings are also `BigDecimal`s. In the case where we set the type of the field to `BigDecimal`, it does find the document, in this case I don't think it should...

What do you think @p-mongo?

p-mongo: This caveat can go into the BigDecimal field documentation. If users query on the field, they are would need to either query with both value types (BigDecimal and string, using e.g. `$in`) or they would need to migrate their data to Decimal128 storage.

Neilshweky: done

Neilshweky:
The tests in this PR will not work until BSON is upgraded to 4.14.0

lib/mongoid/extensions/big_decimal.rb
comandeo: I am inclined to agree with @johnnyshields here, we should not fail if we cannot demongoize value.

Imagine we have the following document:
```ruby
class Item
  include Mongoid::Document
  
  field :name, type: String
  field :kind, type: String
  field :amount, type: BigDecimal
end
```

Let's say our database contains following records:
```
[
  {
    _id: ObjectId("61d8889af7b3f3f5d961f974"),
    name: 'name1',
    kind: 'x',
    amount: Decimal128("1")
  },
  {
    _id: ObjectId("61d8889af7b3f3f5d961f975"),
    name: 'name2',
    kind: 'x',
    amount: 'error: bad input, cannot parse provided data'
  },
  {
    _id: ObjectId("61d8889af7b3f3f5d961f976"),
    name: 'name3',
    kind: 'y',
    amount: Decimal128("2")
  }
]
```

This is a legitimate situation. It can happen, for example, if there is another application that writes to the same collection, and something went wrong. And even if something went wrong, we still may want to work with the data.

Now we try to query the data:
```ruby
# A dumb example to illustrate the problem.
Item.where(kind: 'x').map do |i|
  if i.amount.nil?
    "<p>BAD RECORD</p>" # Or whatever I want to to do handle the problem.
  else
    css_class = if i.amount > 0
                  'green'
                else
                  'red'
                end
    "<p class='#{css_class}'>#{i.name}: #{i.amount}</p>"
  end  
end
```
Proposed solution will throw an exception in the middle of execution. Returning `nil` will give a user a chance to handle broken data.

As a user, I would probably expect that the query returns me requested data even if some fields cannot be deserialized. Setting the problematic field to nil seems to be a good indication of a problem to me, and the problem can be handled on application layer.

I do not think that we should return 0 here, though. Zero is a valid value.

Neilshweky: okay, I think that's a fair solution.

lib/mongoid/extensions/big_decimal.rb
johnnyshields: Probably none in practice, but the docs state that SomeObject.mongoize(...) class method should be a catch-all for all possible types, even ones that do not cleanly map to SomeObject.

lib/mongoid/extensions/big_decimal.rb
johnnyshields: I'm fine with either `nil` or `0`, but again would urge to look at what other classes do an and **keep it consistent**. (If there needs to be a follow-up PR to change BigDecimal + other classes uniformly, that's also fine...)

comandeo: I am afraid we are not consistent already:
```
irb(main):015:0> Date.demongoize('i am so bad')
date.rb:48:in `demongoize': undefined method `year' for "i am so bad":String (NoMethodError)

irb(main):018:0> Time.demongoize('so do i')
time.rb:57:in `demongoize': undefined method `getlocal' for "so do i":String (NoMethodError)

irb(main):020:0> DateTime.demongoize('i am also bad')
time.rb:57:in `demongoize': undefined method `getlocal' for "i am also bad":String (NoMethodError)

irb(main):021:0> Float.demongoize('i am sooooooo bad')
=> 0.0

irb(main):022:0> Integer.demongoize('could not be worse')
=> 0

irb(main):023:0> Range.demongoize('and yet it can')
=> nil..

irb(main):024:0> Set.demongoize('i have no idea what i am doing')
ruby/3.1.0/set.rb:280:in `do_with_enum': value must be enumerable (ArgumentError)

irb(main):026:0> Regexp.demongoize(42).class.name
=> "Integer"

irb(main):028:0> Hash.demongoize('i give up')
=> "i give up"
```
We do have to make all these consistent later, that is true.

johnnyshields: Since the two existing numeric types (`Float` and `Integer`) currently return `0`, I would be inclined to say `BigDecimal` should return `0`. Fine to raise a follow-up ticket/PR if we wish to change it later.

Neilshweky: I think for now it's better practice to return nil, and we should file a ticket to follow upon Float and Integer.

lib/mongoid/extensions/big_decimal.rb
Neilshweky: Okay, I'm willing to change that. Done.

spec/mongoid/criteria_spec.rb
```diff
@@ -3422,20 +3422,139 @@ def self.ages; self; end
...
+          end
+
+          it "finds the document by the big decimal value" do
+            expect(from_db).to eq(band)
+          end
         end
 
-        let(:from_db) do
-          Band.where(sales: sales).first
+        context 'when map_big_decimal_to_decimal128 was false and is now true' do
```
Neilshweky, line 3477: This test does not work. It asserts that a BigDecimal that was originally stored as a string, can be queried with a BigDecimal, with `map_big_decimal_to_decimal128` turned on. We have to fix this. Here are the solutions I can think of:
1. Change the implementation of searching for a `BigDecimal` field to search for both the string representation and the BigDecimal representation.
A problem arises when some people purposely have `BigDecimal`s and Strings stored in the database under the same field, and don't want querying for a `BigDecimal` to return the string. But I think this is a pretty niche case.
2. We can just leave it how it is. Instruct the user how to query for both string and `BigDecimal` representations of their data if that's what they want. This is a little annoying for the user.
3. I also wrote up this quick little query that changes all values in a db to a ``BigDecimal`` but this seems a little reckless, especially for big data sets:
```
db.bands.updateMany({ 
  "field": { "$exists": true } 
}, [
  { "$set": { 
    "field": { "$toDecimal": "$field" } 
  } }
])
```

p-mongo: The first idea would be a really good one but I expect it to be either difficult or perhaps not even possible.

There are two contexts where attribute values are used: implicit equality (e.g. `{field: 42}`) and operators (e.g. `{field: {$eq: 42}}`). While the implicit equality can be transformed to e.g. `{field: {$in: ['42', Decimal128('42')]}}`, this won't work in the operator context.

`mongoize` currently doesn't have context knowledge at all. There is probably a way to communicate some contexts to it but Mongoid does generally permit users to provide MQL in a bunch of places and I'm thinking those might be impossible to change programmatically short of doing much more rewriting of MQL than Mongoid is presently doing.

p-mongo: I think at this point what we need is migration guidance and documentation for how to move from string to bigdecimal field storage. The migration query above is excellent and can go into the migration guidance. Additionally it would be good to have a similar snippet for how to query on individual fields with type conversion into bigdecimal, e.g. if I want to find all documents where the value is > 0 with both types how I would do that.

Neilshweky: Okay! I can't put that all into the documentation PR.... That leaves us with no further code changes then.

lib/mongoid/extensions/big_decimal.rb
p-mongo: I created https://jira.mongodb.org/browse/MONGOID-5221 to address the demongoization issue across all types.

lib/mongoid/extensions/big_decimal.rb
Neilshweky: It's true... I guess that's why this PR was cancelled all these years ago...

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -46,7 +50,13 @@ module ClassMethods
...
+          unless object.nil?
+            if object.is_a?(BSON::Decimal128)
+              object.to_big_decimal
+            elsif object.numeric?
+              BigDecimal(object.to_s)
+            end
+          end
         end
 
         # Mongoize an object of any type to how it's stored in the db as a String.
```
p-mongo, line 62: I think this needs to be updated, specifically the `as a String` part?

Neilshweky: done

spec/mongoid/criteria_spec.rb
p-mongo: This test then can be changed to assert that the document is *not* found when its type in the database is different from how Mongoid is presently configured, for our future reference.

p-mongo (commented):
LGTM, I'll leave this unapproved as a reminder that this PR is blocked on bson release.

p-mongo (approved):

johnnyshields:
Looks good to me

comandeo (approved):

lib/mongoid/extensions/big_decimal.rb
p-mongo: I am unresolving this discussion thread because it's referenced from https://jira.mongodb.org/browse/MONGOID-5221.

johnnyshields: I've crossed out my earlier comment here--BigDecimal correctly returns `nil` today for non-numeric string, we should do the same for Integer/Float. It's ready to merge in PR: https://github.com/mongodb/mongoid/pull/5220

PR #5126, opened 2022-01-06
MONGOID-5213 Document changes to BigDecimal type and addition of global flag

Neilshweky:


docs/reference/fields.txt
```diff
@@ -99,7 +99,6 @@ You can safely omit type specifications when:
 
 Types that are not supported as dynamic attributes since they cannot be cast are:
 
-- ``BigDecimal``
```
p-mongo, line 102: Is there a test where a BigDecimal value is written to & read from a dynamic field? I'm expecting no based on this not being possible previously.

p-mongo: Could be done in a separate PR.

p-mongo (approved):

docs/reference/fields.txt
Neilshweky: Not that I can find, adding one in the MONGOID-4237 PR

Neilshweky: done

docs/reference/fields.txt
```diff
@@ -397,6 +396,56 @@ matches strings containing "hello" before a newline, besides strings ending in
...
+  may result in a loss of precision.
+
+There is an additional caveat when storing a ``BigDecimal`` in a field with no
+type (i.e. a dynamically typed field) and ``Mongoid.map_big_decimal_to_decimal128``
+is false. In this case, the ``BigDecimal`` is stored as a string, and since a
+dynamic field is being used, querying for that field with a ``BigDecimal`` will
+not find the string for that ``BigDecimal``, since the query is looking for a
+``BigDecimal``. In order to query for that string, the ``BigDecimal`` must
+first be converted to a string with ``to_s``. Note that this is not a problem
+when the field has type ``BigDecimal``.
```
p-mongo, line 440: Would `BigDecimal`-typed field have the same issue in the other direction - when querying existing data that is stored as strings after `Mongoid.map_big_decimal_to_decimal128` is set to true, nothing would be found?

Neilshweky: Yes that is true as well... How would you like to handle this case?

docs/reference/fields.txt
```diff
@@ -397,6 +396,56 @@ matches strings containing "hello" before a newline, besides strings ending in
 This is because the meaning of ``$`` is different between PCRE and Ruby
 regular expressions.
 
+BigDecimal Fields
+-----------------
+
+The ``BigDecimal`` field type is used to store numbers with increased precision.
+
+The ``BigDecimal`` field type stores its values in two different ways under the
+hood, depending on the value of the ``Mongoid.map_big_decimal_to_decimal128``
```
johnnyshields, line 405: `under the hood` --> `in the database`

Neilshweky: I kind of like it the way it is tbh...

johnnyshields: "Under the hood":
- Does not clearly indicate under which "hood" -- is it in Ruby's memory or the database?
- Is idiomatic English which does not make sense to non-native speakers (Mongoid is not a car :) )

It's good to think about these points when writing technical documentation.

docs/reference/fields.txt
```diff
@@ -397,6 +396,56 @@ matches strings containing "hello" before a newline, besides strings ending in
...
+first be converted to a string with ``to_s``. Note that this is not a problem
+when the field has type ``BigDecimal``.
+
+If you wish to avoid using ``BigDecimal`` altogether, you can set the field
+type to ``BSON::Decimal128``. This will allow you to keep track of trailing
+zeroes and signed ``NaN`` values.
+
+.. note::
+  In Mongoid 7.6, the global config ``Mongoid.map_big_decimal_to_decimal128``
+  will be defaulted to true instead of false.
```
johnnyshields, line 448: Following Semver, this defaulting should take place in Mongoid 8.0, NOT 7.6

Neilshweky: @p-mongo and I discussed this. He didn't want to release a major version for this... @p-mongo thoughts?

docs/reference/fields.txt
Neilshweky: Alright, if you really think so I'll change it

Neilshweky: done

docs/reference/fields.txt
johnnyshields: It should be documented in the release notes and options documentation for `map_big_decimal_to_decimal128`. It should make clear that when switching the flag, it will break the ability to query against existing data, and that all data must be migrated.

**Fortunately**, in the real-world, applications are very unlikely to be doing any queries on BigDecimal as string today, because it's pretty useless--you'd have to match against exact values. In other worlds, if you have `field :money_amt, type: BigDecimal`, then `.gt(money_amt: 100)` will NOT work today. The only possible case is something like `.ne(money_amt: nil)` which will work regardless of string or decimal128.

docs/reference/fields.txt
Neilshweky: Hey we've seen this and we're reconsidering how we want to version this... Thanks!

p-mongo (approved):
:+1:

comandeo (approved):

PR #5133, opened 2022-01-28
Fix MONGOID-5172 Criteria#any_of with multiple arguments incorrectly adds conditions to existing top-level $or if one is already present


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-5172

johnnyshields:
🎉

docs/release-notes/mongoid-7.5.txt
```diff
@@ -119,6 +119,61 @@ Embedded associations (``embeds_one`` and ``embeds_many``):
 +---------------------------------------+---------------------------------------+
 
 
+``any_of`` Adds Multiple Arguments As Top-Level Conditions
+----------------------------------------------------------
+
+**Breaking change:** When ``any_of`` is invoked with multiple conditions, the
```
Neilshweky, line 125: Did we conclude that we're going to release this breaking change in 7.5? I thought it had to be in the next major version?

Neilshweky (approved):
LGTM, just not sure about which version to release this in.

docs/release-notes/mongoid-7.5.txt
p-mongo: Yes, we will redo the versions but for now since everything is referencing 7.5 it would be confusing to have a mix of 7.5 and 8.0 references thus I'm continuing with the existing numbers.

comandeo (approved):

docs/release-notes/mongoid-7.5.txt
johnnyshields: It's a good call to put this in 8.0.

PR #5136, opened 2022-02-01
MONGOID-5219 Uniqueness validation for StringifiedSymbol field in an embedded document fails

Neilshweky:
The changes here fix the mongoize in the `create_criteria` method for the uniqueness validator. Previously, the selector was `mongoize`d using `Object`'s `.mongoize` method. This was not type specific, so for the `StringifiedSymbol` type, a `Symbol` was not `mongoize`d into a string. These changes extract the field for the attribute we're trying to `mongoize` and `mongoize` based on the type of that field. 

There is a special case when reading the field from the document for a `localized` field. This field is already `mongoized` by the time it gets to the `create_criteria` function and so I have a special case for it.

MONGOID-5219

### DIFF
--- a/lib/mongoid/validatable/uniqueness.rb
+++ b/lib/mongoid/validatable/uniqueness.rb
@@ -94,9 +94,8 @@ module Mongoid
       # @return [ Criteria ] The criteria.
       def create_criteria(base, document, attribute, value)
         criteria = scope(base.unscoped, document, attribute)
-        field_name = document.aliased_fields[attribute.to_s] || attribute.to_s
-        field = document.fields[field_name]
-        mongoized = field&.localized? ? value.mongoize : field.mongoize(value) 
+        field = document.fields[document.database_field_name(attribute)]
+        mongoized = field.try(:localized?) ? value.mongoize : field.mongoize(value) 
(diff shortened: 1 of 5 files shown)


lib/mongoid/validatable/uniqueness.rb
```diff
@@ -94,7 +94,9 @@ def case_sensitive?
       # @return [ Criteria ] The criteria.
       def create_criteria(base, document, attribute, value)
         criteria = scope(base.unscoped, document, attribute)
-        criteria.selector.update(criterion(document, attribute, value.mongoize))
+        field = document.fields[document.database_field_name(attribute)]
+        mongoized = field.try(:localized?) ? value.mongoize : field.mongoize(value)
```
p-mongo, line 98: Can you please add a comment here explaining the logic, specifically when is the instance mongoize the correct method to call and when the class one is the correct one?

Neilshweky: done

p-mongo (approved):
:+1: Thank you that is a great explanation!

comandeo (approved):
:+1:

PR #5137, opened 2022-02-01
MONGOID-5204 default_scopes duplication, query scope inconsistency

Neilshweky:
Look at the following comment: https://jira.mongodb.org/browse/MONGOID-5204?focusedCommentId=4328701&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4328701

The functionality after this PR is as follows: 
```
(byebug) Sound.where(active: true)
#<Mongoid::Criteria
  selector: {"active"=>true}
  options:  {}
  class:    Sound
  embedded: false>
(byebug) Sound.where(active: true).where(active: true)
#<Mongoid::Criteria
  selector: {"active"=>true}
  options:  {}
  class:    Sound
  embedded: false>
(byebug) Sound.where(active: true).where(active: true).where(active:false)
#<Mongoid::Criteria
  selector: {"active"=>true, "$and"=>[{"active"=>false}]}
  options:  {}
  class:    Sound
  embedded: false>
(byebug) Sound.where(active: true).where(active: true).where(active:false).where(active:true)
#<Mongoid::Criteria
  selector: {"active"=>true, "$and"=>[{"active"=>false}]}
  options:  {}
  class:    Sound
  embedded: false>
```
which makes sense to me.

### DIFF
--- a/lib/mongoid/criteria/queryable/storable.rb
+++ b/lib/mongoid/criteria/queryable/storable.rb
@@ -54,7 +54,7 @@ module Mongoid
               # adding them to the existing hash.
               new_value = selector[field].merge(value)
               selector.store(field, new_value)
-            else
+            elsif selector[field] != value
               add_operator_expression('$and', [{field => value}])
             end
           else
--- a/spec/mongoid/scopable_spec.rb
+++ b/spec/mongoid/scopable_spec.rb
@@ -110,6 +110,21 @@ describe Mongoid::Scopable do
         expect(Band).to be_default_scoping
       end
     end
+    
+    context "when parent class has default scope" do 
+      
+      before do 
+        Sound.send(:include, Ownable)
+      end
+      
+      let (:selector) do 
+        AudibleSound.all.selector
+      end
+      
+      it "the subclass doesn't duplicate the default scope in the selector" do
+        expect(selector).to eq({'active' => true})
+      end
+    end
   end
 
   describe ".default_scopable?" do
--- a/spec/mongoid/scopable_spec.rb
+++ b/spec/mongoid/scopable_spec.rb
@@ -110,13 +110,13 @@ describe Mongoid::Scopable do
         expect(Band).to be_default_scoping
       end
     end
-    
-    context "when parent class has default scope" do 
-      
-      let (:selector) do 
+
+    context "when parent class has default scope" do
+
+      let (:selector) do
         AudibleSound.all.selector
       end
-      
+
       it "the subclass doesn't duplicate the default scope in the selector" do
         expect(selector).to eq({'active' => true})
       end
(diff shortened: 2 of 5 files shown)


braindeaf (commented):
I will note that we added a similar fix (which we rolled into production last week). We ended up with odd case where occasional updates complained of nil.reset_strategies!

If it successfully added the expression it would return self after the new criteria changes had been applied. In this case we had to ensure that the method returned self if no changes were made.

braindeaf:
> I will note that we added a similar fix (which we rolled into production last week). We ended up with odd case where occasional updates complained of nil.reset_strategies!
> 
> If it successfully added the expression it would return self after the new criteria changes had been applied. In this case we had to ensure that the method returned self if no changes were made.

I'm gonna take that back...

```
  return self if selector[field] == value

  add_operator_expression('$and', [{field => value}])
end
```

I returned early which is why it didn't work. Cool, this looks good anyway thank you for looking at it :)

spec/mongoid/scopable_spec.rb
```diff
@@ -110,6 +110,21 @@
         expect(Band).to be_default_scoping
       end
     end
+    
+    context "when parent class has default scope" do 
+      
+      before do 
+        Sound.send(:include, Ownable)
```
p-mongo, line 117: Why is this include here and not in the model definition itself?

p-mongo: A test should not be mutating model classes. Either this test needs a new model class or one of the existing defintions needs to be changed in a way that doesn't break any of the other tests.

lib/mongoid/criteria/queryable/storable.rb
```diff
@@ -54,7 +54,7 @@ def add_field_expression(field, value)
               # adding them to the existing hash.
               new_value = selector[field].merge(value)
               selector.store(field, new_value)
-            else
+            elsif selector[field] != value
```
p-mongo, line 57: :+1:

spec/mongoid/scopable_spec.rb
Neilshweky: I was actually just following an example [here](https://github.com/mongodb/mongoid/blob/5c90efb00347e25d8b5ff0ce27d00ac8c906f0e2/spec/mongoid/fields_spec.rb#L1198) that does the same thing. This is done to allow for access of the subclass that lives in the same file without giving it its own file. I can just put the subclass in its own file instead, no big deal.

Neilshweky: done

p-mongo: I see, this is one of the things in the existing code that we shouldn't be doing going forward.

p-mongo (approved):

comandeo (approved):

PR #5140, opened 2022-02-08
MONGOID-5240 embeds_many does not get reset after clear

Neilshweky:


spec/mongoid/reloadable_spec.rb
```diff
@@ -356,6 +356,71 @@
...
+          Contractor.new(name: 'contractor')
+        end
+
+        let(:building) do
+          Building.create!
+        end
+
+        before do
+          building.contractors << contractor
+          building.contractors.clear
```
p-mongo, line 382: This flow should have intermediate assertions, for example that after `clear` there aren't any contractors.

Neilshweky: done

p-mongo (approved):

comandeo (approved):

PR #5141, opened 2022-02-09
MONGOID-5234 Add Feature Flag: Remove BSON::ObjectId#as_json override

Neilshweky:
MONGOID-5234

### DIFF
--- a/lib/mongoid/extensions.rb
+++ b/lib/mongoid/extensions.rb
@@ -3,6 +3,14 @@
 class BSON::ObjectId
   alias :bson_ruby_as_json :as_json
 
+  # Return a string representation of the object id for use in
+  # application-level JSON serialization. This method is intentionally
+  # different from #as_extended_json.
+  #
+  # @example Get the object id as a JSON-serializable object.
+  #   object_id.as_json
+  #
+  # @return [ String ] The object id as a string.
   def as_json(options = nil)
     if Mongoid::use_bson_ruby_as_json
       bson_ruby_as_json(options)
--- a/lib/mongoid/extensions.rb
+++ b/lib/mongoid/extensions.rb
@@ -4,8 +4,7 @@ class BSON::ObjectId
   alias :bson_ruby_as_json :as_json
 
   # Return a string representation of the object id for use in
-  # application-level JSON serialization. This method is intentionally
-  # different from #as_extended_json.
+  # application-level JSON serialization.
   #
   # @example Get the object id as a JSON-serializable object.
   #   object_id.as_json
--- a/lib/mongoid/extensions.rb
+++ b/lib/mongoid/extensions.rb
@@ -10,9 +10,9 @@ class BSON::ObjectId
   #   object_id.as_json
   #
   # @return [ String ] The object id as a string.
-  def as_json(options = nil)
+  def as_json(*args)
     if Mongoid::use_bson_ruby_as_json
-      bson_ruby_as_json(options)
+      bson_ruby_as_json(*args)
     else
       { "$oid" => to_s }
     end
--- a/lib/mongoid/extensions.rb
+++ b/lib/mongoid/extensions.rb
@@ -1,8 +1,14 @@
 # frozen_string_literal: true
 
 class BSON::ObjectId
+  alias :bson_ruby_as_json :as_json
+
   def as_json(options = nil)
-    { "$oid" => to_s }
+    if Mongoid::use_bson_ruby_as_json
+      bson_ruby_as_json(options)
+    else
+      { "$oid" => to_s }
+    end
   end
 end
 
--- a/lib/mongoid/extensions.rb
+++ b/lib/mongoid/extensions.rb
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
+class BSON::ObjectId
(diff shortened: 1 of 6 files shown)


lib/mongoid/extensions.rb
```diff
@@ -1,5 +1,17 @@
 # frozen_string_literal: true
 
+class BSON::ObjectId
+  alias :bson_ruby_as_json :as_json
+
+  def as_json(options = nil)
```
p-mongo, line 6: This method needs a docstring, e.g. copy pasted from the definition in bson-ruby.

spec/mongoid/config_spec.rb
```diff
@@ -270,6 +270,99 @@
     end
   end
 
+  context 'when the update_embedded_after_nil option is set in the config' do
```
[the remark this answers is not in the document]
p-mongo: Please put this into a separate PR referencing the ticket that these changes go with.

p-mongo (commented):
The documentation in  https://github.com/mongodb/mongoid/blob/master/docs/release-notes/mongoid-7.4.txt#L100 needs to be updated also.

lib/mongoid/extensions.rb
Neilshweky: done

p-mongo: `options` is not documented, and why is the signature of this method different from https://github.com/mongodb/bson-ruby/blob/master/lib/bson/object_id.rb#L75 ?

Neilshweky: I'm not sure I didn't create the method signature... That was created by Durran Jordan 12 years ago. I think I definitely could change it, though, because the * operator is able to take in everything a hash was able to. I don't think that would be a breaking change.

Also options is not documented because options is not used at all in the function. It is also not documented in bson-ruby.  I'm not sure what I would document other than something like "The arguments to the function" without specifying what it's expecting...

johnnyshields:
Really shouldn't be a feature flagged change. Everyone has long since monkey-patched this method.

johnnyshields:
Just a thought. It might be nice to prefix all feature flag names with `feature_` or `deprecated_` so it's clear which ones will be removed in the next major Mongoid version. Otherwise, we risk getting users who "prefer" the legacy behavior which is intended to be removed.

comandeo:
> Just a thought. It might be nice to prefix all feature flag names with `feature_` or `deprecated_` so it's clear which ones will be removed in the next major Mongoid version. Otherwise, we risk getting users who "prefer" the legacy behavior which is intended to be removed.

It's a very good point, thank you! We created a ticket for this task, and will review all the flag names before the release - https://jira.mongodb.org/browse/MONGOID-5247

sriedel:
Just stumbled over this PR as a mongoid user. If it does what I think it does (allow a BSON::ObjectId to be serialized as a simple hexadecimal string out of the box instead of using the `$oid` wrapper): Yay!

On the other hand, the naming and documentation of this feature does not clarify what the feature does and how to use it:
```
      # In Mongoid 7.3.3 and earlier, the BSON::ObjectId#as_json method would
      # default to returning { "$oid" => id.to_s }. Turning off this flag will
      # defer the as_json call to bson-ruby's implementation of 
      # BSON::ObjectId#as_json. (default: true)
```
Questions that pop into my head as a mongoid user who treasures his ignorance of internal implementation details: What does defering the call to `bson-ruby` do? Does that preserve the legacy behavior? Change it to a new behavior? What does the output for each setting of the option look like? Do I as a  user of mongoid have to care that the `#as_json` call is being internally delegated to somewhere else? Surely the change in visible behavior should be enough to know about?

```
      use_bson_ruby_as_json: true
```
The option name describes the internal implementation; it does not make it clear what the actual behavioral change (visible from outside mongoid) actually is. If the internal implementation is refactored to use different methods, this name becomes even more opaque as you'd need to know the history of mongoid/bson to make sense of it.

Something like `serialize_bson_object_id_with_oid_wrapper` would be long-ish but a lot clearer as to the purpose and what is changed by toggling the flag.

lib/mongoid/extensions.rb
p-mongo: There are two as_json implementations:

bson-ruby: https://github.com/mongodb/bson-ruby/blob/master/lib/bson/object_id.rb#L75

Mongoid: https://github.com/mongodb/mongoid/blob/7.3-stable/lib/mongoid/extensions.rb#L5

The bson-ruby one is compatible with the Mongoid one, has a docstring and supports more invocation options. Let's use the bson-ruby implementation.

The "This method is intentionally different from #as_extended_json." sentence I think does not make sense in Mongoid's context and should be removed.

Neilshweky: I removed that part of the comment but I'm not sure what the first point you're trying to make is. The crux of the ticket is to defer to the bson-ruby one, that's what we're adding the feature flag for?

p-mongo: I suggest using the bson-ruby signature: `    def as_json(*args)`

Neilshweky: done

p-mongo (approved):

PR #5142, opened 2022-02-09
MONGOID-5232 Add Feature Flag: Bring Mongoid::Document's === implementation in alignment with Ruby behavior

Neilshweky:


docs/reference/configuration.txt
```diff
@@ -296,6 +296,10 @@ for details on driver options.
       # existing method. (default: false)
       scope_overwrite_exception: false
 
+      # Corrects === behavior to be consistent with Ruby by only calling is_a?
+      # (default: false)
+      triple_equals_uses_is_a: false
```
Neilshweky, line 301: I'll definitely take suggestions on better flag names...

p-mongo (approved):

p-mongo (changes requested):
The documentation in https://github.com/mongodb/mongoid/blob/master/docs/release-notes/mongoid-7.4.txt#L28 needs to be adjusted.

Neilshweky:
@p-mongo done!

PR #5143, opened 2022-02-09
MONGOID-5229 Add Feature Flag: .with_scope should restore previous scope

Neilshweky:


p-mongo:
The original change  (bdef211fa0) had a documentation fragment added. Now that we know that in the default configuration Mongoid doesn't behave in the ideal manner, that fragment should be expanded to say that to have the correct nesting behavior the global toggle needs to be toggled.

Neilshweky:
Done @p-mongo

p-mongo (approved):

PR #5144, opened 2022-02-09
MONGOID-5233 Add Feature Flag: Respect aliased fields in pluck/distinct by having Document.database_field_name recursively consider embedded docs

Neilshweky:


johnnyshields:
Can someone please explain why there is a feature flag here?

PR #5146, opened 2022-02-10
MONGOID-5239 Add Feature Flag: Mongoid::Contextual::Aggregable::Mongo#sum does not return nil as expected for Mongoid::Document#none

Neilshweky:
MONGOID-5239

### DIFF
--- a/lib/mongoid/contextual/none.rb
+++ b/lib/mongoid/contextual/none.rb
@@ -13,6 +13,8 @@ module Mongoid
 
       attr_reader :criteria, :klass
 
+      alias :new_sum :sum
+
       # Get the sum in the null context.
       #
       # @example Get the sum of null context.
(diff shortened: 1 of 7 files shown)


lib/mongoid/contextual/none.rb
```diff
@@ -6,11 +6,27 @@ module Mongoid
...
+      #
+      # @example Get the sum of null context.
+      #     context.sum(_field)
+      #
+      # @param [ Symbol ] _field The field to sum.
+      #
+      # @return [ Integer | String ] If the return_nil_on_aggregate_none is
+      #   turn one, this will always be zero. Otherwise, it will return the
+      #   field name as string.
+      def sum(_field = nil)
```
Neilshweky, line 26: Aliasing the old `sum` and using it here is only necessary for the `sum` function. Before original ticket, the null context did not implement `aggregates`, `avg`, `min`, or `max` at all. Since this is an addition I don't think it needs to be put behind a feature flag.

johnnyshields:
Mongoid team seems to be really going overboard with these feature flags... there needs to be some degree of common sense applied to what is likely to actually affect real-world users. This change for example is a bug-fix... every user expects sum to return nil in this case, and real world apps are really not likely to break as a result.

p-mongo (approved):

PR #5147, opened 2022-02-10
MONGOID-5236 Add Feature Flag: Existing `$and` clause in query is overwritten by `where` condition

Neilshweky:
MONGOID-5236

### DIFF
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -272,6 +272,28 @@ for details on driver options.
       # (default: false)
       fix_embedded_alias_pluck_distinct: false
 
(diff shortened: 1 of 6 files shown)


johnnyshields:
This again is a bug. I'm against having a feature-flag here. Users will be worse-off if they are unaware to enable this feature flag.

To reiterate: feature flags should be reserved for bona-fide "golden path" behavior changes that will surprise existing Mongoid users. They should not be used on bugfixes where users are expecting something, e.g. chaining and/where operators, to work properly.

p-mongo (approved):
This PR needs to be rebased but the contents is good with me.

PR #5152, opened 2022-02-16
MONGOID-5176 Deprecate old MongoDB support

comandeo:


docs/reference/compatibility.txt
```diff
@@ -168,6 +168,9 @@ Please refer to `the driver compatibility page
 <https://docs.mongodb.com/ruby-driver/current/reference/driver-compatibility/>`_
 for driver compatibility matrices.
 
+"D" in a column means support for that MongoDB server version is deprecated
+and will be removed in a next version.
```
Neilshweky, line 172: I think you should move this explanation up a section since "Ruby Compatibility" also uses this convention.

Neilshweky: Oh nvmd it is there already.

p-mongo (approved):

PR #5153, opened 2022-02-16
MONGOID-4476 Writes in callbacks of nested embedded documents are not persisted when top-level document is modified via update_attributes

Neilshweky:
I'm going to finish the explanation here. See my preliminary observations [here](https://jira.mongodb.org/browse/MONGOID-4476?focusedCommentId=4359006&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4359006)

When setting an attribute of a Mongoid Document, the `attributes` hash is updated. My original confusion was how this `attributes hash`, when changed, would automatically update the `delayed_atomic_sets` hash. It turns out, that when setting up the `delayed_atomic_sets` hash, we would store the `attributes` hash itself as the value (and the key is the path). This means, that when the `attributes` hash is updated later, the `delayed_atomic_sets` hash would also update automatically, because they reference the same hash in memory. This was happening for `Galaxy`, the grandparent, and `Star`, the parent. However, `Planet`, the grandchild of `Galax`y, would use the `as_document` function when being stored into the `delayed_atomic_sets` hash. This means that the hash that was stored in the `delayed_atomic_sets` hash, was different than the planet's `attributes` hash. Therefore, when changing the age on the planet, the `attributes` hash would be updated, but the `delayed_atomic_sets` hash would not be updated, leading to that attribute to not be persisted.


comandeo (approved):

p-mongo:
Am I correct in understanding that now the attributes hash is effectively passed by reference in a place where it was previously copied? If so, because normally values should be copied:

- can you please annotate with a comment the line where the value must be passed by reference, e.g. copying the explanation from the ticket/this PR into the code
- can you please add a docstring to the as_attributes method modified in this PR explaining what it does

Neilshweky:
@p-mongo done! I also added a test for the embeds_one grandchild case.

PR #5154, opened 2022-02-17
MONGOID-3552 call after_save/after_remove callbacks at correct time in has_and_belongs_to_many association

Neilshweky:
MONGOID-3552

### DIFF
--- a/lib/mongoid/association/proxy.rb
+++ b/lib/mongoid/association/proxy.rb
@@ -170,9 +170,10 @@ module Mongoid
       end
 
       # Execute the before and after callbacks for the given method.
-      # Returns the result of the given block
       #
       # @param [ Symbol ] name The name of the callbacks to execute.
+      #
+      # @return [ Object ] The result of the given block
       def execute_callbacks_around(name, doc)
         execute_callback :"before_#{name.to_s}", doc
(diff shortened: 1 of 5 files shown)


p-mongo:
Can this PR be edited to  have the title referring to the change made, i.e. either use "fixed XXX" format or describe the actual change made rather than describing the problem which presumably no longer exists?

It's hard to tell presently, are there user-visible behavior changes here? (I would expect so.) If yes there should be a release note added describing them.

lib/mongoid/association/proxy.rb
```diff
@@ -169,6 +169,17 @@ def execute_callback(callback, doc)
         end
       end
 
+      # Execute the before and after callbacks for the given method.
+      # Returns the result of the given block
```
comandeo, line 173: Usually we use `@return` tag to specify what is returned from a method. Maybe we should do same here to be consistent.

comandeo (approved):

p-mongo (approved):

lib/mongoid/association/proxy.rb
Neilshweky: good catch, done!

PR #5158, opened 2022-02-21
MONGOID-4869 persist the empty list in an embeds_many relationship

Neilshweky:


spec/support/models/label.rb
```diff
@@ -44,6 +44,6 @@ def after_validation_stub
 
   private
   def cleanup
-    self.name = self.name.downcase.capitalize
+    self.name = self.name.downcase.capitalize if self.name
```
Neilshweky, line 47: Test-only change that was causing saving a Label to crash when not setting a name.

lib/mongoid/association/embedded/batchable.rb
```diff
@@ -232,7 +237,11 @@ def normalize_docs(docs)
         #
         # @return [ String ] The atomic path.
         def path
-          @path ||= _unscoped.first.atomic_path
+          unless _unscoped.empty?
+            @path ||= _unscoped.first.atomic_path
+          else
```
comandeo, line 242: It is generally not recommended to use `unless` with `else` clause - https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/UnlessElse
I think it would be more readable with `if`.

comandeo (approved):

lib/mongoid/association/embedded/batchable.rb
p-mongo: :+1:

p-mongo: Additionally I find it more readable to perform the assignment once, i.e.:

```
@path ||= if ...
  ...
else
  ...
end
```

lib/mongoid/atomic/paths/embedded/many.rb
```diff
@@ -34,6 +34,25 @@ def position
...
             "#{pos}#{"." unless pos.blank?}#{document._association.store_as}#{locator}"
           end
+
+          class << self
+
+            # Get the position of where the document would go for the given
+            # association. The use case for this function is when trying to
+            # persist an empty list for an embedded association. All of the
+            # existing functions for getting the position to store a document
+            # require passing in a docuent to store, which we don't have when
```
p-mongo, line 44: document

lib/mongoid/association/embedded/batchable.rb
Neilshweky: done

p-mongo (approved):

PR #5159, opened 2022-02-21
MONGOID-4986 Do not call callbacks on unset assoc

comandeo:


Neilshweky:
This ticket has been closed as working as intended... Do you think this PR can be closed @comandeo?

MONGOID-4986

PR #5161, opened 2022-02-23
MONGOID-5039 Multiple polymorphic embeds cause wrong association to be assigned

Neilshweky:


lib/mongoid/association/embedded/embeds_one/binding.rb
```diff
@@ -21,7 +21,10 @@ class Binding
           def bind_one
             _target.parentize(_base)
             binding do
-              _target.do_or_do_not(_association.inverse_setter(_target), _base)
+              begin
+                _target.do_or_do_not(_association.inverse_setter(_target), _base)
+              rescue Mongoid::Errors::AmbiguousRelationship => exception
```
p-mongo, line 26: Can you please add a comment here for future reference summarizing our findings and specifically giving an explanation 1) when this exception would be raised and 2) why ignoring this exception does the right thing?

The exception also doesn't need to be assigned to the variable since you aren't doing anything with it subsequently.

lib/mongoid/association/embedded/embedded_in.rb
```diff
@@ -114,7 +114,14 @@ def polymorphic_inverses(other = nil)
                   rel.as == name &&
                   rel.relation_class_name == inverse_class_name
             end
-            matches.collect { |m| m.name }
+
+            candidates = matches.collect { |m| m.name }
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongoid/association/embedded/embeds_one/binding.rb
Neilshweky: I actually removed this piece of code, put the characterization before the binding and add a check for multiple inverses to the binding.

lib/mongoid/association/embedded/embedded_in/binding.rb
```diff
@@ -19,9 +19,10 @@ class Binding
           #   name.person.bind(:continue => true)
           #   name.person = Person.new
           def bind_one
-            _base._association = _association.inverse_association(_target) unless _base._association
-            _base.parentize(_target)
             binding do
+              check_polymorphic_inverses!(_target)
+              _base._association = _association.inverse_association(_target) unless _base._association
+              _base.parentize(_target)
```
Neilshweky, line 25: This is the crux of the fix. What was happening before was as follows:

Lets say we have:
```
class Customer
  include Mongoid::Document

  embeds_one :home_address, class_name: 'CustomerAddress', as: :addressable
  embeds_one :work_address, class_name: 'CustomerAddress', as: :addressable
end

class CustomerAddress
  include Mongoid::Document

  embedded_in :addressable, polymorphic: true
end
```
and we set the following:
```
customer.work_address = CustomerAddress.new
```
What used to happen is this would get redirected to `embeds_one/binding.rb`'s `bind_one` function. The binding function binds the inverse association to this association. Okay great, now we're trying to bind the embedded_in association to our association. Inside that function, we again `bind_one` which gets redirected here. Before this change, we would again try to bind the inverse association to our association, which would ambiguous since there are two possible inverses.
With this change, the binding function checks if we are already in a binding state, and if we are, don't execute the block. (This is consistent with how `belongs_to`'s binding works). This block actually doesn't need to be executed because the inverse of the inverse association, our association, does not need to be assigned. Our association is being assigned in the `characterize_one` function which is called in the `proxy` right before the original call to `bind_one`. 
Note that if you try to assign `addressable` directly, which would be ambiguous, you will get an error triggered in the `check_polymorphic_inverses!` function.

p-mongo: :+1: Awesome explanation.

p-mongo (approved):

PR #5162, opened 2022-02-24
Use config_override macro to DRY config overrides in tests


p-mongo:


Neilshweky (approved):
I didn't know this macro existed! This is awesome!

PR #5163, opened 2022-02-24
MONGOID-5247 change feature flag names to be reflective of user-visible behavior

p-mongo:


docs/reference/configuration.txt
```diff
@@ -335,58 +341,71 @@ for details on driver options.
...
-      # second update would not work and the value for the embedded document
-      # would remain nil. Take this case:
+      # Maintain legacy behavior of === on Mongoid document classes, which
+      # returns true in a number of cases where Ruby's === implementation would
+      # return false. Note that the behavior of === on Mongoid document
+      # instances differs from both the behavior of === on document classes
+      # and from Ruby's behavior of === on simple object instances regardless
+      # of the value of this option.
+      # (default: true in Mongoid 7, will change to false in Mongoid 8)
+      broken_triple_equals: true
```
p-mongo, line 377: Change to legacy_triple_equals

johnnyshields:
Hmm not sure I like the word `broken_` here. `legacy_` or `deprecated_` is probably better everywhere.

spec/support/macros.rb
```diff
@@ -28,5 +28,15 @@ def config_override(key, value)
         Mongoid.send("#{key}=", existing)
       end
     end
+
+    def with_config_values(key, *values)
+      values.each do |value|
+        context "when #{key} is #{value}" do
+          config_override key, value
+
+          yield
```
Neilshweky, line 37: do you think it's worth including the value as a parameter to `yield` here, so that it can be used when the tested result is not the same in both cases?

Neilshweky (approved):

p-mongo:
@Neilshweky Ready for review

spec/support/macros.rb
p-mongo: That could be useful, do you want me to make this change in this PR?

Neilshweky: It's up to you. This change can also just be made when it's needed.

PR #5166, opened 2022-02-25
MONGOID-4646 has_one - belongs_to with optional:true incorrectly clears association when assigned itself

Neilshweky:


lib/mongoid/association/referenced/has_one/proxy.rb
```diff
@@ -49,17 +49,14 @@ def nullify
...
-            # the target.
-
-            unbind_one
-            if persistable?
-              # TODO can this entire method be skipped if self == replacement?
-              if _association.destructive? && self != replacement
-                send(_association.dependent)
-              else
-                save if persisted?
+            if replacement != self
```
Neilshweky, line 52: There is precedent for this in the EmbedsOne proxy

spec/integration/associations/has_one_spec.rb
```diff
@@ -173,4 +173,42 @@
...
+    let(:post) { HomPost.create! }
+    let(:comment) { HomComment.create! }
+    let(:comment2) { HomComment.create! }
+
+    it "persists it correctly" do
+      post.comment = comment
+      post.reload
+      expect(post.comment).to eq(comment)
+
+      post.comment = comment
```
p-mongo, line 187: Can you add the following test case please:

```
comment2 = Comment.new(id: comment.id)
post.comment = comment2
```

lib/mongoid/association/referenced/has_one/proxy.rb
p-mongo: I think this is fine except the comparison should be for identity (via `equal?`), not equality (via `==`). Looking at `==` implementation in `lib/mongoid/equality.rb`, two objects appear to compare equal if they have the same ids, but in a case like that (which I am hoping my other comment  would reproduce) the rebinding is in fact necessary on both objects.

Assuming my hypothesis is correct, the other place which performs `self != replacement` comparison similarly would need to be changed to use identity comparison.

spec/integration/associations/has_one_spec.rb
Neilshweky: I added this test case and it fails with a dup key exception. This case attempts to insert a new comment (since its new record attribute is true) into the comments collection, and fails since it has the same _id. 

You're right that the `replacement != self` condition does fail, but I think that's okay in this case, since the use case of trying to add a document with the same _id doesn't actually make sense, and we shouldn't go forward with unbinding and rebinding etc. If you want to modify the comment, you should get it from post.comment and update it... 

I think all in all the functionality is correct for this case, we should raise an error and not do anything.

Neilshweky: After thinking this over I think you're right... The only downside to this approach are the extra operations that happen when assigning the same document, that don't share the same reference, twice.

lib/mongoid/association/referenced/has_one/proxy.rb
```diff
@@ -49,20 +49,23 @@ def nullify
...
 
-            unbind_one
-            if persistable?
-              # TODO can this entire method be skipped if self == replacement?
-              if _association.destructive? && self != replacement
-                send(_association.dependent)
-              else
-                save if persisted?
+              if replacement
+                reset_foreign_key!(replacement)
```
Neilshweky, line 65: This solves the problem of the stale foreign key not allowing the update to persist. It sets the foreign key to nil without updating the changed_attributes of the replacement document. This is okay, because inside the binding in the HasOne Proxy, the foreign key is set to the _base's key anyway. This will not affect the outcome of the operation at all, unless the replacement document has the same value for the foreign key as the original document which could only happen in the following three cases:
1. We are assigning the same document twice to the has_one association. This case doesn't make it through this code path, and exits early at the `equal?` condition above.
2. A second object created with the foreign key explicitly set.
3. Our case, where the foreign key is stale
The last two case are where our problem arises and this function solves that. In my opinion, those writes should still work.
There is one possible problem, however. There are multiple places where an error can be raised between the time that we set the foreign key to nil and the foreign key is set to the correct value. If this happens then we just altered the value of the foreign key on the object the user gave to us and the operation failed. Now, this operation was executed with the assumption that the foreign key would be changed, but setting it to nil is not what the user would expect. I'm not sure what the way forward is here.
Note, doing this operation right before we set the foreign key doesn't work, since the binding is used in many places that we don't want to reset the foreign key.

Neilshweky:
@p-mongo I think I'm going to leave it as `self != replacement`... After testing with rails and this implementation, it gives the functionality we want. 

The case that you're talking about when you assign two objects with the same _id ends up trying to insert a new record with the same id, which fails with a duplicate key error. Nothing happens in this case. 

I put this case, and other cases in tests and I think this is working as expected. If the replacement record has the same _id and different attributes (i.e. the same object with different attrs), then it doesn't need to go through the whole unbinding and rebinding, and will just update the attributes in the update part of the binding.

spec/integration/associations/has_one_spec.rb
Neilshweky: see my comment below

docs/reference/associations.txt
```diff
@@ -585,6 +585,58 @@ help of MongoDB projection operation:
   #  "projection"=>{"_id"=>1, "label"=>1}}
   Band.where(started_on: {'$gt' => Time.now - 1.year}).only(:label).map(&:label).compact.uniq
 
+Setting Stale Values on Reference Associations
```
p-mongo, line 588: Referenced

docs/reference/associations.txt
```diff
@@ -585,6 +585,58 @@ help of MongoDB projection operation:
...
+  post.comment = comment1
+  post.reload
+
+Finally, this last assignment attempts to set the ``post_id`` on ``comment1``,
+which should be ``nil`` at this point, but is set to the old ``post_id``.
+During this operation, the ``post_id`` is cleared from ``comment2``, and the
+new ``post_id`` is set on ``comment1``. However, since the ``post_id`` was
+already set on ``comment1``, nothing is persisted, and we end up with both
+comments having a ``nil`` ``post_id``. At this point, running ``post.comment``
+returns ``nil``.
```
p-mongo, line 638: Thank you very much, I appreciate the thorough documentation.

p-mongo (approved):

docs/reference/associations.txt
Neilshweky: done!

PR #5168, opened 2022-03-01
MONGOID-5252 Add ruby 3.1 to compat tables

comandeo:


docs/reference/compatibility.txt
```diff
@@ -89,12 +90,14 @@ is deprecated.
...
      - |checkmark|
      - |checkmark|
+     - |checkmark|
      -
      -
      -
      - |checkmark|
 
    * - 7.3
+     - |checkmark| [#mongoid-7.3-ruby-3.0]_
```
p-mongo, line 100: These footnotes aren't appropriate here because they reference 3.0 right?

comandeo: Actually, footnotes say "Mongoid version 7.x.y or higher is required.", and this is correct for both Ruby 3.0 and Ruby 3.1.

comandeo: Actually, footnotes say "Mongoid version 7.x.y or higher is required.", and this is correct for both Ruby 3.0 and Ruby 3.1.

Neilshweky (approved):

p-mongo (approved):

PR #5170, opened 2022-03-01
MONGOID-5043 'where' query incorrectly works for belongs_to relations with custom primary_key

Neilshweky:


lib/mongoid/fields/foreign_key.rb
```diff
@@ -60,6 +60,8 @@ def evolve(object)
         if object_id_field? || object.is_a?(Document)
           if association.polymorphic?
             association.convert_to_foreign_key(object)
+          elsif object.is_a?(Document) && object.respond_to?(association.primary_key)
+            primary_key_field.evolve(object.send(association.primary_key))
```
Neilshweky, line 64: The idea here is if we're trying to evolve a document, we are going to refer to it by its primary key. When no custom primary key is set, the primary key defaults to `_id`, and this operates exactly like `object.__evolve_object_id__`.

Neilshweky: The reason I can't just get rid of the else condition outright, is because there are situations where the object given to the function is not a Document.

comandeo (approved):

PR #5174, opened 2022-03-10
MONGOID-2586 cannot access parent object in the after_find function

neilshweky:
MONGOID-2586

This PR is ready for review.

### DIFF
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -164,7 +164,7 @@ module Mongoid
       @pending_callbacks ||= []
     end
 
-    # Run the pending callbacks. If the callback is :default, we will apply
+    # Run the pending callbacks. If the callback is :apply_defaults, we will apply
     # the defaults for this document. Otherwise, the callback is passed to the
     # run_callbacks function.
     #


lib/mongoid/factory.rb
```diff
@@ -60,14 +60,16 @@ def build(klass, attributes = nil)
     # @param [ Hash ] selected_fields Fields which were retrieved via
     #   #only. If selected_fields are specified, fields not listed in it
     #   will not be accessible in the returned document.
+    # @param [ true | false ] execute_callbacks Flag specifies whether find
+    #   callbacks should be run.
     #
     # @return [ Document ] The instantiated document.
-    def from_db(klass, attributes = nil, criteria = nil, selected_fields = nil)
+    def from_db(klass, attributes = nil, criteria = nil, selected_fields = nil, execute_callbacks: true)
```
johnnyshields, line 67: Think the code would be clearer if this variable were called `defer_callbacks`.

To me, if `execute_callbacks: false` it implies that the callbacks will never be executed; its not obvious that they will be deferred until later.

Neilshweky: Good point, I was still workshopping names as this is a draft PR but thanks for the idea

lib/mongoid/document.rb
```diff
@@ -98,9 +98,11 @@ def identity
     #   Person.new(:title => "Sir")
     #
     # @param [ Hash ] attrs The attributes to set up the document with.
+    # @param [ true | false ] defer_callbacks Flag specifies whether callbacks
+    #   should be run.
     #
     # @return [ Document ] A new document.
-    def initialize(attrs = nil)
+    def initialize(attrs = nil, defer_callbacks: false)
```
Neilshweky, line 105: Is this okay? This is a public function so I'm not sure if I could add to the method signature.

lib/mongoid/document.rb
```diff
@@ -98,9 +98,11 @@ def identity
     #   Person.new(:title => "Sir")
     #
     # @param [ Hash ] attrs The attributes to set up the document with.
+    # @param [ true | false ] defer_callbacks Flag specifies whether callbacks
+    #   should be run.
     #
     # @return [ Document ] A new document.
-    def initialize(attrs = nil)
+    def initialize(attrs = nil, defer_callbacks = false)
```
p-mongo, line 105: In my opinion this is not ideal, I think a better way would be to use `allocate` and add a parallel method to construct instances internally.

Neilshweky: Oh that is a great idea

Neilshweky: done

Neilshweky: The method signature was changed a little so that I could explicitly pass the block, but it is the same for all intents and purposes.

lib/mongoid/document.rb
```diff
@@ -281,20 +296,39 @@ module ClassMethods
...
         doc = allocate
         doc.__selected_fields = selected_fields
         doc.instance_variable_set(:@attributes, attributes)
-        doc.apply_defaults
-        yield(doc) if block_given?
-        doc.run_callbacks(:find) unless doc._find_callbacks.empty?
-        doc.run_callbacks(:initialize) unless doc._initialize_callbacks.empty?
+
+        if defer_callbacks
+          doc.pending_callbacks.push(*[:default, :find, :initialize])
```
johnnyshields, line 310: What's the reason to splat the array here?

Can't this just be:

```
doc.pending_callbacks.push(:default, :find, :initialize)
```

Neilshweky: Yup, good call.

lib/mongoid/association/accessors.rb
```diff
@@ -43,7 +43,17 @@ def __build__(name, object, association, selected_fields = nil)
       def create_relation(object, association, selected_fields = nil)
         type = @attributes[association.inverse_type]
         target = association.build(self, object, type, selected_fields)
-        target ? association.create_relation(self, target) : nil
+        target = target ? association.create_relation(self, target) : nil
```
p-mongo, line 46: Can this be expanded to a simple `if` statement please, for ease of following?

lib/mongoid/association/accessors.rb
```diff
@@ -43,7 +43,17 @@ def __build__(name, object, association, selected_fields = nil)
...
+
+        # Only need to do this on embedded associations. The pending callbacks
+        # are only added when materializing the documents, which only happens
+        # on embedded associations. There is no call to the database in the
+        # construction of a referenced association.
```
p-mongo, line 51: This comment goes with the condition on line 54, right? If so can this `if` statement also be turned around so that the condition is on top next to the comment.

lib/mongoid/document.rb
```diff
@@ -228,6 +215,34 @@ def becomes(klass)
...
     private
 
+    # Does the construction of a document.
+    #
+    # @param [ Hash ] attrs The attributes to set up the document with.
+    # @param [ true | false ] defer_callbacks Flag specifies whether callbacks
+    #   should be run.
+    #
+    # @return [ Document ] A new document.
+    def construct_document(attrs = nil, defer_callbacks = false)
```
p-mongo, line 225: I would prefer `run_callbacks` (positive and matches the method that eventually will be called), and I would like it to be a keyword argument. If this method is always called with the first argument (`attrs`), the first argument should be required. Finally, even though this method is private, it should still be marked `@api private` because application models include `Mongoid::Document` and thus are able to invoke private methods.

lib/mongoid/document.rb
```diff
@@ -281,20 +296,39 @@ module ClassMethods
       # @param [ Hash ] attrs The hash of attributes to instantiate with.
       # @param [ Integer ] selected_fields The selected fields from the
       #   criteria.
+      # @param [ true | false ] defer_callbacks Flag specifies whether callbacks
+      #   should be run.
       #
       # @return [ Document ] A new document.
-      def instantiate(attrs = nil, selected_fields = nil)
+      def instantiate(attrs = nil, selected_fields = nil, defer_callbacks = false)
```
p-mongo, line 303: If we agree that this method is part of Mongoid's public API, I would prefer its API to stay the same as before (like with `initialize`) and another method added to deal with the callback behavior.

lib/mongoid/document.rb
```diff
@@ -281,20 +296,39 @@ module ClassMethods
...
       end
 
+      # Allocates and constructs a document.
+      #
+      # @param [ Hash ] attrs The attributes to set up the document with.
+      # @param [ true | false ] defer_callbacks Flag specifies whether callbacks
+      #   should be run.
+      #
+      # @return [ Document ] A new document.
+      def construct_document(attrs = nil, defer_callbacks = false)
```
p-mongo, line 327: `@api private`, I think?

lib/mongoid/factory.rb
```diff
@@ -23,16 +23,18 @@ module Factory
     #
     # @param [ Class ] klass The class to instantiate from if _type is not present.
     # @param [ Hash ] attributes The document attributes.
+    # @param [ true | false ] defer_callbacks Flag specifies whether callbacks
+    #   should be run.
     #
     # @return [ Document ] The instantiated document.
-    def build(klass, attributes = nil)
+    def build(klass, attributes = nil, defer_callbacks = false)
```
p-mongo, line 30: If this method and `from_db` are part of public API I'm afraid they would also need to be duplicated.

lib/mongoid/interceptable.rb
```diff
@@ -152,6 +152,30 @@ def _mongoid_run_child_callbacks(kind, children: nil, &block)
       end
     end
 
+    # This is used to store callbacks to be executed later. A good use case for
+    # this is delaying the after_find and after_initialize callbacks until the
+    # associations are set on the document. This can also be used to delay
+    # applying the defaults on a document.
+    #
+    # @return [ Array<Symbol> ] an array of symbols that represent the pending callbacks.
```
p-mongo, line 160: `@api private`

lib/mongoid/interceptable.rb
```diff
@@ -152,6 +152,30 @@ def _mongoid_run_child_callbacks(kind, children: nil, &block)
...
+    def pending_callbacks
+      @pending_callbacks ||= []
+    end
+
+    # Run the pending callbacks. If the callback is :default, we will apply
+    # the defaults for this document. Otherwise, the callback is passed to the
+    # run_callbacks function.
+    def run_pending_callbacks
```
p-mongo, line 168: `@api private`

lib/mongoid/interceptable.rb
```diff
@@ -152,6 +152,30 @@ def _mongoid_run_child_callbacks(kind, children: nil, &block)
...
+      pending_callbacks.each do |cb|
+        if cb == :default
```
p-mongo, line 170: Can `:default` be renamed to `:apply_defaults`? That would make the callers easier to read.

spec/mongoid/document_spec.rb
```diff
@@ -339,23 +318,6 @@ class << self; attr_accessor :name; end
...
-      end
-
-      let(:document) do
-        Band.instantiate("_id" => id, "name" => "Depeche Mode") do |band|
-          band.likes = 1000
-        end
-      end
-
-      it "yields to the block" do
-        expect(document.likes).to eq(1000)
```
p-mongo, line 355: I can't find this test in the green part of the diff, is it in here somewhere?

lib/mongoid/document.rb
johnnyshields: Disagree. As per my previous comment, `run_callbacks` (previously `execute_callbacks`) is not a good name because `run_callbacks = false` implies that callbacks will never be run.

Personally I also prefer default variables to be `nil` or `false`.

lib/mongoid/association/accessors.rb
Neilshweky: done

lib/mongoid/association/accessors.rb
Neilshweky: done

lib/mongoid/document.rb
Neilshweky: yup, done.

lib/mongoid/interceptable.rb
Neilshweky: done

spec/mongoid/document_spec.rb
Neilshweky: I have added this test back, as well as the functionality for giving a block to instantiate.

lib/mongoid/document.rb
Neilshweky: I have flipped this flag and renamed it (back) to `execute_callbacks`

lib/mongoid/document.rb
Neilshweky: done

lib/mongoid/factory.rb
Neilshweky: done

p-mongo: I tried searching for this on github and while most results are irrelevant the second hit on https://github.com/search?l=Ruby&p=2&q=factory+from_db&type=Code shows someone using this method.

Neilshweky: IMO if they're not explicitly labeled `@private api` then they are part of the public API. I duplicated the method to add the new args.

spec/mongoid/document_spec.rb
p-mongo: That was added in 9c4daf4435, not clear to what end. Given the test added in that commit perhaps the goal of the block API is to permit augmenting the created object during instantiation rather than consuming it in the block. Let's keep the block API for now.

lib/mongoid/document.rb
```diff
@@ -228,6 +215,34 @@ def becomes(klass)
 
     private
 
+    # Does the construction of a document.
+    #
+    # @param [ Hash ] attrs The attributes to set up the document with.
+    # @param [ true | false ] execute_callbacks Flag specifies whether callbacks
+    #   should be run.
+    #
+    # @return [ Document ] A new document.
```
p-mongo, line 224: `@api private` please?

lib/mongoid/factory.rb
```diff
@@ -63,12 +79,30 @@ def build(klass, attributes = nil)
...
+    # Execute from_db.
+    #
+    # @param [ Class ] klass The class to instantiate from if _type is not present.
+    # @param [ Hash ] attributes The document attributes.
+    # @param [ Criteria ] criteria Optional criteria object.
+    # @param [ Hash ] selected_fields Fields which were retrieved via
+    #   #only. If selected_fields are specified, fields not listed in it
+    #   will not be accessible in the returned document.
+    # @param [ true | false ] execute_callbacks Flag specifies whether callbacks
+    #   should be run.
```
p-mongo, line 94: To clarify the callback behavior, how about something like:

> Whether this method should invoke the callbacks. If true, the callbacks will be invoked normally. If false, the callbacks will be stored in the +pending_callbacks+ list and caller is responsible for invoking +run_pending_callbacks+ at a later time. Use this option to defer callback execution until the entire object graph containing embedded associations is constructed.

That's a lot of text to duplicate in all docstrings, perhaps one copy of it will suffice somewhere with appropriate references.

spec/mongoid/document_spec.rb
```diff
@@ -347,6 +326,7 @@ class << self; attr_accessor :name; end
 
       let(:document) do
         Band.instantiate("_id" => id, "name" => "Depeche Mode") do |band|
+          byebug
```
p-mongo, line 329: Leftover

p-mongo (approved):

lib/mongoid/document.rb
Neilshweky: done

spec/mongoid/document_spec.rb
Neilshweky: done

lib/mongoid/factory.rb
Neilshweky: I don't like doing references in the code in case that code is changed. I put the description in the code once and left everything else as is. Let me know if that's good enough or what kind of reference I can put everywhere else.

spec/mongoid/document_spec.rb
Neilshweky: Okay great

lib/mongoid/factory.rb
```diff
@@ -63,12 +79,34 @@ def build(klass, attributes = nil)
     #
     # @return [ Document ] The instantiated document.
     def from_db(klass, attributes = nil, criteria = nil, selected_fields = nil)
+      execute_from_db(klass, attributes, criteria, selected_fields, execute_callbacks: true)
+    end
+
+    # Execute from_db.
```
comandeo, line 85: I am not sure this comment is helpful. This method seems to do much more that just calling `from_db`. Maybe we can add some more detailed explanation what is does.

lib/mongoid/interceptable.rb
```diff
@@ -152,6 +152,34 @@ def _mongoid_run_child_callbacks(kind, children: nil, &block)
       end
     end
 
+    # This is used to store callbacks to be executed later. A good use case for
```
comandeo, line 155: I think it would be helpful to explain "later" here; I believe what is meant is "later than according to standard callback order", but I am not sure.

comandeo (approved):
:+1:

lib/mongoid/factory.rb
Neilshweky: Do you have any suggestions? Truth is there is a lot to put here, but I left it all under the from_db method.

lib/mongoid/factory.rb
p-mongo: That is good with me!

p-mongo (approved):
I'm good with the diff as it is. I see how it's not immediately obvious what is going on to someone unfamiliar with the implementation but all of the information is in here as far as I can tell so hopefully it won't be too difficult to figure out in the future.

lib/mongoid/interceptable.rb
Neilshweky: I don't think it is after than according to standard callback order. The thing is, I meant this as more of a general operator and not for a specific use case so it's hard to say what later is. That's why I included an example. But in that example, we're just waiting for the associations to get set, the callbacks should still be called in the correct order. Maybe actually the after_find/after_initialize now gets called after the associations after_add.

lib/mongoid/factory.rb
comandeo: If this method is just an implementation and should not be called itself, it's fine to leave it as it is now. Maybe we can make this method private to emphasize it.

lib/mongoid/interceptable.rb
comandeo: :ok_man:

lib/mongoid/factory.rb
Neilshweky: It is an implementation with an additional parameter added to it. It's only called explicitly within our code and is therefore labeled `@api private`... It didn't make sense to make it actually private since it needed to be called from elsewhere in our code.

PR #5175, opened 2022-03-13
MONGOID-7 broken_and


p-mongo:
https://jira.mongodb.com/browse/MONGOID-7

### DIFF
--- a/docs/release-notes/mongoid-7.4.txt
+++ b/docs/release-notes/mongoid-7.4.txt
@@ -17,6 +17,10 @@ The complete list of releases is available `on GitHub
 please consult GitHub releases for detailed release notes and JIRA for
 the complete list of issues fixed in each release, including bug fixes.
 
+All behavior changes in Mongoid 7.4 must be explicitly requested by changing
+the value of configuration options as detailed below. By default,
+Mongoid 7.4 behaves the same as Mongoid 7.3.
+
 
 Ruby Version Support
 --------------------
@@ -25,11 +29,12 @@ As of version 7.4, Mongoid supports Ruby 2.5+.
 Support for Ruby 2.4 and earlier has been dropped.
 
 
-``===`` Operator Matches Ruby Semantics
----------------------------------------
+Change ``===`` Operator To Match Ruby Semantics
+-----------------------------------------------
 
-**Breaking change:** In Mongoid 7.4, the ``===`` operator works the same way
-as it does in Ruby, and is equivalent to calling ``is_a?`` on the right hand
+In Mongoid 7.4, the ``===`` operator on ``Mongoid::Document`` classes and
+instances can be configured to behave the same way as it does in Ruby,
+and is equivalent to calling ``is_a?`` on the right hand
 side with the left hand side as the argument:
 
 .. code-block:: ruby
@@ -39,14 +44,11 @@ side with the left hand side as the argument:
   # equivalent to:
   instance.is_a?(ModelClass)
 
-Previously, ``===`` returned ``true`` for some cases when the equivalent Ruby
-``===`` implementation returned false.
-
-.. note::
-
-  In order to get this functionality, the ``Mongoid.legacy_triple_equals``
-  global flag must be set to false. If it is set to true, the ``===`` operator
-  will function as it did in Mongoid 7.3.
+In order to get this functionality, the ``Mongoid.legacy_triple_equals``
+option must be set to false. If it is set to true, which is the default for
+Mongoid 7.4, the ``===`` operator will function as it did in Mongoid 7.3:
+``===`` returned ``true`` for some cases when the equivalent Ruby
+``===`` implementation returned false, as per the examples below.
 
 Mongoid 7.4 behavior:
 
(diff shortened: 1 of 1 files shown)


docs/release-notes/mongoid-7.4.txt
```diff
@@ -199,3 +241,93 @@ an index where it wouldn't before:
...
+
+  The value of ``Mongoid.broken_update`` option will change to ``false``
+  by default in Mongoid 8.0.
+
+
+Correct Logical ``and`` Query Generation
+----------------------------------------
+
+Mongoid 7.4 with the ``Mongoid.broken_and`` option set to ``false``
+will preserve existing conditions when using ``and`` to add new conditions
```
Neilshweky, line 299: I think I understand what you're trying to say but maybe something like:
"will unnecessarily add redundant conditions when the same operator is used on the same field multiple times"
Is a little more clear? What do you think?

docs/release-notes/mongoid-7.4.txt
```diff
@@ -199,3 +241,93 @@ an index where it wouldn't before:
...
+
+  Band.where(id: 1).and({year: {'$in' => [2020]}}, {year: {'$in' => [2021]}}).where(id: 2)
+
+Mongoid 7.4 with the ``Mongoid.broken_and`` option set to ``false`` will
+generate the following criteria:
+
+.. code-block:: ruby
+
+  <Mongoid::Criteria
+    selector: {"_id"=>1, "year"=>{"$in"=>[2020]}, "$and"=>[{"_id"=>2}]}
```
Neilshweky, line 313: Is this not correct behavior? If I'm not mistaken it looks like you used the same criteria for broken and working behavior...

docs/release-notes/mongoid-7.4.txt
```diff
@@ -199,3 +241,93 @@ an index where it wouldn't before:
...
+    embedded: false>
+
+In Mongoid 7.3 and earlier, and in 7.4 with the ``Mongoid.broken_and``
+option set to ``true`` (the default), the following criteria would be
+generated instead:
+
+.. code-block:: ruby
+
+  <Mongoid::Criteria
+    selector: {"_id"=>1, "year"=>{"$in"=>[2020]}, "$and"=>[{"_id"=>2}]}
```
Neilshweky, line 325: This looks the same as above

docs/release-notes/mongoid-7.4.txt
```diff
@@ -199,3 +241,130 @@ an index where it wouldn't before:
...
+
+.. code-block:: ruby
+
+  Band.with_scope(year: 2020) do
+    Band.with_scope(active: true) do
+      # ...
+    end
+    
+    # {year: 2020} condition is applied here
+  entd
```
Neilshweky, line 351:
```suggestion
  end
```

docs/release-notes/mongoid-7.4.txt
```diff
@@ -199,3 +241,130 @@ an index where it wouldn't before:
...
+
+.. code-block:: ruby
+
+  Band.with_scope(year: 2020) do
+    Band.with_scope(active: true) do
+      # ...
+    end
+    
+    # No scope is applied here
+  entd
```
Neilshweky, line 365:
```suggestion
  end
```

docs/release-notes/mongoid-7.4.txt
p-mongo: Yes indeed, thank you.

docs/release-notes/mongoid-7.4.txt
p-mongo: Or maybe I misunderstood, broken_and does not deal with *redundant* conditions.

docs/release-notes/mongoid-7.4.txt
```diff
@@ -199,3 +240,130 @@ an index where it wouldn't before:
...
+
+  #<Mongoid::Criteria
+    selector: {"_id"=>1, "year"=>{"$in"=>[2020]}, "$and"=>[{"year"=>{"$in"=>[2021]}}, {"_id"=>2}]}
+    options:  {}
+    class:    Band
+    embedded: false>
+
+In Mongoid 7.3 and earlier, and in 7.4 with the ``Mongoid.broken_and``
+option set to ``true`` (the default), the following criteria would be
+generated instead which omit the {"$in" => [2021]} condition:
```
Neilshweky, line 319: "omits"?

It's weird because, yes, criteria is a plural word, but for all intents and purposes we treat it like singular.... Doesn't actually matter just whatever sounds better

docs/release-notes/mongoid-7.4.txt
p-mongo: I added a hopefully correct output now, does that clarify things at all?

p-mongo: What language would you suggest?

Neilshweky: Something like this maybe? 

> Mongoid 7.4 with the ``Mongoid.broken_and`` option set to ``false`` will avoid adding redundant conditions when using ``and`` to add conditions with the same operator on the same field multiple times.

docs/release-notes/mongoid-7.4.txt
Neilshweky: Actually I think you're right on reading it again. Resolving.

docs/release-notes/mongoid-7.4.txt
Neilshweky: Your language looks good to me!

Neilshweky (approved):

PR #5177, opened 2022-03-14
MONGOID-5246 Flip feature flags for 8.0


p-mongo:


johnnyshields:
@p-mongo I think it's fine to remove the feature flags and the legacy behavior entirely in 8.0. It's allowed according to semver.

p-mongo:
That is allowed by semver, true, however we elected to keep the feature flags into the next major release to assist with upgrading.

Neilshweky (approved):
LGTM

PR #5178, opened 2022-03-14
Add VSCode debugger dependencies

Neilshweky:


p-mongo:
Looking at CI results perhaps debase is not compatible with ruby 2.5 and 3.1? If so it can be conditioned to 2.6...3.0.

p-mongo (approved):
Maybe the issue is lack of dependencies, I conditioned on CI.

PR #5181, opened 2022-03-16
MONGOID-5110 Parent ID missing in after_destroy callback when removing via nested attribute

Neilshweky:


lib/mongoid/association/nested/many.rb
```diff
@@ -134,8 +134,9 @@ def destroy(parent, relation, doc)
         # @param [ Proxy ] relation The association proxy.
         # @param [ Document ] doc The document to delete.
         def destroy_document(relation, doc)
+          res = doc.destroy unless doc.embedded? || doc.destroyed?
           relation.delete(doc)
-          doc.destroy unless doc.embedded? || doc.destroyed?
+          res
```
Neilshweky, line 139: The fix here is switching the order of these lines. We're going to destroy the document first and then afterwards remove it from the relation. This way, the relation still exists in the after_destroy callback.

comandeo (approved):

p-mongo:
:+1:

PR #5182, opened 2022-03-16
debase does not build against ruby 3.1


p-mongo:


gemfiles/standard.rb
```diff
@@ -6,7 +6,7 @@ def standard_dependencies
 
     platform :mri do
       # Debugger for VSCode.
-      unless ENV['CI']
+      if !ENV['CI'] && RUBY_VERSION < '3.1'
```
Neilshweky, line 9: I think make this less than 3.0... It doesn't work for 3.0.3 for me either

p-mongo: Okay I excluded 3.0.

Neilshweky (approved):

PR #5183, opened 2022-03-16
MONGOID-5260 Add _translations fields and dot-navigation to pluck and distinct

Neilshweky:
I have not written the release notes or docs yet but I'm going to use this as scratch space:

Difference between new and old pluck and distinct:

|                                              | new                                                    | old                                          |
|----------------------------------------------|--------------------------------------------------------|----------------------------------------------|
| callbacks                                    | none                                     | none                                         |
| plucking a localized field                   | "deutsch-text"                                         | {"de"=>"deutsch-text", "en"=>"english-text"} |
| plucking an _translations                    | {"de"=>"deutsch-text", "en"=>"english-text"}           | nil                                          |
| plucking a dotted field (records.name)       | "the name"                                             | { name => "the name" }                       |
| plucking dotted localized field (:'desc.de') | 'deutsch-text' | {'de' => 'deutsch-text'}                     |
| plucking dotted translation field (:'desc_transaltions.de') | 'deutsch-text' | nil             |

I've decided not to call the find/initialize callbacks in pluck/distinct for the following reasons: 
1. distinct only gets one value so we don't actually initialize an object to call the callbacks on. I would like pluck/distinct to behave the same in this regard
2. Neither pluck or distinct return the actually objects that we're initializing, and I think that's the reason the callbacks weren't called in the first place.

I am open to having those callbacks called (I don't think I could on distinct, though), but I don't think it's necessary.

### DIFF
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -373,7 +373,7 @@ for details on driver options.
       # and will return embedded values themselves (i.e. without putting them
       # in a hash).
       # (default: true in Mongoid 7, will change to false in Mongoid 8)
-      legacy_pluck_distinct: true
+      #legacy_pluck_distinct: true
 
       # Maintain legacy behavior of === on Mongoid document classes, which
       # returns true in a number of cases where Ruby's === implementation would
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -368,9 +368,10 @@ for details on driver options.
 
       # Maintain legacy behavior of pluck and distinct, which does not demongoize
       # values on returning them. Setting this option to false will cause
-      # pluck and distinct to return demongoized values. This will also allow
-      # passing _translations fields to pluck and distinct and will return
-      # embedded values themselves (i.e. without putting them in a hash).
+      # pluck and distinct to return demongoized values. Setting this option to
+      # false will also allow passing _translations fields to pluck and distinct
+      # and will return embedded values themselves (i.e. without putting them
+      # in a hash).
       # (default: true in Mongoid 7, will change to false in Mongoid 8)
       legacy_pluck_distinct: true
 
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -369,10 +369,10 @@ for details on driver options.
       # Maintain legacy behavior of pluck and distinct, which does not demongoize
       # values on returning them. Setting this option to false will cause
       # pluck and distinct to return demongoized values. Setting this option to
-      # false will also allow passing _translations fields to pluck and distinct
-      # and will return embedded values themselves (i.e. without putting them
-      # in a hash).
-      # (default: true in Mongoid 7, will change to false in Mongoid 8)
+      # false will also allow retrieving *_translations fields from pluck and
+      # distinct and will return embedded values themselves (i.e. without
+      # putting them in a hash).
+      # (default: false)
       #legacy_pluck_distinct: true
 
       # Maintain legacy behavior of === on Mongoid document classes, which
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -368,9 +368,11 @@ for details on driver options.
 
       # Maintain legacy behavior of pluck and distinct, which does not demongoize
       # values on returning them. Setting this option to false will cause
-      # pluck and distinct to return demongoized values.
-      # (default: false)
-      #legacy_pluck_distinct: true
+      # pluck and distinct to return demongoized values. This will also allow
+      # passing _translations fields to pluck and distinct and will return
+      # embedded values themselves (i.e. without putting them in a hash).
+      # (default: true in Mongoid 7, will change to false in Mongoid 8)
+      legacy_pluck_distinct: true
 
       # Maintain legacy behavior of === on Mongoid document classes, which
       # returns true in a number of cases where Ruby's === implementation would
--- a/docs/release-notes/mongoid-7.4.txt
+++ b/docs/release-notes/mongoid-7.4.txt
@@ -56,7 +56,7 @@ Mongoid 7.4 with ``Mongoid.legacy_triple_equals`` set to ``false`` behavior:
 
   class Band
     include Mongoid::Document
-    
+
     has_many :members
   end
 
@@ -65,7 +65,7 @@ Mongoid 7.4 with ``Mongoid.legacy_triple_equals`` set to ``false`` behavior:
 
   class Member
     include Mongoid::Document
-    
+
     belongs_to :band
   end
 
@@ -83,10 +83,10 @@ Mongoid 7.4 with ``Mongoid.legacy_triple_equals`` set to ``false`` behavior:
 
   CoverBand === Band
   # => false
-  
+
   band.members === Array
   # => false
-  
+
   band.members === Mongoid::Association::Referenced::HasMany::Enumerable
   # => false
 
@@ -106,10 +106,10 @@ behavior:
 
   CoverBand === Band
   # => true
-  
+
   band.members === Array
   # => true
-  
+
   band.members === Mongoid::Association::Referenced::HasMany::Enumerable
   # => true
 
@@ -198,55 +198,6 @@ top-level models.
 The ``Mongoid.compare_time_by_ms`` option is set to ``false`` by default
 in Mongoid 7.4 for backwards compatibility. In Mongoid 8 the default value
 will change to ``true``.
-
-
-Respect Field Aliases In Embedded Documents When Using ``distinct`` And ``pluck``
----------------------------------------------------------------------------------
-
-When ``distinct`` and ``pluck`` are used with aliased fields in embedded
-documents, the aliases can be expanded if the ``Mongoid.broken_alias_handling``
-option is set to ``false``. By default, for backwards compatibility, in
-Mongoid 7.4 this option is set to true, yielding Mongoid 7.3 and earlier
-behavior. Given the following definitions:
-
-.. code-block:: ruby
-
-  class Band
-    include Mongoid::Document
-    embeds_many :managers
-  end
-
-  class Manager
-    include Mongoid::Document
-    embedded_in :band
-
-    field :name, as: :n
-  end
-
-Mongoid 7.4 behavior with ``Mongoid.broken_alias_handling`` set to ``false``:
-
-.. code-block:: ruby
-
-  # Expands out to "managers.name" in the query:
-  Band.distinct('managers.n')
-  Band.pluck('managers.n')
-
-Mongoid 7.3 and 7.4 with ``Mongoid.broken_alias_handling`` set to ``true`` behavior:
-
-.. code-block:: ruby
-
-  # Sends "managers.n" without expanding the alias:
-  Band.distinct('managers.n')
-  Band.pluck('managers.n')
-
-.. note::
-
-  The alias expansion for top-level fields has already been done by Mongoid 7.3.
-
-.. note::
-
-  The value of ``Mongoid.broken_alias_handling`` option will change to ``false``
-  by default in Mongoid 8.0.
 
 
 ``count``, ``sum``, ``avg``, ``min``, ``max`` Ignore Sort If Not Limiting/Skipping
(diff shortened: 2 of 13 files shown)


Neilshweky:
@johnnyshields does this PR answer all of your comments?

johnnyshields:
@Neilshweky **nice!!**

I think "plucking dotted localized field (:'desc.de')" return "deutsch-text" (i.e. the value of "de"); it should behave the same as `distinct`.

In addition, please add tests for the case where I18n fallbacks are enabled.

I agree we don't need to call the find/initialize callbacks here; pluck/distinct are expected not to run callbacks at all; they should be lightweight.

Neilshweky:
> @Neilshweky **nice!!**
> 
> I think "plucking dotted localized field (:'desc.de')" return "deutsch-text" (i.e. the value of "de"); it should behave the same as `distinct`.
> 
> In addition, please add tests for the case where I18n fallbacks are enabled.
> 
> I agree we don't need to call the find/initialize callbacks here; pluck/distinct are expected not to run callbacks at all; they should be lightweight.

Done and done!

spec/mongoid/criteria_spec.rb
```diff
@@ -3057,8 +3133,77 @@ class D
...
+          end
+        end
+      end
+
+      context 'when fallbacks are enabled with a locale list' do
+        before(:all) do
+          puts "I18n version: #{I18n::VERSION}"
+
+          require "i18n/backend/fallbacks"
+          I18n::Backend::Simple.send(:include, I18n::Backend::Fallbacks)
```
p-mongo, line 3181: We have evergreen configurations that have fallbacks enabled, those should be used to test fallback behavior instead of enabling callbacks in the test body.

docs/reference/configuration.txt
```diff
@@ -369,7 +369,9 @@ for details on driver options.
 
       # Maintain legacy behavior of pluck and distinct, which does not demongoize
       # values on returning them. Setting this option to false will cause
-      # pluck and distinct to return demongoized values.
+      # pluck and distinct to return demongoized values. This will also allow
+      # passing _translations fields to pluck and distinct and will return
+      # embedded values themselves (i.e. without putting them in a hash).
```
p-mongo, line 374: The blurb you're adding is for the false value of the option, whereas the existing description is for the true value.

spec/mongoid/criteria_spec.rb
johnnyshields: @p-mongo my two cents--_as a separate PR_, it would be nice if fallback tests could be incorporated into the main test suite, rather than making them a separate run config. You could probably make an RSpec helper that mocks the fallback functionality for a certain scope.

I don't see any strong argument for making I18n fallbacks require a "special" test run more so than any other feature.

docs/reference/configuration.txt
Neilshweky: Ah I figured saying "This will also allow" was referring to the previous sentence which was talking about setting it to false. I can make that more clear.

lib/mongoid/fields.rb
```diff
@@ -76,6 +76,42 @@ def extract_id_field(attributes)
...
+            key = fn
+            unless klass.fields.has_key?(fn) || klass.relations.has_key?(fn)
+              if tr = fn.match(/(.*)_translations\z/)&.captures&.first
+                key = tr
+              else
+                key = fn
+              end
+
+            end
+            res.push("#{key}")
```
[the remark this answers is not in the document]
Neilshweky: done

spec/mongoid/criteria_spec.rb
p-mongo: The setup in `spec/integration/i18n_fallbacks_spec.rb` loads the fallbacks module into process space. This is a global action affecting the entire test suite. To be able to test with fallbacks both on and off, there are separate configurations.

Neilshweky:
> @Neilshweky one more thing, can you please also test the case of Localized + Embedded fields. e.g.:
> 
> ```
> (Page embeds many Sections, Section has localized field :content)
> 
> Page.all.pluck('sections.content_translations')
> Page.all.pluck('sections.content')
> Page.all.pluck('sections.content.de')
> ```

Done!

lib/mongoid/fields.rb
```diff
@@ -76,6 +76,42 @@ def extract_id_field(attributes)
...
+
+            if klass.fields.has_key?(fn)
```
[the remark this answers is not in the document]
Neilshweky: done

docs/reference/configuration.txt
Neilshweky: done

spec/mongoid/criteria_spec.rb
Neilshweky: done! (removed include fallbacks)

Neilshweky:
@p-mongo you mentioned that pluck instantiating a new model decreases performance, so I refactored this method to navigate given hash.

Neilshweky:
@p-mongo how do I require a fallbacks deployment?

docs/reference/configuration.txt
```diff
@@ -368,8 +368,11 @@ for details on driver options.
 
       # Maintain legacy behavior of pluck and distinct, which does not demongoize
       # values on returning them. Setting this option to false will cause
-      # pluck and distinct to return demongoized values.
-      # (default: false)
+      # pluck and distinct to return demongoized values. Setting this option to
+      # false will also allow passing _translations fields to pluck and distinct
+      # and will return embedded values themselves (i.e. without putting them
+      # in a hash).
+      # (default: true in Mongoid 7, will change to false in Mongoid 8)
```
p-mongo, line 375: Since the default was already changed to false in master, I suggest leaving it as false here and in the 7.4 branch replacing false with the "true in 7, false in 8" line.

docs/reference/configuration.txt
```diff
@@ -368,8 +368,11 @@ for details on driver options.
 
       # Maintain legacy behavior of pluck and distinct, which does not demongoize
       # values on returning them. Setting this option to false will cause
-      # pluck and distinct to return demongoized values.
-      # (default: false)
+      # pluck and distinct to return demongoized values. Setting this option to
+      # false will also allow passing _translations fields to pluck and distinct
```
p-mongo, line 372: I suggest "retrieving *_translation fields", I think this is more clear.

docs/release-notes/mongoid-7.4.txt
```diff
@@ -390,12 +341,64 @@ all scopes are cleared:
...
+    field :name, as: :n
+  end
+
+Mongoid 7.4 behavior with ``Mongoid.broken_alias_handling`` set to ``false``:
+
+.. code-block:: ruby
+
+  # Expands out to "managers.name" in the query:
+  Band.distinct('managers.n')
+  Band.pluck('managers.n')
```
p-mongo, line 376: These still should be "managers.name" expands to "managers.n" right?

docs/release-notes/mongoid-7.4.txt
```diff
@@ -390,12 +341,64 @@ all scopes are cleared:
...
+
+Mongoid 7.3 and 7.4 with ``Mongoid.broken_alias_handling`` set to ``true`` behavior:
+
+.. code-block:: ruby
+
+  # Sends "managers.n" without expanding the alias:
+  Band.distinct('managers.n')
+  Band.pluck('managers.n')
```
p-mongo, line 384: And here I think the actual usage would be "managers.name" as well (but like you said without alias expansion).

docs/release-notes/mongoid-7.4.txt
```diff
@@ -428,3 +436,95 @@ distinct methods will not be demongoized. For example:
 
   The value of ``Mongoid.legacy_pluck_distinct`` option will change to ``false``
   by default in Mongoid 8.0.
+
+Localized Fields with ``pluck`` and ``distinct``
+````````````````````````````````````````````````
+Mongoid 7.4 with the ``Mongoid.legacy_pluck_distinct`` option set to ``false``
+changes the behavior of using ``pluck`` and ``distinct`` with localized fields.
+Now when inputting a localized field into these functions, the translation for
```
p-mongo, line 444: I would say "when retrieving a localized field". Also Ruby doesn't have functions, it has methods.

lib/mongoid/contextual/mongo.rb
```diff
@@ -420,20 +426,28 @@ def map_reduce(map, reduce)
       #
       # @return [ Array<Object, Array> ] The plucked values.
       def pluck(*fields)
+        # Multiple fields can map to the same field name. For example, plucking
+        # a field and its _translations field map to the same field in the database.
+        # because of this, we need to keep track of the fields inputted.
```
p-mongo, line 431: Suggest s/inputted/requested/

lib/mongoid/contextual/mongo.rb
```diff
@@ -420,20 +426,28 @@ def map_reduce(map, reduce)
...
+        normalized_field_names = []
         normalized_select = fields.inject({}) do |hash, f|
-          hash[klass.database_field_name(f)] = 1
+          db_fn = klass.database_field_name(f)
+          normalized_field_names.push(db_fn)
+
+          if Mongoid.legacy_pluck_distinct
+            hash[db_fn] = 1
```
p-mongo, line 438: Although `1` does work, it looks a bit too perl-y. I suggest `true` since Ruby has booleans.

lib/mongoid/contextual/mongo.rb
```diff
@@ -420,20 +426,28 @@ def map_reduce(map, reduce)
...
 
         view.projection(normalized_select).reduce([]) do |plucked, doc|
-          values = normalized_select.keys.map do |n|
-            res = n =~ /\./ ? doc[n.partition('.')[0]] : doc[n]
+          values = normalized_field_names.map do |n|
             if Mongoid.legacy_pluck_distinct
-              res
-            elsif field = klass.fields[n]
-              field.demongoize(res)
+              n =~ /\./ ? doc[n.partition('.')[0]] : doc[n]
```
p-mongo, line 448: I see that the regexp match was already in the code, but I suggest `n.include?('.')` on the assumption that it is surely faster.

lib/mongoid/contextual/mongo.rb
```diff
@@ -680,6 +694,107 @@ def _session
...
         collection.write_concern.nil? || collection.write_concern.acknowledged?
       end
+
+      # Extracts the value for the given field name from the given attribute
+      # hash.
+      #
+      # @param [ Hash ] attrs The attributes hash.
+      # @param [ String ] field_name The name of the field to extract.
+      #
+      # @param [ Any ] The value for the given field name
```
p-mongo, line 704: I usually use `Object` to indicate any type of value is acceptable.

p-mongo (commented):
Looks very good!

docs/reference/configuration.txt
Neilshweky: done

docs/release-notes/mongoid-7.4.txt
Neilshweky: done

docs/release-notes/mongoid-7.4.txt
Neilshweky: done

lib/mongoid/contextual/mongo.rb
Neilshweky: Done. I think this was originally done this way since it represents a MongoDB projection, which is usually of the form `key: [0,1]`. Both work though.

lib/mongoid/contextual/mongo.rb
Neilshweky: done

lib/mongoid/contextual/mongo.rb
Neilshweky: done

p-mongo (approved):

p-mongo:
If you are using `TEST_I18N_FALLBACKS`, you need to add the test files to the list in https://github.com/mongodb/mongoid/blob/master/.evergreen/run-tests.sh#L102 so that they are actually tested in evergreen.

p-mongo (commented):
:+1:

PR #5188, opened 2022-03-22
MONGOID-4911 Range expansion must use $elemMatch to ensure both $gte and $lte match the same array element

Neilshweky:


lib/mongoid/fields.rb
```diff
@@ -310,17 +348,52 @@ def database_field_name(name)
...
           segment, remaining = key.split('.', 2)
-          segment = aliased_fields[segment]&.dup || segment
+
+          # If it's a reference association and there's still remaining, don't
+          # get the alias for this. This is going to retrieve the _id field for
+          # the association name, and getting another property from the _id field
+          # is not the intended functionality.
+          unless remaining && relations.key?(segment) && !relations[segment].embedded?
+            segment = aliased_fields[segment]&.dup || segment
+          end
```
Neilshweky, line 358: Does this logic make sense here? It fixes my tests but I'm not sure it's the right thing. See the comment above it.

lib/mongoid/contextual/mongo.rb
```diff
@@ -712,15 +712,18 @@ def fetch_and_demongoize(d, meth, klass)
...
+        curr = attrs.dup
+
+        klass.traverse_association_tree(field_name) do |meth, obj, is_field|
           is_translation = false
-          if !k.fields.key?(meth) && !k.relations.key?(meth)
-            if tr = meth.match(/(.*)_translations\z/)&.captures&.first
-              is_translation = true
-              meth = tr
-            end
+          # If no relation or field was found, check if the meth is an
```
p-mongo, line 722: s/relation/assocation/

lib/mongoid/criteria/queryable/selector.rb
```diff
@@ -184,16 +200,71 @@ def evolve_hash(serializer, value)
...
+        #     embeds_one or a hash.
+        # (2) The last fields are used inside the $elemMatch. This one is
+        #     actually optional, and will be ignored if the last field is an
+        #     array or embeds_many. If the last field is an array (1), (2) and
+        #     (3) will look like:
+        #
+        #       "klass.child.gchild.ggchild.field" => {
+        #         { "$elemMatch" => { "$gte" => 6, "$lte" => 10 } }
+        #       }
+        #
```
p-mongo, line 224: :+1:

lib/mongoid/criteria/queryable/selector.rb
```diff
@@ -184,16 +200,71 @@ def evolve_hash(serializer, value)
...
-          value.__evolve_range__(serializer: serializer)
+        #
+        # @param [ String ] key The to store the range for.
+        # @param [ Object ] serializer The optional serializer for the field.
+        # @param [ Range ] value The Range to serialize.
+        #
+        # @return [ Array<String, Hash> ] The store name and serialized Range.
+        def evolve_range(key, serializer, value)
+          v = value.__evolve_range__(serializer: serializer)
+          rels = []
```
p-mongo, line 238: While there are many places in Mongoid that use the word "relation", the correct term is "association" and all new code should use the latter.

lib/mongoid/fields.rb
```diff
@@ -270,38 +270,98 @@ def option(option_name, &block)
...
+      # @param [ Hash ] aliased_associations The alaised associations to begin
+      #   the search with.
+      # @param [ Proc ] block The block takes in three paramaters, the current
+      #   meth, the field or the relation, and whether the second parameter is a
+      #   field or not.
+      #
+      # @return [ Field ] The field found for the given key at the end of the
+      #   search. This will return nil if the last thing found is an association
+      #   or no field was found for the given key.
+      def traverse_association_tree(key, fields, associations, aliased_associations)
```
p-mongo, line 297: `@api private` ?

lib/mongoid/fields.rb
```diff
@@ -270,38 +270,98 @@ def option(option_name, &block)
...
+      # pass back the _id field.
       #
       # @param [ String, Symbol ] name The name to get.
+      # @param [ Hash ] relations The associations.
+      # @param [ Hash ] alaiased_fields The aliased fields.
+      # @param [ Hash ] alaiased_associations The aliased associations.
       #
       # @return [ String ] The name of the field as stored in the database.
-      def database_field_name(name)
+      def database_field_name(name, relations, aliased_fields, aliased_associations)
```
p-mongo, line 364: This is an api change, can this method be `@api private` as well?

lib/mongoid/fields.rb
p-mongo: I think this would be more readable with `if` instead of `unless`.

lib/mongoid/fields.rb
```diff
@@ -372,6 +463,21 @@ def using_object_ids?
...
+      #
+      # @param [ String ] key The key used to search the association tree.
+      # @param [ Proc ] block The block takes in three paramaters, the current
+      #   meth, the field or the relation, and whether the second parameter is a
+      #   field or not.
+      #
+      # @return [ Field ] The field found for the given key at the end of the
+      #   search. This will return nil if the last thing found is an association
+      #   or no field was found for the given key.
+      def traverse_association_tree(key, &block)
```
p-mongo, line 477: `@api private`?

lib/mongoid/fields.rb
Neilshweky: This is actually not an API change since I created a separate `database_field_name` with the old API that calls this

lib/mongoid/fields.rb
Neilshweky: done

lib/mongoid/criteria/queryable/selector.rb
Neilshweky: done

lib/mongoid/contextual/mongo.rb
Neilshweky: done

lib/mongoid/fields.rb
Neilshweky: done

lib/mongoid/fields.rb
Neilshweky: done

PR #5189, opened 2022-03-23
move to mongodb.com/docs

terakilobyte:
We're moving the docs to `mongodb.com/docs`. This updates links found in the docs folder that point to `docs.mongodb.com`.

### DIFF
--- a/docs/additional-resources.txt
+++ b/docs/additional-resources.txt
@@ -13,7 +13,7 @@ Additional Resources
 This page lists some of the third-party guides and blog posts about Mongoid,
 as well as sample Mongoid applications. Additional resources for the driver
 are listed on the `respective driver page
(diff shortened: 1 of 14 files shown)


p-mongo (approved):

p-mongo:
@terakilobyte I merged to master and cherry-picked to 7.4-stable. Mongoid docs go back to 5.4 so this will need to be backported quite a ways back. In 7.3-stable there are conflcts, could you make a PR against 7.3-stable with the required changes and I'll continue with cherry-picking it backwards in history.

terakilobyte:
This won't need to be cherry-picked backward beyond that as we'll keep
redirects to ensure older links resolve correctly. However, we'd like new
versions going forward to use the new URL.


{
  name: "*nathan leniz*",
  company: "MongoDB",
  location: "Olympia, WA"
}

PR #5190, opened 2022-03-24
MONGOID-5267 remove driver argument in queryable constructor

Neilshweky:


p-mongo:
The Marshal.dump/load api needs to be retained, meaning the serialized form should still be the 6-element array with `nil` being written instead of `driver`, and when loading the second element should be retrieved and ignored.

p-mongo:
Actually let's write `:mongo` as Mongoid was previously doing.

Neilshweky:
done @p-mongo

lib/mongoid/criteria/marshalable.rb
```diff
@@ -11,7 +11,7 @@ module Marshalable
       #
       # @return [ Array<Object> ] The dumped data.
       def marshal_dump
-        data = [ klass, driver, inclusions, documents, strategy, negating ]
+        data = [ klass, :mongo, inclusions, documents, strategy, negating ]
```
p-mongo, line 14: Yes, but the `load` method shouldn't set `@driver` anymore. Could you also add a note that `:mongo` is written here for backwards compatibility with Mongoid 7 and earlier serializations please?

Neilshweky: done and done

p-mongo (approved):
:+1:  Thank you!

PR #5193, opened 2022-03-24
MONGOID-5261 Partial PR for using config_override

p-mongo:
https://jira.mongodb.com/browse/MONGOID-5261

### DIFF
--- a/spec/mongoid/traversable_spec.rb
+++ b/spec/mongoid/traversable_spec.rb
@@ -336,6 +336,8 @@ describe Mongoid::Traversable do
     end
 
     context "when the discriminator key is changed at the global level" do
+      config_override :discriminator_key, 'hello'
+
       before do
         Mongoid.discriminator_key = "hello"
       end
@@ -904,10 +906,9 @@ describe Mongoid::Traversable do
     end
 
     context "when the global discriminator key conflicts with mongoid's internals" do
-
-      after do
-        Mongoid.discriminator_key = "_type"
-      end
+      # This is not an override, it is meant to restore the original value
+      # after the test only.
+      config_override :discriminator_key, '_type'
 
       [:_association, :invalid].each do |meth|
         context "when the field is named #{meth}" do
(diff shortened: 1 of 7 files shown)


spec/mongoid/traversable_spec.rb
```diff
@@ -904,10 +906,9 @@ class GlobalDuplicateDiscriminatorKeyChild < GlobalDuplicateDiscriminatorKeyPare
     end
 
     context "when the global discriminator key conflicts with mongoid's internals" do
-
-      after do
-        Mongoid.discriminator_key = "_type"
-      end
+      # This is not an override, it is meant to restore the original value
+      # after the test only.
+      config_override :discriminator_key, '_type'
```
Neilshweky, line 911: Why is this an improvement over the other code? Seems like just setting it afterwards is better since it's the intended outcome.

spec/mongoid/traversable_spec.rb
```diff
@@ -336,6 +336,8 @@
     end
 
     context "when the discriminator key is changed at the global level" do
+      config_override :discriminator_key, 'hello'
+
       before do
         Mongoid.discriminator_key = "hello"
       end
```
Neilshweky, line 343: This should be removed, no?

Neilshweky: and the after below

Neilshweky (commented):
There are still tons of references to config options that haven't been changed to use config_override. I gather that's why it's a partial PR. Is your goal to get all of them? If you want to file a ticket I can also complete with you started.

spec/mongoid/traversable_spec.rb
p-mongo: You're right, I missed this.

p-mongo:
The ticket is to use `config_override` everywhere, but this turns out to be somewhat complicated because some tests actually alter the config in the middle of the test (e.g. the discriminator ones) which `config_override` doesn't presently handle well. In order for this PR to not bitrot I decided to get it into master since the contents here is all desired changes.

Neilshweky (approved):

spec/mongoid/traversable_spec.rb
Neilshweky: ah

comandeo (approved):
:clap:

PR #5195, opened 2022-03-28
MONGOID-5217 Add _previously_was methods

comandeo:


johnnyshields:
Any chance to add these on the model as well:

```
previously_new_record?
previously_persisted?
```

johnnyshields:
Great work 👍

docs/release-notes/mongoid-8.0.txt
```diff
@@ -248,3 +248,31 @@ Mongoid 7 output:
 
 Notice that in 7 ``attribute_was(:age)`` returns the old attribute value,
 while in 8.0 ``attribute_was(:age)`` returns the new value.
+
+``*_previously_was``, ``previously_new_record?``, and ``previously_persisted?`` helpers
+---------------------------------------------------------------------------------------
+
+Mongoid 8.0 introduces ActiveModel compatible ``*_previously_was`` helpers,
```
p-mongo, line 255: ActiveModel-compatible

docs/release-notes/mongoid-8.0.txt
```diff
@@ -248,3 +248,31 @@ Mongoid 7 output:
...
+as well as ActiveRecord compatible ``previously_new_record?`` and
```
p-mongo, line 256: ActiveRecord-compatible

docs/release-notes/mongoid-8.0.txt
```diff
@@ -248,3 +248,31 @@ Mongoid 7 output:
...
+``previously_persisted?`` helpers.
```
p-mongo, line 257: Suggest replacing `.` with `:`

spec/integration/callbacks_spec.rb
```diff
@@ -360,4 +360,66 @@ class CBIntSpecProfile
...
+          @previously_new_record_value = doc.previously_new_record?
+        end
+      end
+
+      person = PreviouslyNewRecordPerson.create!(title: "title", age: 55)
+      expect(person.previously_new_record_value).to be_truthy
+      person.title = "New title"
+      person.save!
+      expect(person.previously_new_record_value).to be_falsey
+    end
```
Neilshweky, line 423: I think it also makes sense to test if the document was previously persisted inside a callback. Just because the document was destroyed in the database, doesn't mean it's not still around in the code right?

spec/mongoid/stateful_spec.rb
```diff
@@ -40,6 +40,25 @@
     end
   end
 
+  describe '#previously_new_record?' do
+    it "returns correct values" do
+      person = Person.new("_id" => BSON::ObjectId.new)
```
Neilshweky, line 45: Why not just `Person.new`?

lib/mongoid/reloadable.rb
```diff
@@ -28,6 +28,9 @@ def reload
       @attributes = reloaded
       @attributes_before_type_cast = {}
       changed_attributes.clear
+      previous_changes.clear
+      previous_attributes.clear
```
p-mongo, line 32: These mutate the hashes, to go with my next comment.

lib/mongoid/changeable.rb
```diff
@@ -94,6 +95,10 @@ def previous_changes
       @previous_changes ||= {}
     end
 
+    def previous_attributes
```
p-mongo, line 98: Is this part of public API? If yes it needs a docstring and I suspect the mutation is problematic in the following scenario:

```
band = Band.create!
band.save!
attrs = band.previous_attributes
band.name = 'foo'
band.save! # does this mutate attrs I obtained earlier?
```

p-mongo: I suppose better example would be

```
band = Band.create!
band.save!
attrs = band.previous_attributes
band.reload # does this mutate attrs I obtained earlier?
```

lib/mongoid/stateful.rb
```diff
@@ -31,6 +48,15 @@ def persisted?
       !new_record? && !destroyed?
     end
 
+    # Checks if the document was previously saved to the database
+    # but now it has been deleted.
+    #
+    # @return [ true, false ] True if was persisted but now destroyed,
+    #   otherwise false.
+    def previously_persisted?
+      !new_record? && destroyed?
```
johnnyshields, line 57: The description of this method should be:

```
    # Returns true unless the document was just created -- that is, the object already persisted prior to its last save.
```

I believe the logic should be:

````
  !previously_new_record?
```

I'm pretty sure we don't need to check for `destroyed?` here. The fact that the object was destroyed on the current action shouldn't influence its *previous* state.

p-mongo: As far as I can tell the current docstring matches what is described in https://sapidlabs.com/rails/2021/08/28/rails-7-adds-previously-persisted.html. @johnnyshields are you suggesting an alternate way of communicating the behavior described in this post for rails? Seems to me that your suggestion is to have different behavior.

johnnyshields: You're right, please disregard my comment. OK, makes sense to follow the Rails way.

spec/integration/callbacks_spec.rb
comandeo: Added a test case below. I am not sure how relevant the use case is, but an additional test is helpful.

spec/mongoid/stateful_spec.rb
comandeo: You're right, thank you. Fixed.

docs/release-notes/mongoid-8.0.txt
comandeo: :heavy_check_mark:

docs/release-notes/mongoid-8.0.txt
comandeo: :heavy_check_mark:

docs/release-notes/mongoid-8.0.txt
comandeo: :heavy_check_mark:

lib/mongoid/reloadable.rb
comandeo: Good point, thank you. Replaced `clear` calls with empty hash assignments. I think it should solve the indicated problem.

lib/mongoid/changeable.rb
comandeo: See my answer to the comment above. In addition, I made this method `private`, and added a docstring.

lib/mongoid/changeable.rb
```diff
@@ -291,18 +318,21 @@ def create_dirty_default_change_check(name, meth)
...
+      # Creates the dirty change previous value accessors.
       #
       # @example Create the accessor.
-      #   Model.create_dirty_previous_value_accessor("name", "alias")
+      #   Model.create_dirty_previous_value_accessors("name", "alias")
       #
       # @param [ String ] name The attribute name.
       # @param [ String ] meth The name of the accessor.
-      def create_dirty_previous_value_accessor(name, meth)
+      def create_dirty_previous_value_accessors(name, meth)
```
p-mongo, line 328: If this is a private API it should be marked with `@api private`, otherwise I don't think it can be changed like this.

p-mongo (approved):
I don't see test coverage for the mutation of previous values hash by `reload` - if it's in here and I missed I apologize, otherwise I would prefer it to be added.

lib/mongoid/changeable.rb
comandeo: This method is in a private section. I believe this assumes that such API is private.
On the other hand, there is no need to rename this method, to be honest. I'll revert the change.

johnnyshields:
@comandeo nice work on this PR.

By the way, did we ever rollback the behavior of `attribute_was` in the after_* callbacks?

We can keep the `attribute_previously_was` behavior, however, at least for Mongoid 7.4 release, `attribute_was` needs to preserve its behavior in the after_* backs (i.e. same value as in before_* callbacks). It would be really great to have a feature flag for it, but at a minimum, we need to document it as a breaking change.

comandeo:
Thank you @johnnyshields!

We did not release changes in callbacks/`attribute_was` behavior in 7.4. The plan is to release them in 8.0. They are documented as breaking changes - https://www.mongodb.com/docs/mongoid/master/release-notes/mongoid-8.0/#changeable-module-behavior-made-compatible-with-activemodel-dirty

As to the feature flag: we were investigating a possibility to implement a flag for callbacks order, and it turned out to be quite big and complex. That's why we did not release it in 7.4.

However, we need to take a look it we can add a feature flag for `attribute_was` behavior. If it can be decoupled from other callback related changes, we should do it.

johnnyshields:
Thank you for this comment!

A feature flag just for `attribute_was` behavior (specifically whether or not it reverts in after_* callbacks) would be a big help. The `previously_was` stuff should not be feature flagged, as it is all newly introduced methods.

johnnyshields:
Oh, and I see you just released Mongoid 7.4. Congrats team a lot of work went into this one 🎉

PR #5196, opened 2022-03-28
MONGOID-5034 When excluded from query, embeds many relations are returned as empty array instead of raising MissingAttributeError

Neilshweky:


johnnyshields:
Great 👍

p-mongo (approved):

PR #5200, opened 2022-03-29
MONGOID-5269 Persist writing an attribute after a removal correctly

Neilshweky:


spec/mongoid/attributes_spec.rb
```diff
@@ -2214,4 +2214,19 @@
...
     end
   end
+
+  context "when an attribute is removed then set" do
+    let(:cat) { Cat.create!(name: "Neil") }
+
+    before do
+      cat.remove_attribute("name")
+      cat.name = "Nissim"
+      cat.save
```
p-mongo, line 2224: `save!`

p-mongo (approved):

PR #5203, opened 2022-04-04
Update MRSS with rspec retry stuff

Neilshweky:
After merging to MRSS I need to create a PR to update it on Mongoid, no? 

I'm aware this maybe should be done before it's merged into the MRSS repo, but the code added had nothing to do with Mongoid.


p-mongo:
If you just want to update MRSS like this no need to request review, if CI is green go ahead and merge.

PR #5204, opened 2022-04-04
MONGOID-5274 Add pending test for Issues with #touch method

Neilshweky:


spec/mongoid/touchable_spec.rb
```diff
@@ -584,5 +584,118 @@
...
+
+        let!(:start_time) { Timecop.freeze(Time.at(Time.now.to_i)) }
+
+        let(:update_time) do
+          Timecop.freeze(Time.at(Time.now.to_i) + 2)
+        end
+
+        after do
+          Timecop.return
+        end
```
johnnyshields, line 600: In my version I removed Timecop, because I wanted to make sure various timestamps are actually synchronized when touching. Timecop (because it synchronizes time) will lead to false positives.

spec/mongoid/touchable_spec.rb
```diff
@@ -584,5 +584,118 @@
...
+
+        it "does not update the parent's timestamp" do
+          building.updated_at.should == start_time
+          building.reload.updated_at.should == start_time
+        end
+      end
+
+      context "when touch is false on belongs_to" do
+        let(:parent_cls) { TouchableSpec::Referenced::Building }
```
johnnyshields, line 628: For these types of feature-related specs I think we should move way from "real world" class names (Building, Hospital, etc.) and have the class names be things like EmbedsOne, EmbeddedIn, etc. which describe their actual relationship.

spec/mongoid/touchable_spec.rb
Neilshweky: Can you explain the case specifically where timecop will lead to false positives?

spec/mongoid/touchable_spec.rb
Neilshweky: The problem with that is that the naming space gets too cluttered. How many "NoParent" classes are there going to be, you know? It's also not that big of a deal to just look at the class definition.

p-mongo:
Is there mergeable code in this PR? If yes can you please retitle it to reflect the changes being made.

Neilshweky:
> Is there mergeable code in this PR? If yes can you please retitle it to reflect the changes being made.

This was previously being blocked by MONGOID-5275, but I just explicitly specified the class_name on the models so we don't have to worry about that anymore. I have added a pending test... I think this can be merged as is, and then we can decide if we want to fix this later.

Neilshweky:
After we merge this we can probably just close MONGOID-5274 as a duplicate of MONGOID-5016

spec/mongoid/touchable_spec.rb
```diff
@@ -584,5 +584,117 @@
...
+      shared_examples "does not update the parent" do
+
+        let!(:start_time) { Timecop.freeze(Time.at(Time.now.to_i)) }
+
+        let(:update_time) do
+          Timecop.freeze(Time.at(Time.now.to_i) + 2)
+        end
+
+        after do
+          Timecop.return
```
p-mongo, line 651: Can this return be moved up to the topmost context added in this PR?

spec/mongoid/touchable_spec.rb
```diff
@@ -584,5 +584,117 @@
...
+          expect(floor.building).to_not be nil
+        end
+
+        it "updates the parent's timestamp" do
+          building.updated_at.should == update_time
+          building.reload.updated_at.should == update_time
+        end
+      end
+
+      [ "save!".to_sym, :destroy, :touch].each do |meth|
```
p-mongo, line 623: `:save!` should work no?

spec/mongoid/touchable_spec_models.rb
```diff
@@ -6,15 +6,15 @@ class Building
       include Mongoid::Document
       include Mongoid::Timestamps
 
-      embeds_many :entrances
-      embeds_many :floors
+      embeds_many :entrances, class_name: "TouchableSpec::Embedded::Entrance"
+      embeds_many :floors, class_name: "TouchableSpec::Embedded::Floor"
```
p-mongo, line 10: Is this needed because Mongoid does not resolve class names correctly?

Neilshweky: I believe so

spec/mongoid/touchable_spec.rb
Neilshweky: yes

p-mongo (approved):

PR #5205, opened 2022-04-04
MONGOID-5275 Embedded Associations inside submodules cannot find inverses

Neilshweky:
MONGOID-5275

### DIFF
--- a/lib/mongoid/association/embedded/embeds_many.rb
+++ b/lib/mongoid/association/embedded/embeds_many.rb
@@ -168,9 +168,8 @@ module Mongoid
 
         def determine_inverses(other)
           matches = relation_class.relations.values.select do |rel|
-            assoc_class = rel.relation_class rescue next
             relation_complements.include?(rel.class) &&
-              assoc_class == inverse_class
+              rel.relation_class_name_with_module.sub(/\A::/, '') == inverse_class_name.sub(/\A::/, '')
           end
           if matches.size > 1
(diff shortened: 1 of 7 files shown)


johnnyshields:
@Neilshweky I'll be surprised is the fix is as simple as you have it here. I suspect we may need to rework the relation binding, but if it really is this simple, then great!

Neilshweky:
> @Neilshweky I'll be surprised is the fix is as simple as you have it here. I suspect we may need to rework the relation binding, but if it really is this simple, then great!

Oh I don't think it is but I threw it up in two min just to see what would happen. I'll learn more about the problem from the errors I get.

lib/mongoid/association/embedded/embeds_many.rb
```diff
@@ -168,9 +168,9 @@ def polymorphic_inverses(other = nil)
 
         def determine_inverses(other)
           matches = relation_class.relations.values.select do |rel|
+            assoc_class = rel.relation_class rescue next
```
Neilshweky, line 171: I'm making an assumption here, that if the class doesn't exist, we should not consider this association as a candidate for an inverse.

comandeo (approved):

p-mongo:
Is this waiting for the association target class epic?

Neilshweky:
> Is this waiting for the association target class epic?

If you mean the class autoloading epic, yes.

PR #5206, opened 2022-04-05
Fix MONGOID-5280 Mongoid 7.4 inserts only one embedded document when multiple documents are created with nil _id

Neilshweky:
MONGOID-5280

### DIFF
--- a/lib/mongoid/traversable.rb
+++ b/lib/mongoid/traversable.rb
@@ -178,7 +178,7 @@ module Mongoid
         to_expand = []
         expanding.each do |child|
           next if expanded[child]
-          expanded[child] = true
+          expanded[child] = true if child.id
           children << child
           to_expand += child._children
         end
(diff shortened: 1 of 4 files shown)


lib/mongoid/traversable.rb
```diff
@@ -178,7 +178,7 @@ def collect_descendants
         to_expand = []
         expanding.each do |child|
           next if expanded[child]
-          expanded[child] = true
+          expanded[child] = true if child.id
```
Neilshweky, line 181: The idea here is that documents are compared by their `id`. If the `id` for multiple documents is `nil`, then they will all compare equally. In this case, if multiple documents had a `nil` `id`, they were being dropped by the line above this one.

johnnyshields: We should prefer `._id` to `.id`

PR #5211, opened 2022-04-11
MONGOID-5270 initial implementation of tally

Neilshweky:
This is an initial implementation. The tests don't work and need to be fixed. This ticket is being put on hold for now as 
we prioritize more urgent work for 8.0.

### DIFF
--- a/spec/mongoid/contextual/mongo_spec.rb
+++ b/spec/mongoid/contextual/mongo_spec.rb
@@ -812,6 +812,314 @@ describe Mongoid::Contextual::Mongo do
     end
   end
 
+  describe "#tally" do
+    before do
+      Band.create!(origin: "tally", name: "Depeche Mode", years: 30, sales: "1E2", label: Label.new(name: "Atlantic"))
+      Band.create!(origin: "tally", name: "New Order", years: 30, sales: "2E3", label: Label.new(name: "Atlantic"))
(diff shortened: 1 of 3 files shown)


spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -812,6 +812,314 @@
...
+      let(:criteria) { Band.where(origin: "tally2") }
+
+      let(:tally) do
+        criteria.tally("labels")
+      end
+
+      it "returns the correct hash" do
+        expect(tally).to eq(
+          [1, 2] => 2,
+          [1, 3] => 1
```
johnnyshields, line 947: This doesn't look right. For all cases, the expected behavior should match `criteria.pluck(:field).tally` as close as possible. Here, the keys of the tally should be the label objects. (There's not a lot of practical use case, but still it should be congruent with pluck.)

johnnyshields:
Pretty cool, this will be useful!

spec/mongoid/contextual/mongo_spec.rb
Neilshweky: Correct, that's why this is a draft PR. The tests I have written so far are incomplete some of them are incorrect. I'm leaving this up as a note to myself for when I pick this up in the future.

PR #5212, opened 2022-04-11
MONGOID-4686 remove Document#to_a

Neilshweky:


p-mongo (approved):
This needs to be added to the release notes though.

p-mongo:
And we should probably have a deprecation notice in 7.x.

docs/release-notes/mongoid-8.0.txt
```diff
@@ -276,3 +276,9 @@ as well as ActiveRecord-compatible ``previously_new_record?`` and
 
   user.destroy
   user.previously_persisted?   # => true
+
+
+Removed Document#to_a method
+----------------------------
+
+The ``Document#to_a`` method has been removed in Mongoid 8.
```
Neilshweky, line 284: Does anything else need to be written here?

p-mongo: I think this is quite sufficient, thank you.

p-mongo (approved):

PR #5216, opened 2022-04-11
MONGOID-3267 Allow associations to be used inside defaults

Neilshweky:
MONGOID-3267

Optional reading: The latest change. After deferring the defaults we had the following case:
```
class Band
  include Mongoid::Document
  has_and_belongs_to_many :songs
  field :name
end

class Song
  include Mongoid::Document
  has_and_belongs_to_many :bands
  field :_id, type: String, overwrite: true, default: ->{ "#{name}-hello" }

  field :band_name, default: -> { bands.first.name if bands.first }
  field :favorite_song_id, default: -> { bands.first.song_ids.first if bands.first }
  field :name
end
band = Band.create(name: "Molejo")
song = band.songs.create(name: "Cilada")
``` 
The last command yielded the following order of operations:
1. Build a `song` object
2. Defer the callbacks and defaults
3. Push the `song._id` to the band's foreign key (`song_ids`). Note that `song._id` is being set in the default, therefore `nil` is being pushed here.
4. Set the association. Push song to the band's songs.
5. Apply the defaults
6. Call the pending callbacks
The problem here is evident. The `song._id` is not being set after it is being push to the band's foreign key. 

Solution:
Call song's defaults before pushing the song `_id` to the band's foreign key, i.e. move step 3 until after step 5. Don't change the time where we call the callbacks.

Problem: 
Now that the song._id is being pushed after the defaults are called, the song._id will not be available in the band's song_ids at the time that that song's defaults are called. I am **okay with this**. This is the current functionality in 7.4.x, so no functionality is lost. Also, if the user wants that `song_id`, they can access it using `_id` on the current song.

### DIFF
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -164,6 +164,11 @@ module Mongoid
       @pending_callbacks ||= [].to_set
     end
 
+    def pending_callbacks=(value)
+      @pending_callbacks = value
+    end
+
+
     # Run the pending callbacks. If the callback is :apply_defaults, we will apply
     # the defaults for this document. Otherwise, the callback is passed to the
     # run_callbacks function.
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -171,8 +171,8 @@ module Mongoid
     # @api private
     def run_pending_callbacks
       pending_callbacks.each do |cb|
-        if cb == :apply_defaults
-          self.apply_defaults
+        if [:apply_defaults, :apply_post_processed_defaults].include? cb
+          send(cb)
         else
           self.run_callbacks(cb, with_children: false)
         end
--- a/lib/mongoid/interceptable.rb
+++ b/lib/mongoid/interceptable.rb
@@ -164,6 +164,7 @@ module Mongoid
       @pending_callbacks ||= [].to_set
     end
 
+    # @api private
     def pending_callbacks=(value)
       @pending_callbacks = value
     end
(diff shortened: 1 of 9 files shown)


johnnyshields:
Nice!

lib/mongoid/interceptable.rb
```diff
@@ -161,7 +161,11 @@ def _mongoid_run_child_callbacks(kind, children: nil, &block)
     #
     # @api private
     def pending_callbacks
-      @pending_callbacks ||= []
+      @pending_callbacks ||= [].to_set
+    end
+
+    def pending_callbacks=(value)
```
comandeo, line 167: Should we add `@api private` here?

lib/mongoid/interceptable.rb
```diff
@@ -171,8 +175,8 @@ def pending_callbacks
     # @api private
     def run_pending_callbacks
       pending_callbacks.each do |cb|
-        if cb == :apply_defaults
-          self.apply_defaults
+        if [:apply_defaults, :apply_post_processed_defaults].include? cb
```
comandeo, line 178: I believe we use parenthesis for method calls.

comandeo (approved):
:+1:

lib/mongoid/interceptable.rb
Neilshweky: done

lib/mongoid/interceptable.rb
Neilshweky: done

p-mongo:
Is the belongs_to test case given in the ticket description (https://jira.mongodb.org/browse/MONGOID-3267) somewhere in this PR?

p-mongo:
> Now that the song._id is being pushed after the defaults are called, the song._id will not be available in the band's song_ids at the time that that song's defaults are called. 

Please add a test case for this with the current behavior.

Neilshweky:
@p-mongo I have added the tests that you requested

p-mongo (approved):

PR #5219, opened 2022-04-12
MONGOID-4953 Update the updated_at field when updating a HABTM association

Neilshweky:
MONGOID-4953

### Tl;dr 

**This PR has been modified to be only tests and docs.** See the Jira ticket for more details. 

The only differences between rails and mongoid is that rails does not update updated_at when updating the foreign keys on a HABTM relationship. 

Mongoid does update the updated_at, however I think this is correct behavior, since rails does not update the old document, it adds a new one, while mongoid updates the `_ids` array under the modified document. 

### Action Items for this PR

Make it such that `user.update(addresses: [address])` updates the updated_at field on the user. 

@p-mongo  Please see the following comment: 

The address' updated_at field  is not updated, since touching every element in a HABTM association would be inefficient. This update can be rolled into the add_to_set call that updates the `user_ids` on every Address but I think this should be its own ticket.

### Testing 
I am going to do a deep dive on what operations cause the updated_at to change on both rails and mongoid:
We have the following relationships:
```
User HABTM Addresses
User has_many Accounts
User has_one Pet
```
Assume we have objects that have been persisted to the database user, address, account, pet, but none of them have been assigned to an association. Note that HABTM relationships don't accept a `touch` option. 

### Mongoid
- touch: true
  - `user.update(addresses: [address])`: user **not** updated, address updated 
  - `user.update(accounts: [account])`: user updated, accounts updated
  - `user.update(pet: pet)`: user updated, pet updated
  - `account.update(user: user)`: user updated, account updated
  - `pet.update(user: user)`: user updated, pet updated

  - `user.update(address_ids: [address.id])`: user updated, address **not** updated
  - `user.update(account_ids: [account.id])`: user updated, account **not** updated
  - `account.update(user_id: user.id)`: user **not** updated, account updated
  - `pet.update(user_id: pet.id)`: user **not** updated, pet updated
- touch: false
  - `user.update(addresses: [address])`: no touch option, same as `touch: true`
  - `user.update(accounts: [account])`: user **not** updated, accounts updated
  - `user.update(pet: pet)`: user **not** updated, pet updated
  - `account.update(user: user)`: user **not** updated, account updated
  - `pet.update(user: user)`: user **not** updated, pet updated

  - `user.update(address_ids: [address.id])`: no touch option, same as `touch: true`
  - `user.update(account_ids: [account.id])`: user **not** updated, account **not** updated
  - `account.update(user_id: user.id)`: user **not** updated, account updated
  - `pet.update(user_id: pet.id)`: user **not** updated, pet updated

### Rails
- touch: true
  - `user.update(addresses: [address])`: user **not** updated, address **not** updated 
  - `user.update(accounts: [account])`: user updated, accounts updated
  - `user.update(pet: pet)`: user updated, pet updated
  - `account.update(user: user)`: user updated, account updated
  - `pet.update(user: user)`: user updated, pet updated

  - `user.update(address_ids: [address.id])`: user **not** updated, address **not** updated
  - `user.update(account_ids: [account.id])`: user updated, account **not** updated
  - `account.update(user_id: user.id)`: user **not** updated, account updated
  - `pet.update(user_id: pet.id)`: user **not** updated, pet updated
- touch: false
  - `user.update(addresses: [address])`: no touch option, same as `touch: true`
  - `user.update(accounts: [account])`: user **not** updated, accounts updated
  - `user.update(pet: pet)`: user **not** updated, pet updated
  - `account.update(user: user)`: user **not** updated, account updated
  - `pet.update(user: user)`: user **not** updated, pet updated

  - `user.update(address_ids: [address.id])`: no touch option, same as `touch: true`
  - `user.update(account_ids: [account.id])`: user **not** updated, account **not** updated
  - `account.update(user_id: user.id)`: user **not** updated, account updated
  - `pet.update(user_id: pet.id)`: user **not** updated, pet updated

### Differences
  - `user.update(addresses: [address])`: updating HABTM foreign keys don't trigger any updates on rails
  - `user.update(address_ids: [address.id])`: Again HABTM foreign keys don't trigger updates on rails
 
This ticket should only change HABTM to update the updated_at when updating the foreign keys.

### DIFF
--- a/lib/mongoid/timestamps/updated.rb
+++ b/lib/mongoid/timestamps/updated.rb
@@ -39,6 +39,16 @@ module Mongoid
       def able_to_set_updated_at?
         !frozen? && !timeless? && (new_record? || changed?)
       end
+
+      # Set the updated at for a change that's already happened. This is used
+      # to set the updated_at for updating the foreign key of a HABTM association.
+      #
+      # @api private
+      def set_updated_at_changed
+        if !frozen? && !timeless?
+          self.updated_at = Time.now.utc unless updated_at_changed?
+        end
+      end
     end
   end
 end
--- a/lib/mongoid/timestamps/updated.rb
+++ b/lib/mongoid/timestamps/updated.rb
@@ -39,16 +39,6 @@ module Mongoid
       def able_to_set_updated_at?
         !frozen? && !timeless? && (new_record? || changed?)
       end
-
-      # Set the updated at for a change that's already happened. This is used
-      # to set the updated_at for updating the foreign key of a HABTM association.
-      #
-      # @api private
-      def set_updated_at_changed
-        if !frozen? && !timeless?
-          self.updated_at = Time.now.utc unless updated_at_changed?
-        end
-      end
     end
   end
 end
--- a/spec/mongoid/timestamps_spec.rb
+++ b/spec/mongoid/timestamps_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "spec_helper"
+require_relative './timestamps_spec_models'
 
 describe Mongoid::Timestamps do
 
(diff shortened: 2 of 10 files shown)


johnnyshields:
If `touch: true` on the relationship, will that still update the timestamp? (If so, is there a test for it?)

Neilshweky:
> If `touch: true` on the relationship, will that still update the timestamp? (If so, is there a test for it?)

Good question... So HABTM relationships actually don't have a touch option in rails, and I don't think we have it either. For has_many and has_one relationships, you're right that is a good thing to consider. I threw this up very quickly as a proof of concept PR so it's not exactly finished.

johnnyshields:
@Neilshweky as you are aware, HABTM in ActiveRecord is fundamentally different because it uses a join table rather than array foreign keys. In AR one can add timestamps to the join table to understand when the relationship itself was created.

I suggest Mongoid HABTM should not rely too literally on ActiveRecord as its model.

Personally I always use HABTM with the `inverse_of: nil` option, and I expect many experienced Mongoid devs are doing the same. The HABTM double-sided updates don't use transactions, so in production scenario its potentially dangerous if one side of the update fails--I always find a small fraction of my millions of HABTM docs become out-of-sync. We could use transactions now on HABTM to remedy this, but one of my top wishlist items is a "belongs_to_many / has_many" relationship which works the same as a "belongs_to / has_many" but with an array foreign key on only one side. But I digress...

lib/mongoid/timestamps/updated.rb
```diff
@@ -39,6 +39,16 @@ def set_updated_at
...
         !frozen? && !timeless? && (new_record? || changed?)
       end
+
+      # Set the updated at for a change that's already happened. This is used
+      # to set the updated_at for updating the foreign key of a HABTM association.
+      #
+      # @api private
+      def set_updated_at_changed
+        if !frozen? && !timeless?
+          self.updated_at = Time.now.utc unless updated_at_changed?
```
p-mongo, line 49: This can stay the way it is but I filed https://jira.mongodb.org/browse/MONGOID-5295.

spec/mongoid/timestamps_spec.rb
```diff
@@ -107,4 +108,270 @@
...
+      context "when has_one association" do
+
+        context "when updating the association itself" do
+          before do
+            user.update(pet: pet)
+          end
+
+          it "updates the timestamps correctly" do
+            user.updated_at.should == update_time
+            pet.updated_at.should == update_time
```
p-mongo, line 220: I think all of these tests need to `reload` before inspecting `updated_at`.

lib/mongoid/timestamps/updated.rb
```diff
@@ -39,6 +39,16 @@ def set_updated_at
       def able_to_set_updated_at?
         !frozen? && !timeless? && (new_record? || changed?)
       end
+
+      # Set the updated at for a change that's already happened. This is used
```
p-mongo, line 43: When I read the phrase "already happened" I think that this means the persistence operation has completed, which means the `updated_at` write below is not persisted with the association operation which seems wrong.

I'm not sure of the exact language to clarify this but another sentence regarding this would be appreciated.

p-mongo (commented):
Could you find some place in documentation (associations page?) to add the note that HABTM updates `updated_at` unlike Rails due to the way Mongoid implements HABTM association storage?

spec/mongoid/timestamps_spec.rb
Neilshweky: Can you explain why?

My understanding of the reload is it only needs to be used if outside sources modify the data. If mongoid modifies the data, a reload shouldn't be necessary. Is this a correct understanding?

Additionally, rails does not require a reload when updating the `updated-at`

lib/mongoid/timestamps/updated.rb
Neilshweky: I mean, that is ultimately what is happening. The updated_at write is happening independently of the foreign key write (which is happening on the line right above it.)

I'm not sure how to tie them together...

Neilshweky: I'm referring to proxy.rb lines 32 and 33

Neilshweky:
> Could you find some place in documentation (associations page?) to add the note that HABTM updates `updated_at` unlike Rails due to the way Mongoid implements HABTM association storage?

done

p-mongo (approved):

PR #5221, opened 2022-04-13
MONGOID-5263: Run I18n fallback tests inline with other tests

johnnyshields:
**This PR only affects specs, it does not change any lib code**

MONGOID-5263: Run I18n fallback tests inline with other tests; do not require a separate test run. This ensures cross-platform/version coverage of I18n fallbacks tests, which we do not have today.

Note that this does NOT permanently install fallbacks; I18n backend is reverted to non-fallbacks after test run. This is based on the best practice of the I18n gem tests itself. See: https://github.com/ruby-i18n/i18n/blob/789d12ed265f23195fa018d7e7bf0c873c84f87e/test/backend/fallbacks_test.rb

By the way, the current specs are already [accidentally requiring I18n fallbacks when they should not](https://github.com/mongodb/mongoid/blob/master/spec/mongoid/fields/localized_spec.rb#L121). This PR also fixes that.

Lastly, there are an number of cases where various specs inadvertently mutate the I18n object state. I've fixed all of them by making a global `before` block which resets the I18n state (I18n.locale, etc.)


johnnyshields:
Failing tests are due to [MONGOID-5290](https://jira.mongodb.org/browse/MONGOID-5290) -- please decide what to do.

p-mongo (changes requested):

We do not want to toggle fallbacks on and off. We want to have separate configurations for having fallbacks present and absent, as is the case presently.

The existing fallbacks configuration explicitly executes 3 spec files. This configuration can be expanded to run the entire test suite with fallbacks enabled.

johnnyshields:
The very limited functionality which is actually affected by the I18n.fallbacks feature (i.e. `localize` option) has 100% test coverage of both fallbacks-on and fallbacks-off as a result of this PR?

Can you please clarify how adding full dedicated test runs for I18n.fallbacks further increases coverage beyond what I have here?

johnnyshields:
To further clarify:

> we know that fallbacks are either present or absent, without possibility of one test's changes leaking into other tests

This is also obvious in my tests here--it's in the RSpec test text (e.g. `context "with fallbacks"`).

> without possibility of one test's changes leaking into other tests

In my PR, there is **no possibility** of changes leaking to other tests (unlike the code before my PR, where it was leaking!) Please read my PR carefully.

> you can retain and expand an explicit list of test files that is being run

There's no need. It's done inline in the test suite, with `context` blocks, as we do with all other tests.

johnnyshields:
@p-mongo please review and respond to my comments above, and please also take a second look at the code in the PR. I am strongly of the opinion that my approach here to test I18n is the best way forward (and is what is what is followed Rails, etc.)

p-mongo:
I understand your opinion but we want separate test configurations with fallbacks on and off.

johnnyshields:
@p-mongo with all due politeness, as I strongly feel otherwise, I will leave this PR here as-is. This PR fixes existing issues and greatly improves test coverage without necessitating additional test runs. I urge you to reconsider and merge this PR in it's current form, following the best practices established by Rails and other libraries.

johnnyshields:
Requested change is done here: https://github.com/mongodb/mongoid/pull/5402.

Closing this PR.

johnnyshields:
Merged into [Mongoid Ultra](https://github.com/tablecheck/mongoid-ultra) 💯

PR #5222, opened 2022-04-13
MONGOID-5289 - Faster Regex for numeric mongoization + bug fixes

johnnyshields:
Fixes MONGOID-5289

This PR contains two performance changes and a bug fix to `__numeric__` method used in Mongoization:
- `.match?` is faster than `=~`
- 50% faster regex which is functionally equivalent to the old one.
- Fixes bugs where `foo.` and `foo.000` both match the current regex.
- Empty string behavior is preserved from existing code.

```ruby
Benchmark.bm do |bm|
  bm.report { x.times { '21329.2328'.match?(/(\A[-+]?[0-9]+\z)|(\.0+\z)|(\.\z)/) } }
  bm.report { x.times { '21329.2328'.match?(/\A[-+]?[0-9]*[0-9.]0*\z/) } }
end
#       user     system      total        real
#   0.531000   0.000000   0.531000 (  0.538161)
#   0.282000   0.000000   0.282000 (  0.270321)
```

Note that `.` inside `[ ]` doesn't need to be escaped.

### DIFF
--- a/lib/mongoid/criteria/queryable/extensions/numeric.rb
+++ b/lib/mongoid/criteria/queryable/extensions/numeric.rb
@@ -51,7 +51,7 @@ module Mongoid
             #
             # @return [ Object ] The converted number.
             def __numeric__(object)
-              object.to_s =~ /(\A[-+]?[0-9]+\z)|(\.0+\z)|(\.\z)/ ? object.to_i : Float(object)
+              object.to_s.match?(/\A[-+]?[0-9]*[0-9.]0*\z/) ? object.to_i : Float(object)
             end
 
             # Evolve the object to an integer.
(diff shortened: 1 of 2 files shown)


p-mongo:
The existing regexp is faster when captures are replaced with non-capturing groups.

````
irb(main):001:0> Benchmark.realtime { 1000.times { '1234.1324'.match?(/\A[-+]?[0-9]*[0-9.]0*\z/) } }
=> 0.000783171970397234
irb(main):002:0> Benchmark.realtime { 1000.times { '1234.1324'.match?( /(\A[-+]?[0-9]+\z)|(\.0+\z)|(\.\z)/) } }
=> 0.0021368409506976604
irb(main):003:0> Benchmark.realtime { 1000.times { '1234.1324'.match?( /(?:\A[-+]?[0-9]+\z)|(?:\.0+\z)|(?:\.\z)/) } }
=> 0.0003721569664776325
```

PR #5224, opened 2022-04-14
Fix MONGOID-5089 HABTM inconsistencies

Neilshweky:
MONGOID-5089 fixed the aliased_associations hash to include the foreign key field. For example if a `Bag` has_many `Tag`s, there would be an aliased association from `tag_ids => tag`. Before that PR the aliased associations hash would have `nil => tag`. However, this was only done for has_one and has_many associations. This PR fixes this for HABTM associations.

## Post fix results

Take the following models:
```
User HABTM Addresses 
User has_many Accounts
User has_one Pet
```
- HABTM: `User.aliased_associations` has `address_ids => addresses`, `Address.aliased_associations` has `user_ids => users`
- has_many: `User.aliased_associations` has `account_ids => accounts`, `Account.aliased_associations`  does not have `user_id`. `Account.aliased_fields` has `user => user_id`.
- has_one: `User.aliased_associations` does not have `pet_id`, `Pet.aliased_associations`  does not have `user_id`. `Pet.aliased_fields` has `user => user_id`.

Is this working as intended? Or should I add the `user_id => user` aliased_association for the belongs_to association as well?

### DIFF
--- a/spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
+++ b/spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
@@ -3789,4 +3789,11 @@ describe Mongoid::Association::Referenced::HasAndBelongsToMany::Proxy do
       expect(signature.contracts.first.signature_ids).to eq([signature.id])
     end
   end
+
+  context "when there is a foreign key in the aliased associations" do
+    it "has the correct aliases" do
+      expect(Dog.aliased_associations["breed_ids"]).to eq("breeds")
+      expect(Breed.aliased_associations["dog_ids"]).to eq("dogs")
+    end
+  end
 end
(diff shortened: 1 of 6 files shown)


p-mongo:
`aliased_associations` is not currently documented, can you add a docstring to it with your current idea of what it's supposed to contain/do?

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
```diff
@@ -3770,4 +3770,11 @@ class Distributor
       expect(p2.d_ids).to match_array([d2.id])
     end
   end
+
+  context "when there is a foreign key in the aliased associations" do
+    it "has the correct aliases" do
+      expect(Dog.aliased_associations["breed_ids"]).to eq("breeds")
+      expect(Breed.aliased_associations["dog_ids"]).to eq("dogs")
```
Neilshweky, line 3777: > In `Address.aliased_associations has user_ids => user` why is user singular?

@p-mongo if you look here, breeds and dogs are not singular? Where are you seeing user being singular?

p-mongo:
This PR gained a conflict with something in master it looks like.

Neilshweky:
@p-mongo  I fixed the merge conflicts

p-mongo (approved):

PR #5225, opened 2022-04-14
Fix MONGOID-4484 Clarify that association assignment saves documents


p-mongo:


Neilshweky (approved):
This is great!

docs/reference/associations.txt
```diff
@@ -1255,30 +1255,101 @@ The child may become orphaned if it is ordinarily only referenced via the parent
...
+  
+    # Saves the band and the album.
+    band.albums = [Album.new]
+
+- ``push``, ``<<``:
+
+  .. code-block:: ruby
+  
+    band.albums << Album.new
+    band.albums.push(Album.new)
```
johnnyshields, line 1351: Is there a reason why these methods don't follow autosave? I personally think this is pretty confusing, however, given the apps relying on this behavior, it might be hard to change.

Perhaps we should clarify that the workaround to not use autosave is to do:

```ruby
new_album = Album.new
band.album_ids << new_album._id
new_album.save!
```

I tend to do this workaround a lot in my apps.

docs/reference/associations.txt
```diff
@@ -1255,30 +1255,101 @@ The child may become orphaned if it is ordinarily only referenced via the parent
...
+  album.band_id
+  # => BSON::ObjectId('62576b4b53aefe178b65b8e3')
+  
+  album.band
+  # => #<Band _id: 62576b4b53aefe178b65b8e3, >
+
+The autosaving functionality is automatically added to an association when
+using ``accepts_nested_attributes_for``, so that the application does not
+need to track which associations were modified when processing a form
+submission.
```
johnnyshields, line 1330: Does `accepts_nested_attributes_for` coerce `autosave: true` in all cases? Or does the autosave only happen when passing in `:album_attributes` as a parameter to `new / assign_attributes`? I think we could clarify the language here.

docs/reference/associations.txt
```diff
@@ -1255,30 +1255,101 @@ The child may become orphaned if it is ordinarily only referenced via the parent
...
+
+Embedded associations always autosave, because they are stored as part of the
+parent document.
```
johnnyshields, line 1333: Assuming `Person embeds_many :addresses`, does this mean that doing:

```ruby
Person.addresses << Address.new
```

saves the parent document? If so, this is quite undesirable, as embedded docs are meant to behave like fields. In other words, you wouldn't want setting a regular field to autosave the parent. If the above is the case, in my opinion `autosave` *should* be supported on embedded relations, and its default value should be false.

johnnyshields:
Just a general comment here as a companion to my above comments. As an app implementer, I want to perform persistence **as few times as possible** in my code (e.g. for any controller action, etc.) `Autosave` is particularly toxic because it causes objects to be saved in an intermediate, i.e. half completed state, and it's not always obvious when Mongoid is choosing to save for me. For example:

```ruby
1    band = Band.new(some_attributes) # possibly saves here
2    band.name = "Pink Floyd"
3    album = Album.new("Dark Side of the Moon") 
4    band.albums << album     # saves both band and album here
5    band.rating = 10
6    album.tracks = 13
7    album.save!    # saves abum here
8    band.save!     # saves band here
```

Any saving done on lines 1 or 4 is both an unnecessary performance hit and "surprising" behavior. Moreover, suppose something fails before line 7/8, my `band/album` are in an indeterminate "partially persisted" state.

As an app implementer, I want to have full control over when objects persist (i.e. by calling `save!`), and catch errors/rollback etc. as well. I can understand how newbies might *think* autosaving is a convenience, but as one's app grows in complexity, it quickly becomes an anti-pattern one wishes one had never (unwittingly) adopted.

For these reason I set `autosave: false` on all relations, and use workarounds above (i.e. setting _ids rather than objects) to avoid any autosaving.

docs/reference/associations.txt
```diff
@@ -1255,30 +1255,101 @@ The child may become orphaned if it is ordinarily only referenced via the parent
...
+A non-exhaustive list of these operations is as follows:
+
+- Assignment to the association:
+
+  .. code-block:: ruby
+  
+    # Saves the band and the album.
+    band.albums = [Album.new]
+
+- ``push``, ``<<``:
```
johnnyshields, line 1346: It's also worth mentioning that `.clear` and `.destroy` on embedded relations also persist the parent.

As expressed in other comments, for me any such persistence is both surprising and unwanted. I ended up doing this hack, which after much trial and error magically worked to modify the doc without persistence.

```ruby
# Cart embeds many orders
cart[:orders] = []
cart.orders.reject! {|_| true }
```

p-mongo: Filed https://jira.mongodb.org/browse/MONGOID-5294 for this.

docs/reference/associations.txt
p-mongo: > Is there a reason why these methods don't honor the autosave setting? 

Don't know.

> Perhaps we should clarify that the workaround to not use autosave is to do:

You are welcome to use it but because you no longer have references to objects in the tree the object tree is essentially broken and for this reason I personally wouldn't put it in official documentation as a recommendation.

> the current behavior uses atomic push method, which doesn't call callbacks on the band in the above example?

If you think this is a bug please feel free to create a ticket.

docs/reference/associations.txt
p-mongo: > autosave should be supported on embedded relations, and its default value should be false.

This does not make sense to me.

docs/reference/associations.txt
p-mongo: Please feel free to investigate.

docs/reference/associations.txt
johnnyshields: > atomic push... If you think this is a bug please feel free to create a ticket.

Not a "bug", but rather your documentation in this PR uses the word "save" which is misleading if it's actually an atomic push. "save" implies that callbacks are fired, as per the `.save` method behavior.

p-mongo:
Hi @johnnyshields ,

You bring up some interesting points. Unfortunately they will be lost in this PR, the appropriate place for them is a jira ticket.

I'm not sure why some association operations immediately save documents, perhaps this is how Rails works or there was another reason for this behavior. Regardless, this is existing Mongoid behavior and changing it requires careful consideration.

Removing autosaving functionality is unlikely to happen. One of Mongoid's use cases is to be easier to use than driver, if you want 100% control over your operations we already provide the driver for that use case.

docs/reference/associations.txt
johnnyshields: > the object tree is essentially broken

Hmmm... this comment reveals an important philosophical difference between "Mongoid in theory" and "Mongoid in the real world". The object tree **will always be broken**, whether you want it to be or not. Because errors can happen at anytime (i.e. mid-controller action), you will always have some hanging document references.

Moreover, as per my comments below, autosave exacerbates the problem because you end up doing a document save in a partial state, i.e. the relation you are adding, but not the attributes. It's far safer to commit the entire document (all relations + all attributes) at once, otherwise you end up with partially-saved documents.

johnnyshields:
@p-mongo thanks for reading. I've raised MONGOID-5296 and MONGOID-5297 to track this.

> One of Mongoid's use cases is to be easier to use than driver

Autosaving makes things much harder, not easier. It would be far easier if `=` and `<<` assigns, and `.save` saves.

> If you want 100% control over your operations we already provide the driver for that use case.

This is not a practical option. One can't simply mix Mongoid and driver code for these sorts of cases; my app relies entirely on Mongoid's abstractions.

docs/reference/associations.txt
johnnyshields: > This does not make sense to me.

I mean to say "Embedded relations should not autosave. If we must preserve legacy behavior, support an `autosave` attribute with default `false`"

Raised https://jira.mongodb.org/browse/MONGOID-5297

docs/reference/associations.txt
johnnyshields: Shouldn't this investigation should be done before we write the docs? :)

Raised MONGOID-5298 for this.

johnnyshields: P.S. sorry if I seem pedantic here, many of the things in your documentation even as a 10+ year user of Mongoid I didn't understand exactly how they work (e.g. this specific point.) There be a lot of "magic" here...

docs/reference/associations.txt
johnnyshields: OK, I realize I now mistook the meaning of this line. Please disregard my comment, I thought this was related to "saving associations on assignment"

docs/reference/associations.txt
p-mongo: I don't follow. Even if there are some other issues with referential integrity/unspecified errors (?) this isn't a reason to intentionally break object tree.

docs/reference/associations.txt
p-mongo: To my knowledge everything in this PR is accurate.

docs/reference/associations.txt
Neilshweky: @johnnyshields if you want to create an object for an association without it being persisted to the database, you can use the `build*` methods:
```
For an _many relationship:
User.accounts.build

For a _one relationship:
User.build_pet
```

johnnyshields: @Neilshweky my core point here is I don't understand why:

```ruby
# This doesn't persist
User.accounts.build

# This does persist
account = Account.new
User.accounts << account
```

Most users will find this highly counter intuitive.

johnnyshields: @p-mongo my point is that implementers should be responsible for persisting their own objects (with `save!` method) and maintaining the object tree. Mongoid should not try to "enforce" the object tree by persisting when the user doesn't expect it.

PR #5226, opened 2022-04-15
MONGOID-4734 Test: Updating an association value to object instance when the instance has multiple inverses changes the wrong field

Neilshweky:
MONGOID-4734 

This is working as intended. I added a test to document it.


p-mongo (approved):
Could you duplicate the test so that both orders are tested (assignment to association that is defined first and to the one that is defined second).

PR #5227, opened 2022-04-15
MONGOID-4779 Test: Calculation is different by cache in memory for association document

Neilshweky:
MONGOID-4779

This is working as intended, I have added a test to document this.


johnnyshields:
I believe this is the same issue in principle which https://github.com/mongodb/mongoid/pull/5038 fixes. We should test #5038 on both embedded and referenced docs.

p-mongo (approved):
:+1:

PR #5233, opened 2022-04-19
MONGOID-5299 Undeprecate JRuby

comandeo:


.evergreen/config.yml
```diff
@@ -415,10 +415,10 @@ axes:
   - id: "jruby"
     display_name: JRuby Version
     values:
-      - id: "jruby-9.2"
-        display_name: jruby-9.2
+      - id: "jruby-9.3"
```
p-mongo, line 418: I woud test both 9.2 and 9.3 since it's the same effort as testing one version.

p-mongo (approved):

Neilshweky (approved):

.evergreen/config.yml
comandeo: Done

PR #5234, opened 2022-04-19
MONGOID-5302 Replace Hash#update_values with ActiveSupport's transform_values! 

johnnyshields:
Also removes Array#update_values which appears to be no longer used.


johnnyshields:
Ready to be merged.

p-mongo (approved):

PR #5235, opened 2022-04-20
MONGOID-5308: [WIP] `ActionController::Parameters` are not properly sanitized in all cases (*not a security issue)

johnnyshields:
### TODO

* [ ] https://github.com/mongodb/mongoid/pull/5417 should be merged first, then this needs rebase.
* [ ] Tests for following cases
   - Localized fields
   - Array and Hash fields, including deeply nested Arrays and Hashes
   - ``accepts_nested_attributes_for`` (Was this working before?)

---

`ActionController::Parameters` are not properly sanitized when instantiating embedded/referenced relations and localized fields.

**Fortunately** the current behavior does **NOT** cause a security issue because an MethodMissing error happens in all cases (i.e. Mongoid chokes on processing the `ActionController::Parameters` when it expects a `Hash`)

This PR standardizes and tests the behavior.

### DIFF
--- a/docs/release-notes/mongoid-8.1.txt
+++ b/docs/release-notes/mongoid-8.1.txt
@@ -116,3 +116,38 @@ from the ``_translations`` hash:
 
 See the section on :ref:`Localize :present Field Option <present-fields>` for
 more details on how these are used.
+
+
+Support for Assignment of Nested ``ActionController::Parameters``
+-----------------------------------------------------------------
+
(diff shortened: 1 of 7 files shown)


Neilshweky:
Closing this since the base feature is no longer requested.

johnnyshields:
This is fine to close. Can revisit if/when there is a use case to have it.

PR #5236, opened 2022-04-20
ActiveRecord parity: Prefer #sanitize_forbidden_attributes over #sanitize_for_mass_assignment

johnnyshields:
Converts the current single usage of #sanitize_for_mass_assignment to #sanitize_forbidden_attributes. **The two methods are aliased in ActiveModel**, and ActiveRecord uses the latter only and never the former. There are some gems, for example ProtectedAttributesContinued that work with the latter only.

Although this PR does NOT change any behavior, I've also added more robust test coverage in the affected area.


johnnyshields:
Ready to be merged.

PR #5237, opened 2022-04-20
MONGOID-5328: [WIP] Hash and localized fields assignment should work with `ActionController::Parameters` objects

johnnyshields:
### TODO

* [ ] Tests

---

This PR makes `Object`, `Hash` and `:localized` field assignment work with `ActionController::Parameters` objects, which are Hash-like but (as of Rails 5.x) do not inherit from Hash class. It is necessary to call `#to_h` on the `ActionController::Parameters`object, and `Hash.evolve` looks like the right place to do it.

As a side note, to err on the side of security, I think it is better to call `#to_h` rather than `#to_unsafe_h` on the `ActionController::Parameters`, which will filter out any unpermitted parameters. I need to test how this works in various scenarios (#permit!) etc.


Neilshweky:
@johnnyshields how do you feel about reworking this PR to take into account all of the new code that went into the mongoization epic?

johnnyshields:
@Neilshweky, sure, will be glad to do it.

johnnyshields:
(FYI will wait until dust settles on the mongoize/evolve refactoring before handling)

Neilshweky:
Is there an update on this?

johnnyshields:
@Neilshweky no update unfortunately, I've ended up not needing this for my app so I may abandon it for now.

Neilshweky:
I've opened #5417 to continue this work

johnnyshields:
Thank you! 👍

PR #5240, opened 2022-04-20
MONGOID-5323: Remove deprecated Mongoid::Errors::EagerLoad class

johnnyshields:
It is not used in the code at all.

Don't think this needs to be added to release notes (as no one is likely to be using it), but happy to add it if you'd like.


lib/mongoid/errors/eager_load.rb
```diff
@@ -1,25 +0,0 @@
...
-      Mongoid.deprecate(self, :initialize)
-
-      # Create the new eager load error.
-      #
-      # @example Create the new eager load error.
-      #   EagerLoad.new(:preferences)
-      #
-      # @param [ Symbol ] name The name of the association.
-      def initialize(name)
-        super(compose_message("eager_load", { name: name }))
```
p-mongo, line 21: This message needs to be removed from `lib/config/locales/en.yml` also.

p-mongo:
https://jira.mongodb.org/browse/MONGOID-5323

lib/mongoid/errors/eager_load.rb
johnnyshields: Good catch. Fixed.

p-mongo:
The require needs to be removed from `errors.rb`.

johnnyshields:
> The require needs to be removed from `errors.rb`.

Done

Neilshweky (approved):
This looks good to me, I am re-running the evergreen tests to make sure they work correctly.

Neilshweky:
@johnnyshields this one needs to be rebased as well

johnnyshields:
Rebase done.

Neilshweky (approved):
@p-mongo release notes look good to me

PR #5243, opened 2022-04-20
MONGOID-5304: Add Rubocop static code analyzer

johnnyshields:
This PR adds a static code analyzer (Rubocop) to the project, as required by MongoDB's [Technical and Organizational Security Measures](https://www.mongodb.com/technical-and-organizational-security-measures).

It is integrated with Github CI. To add it to Evergreen, simply run this shell command (only needs to be run once for the entire suite; not dependent on Ruby version, etc.) I've done a draft of Evergreen integration in this PR, but it may not work as I cannot run Evergreen myself.

```
bundle exec rake rubocop
```

---

### How to Use

This PR adds all cops in disabled state to begin with, i.e. "null" mode.

To begin Rubocop after merging this PR, first go to `.rubocop.yml` and enable some cops. You can do this by deleting a line such as:

```yaml
Performance:      # Delete this to enable
  Enabled: false  # performance-related cops
```

Then, generate a `.rubocop_todo.yml` file:

```
bundle exec rubocop --auto-gen-config --exclude-limit=1000
```

Your `.rubocop_todo.yml` will then contain all the cops you need to fix to become compliant. The way I do it is delete one item from the TODO, run autocorrect (`-A`), and then review the diff and raise a PR just for that fix. It's pretty easy to fix, and you'll be amazed at how quickly the code quality improves.

```
bundle exec rubocop -A
```


johnnyshields:
Ready to merge, please check Evergreen setup works.

johnnyshields:
@p-mongo you have marked this as `oleg-responded`, but I do not see any response from you in either Github or in JIRA (MONGOID-5304). I did receive a response from your colleague Bernie Hackett who said "I'm all for adding pre-commit static analysis of pure Ruby"

What is the specific next step here?

johnnyshields:
@p-mongo this is updated as per your comments in https://jira.mongodb.org/browse/MONGOID-5304

p-mongo (changes requested):
After https://github.com/mongodb/bson-ruby/pull/284#pullrequestreview-977317484 is fixed please make this PR same as the bson-ruby one.

johnnyshields:
@p-mongo Changes to the Rubocop file are done as requested (only Security in /lib dir)

Updating the Evergreen config is blocked by: https://jira.mongodb.org/browse/MONGOID-5396

Please either unblock me on that or handle adding Rubocop to Evergreen, thanks!

Rakefile
```diff
@@ -103,3 +103,6 @@ namespace :release do
     end
   end
 end
+
+require 'rubocop/rake_task'
+RuboCop::RakeTask.new
```
p-mongo, line 108: This isn't needed right?

johnnyshields: Not needed, we can run `rubocop` command instead of `rake rubocop`.

johnnyshields:
This is now ready for review.

comandeo (approved):

PR #5245, opened 2022-04-21
fixes errors where getlocal is unrecognized

alilland:
fixes the following issue
```
caught error of type NoMethodError in after callback inside Grape::Middleware::Formatter : undefined method `getlocal' for "2022-04-20T17:20:04.75-07:00":String
NoMethodError: undefined method `getlocal' for "2022-04-20T17:20:04.75-07:00":String
	/Users/aronlilland/.rvm/gems/ruby-2.6.5/gems/mongoid-6.4.8/lib/mongoid/extensions/time.rb:52:in `demongoize'
	/Users/aronlilland/.rvm/gems/ruby-2.6.5/gems/mongoid-6.4.8/lib/mongoid/fields/standard.rb:10:in `demongoize'
	/Users/aronlilland/.rvm/gems/ruby-2.6.5/gems/mongoid-6.4.8/lib/mongoid/fields.rb:418:in `block (2 levels) in create_field_getter'
```

### DIFF
--- a/lib/mongoid/extensions/time.rb
+++ b/lib/mongoid/extensions/time.rb
@@ -54,7 +54,9 @@ module Mongoid
         # @return [ Time ] The object as a date.
         def demongoize(object)
           return nil if object.blank?
-          object = object.getlocal unless Mongoid::Config.use_utc?
+          if object.respond_to?(:getlocal) && !Mongoid::Config.use_utc?
+            object = object.getlocal
+          end
           if Mongoid::Config.use_activesupport_time_zone?
             object = object.in_time_zone(Mongoid.time_zone)
           end


johnnyshields:
Please add a test case which proves this behavior. You can use RSpec mocks to hide the method.

johnnyshields:
@p-mongo it looks like the issue here is that the value persisted to the DB is a string field, but Mongoid has field `type: Time`. Time#demongoize is choking on the string value.

```
"2022-04-20T17:20:04.75-07:00".getlocal   #=> 
undefined method `getlocal' for "2022-04-20T17:20:04.75-07:00":String (NoMethodError)
```

To fix this, we should have Time/Date/etc `demongoize` be able to handle a string value, as per the standard that demongoize should take whatever value happens to be in the DB and "do the right thing" the extent that it is sensible to do so. Here `Time.parse(value) if value.is_a?(String)` would make sense.

johnnyshields:
Filed [MONGOID-5315](https://jira.mongodb.org/browse/MONGOID-5315) for this issue.

johnnyshields:
@p-mongo out of curiosity, what does the "oleg-responded" label mean exactly in terms of next steps? It seems to imply it's "waiting for PR author to respond", however, in this case, the all the necessary info is in [MONGOID-5315](https://jira.mongodb.org/browse/MONGOID-5315), including the code snippet which reproduces the issue.

johnnyshields:
@p-mongo it's exactly this posted above.

```
Time.demongoize("2022-04-20T17:20:04.75-07:00")
  #=> undefined method `getlocal' for "2022-04-20T17:20:04.75-07:00":String (NoMethodError)
```

As I've explained, somehow the user has a String value in the their DB that they later changed to a Time field in Mongoid.

alilland:
the string was accidentally saved by doing the following:

```ruby
class Event
  include Mongoid::Document
  field :timestamp, type: Time
end

event = Event.new
event.timestamp = Time.now.iso8601(3)
event.save
```

then on read the error was being thrown above

p-mongo:
Hi @alilland , thank you for the response.

There are  a number of outstanding mongoization/demongoization issues we are aware of, which are now grouped under https://jira.mongodb.org/browse/MONGOID-5332.

For your situation as I understand it, the first relevant issue is https://jira.mongodb.org/browse/MONGOID-5335 - Mongoid should have stored the value in a Time field as a Time instance, but didn't.

Subsequently, loading string values from the database into Time fields would be covered by https://jira.mongodb.org/browse/MONGOID-5315.

As I understand the fix proposed in this PR it would retain the string value during loading which is not desired (the value should be converted to a Time).

p-mongo:
Closing as per the above comment.

alilland:
It showed up when I had a string saved instead of a time object saved, yes -

Later on today I will describe how the string got saved which may uncover a second loophole that may want to be fixed as well, but it was definitely a user error that caused the Time object to be saved as a string

johnnyshields:
@alilland this will be fixed in the latest master, and I think the latest released version as well.

PR #5258, opened 2022-04-28
MONGOID-5327 document behavior change in release notes


p-mongo:
https://jira.mongodb.com/browse/MONGOID-5327

### DIFF
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -278,6 +278,37 @@ as well as ActiveRecord-compatible ``previously_new_record?`` and
   user.previously_persisted?   # => true
 
 
+Unknown Field Type Symbols/Strings Prohibited
+---------------------------------------------
+
+Mongoid 8 prohibits using symbols and strings as field types when these symbols
+and strings do not map to a known type. Previously such usage would create a
+field of type ``Object``.
+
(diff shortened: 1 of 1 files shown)


docs/release-notes/mongoid-8.0.txt
```diff
@@ -278,6 +278,37 @@ as well as ActiveRecord-compatible ``previously_new_record?`` and
...
+Mongoid 7 behavior:
+
+.. code-block:: ruby
+
+  class User
+    include Mongoid::Document
+
+    field :name, type: :bogus
+    # Equivalent to:
+    field :name
```
Neilshweky, line 308: Is this true? I thought if you leave off the type then it's a "dynamic" field, not that it has type Object.

johnnyshields: It's true.
https://github.com/mongodb/mongoid/blob/master/lib/mongoid/fields.rb#L748

Neilshweky (approved):

docs/release-notes/mongoid-8.0.txt
p-mongo: Dynamic fields are those that aren't explicitly defined as fields - https://www.mongodb.com/docs/mongoid/master/reference/fields/#dynamic-fields

PR #5223, opened 2022-04-13
MONGOID-5290: Require I18n gem version 1.1 or greater

johnnyshields:
See https://jira.mongodb.org/browse/MONGOID-5290

PR #5262, opened 2022-04-29
MONGOID-5337: Fix "taken" I18n message for UniquenessValidator

johnnyshields:
Fixes MONGOID-5337

UniquenessValidator does not work properly with I18n; the error message is always "stuck" in English.

The issue is due to [Rails I18n](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml#L132) and [ActiveRecord](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/locale/en.yml#L11) putting the error under key `errors.messages.taken`, while Mongoid uses `mongoid.errors.messages.taken`. 

Given the way ActiveModel generates errors, the presence of Mongoid's key prevents from loading the key in other locales. Changing Mongoid to use `errors.messages.taken` fixes the issue.


johnnyshields:
Please review.

johnnyshields:
Let's run Evergreen on this.

p-mongo (changes requested):
Hi @johnnyshields ,

All I see in this PR is a change of the message, which isn't what the PR is supposedly about.

To change the message, given the size of the diff involved, please create a separate PR.

For the bug fix, please add a test case (I assume it requires a translation into language other than English).

johnnyshields:
OK, will split into two PRs. Thanks.

johnnyshields:
Message text change split out into https://github.com/mongodb/mongoid/pull/5275. Please merge that one first.

johnnyshields:
Ready for review.

spec/mongoid/validatable/uniqueness_spec.rb
```diff
@@ -2472,4 +2472,32 @@ class SpanishActor < EuropeanActor
       }.to raise_error(Mongo::Error::OperationFailure)
     end
   end
+
+  describe "i18n" do
+
+    before do
+      Dictionary.validates_uniqueness_of :name
```
p-mongo, line 2479: Please use an existing model field or add a new field to a suitable model.

spec/mongoid/validatable/uniqueness_spec.rb
```diff
@@ -2472,4 +2472,32 @@ class SpanishActor < EuropeanActor
...
+    end
+
+    around do |example|
+      I18n.with_locale(:fr) { example.run }
+    end
+
+    before do
+      # Translation key location is as per rails-i18n gem.
+      # See: https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml
+      I18n.backend.store_translations(:fr, { errors: { messages: { taken: 'est déjà utilisé(e)' } } })
```
p-mongo, line 2493: This either needs to be configured globally for the entire test suite or reset if configured locally in this test.

spec/mongoid/validatable/uniqueness_spec.rb
```diff
@@ -2472,4 +2472,32 @@ class SpanishActor < EuropeanActor
...
+    end
+
+    it "uses the correct language translation" do
```
p-mongo, line 2496: This is missing the `context` describing the context, i.e. "when the taken validation message is translated"

lib/config/locales/en.yml
```diff
@@ -1,4 +1,9 @@
 en:
+  # For ActiveRecord parity.
```
p-mongo, line 2: The message should be here because this is where it's needed for Mongoid, not because this is where AR keeps it.

p-mongo (changes requested):

spec/mongoid/validatable/uniqueness_spec.rb
johnnyshields: @p-mongo I am following the convention used throughout this spec. Changes to the spec convention should be a separate ticket.

spec/mongoid/validatable/uniqueness_spec.rb
johnnyshields: Done.

spec/mongoid/validatable/uniqueness_spec.rb
johnnyshields: Done.

johnnyshields:
Responded to all comments.

spec/mongoid/validatable/uniqueness_spec.rb
p-mongo: This PR is adding wrong code that someone will have to fix in the future. Instead of doing that please add the correct code that won't need to be changed.

johnnyshields: In my philosophy, within a given code file it is better to be consistently wrong than inconsistently right.

However, I doubt I am going to be able to convince you of this here, so I've done as you've asked.

spec/mongoid/validatable/uniqueness_spec.rb
```diff
@@ -2472,4 +2472,31 @@ class SpanishActor < EuropeanActor
...
+      end
+
+      before do
+        # Translation key location is as per rails-i18n gem.
+        # See: https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml
+        I18n.backend.store_translations(:fr, { errors: { messages: { taken: 'est déjà utilisé(e)' } } })
+      end
+
+      after do
+        I18n.backend.instance_variable_get(:@translations).delete(:fr)
```
p-mongo, line 2491: I'm not thrilled with this, @Neilshweky what are your thoughts?

Neilshweky: I agree... As per removing a translation I think this would be a little better:
```suggestion
        I18n.backend.translations.delete(:fr)
```

johnnyshields:
Hi, looks like you've tweaked my spec to your liking. Let's merge?

spec/mongoid/validatable/uniqueness_spec.rb
```diff
@@ -2472,4 +2472,31 @@ class SpanishActor < EuropeanActor
...
+      end
+
+      before do
+        # Translation key location is as per rails-i18n gem.
+        # See: https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/en.yml
+        I18n.backend.store_translations(:fr, { errors: { messages: { taken: 'est déjà utilisé(e)' } } })
+      end
+
+      after do
+        I18n.backend.translations.delete(:fr)
```
p-mongo, line 2491: This needs to `send(:translations)` for i18n 1.0 where `translations` is not public.

johnnyshields: Done. Please also accept [my other PR]
(https://github.com/mongodb/mongoid/pull/5223) which sets the minimum I18n version to 1.1, it will make everyone's life much easier.

johnnyshields:
Let's merge?

p-mongo (approved):

Neilshweky (approved):

PR #5263, opened 2022-05-03
MONGOID-5330 Operator methods overwrite previous conditions

Neilshweky:
This PR changes certain functions to combine using $and instead of replacement. For example, given the following query: 
```
Person.gt(age: 2).gt(age: 1)
```
We currently get:
```
{"age"=>{"$gt"=>1}}
```
And we should expect:
```
{"age"=>{"$gt"=>2}, "$and"=>[{"age"=>{"$gt"=>1}}]}
```
Now, I have enumerated the functions below that I believe require this change, and I have also added a list of functions that I believe should not get this change. 

Functions to fix:
- gt
- gte
- lt
- lte
- ne
- near 
- near_sphere 
- eq
- mod
- elem_match

Note that these functions all currently use the ``__override__`` strategy when adding to the criteria.

No longer doing: 

Previously, I thought to not make this change for `eq` and `mod` for the following reason:

> The logic behind the list of functions to _not_ fix is that, these functions don't make sense to combine with an `and`. For example, combining two `eq` clauses together with an `and` will almost always cause no results to be returned. (I'm trying to think of a situation were two `eq`s combined with an `and` return results, but I can't think of one).

Johnny has provided a use case for anding eq and mod below, so I have included them.

### DIFF
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -258,7 +258,7 @@ for details on driver options.
       # feature flag to true will cause those operators to be combined using an
       # and. Setting this feature flag to false will cause the later chained
       # operators to overwrite the earlier ones. (default: false)
-      and_chained_operators: false
+      #and_chained_operators: true
 
       # Application name that is printed to the MongoDB logs upon establishing
       # a connection in server versions 3.4 or greater. Note that the name
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -254,12 +254,6 @@ for details on driver options.
 
     # Configure Mongoid-specific options. (optional)
     options:
-      # When chaining the same operators that use the same field, setting this
-      # feature flag to true will cause those operators to be combined using an
-      # and. Setting this feature flag to false will cause the later chained
-      # operators to overwrite the earlier ones. (default: false)
-      #and_chained_operators: true
-
       # Application name that is printed to the MongoDB logs upon establishing
       # a connection in server versions 3.4 or greater. Note that the name
       # cannot exceed 128 bytes in length. It is also used as the database name
@@ -400,6 +394,12 @@ for details on driver options.
       # is used, the return value will be the hexadecimal ObjectId string only.
       # (default: false)
       #object_id_as_json_oid: true
+
+      # When chaining the same operators that use the same field, setting this
+      # feature flag to false will cause those operators to be combined using an
+      # and. Setting this feature flag to true will cause the later chained
+      # operators to overwrite the earlier ones. (default: false)
+      #overwrite_chained_operators: false
 
       # Preload all models in development, needed when models use
       # inheritance. (default: false)
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -254,6 +254,12 @@ for details on driver options.
 
     # Configure Mongoid-specific options. (optional)
     options:
+      # When chaining the same operators that use the same field, setting this
+      # feature flag to true will cause those operators to be combined using an
+      # and. Setting this feature flag to false will cause the later chained
+      # operators to overwrite the earlier ones. (default: false)
+      and_chained_operators: false
+
       # Application name that is printed to the MongoDB logs upon establishing
       # a connection in server versions 3.4 or greater. Note that the name
       # cannot exceed 128 bytes in length. It is also used as the database name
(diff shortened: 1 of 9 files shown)


johnnyshields:
Overwriting conditions when chaining criteria methods is unexpected by users and dangerous. It is better to combine them with `$and` like you have in the example. That being said, any change to query logic needs a feature flag.

johnnyshields:
Should work for `$eq` and `$mod` as well.

```ruby
class MyModel
  include Mongoid::Document

  field :num, type: Integer
  field :ary, type: Array
end

m1 = MyModel.create!(ary: [1, 2])
m2 = MyModel.create!(ary: [2, 3, 4])
MyModel.where('$and' => [{ary: {'$eq' => 2}}, {ary: {'$eq' => 4}}]).count #=> 1

m3 = MyModel.create!(num: 18)
MyModel.where('$and' => [{num: {'$mod' => [6, 0]}}, {num: {'$mod' => [10, 8]}}]).count #=> 1
```

johnnyshields:
Thank you for including `$eq` and `$mod` 🙇‍♂️ 

One more note which is certainly _out-of-scope_ for this PR. Syntax-wise, I suspect most Mongoid users would prefer the generated query in this form:

```ruby
{"$and"=>[{"age"=>{"$gt"=>2}}, {"age"=>{"$gt"=>1}}]}
```

instead of this:

```ruby
{"age"=>{"$gt"=>2}, "$and"=>[{"age"=>{"$gt"=>1}}]}
```

The latter feels ambiguous and non-hierarchical; reading left-to-right not clear that the two `"age"` conditions are `$and`'ed together. Although it "works" in terms of passing specs, my hunch is that this non-hierarchical structure is the source of a good bit of [counter-intuitive query behavior](https://jira.mongodb.org/browse/MONGOID-5272).

johnnyshields:
Also, how about $in, $nin, $all, $elemMatch, [et cetera](https://www.mongodb.com/docs/manual/reference/operator/query/)?

p-mongo:
I think all operators should work the same for the scope of this ticket. Mongoid does have https://www.mongodb.com/docs/mongoid/master/reference/queries/#merge-strategies which can be considered but the default behavior should be that query conditions are additive and all operators behave the same.

Neilshweky:
> I think all operators should work the same for the scope of this ticket. Mongoid does have https://www.mongodb.com/docs/mongoid/master/reference/queries/#merge-strategies which can be considered but the default behavior should be that query conditions are additive and all operators behave the same.

I think that's what this ticket does! Can you point to what specifically you are referring to with the first part of this comment?

lib/mongoid/criteria/queryable/mergeable.rb
```diff
@@ -52,6 +52,27 @@ def reset_strategies!
           self
         end
 
+        # Merge criteria with operators using the and function.
```
p-mongo, line 55:
```suggestion
        # Merge criteria with operators using the and method.
```

lib/mongoid/criteria/queryable/mergeable.rb
```diff
@@ -52,6 +52,27 @@ def reset_strategies!
           self
         end
 
+        # Merge criteria with operators using the and method.
```
p-mongo, line 55:
```suggestion
        # Merge criteria with operators using the and operator.
```

lib/mongoid/criteria/queryable/mergeable.rb
```diff
@@ -52,6 +52,27 @@ def reset_strategies!
           self
         end
 
+        # Merge criteria with operators using the and operator.
+        #
+        # @param [ Hash ] The criterion to add to the criteria.
```
p-mongo, line 57:
```suggestion
        # @param [ Hash ] criterion The criterion to add to the criteria.
```

p-mongo (approved):

docs/reference/configuration.txt
```diff
@@ -254,6 +254,12 @@ for details on driver options.
 
     # Configure Mongoid-specific options. (optional)
     options:
+      # When chaining the same operators that use the same field, setting this
+      # feature flag to true will cause those operators to be combined using an
+      # and. Setting this feature flag to false will cause the later chained
+      # operators to overwrite the earlier ones. (default: false)
+      #and_chained_operators: true
```
p-mongo, line 261: I would rename the option to `overwrite_chained_operators` and invert it to make it be in line with how other options behave (the false value is default behavior of 8.0+).

p-mongo (approved):

p-mongo:
Actually if you don't mind I would much prefer `overwrite_chained_operators`.

p-mongo (approved):
Thank you!

comandeo (approved):

PR #5267, opened 2022-05-04
MONGOID-5314: Support aliases on partial_filter_expression

johnnyshields:


lib/mongoid/association/relatable.rb
```diff
@@ -411,7 +411,7 @@ def default_inverse
       def namespace_hierarchy(mod)
         parent = Object
         hier = [parent]
-        mod.name.split('::').each do |part|
+        mod.name&.split('::')&.each do |part|
```
johnnyshields, line 414: @Neilshweky this logic breaks with anonymous classes (e.g. as used in specs.) Adding this safe navigation `&.` check seems to resolve the issue.

lib/mongoid/indexable/specification.rb
```diff
@@ -46,7 +46,7 @@ def initialize(klass, key, opts = nil)
         @klass = klass
         @key = normalize_key(key)
         @fields = @key.keys
-        @options = normalize_options(options.dup)
+        @options = normalize_options(options)
```
johnnyshields, line 49: `.dup` not needed here, since `normalize_options` does not modify the arg.

lib/mongoid/indexable/specification.rb
```diff
@@ -93,9 +93,16 @@ def normalize_key(key)
...
         options = {}
+
         opts.each_pair do |option, value|
+          option = option.to_sym
           options[MAPPINGS[option] || option] = value
         end
+
+        if options[:partial_filter_expression]
+          options[:partial_filter_expression] = normalize_key(options[:partial_filter_expression])
+        end
```
johnnyshields, line 104: This is the fix.

p-mongo (changes requested):

As I noted in https://jira.mongodb.org/browse/MONGOID-5314?focusedCommentId=4526185&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4526185, the behavior implemented by the server is more complex than just top-level operators. Specifically, there should be test cases in this PR for:

- Top level key-value mappings without operators where the value isn't a hash
- Top level key-value mappings where the value is a hash
- Operators like you have now
- Operators nested under `$and`
- Non-operators nested under `$and`

Please carefully review https://www.mongodb.com/docs/manual/reference/command/createIndexes/ for other cases that might need storage field name expansion. 

This PR should also add documentation stating:

- That Mongoid expands storage field names in `:partial_index_expression`
- That this expansion is performed according to behavior of server 6.0, and future server versions may change how they interpret the partial index expressions as well as what expressions are allowed, and Mongoid may not be able to follow those changes
-

p-mongo:
Please also extract unrelated changes to their own PRs.

johnnyshields:
Good call, looks like we also need to add for:
- weights
- wildcardProjection

These are both flat key-value structures.

p-mongo:
Please also refer to https://www.mongodb.com/docs/v6.0/reference/command/createIndexes/ rather than https://www.mongodb.com/docs/manual/reference/command/createIndexes/ which is for 5.0.

johnnyshields:
This is still in my queue. Will do in 1-2 weeks.

johnnyshields:
@p-mongo this is now ready for review.

Note that I have added `wildcard_projection` as an index option which was missing from Mongoid and is needed for this PR.

Re: your comments on what to test:

1. Top level key-value mappings without operators where the value isn't a hash
   - These are covered by the existing tests; the code changes in this PR doesn't affect these.

2. Top level key-value mappings where the value is a hash
   - There are three of these: `partial_filter_expression, weights, wildcard_projection`. All are covered in this PR.

3. Operators like you have now
   - This is covered.

4. Operators nested under $and
   - This is covered. You'll see in the code that $and is a special case.

5. Non-operators nested under $and
   - AFAIK all expressions under $and must have a nested operator. Can you give an example of this?

johnnyshields:
@p-mongo this is awaiting further response from you. Please kindly remove `oleg-responded` label.

johnnyshields:
@p-mongo I have addressed all your comments in this PR and am waiting on further review from you.

johnnyshields:
Ready for final review

johnnyshields:
Please mark as `oleg-todo`

lib/mongoid/indexable/specification.rb
```diff
@@ -93,9 +93,18 @@ def normalize_key(key)
...
         opts.each_pair do |option, value|
+          option = option.to_sym
           options[MAPPINGS[option] || option] = value
         end
+
+        %i[partial_filter_expression weights wildcard_projection].each do |key|
+          options[key] = normalize_key(options[key]) if options[key].is_a?(Hash)
+        end
+
+        options.dig(:partial_filter_expression, :'$and')&.map! { |expr| normalize_key(expr) }
```
p-mongo, line 106: This looks to me like it would mutate user input.

p-mongo (changes requested):
For documenting the current behavior of Mongoid, there should be tests for :+1: 

top-level operator other than `$and` (e.g. `$or`)
`$and` nested under another `$and`

I am also not seeing the user documentation requested earlier.

lib/mongoid/indexable/specification.rb
johnnyshields: You're right. I've committed a fix.

johnnyshields:
@p-mongo fix + documentation added as requested. This is ready for final review.

docs/reference/indexes.txt
```diff
@@ -82,6 +82,18 @@ Indexes can be scoped to a specific database:
       index({ ssn: 1 }, { database: "users", unique: true, background: true })
     end
 
+You may use aliased field names in index definitions. Field aliases
+will also be resolved on the following options: ``partial_filter_expression``,
```
p-mongo, line 86: The example is using storage field name rather than a field alias right?

johnnyshields: In the example:

```ruby
      field :a, as: :age
      index({ age: 1 }, { partial_filter_expression: { age: { '$gte' => 20 } })
```

`:a` is the storage field name, `:age` is the alias, so the example illustrates how to use the alias in the Ruby index definition.

p-mongo: Aliases are defined in https://www.mongodb.com/docs/mongoid/master/reference/fields/#field-aliases. Are they supported in the implementation added by this PR?

p-mongo:
Where are the $and/$and and $or tests?

docs/reference/indexes.txt
johnnyshields: Yes, because they are returned by `database_field_name`

```ruby
class Band
  include Mongoid::Document
  field :name, type: String
  alias_attribute :n, :name
end

Band.database_field_name(:n) #=> "name"
```

spec/mongoid/indexable_spec.rb
```diff
@@ -591,22 +608,221 @@ def self.hereditary?
...
+        let(:expected) do
+          {
+            '$and': [
+              { 'a': { '$gte' => 0 } },
+              { 'a': { '$type' => 16 } }
+            ]
+          }
+        end
+
+        it "resolves alias on $and option" do
```
johnnyshields, line 685: Here's the test for `$and`

johnnyshields:
The test for $and is on spec/mongoid/indexable_spec.rb line 685 (see above).

Partial index expressions [do not support an $or option](https://www.mongodb.com/docs/manual/core/index-partial/)

docs/reference/indexes.txt
p-mongo: Storage field names and field aliases are separate features, if both are supported they need to be both mentioned in the docs and there need to be test cases for each of them.

p-mongo:
They don't, the tests in Mongoid are still required so that 1) we know what Mongoid would do in those situations and 2) if this functionality is implemented in the server in the future, Mongoid is hopefully compatible with it.

johnnyshields:
@p-mongo we're now entering the territory of preemptively supporting things that the server doesn't actually support today. I'm trying to guess what you are expecting here. Here's my best guess:

- `:partial_filter_expression` should recursively support `$and`, `$or`, and `$nor` (N-levels deep).
- Other options are not expected to support these, and are only one level.

I've done a commit which does this. Note that, given the need for recursion, the code now becomes much cleaner if we `.deep_dup` the options hash at the beginning, then use bang `!` methods to mutate the content of the dupe thereafter.

Please kindly review code and confirm my guess is correct. If so, I will complete the specs and finish the PR.

docs/reference/indexes.txt
johnnyshields: I'll do it, but bear in mind we have 1000s of tests for alias-related features today, and *literally none* of them test the case of `alias_attribute` being different than `:as`. The goal post seems to constantly be moving here.

johnnyshields:
@p-mongo please confirm the approach in this PR is correct, thanks!

p-mongo:
@johnnyshields The request was for *adding tests* for the cases I listed. The $and/$and and $or cases *are not going to be officially supported* by Mongoid, because the server does not implement them. The tests are required so that we can verify that Mongoid is behaving sensibly in these situations, in the event that they are supported by the server in the future. Hopefully this clarifies the difference between having tests for X and supporting X.

With respect to the implementation, I think if you are going to handle 3 operators when the server permits 1 you should handle all of them, i.e. if the key starts with `$` assume it is an operator and behave the same as in the $and case. Do you see any issues with this approach?

johnnyshields:
@p-mongo sorry, I misunderstood you earlier comment about just adding tests.

Please kindly clarify what you'd like me to do here:

A) Go back to the earlier implementation and match *exactly what the server supports* today (i.e. one-level of `$and` operator) + add tests for the cases you mentioned, i.e. verifying that Mongoid does *not* normalize those cases.

B) Do something like the current approach, and normalize any operator beginning with `$` N-levels deep + add tests.

Either way is fine by me.

p-mongo:
Which behavior makes more sense to you?

johnnyshields:
@p-mongo Hmmm I really don't have a strong preference.

I would go with (B) but limit the operator to only `$and/$or/$nor` rather than all `$` because:
(a) potential dollar-sign named field shenanigans
(b) MongoDB could introduce some new `$foo` operator that takes a non-field hash arg which shouldn't be normalized

I think this would be future-proof without being *too far* ahead...

But please make the final call and I will do as you prefer here.

johnnyshields:
@p-mongo greenlight to proceed as per above? Would like to finish this PR.

lib/mongoid/indexable/specification.rb
```diff
@@ -63,39 +63,73 @@ def name
...
+        %i[weights wildcard_projection].each do |key|
+          normalize_aliases!(options[key])
+        end
+
+        recursive_normalize_conditionals!(options[:partial_filter_expression])
+
+        options
+      end
+
+      # TODO: DOCS.
```
Neilshweky, line 110: Reminder:

lib/mongoid/indexable/specification.rb
```diff
@@ -63,39 +63,73 @@ def name
...
-          options[MAPPINGS[option] || option] = value
+      def normalize_options!(options)
+
+        options.transform_keys! do |option|
+          option = option.to_sym
+          MAPPINGS[option] || option
+        end
+
+        %i[weights wildcard_projection].each do |key|
+          normalize_aliases!(options[key])
```
Neilshweky, line 102: Can you explain why `recursuve_normalize_conditionals` is not called here?

johnnyshields: Because these options do not support any nesting; their field hash is required to be one level.

lib/mongoid/indexable/specification.rb
johnnyshields: Yep, waiting on Oleg's reply to this comment: https://github.com/mongodb/mongoid/pull/5267#issuecomment-1207218570

lib/mongoid/indexable/specification.rb
```diff
@@ -63,39 +63,73 @@ def name
...
       #
-      # @param [ Hash ] opts The index options.
+      # @param [ Hash ] options The index options.
       #
       # @return [ Hash ] The normalized options.
-      def normalize_options(opts)
-        options = {}
-        opts.each_pair do |option, value|
-          options[MAPPINGS[option] || option] = value
+      def normalize_options!(options)
```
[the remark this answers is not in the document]
johnnyshields: The reason to do modifications in place is that `recursive_normalize_conditionals!` becomes a lot cleaner; we no longer have to worry about returning values in the recursion tree. Given this design decision, I thought it would be more consistent to use in-place modification everywhere in this class and do the necessary `dup` call at the beginning. Ultimately it makes for more terse code.

lib/mongoid/indexable/specification.rb
```diff
@@ -63,39 +63,73 @@ def name
 
       private
 
-      # Normalize the spec, in case aliased fields are provided.
+      # Normalize the spec in place, in case aliased fields are provided.
```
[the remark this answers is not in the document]
johnnyshields: See https://github.com/mongodb/mongoid/pull/5267#discussion_r958781671. Ultimately nothing "has to be done in place", it's certainly possible to change the code so that returns and assignment is used everywhere, but it's gonna be 3x the lines of code, less readable, and more brittle IMHO.

Neilshweky:
> @p-mongo Hmmm I really don't have a strong preference.
> 
> I would go with (B) but limit the operator to only `$and/$or/$nor` rather than all `$` because:
> 
> 1. potential dollar-sign named field shenanigans
> 2. MongoDB could introduce some new `$foo` operator that takes a non-field hash arg which shouldn't be normalized
> 
> I think this would be future-proof without making too many assumptions... But please make the final call and I will do as you prefer here.

Hey @johnnyshields... Just catching up on this PR. I agree with @p-mongo in that we should support recursing over arbitrary operators (i.e. anything that starts with a `$`). 

I personally am not worried about your latter concern, and in terms of your first concern, our support for dollar-field names is limited, and I am okay with sacrificing the ability to expand the aliases of dollar fields in this case. 

To me, the most effective way of supporting server functionality as time goes on, is to support arbitrary operators.

lib/mongoid/indexable/specification.rb
Neilshweky: I see that. In general I think I would prefer a more general solution, i.e. one that is not specific to current server syntax. I am okay with specifying `weights`, `wildcard_projection` and `partial_filter_expression` but I think in general it would be a good idea to automatically recurse N-levels deep in any hash, in case the server syntax changes. This also would make for a slimmer and more easily readable recursive implementation.

johnnyshields:
@Neilshweky OK, I will change the PR to recurse N-levels deep for any `$__` key and for all 3 options `partial_filter_expression weights wildcard_projection`. Works for me.

lib/mongoid/indexable/specification.rb
Neilshweky: The two things I think that should change in favor of a more general solution:
1. always recurse N-levels deep
2. don't specify `$and`, `$or`, `$nor`, take any arbitrary operator (string beginning with $) into account

lib/mongoid/indexable/specification.rb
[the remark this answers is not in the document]
Neilshweky: Personally, I prefer things not to be done in place if it can be avoided, (side note: I think if you simplify the recursion as I have pointed out below, I have a feeling this can be done with a simple `reduce`), however I'm not gonna hold up this PR just because of that.

johnnyshields:
@Neilshweky changes done; this is ready for final review.

spec/mongoid/indexable/specification_spec.rb
```diff
@@ -80,25 +80,79 @@
...
+      end
+
+      it "normalizes the options" do
+        expect(spec.options).to eq(background: true, unique: true)
+      end
     end
 
-    it "normalizes the options" do
-      expect(spec.options).to eq(background: true, unique: true)
+    context "with aliased field options" do
```
Neilshweky, line 107: Can you add a _bunch_ more tests than just the one you have here? Some thoughts:

1. What happens when it's not under one of the three keys
2. What happens when we have an arbitrary $
3. What happens when giving a hash of arbitrary depth

Try to also make separate tests with the different keys so that we're not testing it all at once (you can keep the one with all 3 just add more) this way if any of them break we'll know which easily...

lib/mongoid/indexable/specification.rb
```diff
@@ -63,39 +63,77 @@ def name
...
+
+        %i[partial_filter_expression weights wildcard_projection].each do |key|
+          recursive_normalize_conditionals!(options[key])
+        end
+
+        options
+      end
+
+      # Recursively normalizes the nested elements of an options hash in-place,
+      # to account for conditional operators such as $and, $or, and $nor.
```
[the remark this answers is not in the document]
johnnyshields: How about:

```
      # Recursively normalizes the nested elements of an options hash in-place,
      # to account for $and operator, and other potential $-prefixed operators
      # which may be supported by MongoDB in the future.
```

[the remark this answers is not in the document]
Neilshweky: Looks good to me!

spec/mongoid/indexable/specification_spec.rb
johnnyshields: @Neilshweky I've taken a second look at this. When considering this spec plus the many specs in spec/mongoid/indexable_spec.rb (below) which are testing the same thing, I really don't think there are more cases to add, I've covered all the canonical "real" cases and a good cross section of the possible non-sense.

If you can give me a list of specific extra scenarios you have in mind--and please double-check my specs don't cover them already--I'll be happy to add them. Otherwise, let's mark this done.

johnnyshields:
@p-mongo this is ready for final review. I've added specs covering aliases via both `:as` and `alias_attribute`.

docs/reference/indexes.txt
johnnyshields: Specs for `alias_attribute` cases are added.

johnnyshields:
@p-mongo kind reminder on this. It's marked "Oleg responded" but it's actually waiting for your final review/response.

spec/mongoid/indexable/specification_spec.rb
```diff
@@ -154,6 +156,68 @@
...
+        expect(spec.klass).to eq(Band)
+      end
+
+      it "normalizes the key" do
+        expect(spec.key).to eq(name: 1, title: 1, y: -1)
+      end
+
+      it "normalizes the options" do
+        expect(spec.options).to eq(partial_filter_expression: {
+          '$foo': { y: { years: { '$lte' => 50 } } },
```
Neilshweky, line 217: Can you explain to me why you've only decided to expand the alias of one level below the operator?

johnnyshields: Because there is no real-world syntax in MongoDB which allows fields to be used this way.

I'd like to again ask that the focus here be on *real-world behavior*, and not nonsensical hypotheticals. This PR covers every real-world case and more of the nonsense than it should, in my opinion.

Let's merge.

johnnyshields:
@Neilshweky let's merge this. This fixes a bug that bit me in a fairly severe way in production, making my partial filter index that I "assumed to work" but actually did nothing, and causing bad query performance a result. The tests are more than adequate to verify the change.

johnnyshields:
@alexbevi would like to ask that this be reviewed soon, the bug here can lead to some pretty nasty unexpected behavior that has affected me in prod. Previous Mongo team has already reviewed extensively it so hopefully it is a straight-shot to merge.

PR #5268, opened 2022-05-09
MONGOID-5344 Using concat on has_and_belongs_to_many ids default value is not working the first time

Neilshweky:


spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
```diff
@@ -3796,4 +3796,31 @@ class Distributor
...
+
+  # This test is for MONGOID-5344 which tests that the initial call to
+  # signature_ids refers to the same array as subsequent calls to signature_ids.
+  # Prior to the change in that ticket, this test broke because the array
+  # returned from write_attribute (which is triggered the first time the
+  # foreign key array is referenced, to set the default), refers to a different
+  # array to the one stored in the attributes hash. This happened because,
+  # when retrieving a document from the database, the attributes hash is actually
+  # a BSON::Document, which applies a transformation to the array before
+  # storing it.
```
Neilshweky, line 3808: Look here for an explanation of the fix in this ticket.

The fix here is to convert the attributes BSON::Document to a Hash when retrieving from the db.

PR #5269, opened 2022-05-09
MONGOID-5336: User-defined symbol field aliases

johnnyshields:


docs/reference/fields.txt
```diff
@@ -897,6 +897,17 @@ can use in our model class as follows:
...
   end
 
+You may optionally declare a symbol mapping for the new field type in an initializer:
+
+.. code-block:: ruby
+
+  # in /config/initializers/mongoid_custom_fields.rb
+
+  Mongoid::Fields.configure do
+    type :point, Point
```
p-mongo, line 907: `type` is ambiguous here, for example if we wanted to require the allowed set of types to be defined before use that definition could potentially need `type`. What do you think about `alias` instead?

johnnyshields: `alias` is a ruby keyword, so should avoid it. how about `define_type` or `field_type`?

p-mongo: You're right, I overlooked that. This is not a type definition but an alias definition, so what about `type_alias`?

johnnyshields: The word `alias` implies like-for-like, e.g. symbol-for-symbol `type_alias :bool, :boolean`.

I think `define_type` is best.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,67 @@
...
+        time: Time,
+        time_with_zone: ActiveSupport::TimeWithZone
+      }.with_indifferent_access.freeze
+
+      def get(value)
+        value = value.to_sym if value.is_a?(String)
+        mapping[value] || handle_unmapped_type(value)
+      end
+
+      def define(symbol, klass)
```
p-mongo, line 43: All of the public methods in this module need docstrings.

The `klass` must be either checked to be a `Class` instance with other types raising an `ArgumentError` or a Mongoid-specific exception, or additionally a symbol/string which is already mapped to a class. The documentation needs to state which of these behaviors happens.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,67 @@
...
+module Mongoid
+  module Fields
+
+    # Singleton module which contains a cache for field type definitions.
+    # Custom field types can be configured.
+    module FieldTypes
+      extend self
+
+      # For fields defined with symbols use the correct class.
+      DEFAULT_MAPPING = {
```
p-mongo, line 12: This should be `@api private`.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Fields
+
+    # Singleton module which contains a cache for field type definitions.
+    # Custom field types can be configured.
+    module FieldTypes
+      extend self
```
p-mongo, line 9: Please use `class << self` instead to define the methods only in the correct scope.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Fields
+
+    # Singleton module which contains a cache for field type definitions.
```
p-mongo, line 6: This is a mapping of field type string/symbol aliases to the underlying classes, not a cache.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,67 @@
...
+        set: Set,
+        string: String,
+        stringified_symbol: Mongoid::StringifiedSymbol,
+        symbol: Symbol,
+        time: Time,
+        time_with_zone: ActiveSupport::TimeWithZone
+      }.with_indifferent_access.freeze
+
+      def get(value)
+        value = value.to_sym if value.is_a?(String)
```
p-mongo, line 39: The mapping is already indifferently accessible, the conversion to symbol is not needed.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      def define(symbol, klass)
+        mapping[symbol.to_sym] = klass
+      end
+
+      delegate :delete, to: :mapping
+
+      private
+
+      def mapping
```
p-mongo, line 51: This method needs a docstring explaining what it is (mapping of aliases to classes possibly modified by the application).

spec/mongoid/errors/invalid_field_type_spec.rb
```diff
@@ -17,13 +17,13 @@
 
       it "contains the problem in the message" do
         expect(error.message).to include(
-          "Invalid field type :stringgy for field 'first_name' on model 'Person'."
+          "Invalid field type :stringgy for field :first_name on model 'Person'."
```
p-mongo, line 20: Please undo these changes.

lib/config/locales/en.yml
```diff
@@ -174,12 +176,21 @@ en:
             Refer to:
             https://docs.mongodb.com/mongoid/current/reference/fields/#custom-field-options"
         invalid_field_type:
-          message: "Invalid field type %{type_inspection} for field '%{field}' on model '%{klass}'."
-          summary: "Model '%{klass}' defines a field '%{field}' with an unknown type value
-            %{type_inspection}."
-          resolution: "Please provide a valid type value for the field.
+          message: "Invalid field type %{type_inspection} for field :%{field} on model '%{klass}'."
```
p-mongo, line 179: Please review the language changes as well and undo the quote removal here.

docs/reference/fields.txt
```diff
@@ -897,6 +897,17 @@ can use in our model class as follows:
     field :location, type: Point
   end
 
+You may optionally declare a symbol mapping for the new field type in an initializer:
```
p-mongo, line 900: This permits referencing the alias as a string also, not just as a symbol.

p-mongo (changes requested):

lib/mongoid/fields/field_types.rb
johnnyshields: Will fix this. By the way, [Rubocop's Style/Documentation](https://docs.rubocop.org/rubocop/cops_style.html#styledocumentation) cop can automatically check whether docs are present, which will save time in future code reviews 👍

lib/mongoid/fields/field_types.rb
johnnyshields: This is a matter of preference, but will do as you say. Rubocop [has a cop for this](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ModuleFunction) that can enforce your preferred way.

lib/mongoid/fields/field_types.rb
johnnyshields: Good catch, thanks.

docs/reference/fields.txt
johnnyshields: To avoid confusion, I have removed word "symbol" and just said "You may optionally declare a mapping..."

johnnyshields:
@p-mongo I believe I have resolved all your requested changes. Please review again, thanks.

lib/config/locales/en.yml
johnnyshields: Resolved.

spec/mongoid/errors/invalid_field_type_spec.rb
johnnyshields: Resolved.

lib/mongoid/fields/field_types.rb
johnnyshields: Resolved.

lib/mongoid/fields/field_types.rb
johnnyshields: Done.

lib/mongoid/fields/field_types.rb
johnnyshields: Done.

docs/reference/fields.txt
johnnyshields: Changed to `define_type`

lib/config/locales/en.yml
```diff
@@ -175,11 +177,20 @@ en:
             https://docs.mongodb.com/mongoid/current/reference/fields/#custom-field-options"
         invalid_field_type:
           message: "Invalid field type %{type_inspection} for field '%{field}' on model '%{klass}'."
-          summary: "Model '%{klass}' defines a field '%{field}' with an unknown type value
-            %{type_inspection}."
-          resolution: "Please provide a valid type value for the field.
+          summary: "Model '%{klass}' declares a field '%{field}' with an unknown :type value
```
p-mongo, line 180: Remove : in front of type please.

lib/config/locales/en.yml
```diff
@@ -175,11 +177,20 @@ en:
...
+            %{type_inspection}. This value is neither present in Mongoid's default type mapping,
+            nor defined in a custom field type mapping."
+          resolution: "Please provide a valid :type value for the field. If you
```
p-mongo, line 183: And here.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,91 @@
...
+        # Defines a field type mapping, for later use in field :type option.
+        #
+        # @example
+        #   Mongoid::FieldTypes.define(:point, Point)
+        #
+        # @param [ Symbol ] symbol the symbol identifier of the defined type
+        # @param [ Class ] klass the class of the defined type, which must
+        #   include mongoize and related methods.
+        def define(symbol, klass)
+          mapping[symbol.to_sym] = klass
```
p-mongo, line 60: Conversion to symbol is redundant because symbol is already a symbol.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,91 @@
...
+        end
+
+        # Defines a field type mapping, for later use in field :type option.
+        #
+        # @example
+        #   Mongoid::FieldTypes.define(:point, Point)
+        #
+        # @param [ Symbol ] symbol the symbol identifier of the defined type
+        # @param [ Class ] klass the class of the defined type, which must
+        #   include mongoize and related methods.
```
p-mongo, line 58: This should simply say `mongoize, demongoize and evolve methods` which takes only two more words.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,91 @@
...
+        # @return [ ActiveSupport::HashWithIndifferentAccess<Symbol, Class> ] The memoized field mapping.
+        def mapping
+          @mapping ||= DEFAULT_MAPPING.dup
+        end
+
+        # Handles fallback for case where mapping does not contain the
+        # requested type.
+        #
+        # @return [ Class | nil ] The class to use as a fallback, or nil.
+        def handle_unmapped_type(type)
```
p-mongo, line 78: The caller of this method does not permit its argument to be nil per its docstring, either the caller's docstring is wrong or nil can't be passed here.

p-mongo: And in fact type should be a string/symbol, not class, per the docstring.

spec/mongoid/errors/invalid_field_type_spec.rb
```diff
@@ -23,7 +23,7 @@
 
       it "contains the summary in the message" do
         expect(error.message).to include(
-          "Model 'Person' defines a field 'first_name' with an unknown type value :stringgy."
+          "Model 'Person' declares a field 'first_name' with an unknown :type value :stringgy."
```
p-mongo, line 26: Remove : in front of type

spec/mongoid/errors/invalid_field_type_spec.rb
```diff
@@ -41,14 +41,14 @@
 
       it "contains the summary in the message" do
         expect(error.message).to include(
-          %q,Model 'Person' defines a field 'first_name' with an unknown type value "stringgy".,
+          %q,Model 'Person' declares a field 'first_name' with an unknown :type value "stringgy".,
```
p-mongo, line 44: And here

spec/mongoid/errors/invalid_field_type_spec.rb
```diff
@@ -41,14 +41,14 @@
...
         )
       end
     end
 
     it "contains the resolution in the message" do
       expect(error.message).to include(
-        'Please provide a valid type value for the field.'
+        'Please provide a valid :type value for the field.'
```
p-mongo, line 51: And here

spec/mongoid/fields/field_types_spec.rb
```diff
@@ -0,0 +1,129 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe Mongoid::Fields::FieldTypes do
+
+  after do
+    described_class.instance_variable_set(:@mapping, described_class::DEFAULT_MAPPING.dup)
```
p-mongo, line 8: You are assuming initial state to be the unmodified state.

At a minimum, this logic can be duplicated into a `before` hook, or it can be rewritten to use an `around` hook, read the current mapping, set the default mapping, run the test, restore previous mapping.

spec/mongoid/fields/field_types_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+      let(:type) { String }
+
+      it 'uses the module type' do
+        is_expected.to eq String
+      end
+    end
+
+    context 'when value is the module Boolean' do
+      let(:type) do
+        stub_const('Boolean', Module.new)
```
p-mongo, line 74: What is the purpose of this stub?

p-mongo (changes requested):

lib/mongoid/fields/field_types.rb
johnnyshields: This is the `@return` value, but sure, I'll improve the doc.

spec/mongoid/fields/field_types_spec.rb
johnnyshields: The constant Boolean does not exist in the global scope. If I comment out this line:

```
  0) Mongoid::Fields::FieldTypes.get when value is the module Boolean returns Mongoid::Boolean type
     Failure/Error: Boolean

     NameError:
       uninitialized constant Boolean
     # ./spec/mongoid/fields/field_types_spec.rb:78:in `block (4 levels) in <top (required)>'
```

spec/mongoid/fields/field_types_spec.rb
johnnyshields: Done. I have used an around hook which resets the value both before and after.

spec/mongoid/errors/invalid_field_type_spec.rb
johnnyshields: Done.

spec/mongoid/errors/invalid_field_type_spec.rb
johnnyshields: Done.

spec/mongoid/errors/invalid_field_type_spec.rb
johnnyshields: Done.

lib/mongoid/fields/field_types.rb
johnnyshields: Done.

lib/mongoid/fields/field_types.rb
johnnyshields: Done.

lib/mongoid/fields/field_types.rb
johnnyshields: Done.

lib/config/locales/en.yml
johnnyshields: Done.

lib/config/locales/en.yml
johnnyshields: Done.

johnnyshields:
Have done all requested second round changes. Appreciate if you can resolve the comments once you check them.

lib/mongoid/fields.rb
```diff
@@ -14,27 +15,6 @@ module Fields
     StringifiedSymbol = Mongoid::StringifiedSymbol
     Boolean = Mongoid::Boolean
 
-    # For fields defined with symbols use the correct class.
-    TYPE_MAPPINGS = {
```
p-mongo, line 18: This is used in client code, e.g. https://github.com/shoplineapp/mongoid-client-side-encryption/blob/master/lib/mongoid_client_side_encryption/models/field.rb#L6

johnnyshields: I've now deprecated TYPE_MAPPINGS. They will need to migrate at some point in the future.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,92 @@
...
+    # Singleton module which contains a mapping of field types to field class.
+    # Custom field types can be configured.
+    #
+    # @api private
+    module FieldTypes
+
+      # The default mapping of field type symbol/string aliases to classes.
+      #
+      # @api private
+      DEFAULT_ALIASES = {
```
johnnyshields, line 15: This is a mapping, not an alias. Alias has a specific meaning in Ruby. It was called a mapping previously and I don't think we should change it.

p-mongo: It's a mapping from aliases to classes, as per the comment.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,92 @@
...
+        #   alias of the defined type. The alias will be accessible as a string
+        #   or a symbol regardless of the type passed to this method.
+        # @param [ Class ] klass the class of the defined type, which must
+        #   include mongoize, demongoize, and evolve methods.
+        def define_alias(field_type_alias, klass)
+          unless field_type_alias.is_a?(String) || field_type_alias.is_a?(Symbol)
+            raise Mongoid::Errors::InvalidFieldTypeAliasName.new(field_type_alias)
+          end
+          unless klass.is_a?(Class)
+            raise Mongoid::Errors::InvalidFieldTypeAliasValue.new(field_type_alias, klass)
```
p-mongo, line 76: These two exceptions need to be added.

lib/mongoid/fields/field_types.rb
```diff
@@ -0,0 +1,92 @@
...
+        # @return [ Class | nil ] The underlying field type class, or nil if
+        #   string or symbol was passed and it is not mapped to any class.
+        def get(field_type)
+          case field_type
+          when Class
+            type
+          when Symbol, String
+            mapping[field_type]
+          else
+            raise Mongoid::Errors::InvalidFieldType.new(self.name, field, field_type)
```
p-mongo, line 57: InvalidFieldType's language should be changed to cover the case of invalid type (i.e. an integer instead of symbol/string), UnknownFieldType should be added to replace the current usage of InvalidFieldType when the alias isn't defined.

p-mongo (changes requested):

lib/mongoid/fields/field_types.rb
johnnyshields: Yes, the constant itself is a *mapping* so its name should be `DEFAULT_MAPPING` (or `DEFAULT_SYMBOL_TO_CLASS_MAPPING`, but that's rather verbose and not necessary given the context.)

spec/mongoid/fields/field_types_spec.rb
johnnyshields: Please mark resolved.

spec/mongoid/fields/field_types_spec.rb
johnnyshields: Please mark resolved.

spec/mongoid/errors/invalid_field_type_spec.rb
johnnyshields: Please mark resolved.

spec/mongoid/errors/invalid_field_type_spec.rb
johnnyshields: Please mark resolved.

spec/mongoid/errors/invalid_field_type_spec.rb
johnnyshields: Please mark resolved.

lib/mongoid/fields/field_types.rb
johnnyshields: Please mark resolved.

lib/mongoid/fields/field_types.rb
johnnyshields: Please mark resolved.

lib/config/locales/en.yml
johnnyshields: Please mark resolved.

lib/config/locales/en.yml
johnnyshields: Please mark resolved.

lib/mongoid/fields/field_types.rb
johnnyshields: Please mark resolved.

lib/mongoid/fields/field_types.rb
johnnyshields: > InvalidFieldType's language should be changed to cover the case of invalid type
If such an error is not actually being called today, we should just rename `InvalidFieldType` to `UnknownFieldType`. After merging this PR, there won't be any error class `InvalidFieldType`. We can add in a future PR if/when it's actually triggered.

p-mongo:
I like the block configuration syntax but it should be on top-level Mongoid module, e.g. `Mongoid.configure do ...` not nested under fields.

The actual changes pertaining to alias definition I think are mostly OK.

I'm not seeing the purpose of maintaining two mappings - the "default" Mongoid one and the active one. Can you explain this part please?

johnnyshields:
> I like the block configuration syntax but it should be on top-level Mongoid module.

Sure, will be glad to do this change.

> I'm not seeing the purpose of maintaining two mappings - the "default" Mongoid one and the active one. Can you explain this part please?

`Mongoid::Fields::FieldTypes::DEFAULT_MAPPING` is an immutable constant. The active mapping `Mongoid::Fields::FieldTypes.mapping` (singleton) is initialized to a `dup` of this constant, and is mutable so users can add their own types. It would of course be possible to inline the constant into the `.mapping` method, but I thought it was cleaner to keep it separate (plus we need it for the deprecated `Mongoid::Fields::TYPE_MAPPINGS` constant.)

Does this answer your question?

johnnyshields:
@p-mongo config change done, ready for a final review.

Note the config will look like this:

```ruby
  Mongoid.configure do |config|
    config.define_field_type :point, Point
  end
```

To keep consistency, I'm also adding `config.define_field_option` in this PR.

I've raised https://jira.mongodb.org/browse/MONGOID-5422 to support doing `configure` without the `|config|` arg. I'm pretty sure it can be done in a backwards compatible manner.

johnnyshields:
Tests are green 👍

p-mongo (changes requested):

docs/reference/fields.txt
```diff
@@ -1010,8 +1021,10 @@ specifiying its handler function as a block:
 
   # in /config/initializers/mongoid_custom_fields.rb
 
-  Mongoid::Fields.option :required do |model, field, value|
-    model.validates_presence_of field if value
+  Mongoid.configure do |config|
+    config.define_field_option :required do |model, field, value|
```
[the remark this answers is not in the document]
johnnyshields: Done

docs/reference/fields.txt
```diff
@@ -915,6 +915,17 @@ can use in our model class as follows:
...
   end
 
+You may optionally declare a mapping for the new field type in an initializer:
+
+.. code-block:: ruby
+
+  # in /config/initializers/mongoid_custom_fields.rb
+
+  Mongoid.configure do |config|
+    config.define_field_type :point, Point
```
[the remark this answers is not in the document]
johnnyshields: I've gone with `config.field_type`. Its shorter and is more in line with the language in MONGOID-5407 / spec of MONGOID-5409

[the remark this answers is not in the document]
p-mongo: You are not implementing 5409. For Mongoid as it exists today, the language you want to use is incorrect.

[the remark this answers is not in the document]
johnnyshields: Hmm... you're right. As the desirable config parameter name here is likely to change in the future, let's hold-off on merging this ticket until after MONGOID-5409 is resolved. We can see how things look at that time and revisit this decision.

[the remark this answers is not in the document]
p-mongo: There is no timeline for implementing 5409.

lib/mongoid/fields/field_types.rb
johnnyshields: Resolving as this is outdated.

lib/mongoid/fields/field_types.rb
johnnyshields: Resolving as this is outdated.

johnnyshields:
Please kindly remove the "Oleg Responded" / "Oleg Todo" flags on this and all other PRs.

PR #5270, opened 2022-05-09
MONGOID-5345: Fix namespace_hierarchy method for anonymous classes

johnnyshields:
@Neilshweky module.name is nil with anonymous classes (e.g. created with Class.new, as is sometimes done in specs), hence this line breaks.

Adding this safe navigation &. check here resolves the issue.

This code demonstrates the issue:

```ruby
      Class.new do
        include Mongoid::Document

        embeds_many :addresses  # causes error
      end
```


johnnyshields:
Ready to merge, please run Evergreen.

p-mongo:

Please see my comment in https://jira.mongodb.org/browse/MONGOID-5345?focusedCommentId=4539265&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4539265.

johnnyshields:
@p-mongo I replied to your comment.

johnnyshields:
Let's run Evergreen and merge please.

lib/mongoid/association/relatable.rb
```diff
@@ -411,7 +411,7 @@ def default_inverse
       def namespace_hierarchy(mod)
         parent = Object
         hier = [parent]
-        mod.name.split('::').each do |part|
+        mod.name&.split('::')&.each do |part|
```
p-mongo, line 414: Can you please change this to 

```
if mod.name
  ...
end
```

instead of try-traversing the entire chain.

p-mongo: + add a comment that `mod.name` is nil for anonymous classes, and for those the hierarchy is just `[Object]`.

spec/mongoid/association/relatable_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  context "anonymous classes" do
+    let(:klass) do
+      Class.new do
+        include Mongoid::Document
+        embeds_many :addresses
+      end
+    end
+
+    it "allows relations" do
```
p-mongo, line 15: This should be  something like 

```
context 'when an association is defined' do
  it 'loads successfully' do
```

Also, this is an integration test, it should be under `spec/integration` probably in `spec/integration/associations` and should have a label like "when an association is defined in an anonymous class" ...

p-mongo (changes requested):

lib/mongoid/association/relatable.rb
johnnyshields: Both above done.

johnnyshields:
Comments addressed, please review again.

Neilshweky (approved):

p-mongo (approved):

PR #5272, opened 2022-05-12
MONGOID-5351 Implement Dots and Dollars in Mongoid

Neilshweky:
CI failure are unrelated and appear on other PRs


lib/mongoid/fields.rb
```diff
@@ -241,6 +241,28 @@ def using_object_ids?
...
+
+    # Validate whether or not the field starts with a dollar sign ($) or
+    # contains a dot/period (.).
+    #
+    # @raise [ ProhibitedSetter ] If contains dots or starts with a dollar.
+    #
+    # @param [ String ] name The field name.
+    def validate_dots_and_dollars!(name)
+      if dot_dollar_field?(name)
+        raise Errors::ProhibitedSetter.new(self.class, name)
```
johnnyshields, line 262: How about `ProbihitedFieldName` or `ProbihitedFieldNameInSetter`

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,39 @@ requiring the usage of the aggregation pipeline for both queries and updates,
...
+    class User
+      include Mongoid::Document
+      field :"first.last", type: String
+      field :"$_amount", type: Integer
+    end
+
+    user = User.first
+    user.send(:"first.last")
+    # => Mike.Trout
+    user.send(:"$_amount")
```
johnnyshields, line 1404: Defining Ruby methods which do not adhere to [Ruby's naming conventions](https://docs.ruby-lang.org/en/2.4.0/syntax/methods_rdoc.html#label-Method+Names) and require using the Kernel `.send` is an anti-pattern/code smell. I have discussed many times that we should not abuse Ruby's Kernel conventions, and this case is no different.

Instead:
- Do not create getters/setters if they do not adhere to Ruby's naming convention.
- Define an alternate Mongoid method to access these, for example `#read_attribute` which is already in existence may work, or maybe we add a new method like `#get_attribute`.

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,39 @@ requiring the usage of the aggregation pipeline for both queries and updates,
...
+.. code::
+
+    class User
+      include Mongoid::Document
+      field :"first.last", type: String
+      field :"$_amount", type: Integer
+    end
+
+    user = User.new
+    user.set(:"first.last=", "Shohei.Ohtani")
```
johnnyshields, line 1420: Just FYI -- note that `user.set` is an atomic persistence operation which persists directly to the DB. In the future when it is possible to write dot/dollar fields, we should have some way to set the value just on the model, and have it persist when we call `.save!`.

Neilshweky: This is supposed to be `send` good catch

johnnyshields: OK. See above comment re: use of `send`.

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,39 @@ requiring the usage of the aggregation pipeline for both queries and updates,
 applications should avoid using dots in field names and starting field names
 with the dollar sign if possible.
 
-The Ruby driver `currently prohibits
-<https://jira.mongodb.org/browse/RUBY-2528>`_ inserting documents whose
-field names contain dots or begin with the dollar sign. However, if such
-documents are inserted using other software, Mongoid and the Ruby driver
-provide limited support for retrieving and operating on these documents.
+Mongoid, starting in version 8, now allows users to access fields that begin with
```
p-mongo, line 1389: This works in previous Mongoid versions also, doesn't it? If so I would remove the "starting in version 8" part.

docs/reference/fields.txt
p-mongo: @johnnyshields We already discussed your position on `send` etc. in the previous dots/dollars PRs, please do not bring this up in every PR going forward. The `read_attribute` addition I agree with.

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,39 @@ requiring the usage of the aggregation pipeline for both queries and updates,
...
+      include Mongoid::Document
+      field :"first.last", type: String
+      field :"$_amount", type: Integer
+    end
+
+    user = User.first
+    user.send(:"first.last")
+    # => Mike.Trout
+    user.send(:"$_amount")
+    # => 42650000
```
p-mongo, line 1405: I think it is a good idea to add a `read_attribute` example here, assuming it works, e.g.

> It is also possible to use `read_attribute` to access these fields:

```
user.read_attribute("first.last")
# => Mike.Trout
```

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,39 @@ requiring the usage of the aggregation pipeline for both queries and updates,
...
+
+Since the update operations will not function correctly on these fields, as
```
p-mongo, line 1407: I would reword this to be something like:

> Due to [server limitations](https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/), updating and replacing fields containing dots and dollars requires using special operators. For this reason, calling setters on these fields is prohibited and will raise an error:

This way we are explaining to users that the issue is in the server, not Mongoid.

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,39 @@ requiring the usage of the aggregation pipeline for both queries and updates,
...
+
+    class User
+      include Mongoid::Document
+      field :"first.last", type: String
+      field :"$_amount", type: Integer
+    end
+
+    user = User.new
+    user.send(:"first.last=", "Shohei.Ohtani")
+    # raises a ProhibitedSetter error
```
p-mongo, line 1421: What do you think about `InvalidDotDollarAssignment`?

lib/mongoid/attributes.rb
```diff
@@ -121,6 +121,7 @@ def read_attribute_before_type_cast(name)
     # @raise [ Errors::ReadonlyAttribute ] If the field cannot be removed due
     #   to being flagged as reaodnly.
     def remove_attribute(name)
+      validate_dots_and_dollars!(name.to_s)
```
p-mongo, line 124: I would suggest something like `validate_writable_field_name!` to reflect the purpose of the method better rather than its implementation.

lib/mongoid/fields.rb
```diff
@@ -241,6 +241,28 @@ def using_object_ids?
...
     end
 
+    # Does this field start with a dollar sign ($) or contain a dot/period (.)?
+    #
+    # @param [ String ] name The field name.
+    #
+    # @return [ true, false ] If this field is dotted or dollared.
+    def dot_dollar_field?(name)
+      n = aliased_fields[name] || name
+      fields.key?(n) && (n.include?('.') || n.start_with?('$'))
```
p-mongo, line 251: Can you explain why the `key?` condition is here please? Wouldn't it allow writes to dotted fields that aren't already persisted?

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,39 @@ requiring the usage of the aggregation pipeline for both queries and updates,
...
+    user.send(:"$_amount=", 8500000)
+    # raises a ProhibitedSetter error
+
```
p-mongo, line 1424: I think in here should also be either an example of how to actually write these fields using the driver or a link to driver documentation which explains it, along with stating that driver 2.18+ is needed for these writes.

docs/reference/fields.txt
Neilshweky: Not exactly... Technically accessing fields that start with a `$` would work, but accessing dotted fields would not.

docs/reference/fields.txt
Neilshweky: done

docs/reference/fields.txt
Neilshweky: done

docs/reference/fields.txt
Neilshweky: I like that better

lib/mongoid/attributes.rb
Neilshweky: done

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,47 @@ requiring the usage of the aggregation pipeline for both queries and updates,
...
+    class User
+      include Mongoid::Document
+      field :"first.last", type: String
+      field :"$_amount", type: Integer
+    end
+
+    user = User.new
+    user.send(:"first.last=", "Shohei.Ohtani")
+    # raises a InvalidDotDollarAssignment error
```
p-mongo, line 1429:
```suggestion
    # Raises Mongoid::Errors::InvalidDotDollarAssignment
```

docs/reference/fields.txt
```diff
@@ -1386,8 +1386,47 @@ requiring the usage of the aggregation pipeline for both queries and updates,
...
+    user.send(:"$_amount=", 8500000)
+    # raises a InvalidDotDollarAssignment error
```
p-mongo, line 1431:
```suggestion
    # Raises Mongoid::Errors::InvalidDotDollarAssignment
```

lib/mongoid/errors/invalid_dot_dollar_assignment.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    # This error is raised when trying to use the setter for a field that starts
+    # with a dollar sign ($) or contains a dot/period (.).
+    class InvalidDotDollarAssignment < MongoidError
+
+      # Create the new error.
+      #
+      # @param [ Class ] klass The class of the document.
+      # @param [ Class ] attr The attribute attempted to be written.
+      def initialize(klass, attr)
```
p-mongo, line 14: `@api private`

lib/mongoid/fields.rb
```diff
@@ -241,6 +241,28 @@ def using_object_ids?
       self.class.using_object_ids?
     end
 
+    # Does this field start with a dollar sign ($) or contain a dot/period (.)?
+    #
+    # @param [ String ] name The field name.
+    #
+    # @return [ true, false ] If this field is dotted or dollared.
+    def dot_dollar_field?(name)
```
p-mongo, line 249: `@api private`

lib/mongoid/fields.rb
```diff
@@ -241,6 +241,28 @@ def using_object_ids?
...
+      fields.key?(n) && (n.include?('.') || n.start_with?('$'))
+    end
+
+    # Validate whether or not the field starts with a dollar sign ($) or
+    # contains a dot/period (.).
+    #
+    # @raise [ InvalidDotDollarAssignment ] If contains dots or starts with a dollar.
+    #
+    # @param [ String ] name The field name.
+    def validate_writable_field_name!(name)
```
p-mongo, line 260: `@api private`

p-mongo (approved):

lib/mongoid/errors/invalid_dot_dollar_assignment.rb
Neilshweky: done

lib/mongoid/fields.rb
Neilshweky: done

comandeo (approved):

PR #5273, opened 2022-05-13
MONGOID-5330 reverse feature flag for 7.5

Neilshweky:


p-mongo (approved):
Reverse for 8.0?

Neilshweky:
> Reverse for 8.0?

This is actually reversing the feature flag for 7.5 as you can see the base of the PR is 7.5-stable. The value as true for overwrite_chained_operators is the current functionality in Mongoid.

p-mongo:
:+1:

docs/reference/configuration.txt
```diff
@@ -395,8 +395,8 @@ for details on driver options.
       # When chaining the same operators that use the same field, setting this
       # feature flag to false will cause those operators to be combined using an
       # and. Setting this feature flag to true will cause the later chained
-      # operators to overwrite the earlier ones. (default: false)
-      #overwrite_chained_operators: false
+      # operators to overwrite the earlier ones. (default: true)
```
johnnyshields, line 398: Please follow the comment format used elsewhere in this file, if this is indeed the case:

```
      # (default: true in Mongoid 7, will change to false in Mongoid 8)
```

comandeo (approved):

PR #5274, opened 2022-05-16
MONGOID-5352 assign_attributes is working incorrectly since 7.3.4

Neilshweky:
MONGOID-5352

Should we backport this to 7.3-stable and 7.4-stable?

CI failure are unrelated and appear on other PRs

### DIFF
--- a/lib/mongoid/association/embedded/batchable.rb
+++ b/lib/mongoid/association/embedded/batchable.rb
@@ -248,9 +248,9 @@ module Mongoid
           end
         end
 
-        # Retrieve the path and invalidate the cache afterwards. This method
-        # is used for when the association has not been set on the document
-        # yet, which can cause path and atomic_paths to be calculated
+        # Clear the cache for path and atomic_paths. This method is used when
+        # the path method is used, and the association has not been set on the
+        # document yet, which can cause path and atomic_paths to be calculated
         # incorrectly later.
         #
         # @api private
--- a/lib/mongoid/association/embedded/batchable.rb
+++ b/lib/mongoid/association/embedded/batchable.rb
@@ -85,7 +85,7 @@ module Mongoid
         def batch_replace(docs)
           if docs.blank?
             if _assigning? && !empty?
-              _base.delayed_atomic_sets.delete(path)
+              _base.delayed_atomic_sets.delete(path_without_cache)
               _base.add_atomic_unset(first)
               target_duplicate = _target.dup
               pre_process_batch_remove(target_duplicate, :delete)
@@ -97,7 +97,7 @@ module Mongoid
             _base.delayed_atomic_sets.clear unless _assigning?
             docs = normalize_docs(docs).compact
             _target.clear and _unscoped.clear
-            _base.delayed_atomic_unsets.delete(path)
+            _base.delayed_atomic_unsets.delete(path_without_cache)
             inserts = execute_batch_set(docs)
             add_atomic_sets(inserts)
           end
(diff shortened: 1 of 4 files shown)


lib/mongoid/association/embedded/batchable.rb
```diff
@@ -239,10 +239,11 @@ def normalize_docs(docs)
         #
         # @return [ String ] The atomic path.
         def path
-          @path ||= if _unscoped.empty?
+          return @path if @path
+          if _unscoped.empty?
             Mongoid::Atomic::Paths::Embedded::Many.position_without_document(_base, _association)
```
Neilshweky, line 244: The idea here is we don't want to set the cache if we are calculating the position without a document, since this method is often called before the association parent is set. This was causing the path to be set incorrectly for later on.

spec/mongoid/association/embedded/embeds_many/proxy_spec.rb
```diff
@@ -4699,4 +4699,24 @@ class DNS::Record
...
+    let(:post) { EmmPost.create! }
+
+    before do
+      post.assign_attributes(company_tags: [{id: BSON::ObjectId.new, title: 'a'}],
+        user_tags: [{id: BSON::ObjectId.new, title: 'b'}])
+      post.save
+      post.reload
+      post.assign_attributes(company_tags: [{id: BSON::ObjectId.new, title: 'c'}],
+        user_tags: [])
+      post.save
```
p-mongo, line 4713: `save!` please

lib/mongoid/association/embedded/batchable.rb
```diff
@@ -246,6 +246,21 @@ def path
...
 
+        # Retrieve the path and invalidate the cache afterwards. This method
+        # is used for when the association has not been set on the document
+        # yet, which can cause path and atomic_paths to be calculated
+        # incorrectly later.
+        #
+        # @api private
+        #
+        # @return [ String ] The atomic path.
+        def path_without_cache
```
p-mongo, line 257: I think this is a confusing method, can it be changed to just clear the cache and the caller look like

```
            _base.delayed_atomic_unsets.delete(path)
￼            clear_atomic_path_cache
￼```

?

**Comment by johnnyshields** [comment]
@Neilshweky great catch. Posted comment on ticket, echoing here:

```
Please double check if there is any impact to these:

setters: post.company_tags =
accepts_nested_attributes_for: post.assign_attributes(company_tags_attributes: ...)
```

File: lib/mongoid/association/embedded/batchable.rb (hunk 3 above)
Comment (Neilshweky) [reply lib/mongoid/association/embedded/batchable.rb:257]: done

**Review by p-mongo — APPROVED** [review]

**Comment by p-mongo** [comment]
I think this should be backported to 7.3 & 7.4 if the commit backports.

File: lib/mongoid/association/embedded/batchable.rb [hunk 4]
```diff
@@ -85,7 +85,8 @@ def batch_remove(docs, method = :delete)
         def batch_replace(docs)
           if docs.blank?
             if _assigning? && !empty?
-              _base.delayed_atomic_sets.clear
+              _base.delayed_atomic_sets.delete(path)
```
jasonpenny, line 88: > I think this should be backported to 7.3 & 7.4 if the commit backports.

@p-mongo I see this change was backported to 7.3 & 7.4, but it is **not** on the `7.5-stable` branch

Can this be updated on 7.5, also?

Neilshweky: @comandeo just making sure you see this!

jasonpenny: https://github.com/mongodb/mongoid/pull/5646 cherry-picks the change to `7.5-stable`

PR #5276, opened 2022-05-18
MONGOID-4843 don't populate .changed_attributes setting has_and_belongs_to_many association

Neilshweky:
MONGOID-4843

### DIFF
--- a/docs/reference/crud.txt
+++ b/docs/reference/crud.txt
@@ -41,17 +41,17 @@ operations with examples.
 
        *Insert a document or multiple documents into the database, raising an
        error if a validation or server error occurs.*
-       
+
        *Pass a hash of attributes to create one document with the specified
        attributes, or an array of hashes to create multiple documents.
        If a single hash is passed, the corresponding document is returned.
        If an array of hashes is passed, an array of documents corresponding
        to the hashes is returned.*
-       
+
        *If a block is given to* ``create!`` *, it will be invoked with each
        document as the argument in turn prior to attempting to save that
        document.*
-       
+
        *If there is a problem saving any of the documents, such as
        a validation error or a server error, an exception is raised
        and, consequently, none of the documents are returned.
@@ -78,12 +78,12 @@ operations with examples.
 
        *Instantiate a document or multiple documents and, if validations pass,
        insert them into the database.*
-       
+
        ``create`` *is similar to* ``create!`` *but does not raise
        exceptions on validation errors. It still raises errors on server
        errors, such as trying to insert a document with an* ``_id`` *that
        already exists in the collection.*
-       
+
        *If any validation errors are encountered, the respective document
        is not inserted but is returned along with documents that were inserted.
        Use* ``persisted?`` *,* ``new_record?`` *or* ``errors`` *methods
@@ -105,13 +105,13 @@ operations with examples.
           Person.create(first_name: "Heinrich") do |doc|
             doc.last_name = "Heine"
           end # => Person instance
-          
+
           class Post
             include Mongoid::Document
-            
+
             validates_uniqueness_of :title
           end
-          
+
           posts = Post.create([{title: "test"}, {title: "test"}])
           # => array of two Post instances
           posts.map { |post| post.persisted? } # => [true, false]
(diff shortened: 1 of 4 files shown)


lib/mongoid/fields.rb
```diff
@@ -600,7 +600,7 @@ def create_field_getter(name, meth, field)
               write_attribute(name, field.eval_default(self))
             else
               value = field.demongoize(raw)
-              attribute_will_change!(name) if value.resizable?
+              attribute_will_change!(name) if value.resizable? && !field.foreign_key?
```
Neilshweky, line 603:
```suggestion
              attribute_will_change!(name) if value.resizable?
```

docs/reference/crud.txt
```diff
@@ -863,6 +863,12 @@ the database up to the time it is saved. Any persistence operation clears the ch
    # Get the previous value for a field.
    person.name_was # "Alan Parsons"
 
+.. note::
+
+  Setting the associations on a document does not cause the ``changed`` or
+  ``changed_attributes`` hashes to be modified. This is true for all associations
+  whether referenced or embedded.
```
johnnyshields, line 870: FYI--In my app, we use a gem hack https://github.com/tablecheck/mongoid_relations_dirty_tracking that allows us to track this. For **embedded** (which are expected to behave like fields), there is a strong argument to add change tracking to Mongoid core in the future. (Not needed for referenced.)

spec/mongoid/changeable_spec.rb
```diff
@@ -1012,12 +1012,29 @@
         person.user_accounts << user_account
       end
 
-      it 'returns a hash of changes' do
-        pending 'https://jira.mongodb.org/browse/MONGOID-4843'
+      it 'should not add to the changes or changed_attributes hash' do
+        person.changes.should == {}
+        person.changed_attributes.should == {}
+      end
```
johnnyshields, line 1018: Foreign key changes on `_ids` must continue to be tracked for both BelongsTo and HasAndBelongsToMany. The correct spec here is:

docs/reference/crud.txt
Neilshweky: I agree, would you mind creating a ticket for this?

johnnyshields: I will do so, thanks!

spec/mongoid/changeable_spec.rb
Neilshweky: Hi @johnnyshields For referenced associations the `changes` hash was always the empty hash on change to the association, for every mongoid version that I looked at. The spec you provided would actually be a change to the current Mongoid functionality. Here's a code snippet if you'd like to verify that:
```
class User
  include Mongoid::Document

  has_and_belongs_to_many :addresses
end

class Address
  include Mongoid::Document

  has_and_belongs_to_many :users
end

a = Address.create!
u = User.create!

u.addresses << a
p u.changed
```

docs/reference/crud.txt
johnnyshields: https://jira.mongodb.org/browse/MONGOID-5272

spec/mongoid/changeable_spec.rb
johnnyshields: @Neilshweky at least if I use `u.address_ids` for assignment, I see that it does appear in `u.changes`:

```ruby
u.address_ids << a._id
=> [BSON::ObjectId('6286663e071daa69cc155284')]

u.changes
=> {"address_ids"=>[nil, [BSON::ObjectId('6286663e071daa69cc155284')]]}
```

johnnyshields: ah... ok. I keep forgetting `u.addresses` persists immediately, while `u.address_ids` does not. Hence `u.addresses` doesn't track the change (not even in `previous_changes`).

As long as  `u.address_ids << a._id` continues to track the change, we're fine.

We've discussed in other threads but the immediate persistence of relations when using `<<` is one thing we should really change. It causes so much confusion...

johnnyshields: Thank you for taking time to make this script!

Neilshweky: Ah yes, I was talking about the association itself while you were talking about the `_ids` field. Apologies for the confusion... The address_ids does still continue to be populated in the changes hash after this PR.

p-mongo:
@Neilshweky Please retitle this PR to reflect the change being made.

spec/mongoid/changeable_spec.rb
Neilshweky: done!

p-mongo (approved):

PR #5277, opened 2022-05-19
MONGOID-3624 read_attribute_for_validation uses getter

Neilshweky:
The user's issue in this case was that `read_attribute_for_validation` uses the getter, and they were suggesting a work around. I misunderstood this to mean that they wanted this method to work with mongoized instead of demongoized values, whereas they just didn't want us to use the getter, so that they can override it.

In this PR, I change `read_attribute_for_validation` to not use the getter, but still validate on the demongoized value.

### DIFF
--- a/lib/mongoid/validatable.rb
+++ b/lib/mongoid/validatable.rb
@@ -69,7 +69,7 @@ module Mongoid
       elsif fields[attribute].try(:localized?)
         attributes[attribute]
       else
-        send(attr)
+        read_attribute(attr)
       end
     end
 
--- a/spec/mongoid/validatable_spec.rb
+++ b/spec/mongoid/validatable_spec.rb
@@ -62,6 +62,21 @@ describe Mongoid::Validatable do
         expect(value).to be_empty
       end
     end
+
+    context "when validating a mongoizable field" do
+
+      let(:princess) do
+        Princess.new(family: :fam)
+      end
+
+      let(:value) do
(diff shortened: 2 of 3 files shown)


johnnyshields:
Hmm... I'm not sure this needs to be changed.

The change is _probably_ fine, but we should also check what ActiveRecord in the case a getter is overridden.

As this has potential to break apps, it needs a feature flag.

johnnyshields:
Also, please add an integration spec which demonstrates the behavior when a getter is actually overridden (e.g. using alias_method_chain or the like)

Neilshweky:
Yeah @johnnyshields after looking at the tests that are now failing, it seems like the functionality to override the getter for validation purposes seems to be deliberate, and I think this is not worth changing.

Also after looking at what rails does, it seems like they also deliberately use the getter for validation:
```
class User < ActiveRecord::Base
  self.table_name = 'user'

  validates_length_of :name, minimum: 5

  def name
    super + "llo!"
  end
end

u = User.create!(name: "he") # does not raise a validation error
```

johnnyshields:
@Neilshweky while we're here, I think it's still worthwhile to add a spec which covers the existing behavior, i.e. that overriding a getter uses the overridden method in validation.

johnnyshields:
I was thinking from integration spec standpoint, i.e. actually overriding a getter method with `Module.prepend`

johnnyshields:
ah but I guess that works too. ok good enough.

PR #5283, opened 2022-05-20
MONGOID-5190 Respect persistence options on auto-saved associations

Neilshweky:


lib/mongoid/clients/options.rb
```diff
@@ -45,6 +45,10 @@ def persistence_context
             PersistenceContext.new(self.class)
       end
 
+      def persistence_context?
+        !!PersistenceContext.get(self) || PersistenceContext.get(self.class)
```
Neilshweky, line 49:
```suggestion
        !!(PersistenceContext.get(self) || PersistenceContext.get(self.class))
```

p-mongo (approved):

PR #5284, opened 2022-05-20
MONGOID-2951 Mark attributes changed when accessed via #read_attribute / subscript operator

Neilshweky:


spec/mongoid/attributes_spec.rb
```diff
@@ -2684,4 +2684,18 @@
       end
     end
   end
+
+  context "when modifiying a hash referenced with the [] notation" do
+    let(:church) { Church.create!(location: { x: 1 }) }
+
+    before do
+      church[:location].merge!(y: 2)
```
Neilshweky, line 2692: The idea behind this change, is that the following test would work on master if this was `church.location`, i.e. using the dot notation instead of `[]`. This happens because, the dot notation uses the field getter, which contains the following code: 
```
attribute_will_change!(name) if value.resizable?
```
which marks the hash as changed right before it changes it. The `[]` notation however, does not contain that line. Therefore, the hash is not marked as changed before the update, and the change to the hash is not persisted. This PR adds the above line of code to the read_attribute method.

lib/mongoid/attributes.rb
```diff
@@ -86,7 +86,9 @@ def has_attribute_before_type_cast?(name)
     def read_attribute(name)
       field = fields[name.to_s]
       raw = read_raw_attribute(name)
-      field ? field.demongoize(raw) : raw
+      value = field ? field.demongoize(raw) : raw
+      attribute_will_change!(name.to_s) if value.resizable?
```
johnnyshields, line 90: @Neilshweky this seems like the wrong approach. The `read_attribute` method should not mutate any internal state (it is expected to be read only). 

I also think that users should not expect they can mutate objects in Ruby's memory fetched by `read_attribute`, or accessed any other low-level ways such as `instance_variable_get(___)`. Instead, they should be required to use an official setter or `write_attribute`.

Neilshweky: I think I might agree but I would love to see what @p-mongo thinks

p-mongo (approved):
I approve per the much expanded description of https://jira.mongodb.org/browse/MONGOID-2951, this PR is solution C as described there.

johnnyshields:
Let's slow down here and discuss this carefully on [MONGOID-2951](https://jira.mongodb.org/browse/MONGOID-2951)

p-mongo:
@johnnyshields Your proposal to have applications call `attribute_will_change!` or a similar method doesn't make sense. You can simply assign the new value back to the attribute and it will be persisted:

```
arr = model.array_field
arr.delete(:x)
model.array_field = arr
model.save!
```

PR #5287, opened 2022-05-23
MONGOID-4935 Fix Errors::InvalidField gives the wrong method name in description

Neilshweky:


lib/mongoid/errors/invalid_field.rb
```diff
@@ -13,13 +13,14 @@ class InvalidField < MongoidError
       #   InvalidField.new(person, :crazy_method_name)
       #
       # @param [ Class ] klass The document class.
+      # @param [ Symbol ] field The field name.
       # @param [ Symbol ] name The method name.
-      def initialize(klass, name)
+      def initialize(klass, field, name)
```
Neilshweky, line 18: I know this is an API change but it is a major version... I can also make the third parameter optional...

lib/mongoid/errors/invalid_field.rb
```diff
@@ -13,13 +13,14 @@ class InvalidField < MongoidError
...
         super(
           compose_message(
             "invalid_field",
             {
-              name: name,
+              name: field,
```
p-mongo, line 23: Shouldn't the message use both parameters, the method that is being defined and the method that the conflict is over?

Neilshweky: yes... new message:
```
Mongoid::Errors::InvalidField ()
message:
  Defining a field named 'invalid' is not allowed.
summary:
  Defining this field would override the method 'invalid?', which would cause issues with expectations around the original method and cause extremely hard to debug issues. The original method was defined in:
   Object: ActiveModel::Validations
   File: /Users/neil.shweky/.gem/gems/activemodel-7.0.2.3/lib/active_model/validations.rb
   Line: 373
resolution:
  Use Mongoid.destructive_fields to see what names are not allowed, and don't use these names. These include names that also conflict with core Ruby methods on Object, Module, Enumerable, or included gems that inject methods into these or Mongoid internals.
```

p-mongo (approved):

PR #5288, opened 2022-05-24
MONGOID-3291 Clear inverse associations when setting across all association types

Neilshweky:
Let me explain what's going on here with a bit of a simpler example than the one in the ticket:
```
Person has_many Drugs

person1 = Person.create
person2 = Person.create
drug = Drug.create

person1.drugs << drug
person2.drugs << drug

person1.drugs
# => [#<Drug _id: 628d3445d1327ab6cae99fef, person_id: BSON::ObjectId('628d3445d1327ab6cae99fee')>]
# should be []
person2.drugs
# => [#<Drug _id: 628d3445d1327ab6cae99fef, person_id: BSON::ObjectId('628d3445d1327ab6cae99fee')>]
# correct
```
When the drug is reassigned to a different person, the first person's drug array should be cleared. Now, I noticed this is true with some of the other associations, but it is intended with some of them:
- HABTM: elements in a HABTM association can belong to multiple relationships. Therefore assigning the same item to multiple associations should not remove them from their previous assignments.
- belongs_to: 
  - when the inverse is a `has_many`: assigning the same element to multiple `belongs_to` associations is expected and we should not remove the element from its previous assignments. For example:
  ```
  drug1.person = person1
  drug2.person = person1
  person1.drugs
  # => [#<Drug _id: 628d35e2d1327ab6ca4c7937, person_id: BSON::ObjectId('628d35e2d1327ab6ca4c7935')>, #<Drug _id: 628d35e2d1327ab6ca4c7938, person_id: BSON::ObjectId('628d35e2d1327ab6ca4c7935')>]
  ```
    as you can see, both drugs are in the `drugs` array still, and therefore we should not unassign `drug1`'s person.
  - when the inverse is a `has_one`: we should remove the previous assignment's inverse in this case.
- `has_one`: it is already implemented for `has_one` association.
- `embeds_one`: I have implemented similarly to `has_one`
- `embeds_many`: I have implemented similarly to `has_many`
- `embedded_in`: only for `embeds_one` inverse, similarly to `belongs_to`.


spec/mongoid/association/embedded/embedded_in/buildable_spec.rb
```diff
@@ -33,4 +33,58 @@
       end
     end
   end
+
+  context 'when the object is already related to another object' do
```
p-mongo, line 37: s/related to/associated with/

spec/mongoid/association/embedded/embedded_in/buildable_spec.rb
```diff
@@ -33,4 +33,58 @@
...
+      let(:person) do
+        Person.create!
+      end
+
+      before do
+        appointment1.person = person
+        appointment2.person = person
+      end
+
+      it 'does not clear the object of its previous relation' do
```
p-mongo, line 58: association

spec/mongoid/association/embedded/embedded_in/buildable_spec.rb
```diff
@@ -33,4 +33,58 @@
...
+      let(:owner) do
+        Owner.create!
+      end
+
+      before do
+        scribe1.owner = owner
+        scribe2.owner = owner
+      end
+
+      it 'clears the object of its previous relation' do
```
p-mongo, line 84: association

p-mongo (approved):

spec/mongoid/association/embedded/embedded_in/buildable_spec.rb
Neilshweky: done

spec/mongoid/association/embedded/embedded_in/buildable_spec.rb
Neilshweky: done

spec/mongoid/association/embedded/embedded_in/buildable_spec.rb
Neilshweky: done

PR #5290, opened 2022-05-26
MONGOID-2951: Unify #create_field_getter and #read_attribute

johnnyshields:
Follow-up to https://github.com/mongodb/mongoid/pull/5284

### DIFF
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -362,7 +362,7 @@ module Mongoid
     def __read_attribute__(name, field = nil, write_default_value = false)
       raw = read_raw_attribute(name)
       field ||= fields[name.to_s]
-      if write_default_value && lazy_settable?(field, raw)
+      if field && write_default_value && lazy_settable?(field, raw)
         write_attribute(name, field.eval_default(self))
       else
         value = field ? field.demongoize(raw) : raw
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -85,8 +85,7 @@ module Mongoid
     # @return [ Object ] The value of the attribute.
     def read_attribute(name)
       field = fields[name.to_s]
-      raw = read_raw_attribute(name)
-      __read_attribute__(name, field, raw)
+      __read_attribute__(name, field)
     end
     alias :[] :read_attribute
 
@@ -361,7 +360,8 @@ module Mongoid
       value.present?
     end
 
-    def __read_attribute__(name, field, raw)
+    def __read_attribute__(name, field)
+      raw = read_raw_attribute(name)
       value = field ? field.demongoize(raw) : raw
       attribute_will_change!(name) if value.resizable?
       value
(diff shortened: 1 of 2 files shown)


lib/mongoid/fields.rb
```diff
@@ -595,14 +595,7 @@ def create_accessors(name, meth, options = {})
       def create_field_getter(name, meth, field)
         generated_methods.module_eval do
           re_define_method(meth) do
-            raw = read_raw_attribute(name)
-            if lazy_settable?(field, raw)
-              write_attribute(name, field.eval_default(self))
```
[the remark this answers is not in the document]
johnnyshields: OK, did a change. Please check now.

johnnyshields:
By the way, it looks like Github Actions is on the fritz... MongoDB instance not booting or something. Will check details later.

p-mongo:
Hi @johnnyshields ,

Can you please explain what problem this PR is trying to solve. You have several changes going on here and without a clear benefit my preference would be to not make any changes.

johnnyshields:
The problem is that the field getters and `read_attribute` do not use the same code under-the-hood. If the code isn't unified, there are bound to be more issues like MONGOID-2951 in the future.

> You have several changes going on here

There is only one change, which is to extract a common a `__read_attribute__` base method. AFAIK there are no actual changes to any logic/behavior.

p-mongo (changes requested):
This PR takes two methods which each have a purpose but have different implementations, and moves both implementations into a third method.  The third method now doesn't have its own clear purpose. It contains many conditionals and generally is difficult to understand and reason about.

If you want to extract a common helper that doesn't have conditionals that, for example, combines `read_raw_attribute`, `demongoize` and `attribute_will_change!` I think that would be OK.

lib/mongoid/attributes.rb
```diff
@@ -362,5 +360,11 @@ def lookup_attribute_presence(name, value)
       end
       value.present?
     end
+
+    def __read_attribute__(name, field, raw)
+      value = field ? field.demongoize(raw) : raw
+      attribute_will_change!(name) if value.resizable?
+      value
+    end
```
p-mongo, line 368: Now this method doesn't read any attributes, contrary to its  name.

I suggest leaving the existing code to be as it is now.

johnnyshields: Let me think of a better name for this method.

johnnyshields:
Did you review my latest code here? I think it's pretty readable / reasonable and will prevent future issues better than a code comment would.

p-mongo:
I did, this is why I wrote my comment. Your most recent version changes a method not marked private which is not allowed and adds a new method which is also not marked private which then requires a docstring plus adds public API surface. Writing two lines of comments would have taken significantly less time and reviewing that change would have also taken significantly less time than has already gone into this PR.

johnnyshields:
OK, I wasn't aware that was the issue you had with the PR. I've fixed it:
* Reverted changes to `lazy_settable?`
* `__read_attribute__` is already a private method in Ruby

I think the changes here are reasonable, given that we experienced an issue due to code divergence, but I'll go ahead an open a second PR which adds the code comment.

PR #5291, opened 2022-05-26
MONGOID-5360 Warn when defining field types of BSON::Int64/Decimal128 and other unsupported types

Neilshweky:
Here are the warnings that are generated:
```
W, [2022-05-26T12:38:31.842319 #88974]  WARN -- : Using BSON classes, like BSON::Decimal128, as the field type is not supported. In 
BSON <=4, the BSON::Decimal128 type will work as expected for both storing and querying, but will return a BigDecimal on query in 
BSON 5+.
W, [2022-05-26T12:38:31.842455 #88974]  WARN -- : Using BSON classes, like BSON::Int64, as the field type is not supported. Saving 
these types to the database will work as expected, however, querying them will return the corresponding native Ruby types.
``` 
# ------------------------ >8 ------------------------ # Do not modify or remove the line above.
# Everything below it will be ignored.

Requesting a pull to mongodb:master from Neilshweky:MONGOID-5360

Write a message for this pull request. The first block
of text is the title and the rest is the description.

### DIFF
--- a/lib/mongoid/fields.rb
+++ b/lib/mongoid/fields.rb
@@ -731,8 +731,19 @@ module Mongoid
         opts = options.merge(klass: self)
         type_mapping = TYPE_MAPPINGS[options[:type]]
         opts[:type] = type_mapping || unmapped_type(options)
-        unless opts[:type].is_a?(Class)
+        if !opts[:type].is_a?(Class)
           raise Errors::InvalidFieldType.new(self, name, options[:type])
+        else
+          invalid_bson_classes = [ BSON::Boolean, BSON::ByteBuffer, BSON::Code, BSON::CodeWithScope, BSON::DBRef, BSON::DbPointer, BSON::Decimal128, BSON::Document, BSON::Error, BSON::Int32, BSON::Int64, BSON::InvalidKey, BSON::MaxKey, BSON::MinKey, BSON::Timestamp, BSON::Undefined ]
+          if invalid_bson_classes.include?(opts[:type])
+            warn_message = "Using BSON classes, like #{opts[:type]}, as the field type is not supported. "
+            if opts[:type] == BSON::Decimal128
+              warn_message += "In BSON <=4, the BSON::Decimal128 type will work as expected for both storing and querying, but will return a BigDecimal on query in BSON 5+."
+            else
+              warn_message += "Saving these types to the database will work as expected, however, querying them will return the corresponding native Ruby types."
+            end
+            Mongoid.logger.warn(warn_message)
+          end
         end
         return Fields::Localized.new(name, opts) if options[:localize]
         return Fields::ForeignKey.new(name, opts) if options[:identity]
--- a/docs/reference/fields.txt
+++ b/docs/reference/fields.txt
@@ -89,6 +89,13 @@ Mongoid also recognizes the string ``"Boolean"`` as an alias for the
 
 To define custom field types, refer to :ref:`Custom Field Types <custom-field-types>` below.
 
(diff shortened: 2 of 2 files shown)


lib/mongoid/fields.rb
```diff
@@ -731,8 +731,19 @@ def field_for(name, options)
         opts = options.merge(klass: self)
         type_mapping = TYPE_MAPPINGS[options[:type]]
         opts[:type] = type_mapping || unmapped_type(options)
-        unless opts[:type].is_a?(Class)
+        if !opts[:type].is_a?(Class)
           raise Errors::InvalidFieldType.new(self, name, options[:type])
+        else
+          invalid_bson_classes = [ BSON::Decimal128, BSON::Int32, BSON::Int64 ]
```
johnnyshields, line 737: extract this to a constant

lib/mongoid/fields.rb
```diff
@@ -731,8 +731,19 @@ def field_for(name, options)
...
+          if invalid_bson_classes.include?(opts[:type])
+            warn_message = "Using #{opts[:type]}, as the field type is not supported. "
+            if opts[:type] == BSON::Decimal128
+              warn_message += "In BSON <=4, the BSON::Decimal128 type will work as expected for both storing and querying, but will return a BigDecimal on query in BSON 5+."
+            else
+              warn_message += "Saving these types to the database will work as expected, however, querying them will return the corresponding native Ruby types."
```
Neilshweky, line 743:
```suggestion
              warn_message += "Saving this type to the database will work as expected, however, querying them will return the corresponding native Ruby types."
```

lib/mongoid/fields.rb
```diff
@@ -731,8 +731,19 @@ def field_for(name, options)
...
         opts[:type] = type_mapping || unmapped_type(options)
-        unless opts[:type].is_a?(Class)
+        if !opts[:type].is_a?(Class)
           raise Errors::InvalidFieldType.new(self, name, options[:type])
+        else
+          invalid_bson_classes = [ BSON::Decimal128, BSON::Int32, BSON::Int64 ]
+          if invalid_bson_classes.include?(opts[:type])
+            warn_message = "Using #{opts[:type]}, as the field type is not supported. "
+            if opts[:type] == BSON::Decimal128
+              warn_message += "In BSON <=4, the BSON::Decimal128 type will work as expected for both storing and querying, but will return a BigDecimal on query in BSON 5+."
```
p-mongo, line 741: Space before `4`?

lib/mongoid/fields.rb
```diff
@@ -731,8 +731,19 @@ def field_for(name, options)
...
+            else
+              warn_message += "Saving these types to the database will work as expected, however, querying them will return the corresponding native Ruby types."
+            end
+            Mongoid.logger.warn(warn_message)
```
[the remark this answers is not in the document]
p-mongo: This is not necessary, the warnings will be issued only for field definition not for usage.

p-mongo (approved):

lib/mongoid/fields.rb
p-mongo: I am not opposed to this but I don't think this is necessary either.

lib/mongoid/fields.rb
Neilshweky, line 741:
```suggestion
              warn_message += "In BSON <= 4, the BSON::Decimal128 type will work as expected for both storing and querying, but will return a BigDecimal on query in BSON 5+."
```

PR #5294, opened 2022-05-27
Fix MONGOID-5366 #upsert misuses #update_one


p-mongo:


comandeo:
Does it worth mentioning in release notes?

p-mongo:
Yes that is a good idea, I'll type up documentation.

p-mongo:
Because updating would be a behavior change. This PR maintains existing behavior and only prevents users from receiving warnings due to https://github.com/mongodb/mongo-ruby-driver/pull/2515.

comandeo (approved):

PR #5299, opened 2022-06-01
MONGOID-4838 #clone method on document is not safe when embedded children have fields changed/removed

Neilshweky:


lib/mongoid/association/relatable.rb
```diff
@@ -306,6 +306,18 @@ def parent_inclusions
         @parent_inclusions ||= []
       end
 
+      def many?
```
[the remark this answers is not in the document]
p-mongo: Is this public API? I think these end up being user accessible, in which case they need docstrings.

lib/mongoid/copyable.rb
```diff
@@ -19,29 +19,49 @@ module Copyable
     def clone
       # @note This next line is here to address #2704, even though having an
       # _id and id field in the document would cause problems with Mongoid
-      # elsewhere.
+      # elsewhere. Note this is only done on the root document as we want
+      # to maintian the same _id on the embedded documents.
```
p-mongo, line 23: maintain

lib/mongoid/copyable.rb
```diff
@@ -19,29 +19,49 @@ module Copyable
...
+
+    # Create clone of a document of the given klass with the given attributes
+    # hash. This is used recursively so that embedded associations are cloned
+    # safely.
+    #
+    # @param klass [ Class ] The class of the document to create.
+    # @param attrs [ Hash ] The hash of the attributes.
+    #
+    # @return [ Document ] The new document.
+    def clone_with_hash(klass, attrs)
```
p-mongo, line 39: If this method doesn't reference any instance variables can it be a class method?

spec/mongoid/copyable_spec.rb
```diff
@@ -696,5 +696,455 @@
...
+            field :e, type: :string
+            field :f, type: :string
+          end
+
+        end
+
+        after do
+          Object.send(:remove_const, :CloneParent)
+          Object.send(:remove_const, :CloneChild)
+          Object.send(:remove_const, :CloneGrandchild)
```
p-mongo, line 738: Is this copy-pasted from elsewhere? I would prefer to use permanently defined models in spec/support/models etc.

spec/mongoid/copyable_spec.rb
```diff
@@ -696,5 +696,455 @@
...
+            parent = CloneParent.new(a: "1", b: "2")
+            parent.clone_child = CloneChild.new(c: "3", d: "4")
+            parent.save
+          end
+
+          let(:parent) { CloneParent.last }
+          let(:clone) { parent.clone_child.send(method) }
+
+          it "doesn't clone the parent" do
+            expect(clone.clone_parent).to be_nil
```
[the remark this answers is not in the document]
p-mongo: Can this be added to docs please?

lib/mongoid/copyable.rb
p-mongo: I assume maintaining _id on embedded documents through cloning is existing behavior, can it be added to docs as a note please?

p-mongo (approved):

lib/mongoid/association/relatable.rb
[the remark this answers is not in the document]
Neilshweky: done

spec/mongoid/copyable_spec.rb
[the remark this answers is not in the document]
Neilshweky: I added it to the docstring. We don't currently have any documentation on the clone method other than the API docs, which I think is preferable.

lib/mongoid/copyable.rb
Neilshweky: Yes that is existing behavior. I have added it to the docstring also, as explained above, we don't have docs on this either and I think the API docs is enough.

lib/mongoid/copyable.rb
Neilshweky: Sure. I have changed it, let me know if this it what you envisioned.

p-mongo (approved):

PR #5302, opened 2022-06-02
MONGOID-5362 Test return value type of #attributes

Neilshweky:
This PR dpends on #5301.

@p-mongo is this what you envisioned? Would you like me to add more tests? I will remove the feature flag 
changes when backporting.

### DIFF
--- a/spec/mongoid/document_spec.rb
+++ b/spec/mongoid/document_spec.rb
@@ -175,7 +175,7 @@ describe Mongoid::Document do
 
     context "when instantiating a new document" do
       it "returns a Hash" do
-        expect(person.attributes).to be_a(Hash)
+        expect(person.attributes.class).to eq(Hash)
       end
     end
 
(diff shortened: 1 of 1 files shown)


spec/mongoid/document_spec.rb
```diff
@@ -165,13 +165,40 @@ class << self; attr_accessor :name; end
...
+
+    context "when retrieving a document from the database" do
+
+      let(:from_db) { Person.first }
+
+      context "when attributes_as_hash is true" do
+        config_override :attributes_as_hash, true
+
+        it "returns a Hash" do
+          expect(from_db.attributes).to be_a(Hash)
```
p-mongo, line 190: `BSON::Document` should be a `Hash`, I think this line should be

```
          expect(from_db.attributes.class).to be Hash
```

p-mongo: And I think the BSON::Document expectation should similarly be worded using exact comparison.

p-mongo (approved):

p-mongo (approved):

PR #5303, opened 2022-06-02
MONGOID-4160 Don't persist document if associated document is invalid with optional: false

Neilshweky:
Like rails, we fail the entire operation when the `validate: true` flag is passed to the `belongs_to` method and using `save!`.

The problem in this ticket is not the validation not raising, rather, when saving an invalid document to a `belongs_to` association, with `validate: false`, and `save`, the _id is persisted, but the document that that _id points to is not because it is invalid. 

Below is a table of how Mongoid performs: 
|       | belongs_to with validate: true                     | belongs_to with validate: false                                 |
|-------|-----------------------------------------------|------------------------------------------------------------|
| save  | Does not raise an error, nothing is persisted | child is persisted with parent_id, parent is not persisted |
| save! | Raises validation error, nothing is persisted | child is persisted with parent_id, parent is not persisted |

It turns out that rails has the same functionality, where the parent_id is persisted but the parent is not: 
```
pet = Pet.new
pet.user = User.new(id:1)
pet.save
Pet.first.id
# => 1
```
So this is working as intended. The difference is that Mongoid assigns an _id on the document constructor, while rails does on save or on explicitly setting it. 

As this ticket was being discussed we decided to investigate what happens when an invalid document is set on an unpersisted document's belongs_to association. This is with `validate: false` on the `belongs_to` association. Two things could possibly happen:
A) child is persisted with the parent_id, but no parent persisted.
B) Nothing is persisted or error raised if `save!` is used
|                   | belongs_to_required_by_default: true | belongs_to_required_by_default: false |
|-------------------|--------------------------------------|---------------------------------------|
| optional: true    | A                                    | A                                     |
| optional: false   | A, should be B                       | A, should be B                        |
| optional: not set | A, should be B                       | A                                     |

 If validate is true, then B is always the outcome. 

This ticket is for changing an invalid belongs_to to not be persisted, if either `optional` is false, or it is unset with the `belongs_to_required_by_default` flag set to true.

### DIFF
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -351,3 +351,59 @@ Removed ``Array#update_values`` and ``Hash#update_values`` methods
 
 The previously deprecated ``Array#update_values`` and ``Hash#update_values``
 methods have been removed in Mongoid 8.
+
+
+No Longer Persisting Documents with Invalid ``belongs_to`` Associations
+-----------------------------------------------------------------------
+
+In Mongoid 8, if an invalid document is assigned to a ``belongs_to`` association,
+and the base document is saved, if the ``belongs_to`` association had the
+option ``optional: false`` or ``optional`` is unset and the global flag
+``require_belongs_to_by_default`` is true, neither the document nor the
+associated document will be persisted. For example, given the following
+models:
+
+.. code::
+
+  class Parent
+    include Mongoid::Document
+    has_one :child
+    field :name
+    validates :name, presence: true
+  end
+
+  class Child
+    include Mongoid::Document
+
+    belongs_to :parent, autosave: true, validate: false, optional: false
+  end
+
+  child = Child.new
+  parent = Parent.new
+  child.parent = parent # parent is invalid, it does not have a name
+  child.save
+
+In this case, both the child and the parent will not be persisted.
+
+.. note::
+  If ``save!`` were called, a validation error would raised.
+
+If optional is false, then the Child will be persisted with a parent_id, but the
+parent will not be persisted:
+
+.. code::
+
+  child = Child.new
+  parent = Parent.new
+  child.parent = parent # parent is invalid, it does not have a name
+  child.save
+
+  p Child.first
+  # => <Child _id: 629a50b0d1327aad89d214d2, parent_id: BSON::ObjectId('629a50b0d1327aad89d214d3')>
+  p Parent.first
+  # => nil
+
+If you want the functionality of neither document being persisted in Mongoid 7
+and earlier, the ``validate: true`` option can be set on the association. If
+you want the functionality of only the Child being persisted, the ``validate:
+false`` option can be set on the association.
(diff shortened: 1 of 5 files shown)


lib/mongoid/association/referenced/auto_save.rb
```diff
@@ -58,7 +58,7 @@ def self.define_autosave!(association)
                     Array(assoc_value).each do |doc|
                       pc = doc.persistence_context? ? doc.persistence_context : persistence_context
                       doc.with(pc) do |d|
-                        d.save
+                        d.save!
```
Neilshweky, line 61:
```suggestion
                        d.save
```

Neilshweky:
Do foreign keys get persisted if the association is not persisted in rails?

johnnyshields:
Hi Neil, just trying to get my head around the actual change here. Is it correct that the *only* change in this PR is: "we now force validating belongs_to relations when require: true".

If so, perhaps we should force the `:validate` option to be true when `require: true`?

Neilshweky:
> Hi Neil, just trying to get my head around the actual change here. Is it correct that the _only_ change in this PR is: "we now force validating belongs_to relations when require: true".
> 
> (If so, perhaps we should force the `:validate` option to be true when `require: true`?)

We force validating belongs_to associations, i.e. if the associated document is invalid we won't persist either the document or the associated document, if any of the following are true: 
- required: true
- optional: false and required unset
- optional/required unset, Mongoid.belongs_to_required_by_default is true.

docs/release-notes/mongoid-8.0.txt
```diff
@@ -351,3 +351,59 @@ Removed ``Array#update_values`` and ``Hash#update_values`` methods
...
+
+  child = Child.new
+  parent = Parent.new
+  child.parent = parent # parent is invalid, it does not have a name
+  child.save
+
+In this case, both the child and the parent will not be persisted.
+
+.. note::
+  If ``save!`` were called, a validation error would raised.
```
p-mongo, line 389: be raised

docs/release-notes/mongoid-8.0.txt
```diff
@@ -351,3 +351,59 @@ Removed ``Array#update_values`` and ``Hash#update_values`` methods
...
+  parent = Parent.new
+  child.parent = parent # parent is invalid, it does not have a name
+  child.save
+
+  p Child.first
+  # => <Child _id: 629a50b0d1327aad89d214d2, parent_id: BSON::ObjectId('629a50b0d1327aad89d214d3')>
+  p Parent.first
+  # => nil
+
+If you want the functionality of neither document being persisted in Mongoid 7
```
p-mongo, line 406: This behavior is available also in 8, isn't it? If yes I'd say something like "in Mongoid 7 or 8"

spec/mongoid/association/referenced/has_one/proxy_spec.rb
```diff
@@ -1148,4 +1148,101 @@
       expect(Vehicle.aliased_fields["driver"]).to eq("driver_id")
     end
   end
+
+  context "when the document is unpersisted and the association is invalid" do
```
p-mongo, line 1152: "not persisted" or "unsaved" are how we typically refer to this state

p-mongo (approved):

docs/release-notes/mongoid-8.0.txt
Neilshweky: done

PR #5305, opened 2022-06-03
MONGOID-5325 Run app tests against rails 7


p-mongo:


p-mongo:
The new configurations are failing due to ruby 3 but the failures look to be identical to the 6.x failures.

comandeo (approved):

PR #5306, opened 2022-06-06
MONGOID-5216 Run all tests in docker on evergreen


p-mongo:
https://jira.mongodb.com/browse/MONGOID-5216

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -96,6 +96,7 @@ functions:
               export AUTH=${AUTH}
               export SSL=${SSL}
               export APP_TESTS=${APP_TESTS}
+              export DOCKER_DISTRO=${DOCKER_DISTRO}
            EOT
            # See what we've done
            cat expansion.yml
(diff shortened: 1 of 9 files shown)


p-mongo:
Fix for ruby 3.0 app tests: https://github.com/mongoid/mongoid-test-apps/commit/183baa96c66b6023e99f4d5d9b2bf9b52ea3fe54

Changes here:

- replace evergreen config.yml with several erb template files as in driver
- add script to generate config.yml back
- use docker to run all configurations in evergreen
- app tests reference update to incorporate the above commit fixing ruby 3


jruby 9.2 is broken still because current nokogiri demands ruby 2.6 and jruby 9.2 is compatible with 2.5. We could fix this or simply stop testing on 9.2 and just test on 9.3. Per https://github.com/jruby/jruby/releases last 9.2 release was 6 months ago, my guess is it's likely there won't be any more 9.2 releases in the future.

Neilshweky (approved):

PR #5308, opened 2022-06-07
MONGOID-5369: Move Aggregation Pipeline docs

johnnyshields:
Move Aggregation Pipeline docs to its own page (with Map Reduce on the same page at the bottom)

Includes some very light text changes to make the docs more readable in light of the move.


p-mongo (changes requested):

Please only move the aggregation pipeline documentation without making any other changes. Leave map/reduce documentation where it is, etc.

johnnyshields:
OK done. Map/reduce reverted.

docs/reference/aggregation-pipeline.txt
```diff
@@ -0,0 +1,198 @@
...
+    } },
+    { '$match' => {
+      'tours.year' => {'$gte' => 2000},
+      'awards._id' => {'$exists' => true},
+    } },
+    {'$project' => {_id: 1}},
+  ])
+  bands = Band.find(band_ids)
+
+Since the aggregation pipeline is implemented by the MongoDB Ruby Driver
```
p-mongo, line 79: Please return this paragraph to its original version.

docs/reference/aggregation-pipeline.txt
```diff
@@ -0,0 +1,198 @@
...
+and not Mongoid, it returns raw ``BSON::Document`` objects rather than
+``Mongoid::Document`` model instances. The above example projects only
+the ``_id`` field which is then used to load full models. An alternative is
+to not perform such a projection and work with raw fields, which would eliminate
+having to send the list of document ids to Mongoid in the second query
+(which could be large.)
+
+
+.. _aggregation-pipeline-dsl:
```
p-mongo, line 88: Please return this label to what it was.

docs/reference/aggregation-pipeline.txt
```diff
@@ -0,0 +1,198 @@
...
+
+  class Participant
+    include Mongoid::Document
+
+    embedded_in :tour
+
+    field :name, type: String
+  end
+
+We could query which states a participant visited:
```
p-mongo, line 127: Return this to what it was.

p-mongo (changes requested):

docs/reference/aggregation-pipeline.txt
```diff
@@ -0,0 +1,198 @@
...
+``Mongoid::Document`` model instances. The above example projects only
+the ``_id`` field which is then used to load full models. An alternative is
+to not perform such a projection and work with raw fields, which would eliminate
+having to send the list of document ids to Mongoid in the second query
+(which could be large.)
+
+
+.. _aggregation-pipeline-dsl:
+
+Aggregation Pipeline DSL
```
p-mongo, line 90: This heading should also be returned to what it was.

johnnyshields:
@p-mongo my minor changes here add clarity to the documentation, there is no reason to revert them in my opinion. You're welcome to take over this PR and revert them if you'd like.

p-mongo:
Hi @johnnyshields ,

Removing text (as you have done with the headings and labels) does not add context. Removing text could potentially add clarity, but because there isn't a clear diff for the text changes (due to the entirety of text being moved from one file to another) this change is not reviewable.

Renaming labels breaks links to them.

I moved the documentation in https://github.com/mongodb/mongoid/pull/5320 without making further changes.

johnnyshields:
OK, fair enough. I'll see if any of these changes can be merged in as a separate PR later.

PR #5309, opened 2022-06-08
MONGOID-4811 Output controller runtime stats in milliseconds

neilshweky:
For retrieving all posts from the database:
```
> e.duration
# => 0.0004160000244155526 (this is in seconds)
```
Before this change:
```
Completed 200 OK in 73ms (Views: 45.8ms | MongoDB: 0.0ms | Allocations: 54152)
```
After this change:
```
Completed 200 OK in 73ms (Views: 45.8ms | MongoDB: 0.4ms | Allocations: 54152)
```

### DIFF
--- a/lib/mongoid/railties/controller_runtime.rb
+++ b/lib/mongoid/railties/controller_runtime.rb
@@ -62,7 +62,7 @@ module Mongoid
         def started _; end
 
         def _completed e
-          Collector.runtime += e.duration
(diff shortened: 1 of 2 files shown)


p-mongo (approved):

p-mongo:
I adjusted the title of this PR to be more accurate, usually "coerce" is used to refer to the type of the value.

PR #5310, opened 2022-06-08
MONGOID-5374 HABTM fails length validation on create when association is not accessed from child

Neilshweky:
Very preliminary PR here. A few things:

1. As per my comment on MONGOID-5374 this PR implements fix #2, which, as discussed there, is not an optimal solution.
2. This solution should be reworked to move the changes somewhere else instead of the Localizable module. This PR is sort of "proof-of-concept" and needs a few changes in order to be ready to merge if this is the direction we would like to go.

### DIFF
--- a/lib/mongoid/validatable/localizable.rb
+++ b/lib/mongoid/validatable/localizable.rb
@@ -20,6 +20,8 @@ module Mongoid
           value.values.each do |_value|
             super(document, attribute, _value)
           end
+        elsif assoc = document.relations[attribute]
+          super(document, assoc.foreign_key, document.send(assoc.foreign_key))
         else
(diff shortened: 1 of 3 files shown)


lib/mongoid/validatable/localizable.rb
```diff
@@ -20,6 +20,8 @@ def validate_each(document, attribute, value)
           value.values.each do |_value|
             super(document, attribute, _value)
           end
+        elsif assoc = document.relations[attribute]
+          super(document, assoc.foreign_key, document.send(assoc.foreign_key))
```
Neilshweky, line 24: Should move this elsewhere.

PR #5317, opened 2022-06-14
MONGOID-4823 Values for BSON::Binary fields are not converted to Binary

Neilshweky:
`ObjectId`'s return the inputted value from their `mongoize` value on uncastable value. 
```
BSON::ObjectId.mongoize(:"Invalid value")
# => :"Invalid value"
```
I think this should be the desired functionality as when a user calls for a field of type `ObjectId`, they mean for it to function as some sort of unique identifier, and returning `nil` would cause all sorts of problems... Also, it would be massively backwards breaking...

### DIFF
--- a/lib/mongoid/extensions/binary.rb
+++ b/lib/mongoid/extensions/binary.rb
@@ -2,41 +2,43 @@
 
 module Mongoid
   module Extensions
-    module Binary
+    module BSON
+      module Binary
 
-      # Turn the object from the ruby type we deal with to a Mongo friendly
-      # type.
-      #
-      # @example Mongoize the object.
-      #   object.mongoize
-      #
-      # @return [ Object ] The object.
-      def mongoize
-        BSON::Binary.mongoize(self)
-      end
-
-      module ClassMethods
-
-        # Mongoize an object of any type to how it's stored in the db.
+        # Turn the object from the ruby type we deal with to a Mongo friendly
+        # type.
         #
         # @example Mongoize the object.
-        #   BigDecimal.mongoize(123)
+        #   object.mongoize
         #
-        # @param [ Object ] object The object to Mongoize
-        #
-        # @return [ String | BSON::Binary | nil ] A String or Binary
-        #   representing the object or nil.
-        def mongoize(object)
-          return if object.nil?
+        # @return [ Object ] The object.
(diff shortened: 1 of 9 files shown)


lib/mongoid/extensions/binary.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Extensions
+    module Binary
```
p-mongo, line 5: Could this module be `BSONBinary` or similar for clarity?

p-mongo: Or maybe actually nest it in `BSON` submodule to get `Mongoid::Extensions::BSON::Binary`.

lib/mongoid/extensions/binary.rb
```diff
@@ -0,0 +1,48 @@
...
+        def mongoize(object)
+          return if object.nil?
+
+          case object
+          when BSON::Binary
+            object
+          when String
+            BSON::Binary.new(object)
+          else
+            # TODO: MONGOID-5222 raise on the setting of feature flag.
```
p-mongo, line 38: What are your thoughts on treating Symbol as String here?

Neilshweky: I would be okay with that

lib/mongoid/extensions/binary.rb
Neilshweky: done

Neilshweky: Just a note, now when BSON is used in /extensions module it needs to be prefixed with ::

lib/mongoid/extensions/binary.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Extensions
+    module BSON
+      module Binary
```
p-mongo, line 6: The file should also be nested under `bson` subdir.

Neilshweky: After looking at what we've been doing I think I should remove the BSON module since we already have `Mongoid::Extensions::ObjectId`, unless you want me to also move that module, which would be backwards breaking.

PR #5220, opened 2022-04-13
MONGOID-5221/5222: Fix mongoization + demongoization of non-empty string as Integer/Float

johnnyshields:
This is a **partial** fix for [MONGOID-5222](https://jira.mongodb.org/browse/MONGOID-5222) and [MONGOID-5221](https://jira.mongodb.org/browse/MONGOID-5221)

PR #5319, opened 2022-06-15
MONGOID-5222 Make mongoization of uncastable values consistent across types and document

Neilshweky:
I have decided to leave the standardization of the evolve methods for another PR as it seems to be a rather hefty change on its own.

TODO: 
- update docstrings of mongoize methods

Relationship between mongoize/evolve:
- Array uses evolve in mongoize.
- So do Boolean and Hash

New Error message:
```
Mongoid::Errors::InvalidValue ()
message:
  The value 1 cannot be written to a field of type Array
summary:
  The value 1 could not be coerced to type Array
resolution:
  Verify that the value to be set corresponds to the field definition
```

### DIFF
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -434,7 +434,7 @@ for details on driver options.
       # (default: false)
       use_utc: false
 
-      # Raise an error on the assignment of a value that cannot be casted to
+      # Raise an error on the assignment of a value that cannot be cast to
       # the assigning field type. If this flag is off, no error will be raised
       # and nil will be written.
       validate_attribute_types: true
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -434,6 +434,11 @@ for details on driver options.
       # (default: false)
       use_utc: false
 
+      # Raise an error on the assignment of a value that cannot be casted to
+      # the assigning field type. If this flag is off, no error will be raised
+      # and nil will be written.
+      validate_attribute_types: true
+
       # (Deprecated) In MongoDB 4.0 and earlier, set whether to create
       # indexes in the background by default. (default: false)
       background_indexing: false
--- a/docs/reference/fields.txt
+++ b/docs/reference/fields.txt
@@ -1005,6 +1005,14 @@ setter methods for fields of your custom type.
    venue = Venue.new(location: point) # This uses the Point#mongoize instance method.
    venue = Venue.new(location: [ 12, 24 ]) # This uses the Point.mongoize class method.
 
+.. note::
+
+  The ``mongoize`` method should raise an "InvalidValue" error on values that
+  are uncastable to your custom type. Mongoid will handle catching that error
+  and writing ``nil`` if the ``validate_attribute_types`` flag is turned off.
+  See the secion on :ref:`Uncastable Values <uncastable-values>` for more
+  details.
+
 The class method ``demongoize`` does the inverse of ``mongoize``. It takes the raw object
 from the MongoDB Ruby driver and converts it to an instance of your custom type.
 In this case, the database driver returns an ``Array`` and we instantiate a ``Point`` from it.
--- a/docs/reference/fields.txt
+++ b/docs/reference/fields.txt
@@ -825,6 +825,35 @@ Mongoid also defines the ``id`` field aliased to ``_id``. The ``id``
 alias can :ref:`be removed <unalias-id>` if desired (such as to integrate
 with systems that use the ``id`` field to store value different from ``_id``.
 
+.. _uncastable-values:
+
+Assigning Uncastable Values
+---------------------------
+
+In Mongoid 8, Mongoid has standardized the treatment of the assignment of
+"uncastable" values. A value is considered "uncastable" when it cannot be
+coerced to the type of the assigning field. For example:
+
+.. code::
+
+  class User
+    include Mongoid::Document
+
+    field :name, type: Integer
+  end
+
+  User.new(name: [ "hello" ])
+
+Assigning an array to a field of type Integer doesn't work since an array can't
+be coerced to an Integer. To deal with this case, Mongoid has introduced the
+``validate_attribute_types`` flag.
+
+When the ``validate_attribute_types`` flag is turned on, the assignment of uncastable
+values to a field will cause an ``InvalidValue`` error to be raised.
+
+When the ``validate_attribute_types`` flag is turned off, the assignment of uncastable
+values to a field will cause a ``nil`` to be written.
+
 
 .. _customizing-field-behavior:
 
--- a/docs/reference/fields.txt
+++ b/docs/reference/fields.txt
@@ -832,7 +832,7 @@ Assigning Uncastable Values
 
 In Mongoid 8, Mongoid has standardized the treatment of the assignment of
 "uncastable" values. A value is considered "uncastable" when it cannot be
-coerced to the type of the assigning field. For example:
+coerced to the type of the field. For example:
 
 .. code::
 
@@ -1008,7 +1008,7 @@ setter methods for fields of your custom type.
 .. note::
 
   The ``mongoize`` method should raise a ``Mongoid::Errors::InvalidValue`` on values that
-  are uncastable to your custom type. Mongoid will handle catching that error
+  are uncastable to your custom type. Mongoid will handle rescuing that error
   and writing ``nil`` if the ``validate_attribute_types`` flag is turned off.
   See the secion on :ref:`Uncastable Values <uncastable-values>` for more
   details.
--- a/docs/reference/fields.txt
+++ b/docs/reference/fields.txt
@@ -849,7 +849,7 @@ be coerced to an Integer. To deal with this case, Mongoid has introduced the
 ``validate_attribute_types`` flag.
 
 When the ``validate_attribute_types`` flag is turned on, the assignment of uncastable
-values to a field will cause an ``InvalidValue`` error to be raised.
+values to a field will cause a ``Mongoid::Errors::InvalidValue`` exception being raised.
 
 When the ``validate_attribute_types`` flag is turned off, the assignment of uncastable
 values to a field will cause a ``nil`` to be written.
@@ -1007,7 +1007,7 @@ setter methods for fields of your custom type.
 
 .. note::
 
-  The ``mongoize`` method should raise an "InvalidValue" error on values that
+  The ``mongoize`` method should raise a ``Mongoid::Errors::InvalidValue`` on values that
   are uncastable to your custom type. Mongoid will handle catching that error
   and writing ``nil`` if the ``validate_attribute_types`` flag is turned off.
   See the secion on :ref:`Uncastable Values <uncastable-values>` for more
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -47,6 +47,40 @@ changed in Mongoid 8.0:
 Please refer to :ref:`configuration option <configuration-options>` for
 the description and effects of each of these options.
 
+Storing Uncastable Types and the ``validate_attribute_types`` Flag
+------------------------------------------------------------------
+
+In Mongoid 8, Mongoid deals with "uncastable values" based on the values of the
+``validates_attribute_types`` flag. If the ``validates_attribute_types`` flag
+is set to true, an ``InvalidValue`` error is raise, and if it is set to false
+a ``nil`` is written. See the secion on :ref:`Uncastable Values <uncastable-values>`
+for more details.
+
+Some ``mongoize`` methods were also changed to perform consistently with rails
+and the other mongoize methods. The following is a table of the changes in
+functionality:
+
++--------------+------------------------+------------------------+-------------------+
+| Field Type   | Situation              | Previous Functionality | New Functionality |
++==============+========================+========================+===================+
+| Integer/Float| When a non-numeric     | return ``nil``         | return ``42``     |
+|              | string starts with a   |                        |                   |
+|              | number: "42bogus"      |                        |                   |
++--------------+------------------------+------------------------+-------------------+
+| Boolean      | When a non-boolean     | return ``false``       | return ``nil`` or |
+|              | string is assigned:    |                        | raise             |
+|              | "bogus value"          |                        | ``InvalidError``  |
++--------------+------------------------+------------------------+-------------------+
+| Symbol       | When a value that does | return ``nil``         | return ``nil`` or |
+|              | not respond to         |                        | raise             |
+|              | ``to_sym`` is          |                        | ``InvalidError``  |
+|              | assigned: ``[]``       |                        |                   |
++--------------+------------------------+------------------------+-------------------+
+| All Other    | When an uncastable     | undefined behavior,    | return ``nil`` or |
+| Types        | value is assigned      | occasionally raises    | raise             |
+|              |                        | ``NoMethodError``      | ``InvalidError``  |
++--------------+------------------------+------------------------+-------------------+
+
 
 Order of Callback Invocation
 ----------------------------
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -52,7 +52,7 @@ Storing Uncastable Types and the ``validate_attribute_types`` Flag
 
 In Mongoid 8, Mongoid deals with "uncastable values" based on the values of the
 ``validates_attribute_types`` flag. If the ``validates_attribute_types`` flag
-is set to true, an ``InvalidValue`` error is raise, and if it is set to false
+is set to true, a ``Mongoid:Errors::InvalidValue`` exception is raised, and if it is set to false
 a ``nil`` is written. See the secion on :ref:`Uncastable Values <uncastable-values>`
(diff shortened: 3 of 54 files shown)


lib/mongoid/criteria/queryable/extensions/boolean.rb
```diff
@@ -20,7 +20,11 @@ module ClassMethods
             # @return [ true, false ] The boolean value.
             def evolve(object)
               __evolve__(object) do |obj|
-                obj.to_s =~ (/\A(true|t|yes|y|on|1|1.0)\z/i) ? true : false
+                if obj.to_s =~ (/\A(true|t|yes|y|on|1|1.0)\z/i)
+                  true
+                elsif obj.to_s =~ (/\A(false|f|no|n|off|0|0.0)\z/i)
+                  false
```
Neilshweky, line 26: This is a behavior change, before everything defaulted to false, now it defaults to nil.

lib/mongoid/extensions/float.rb
```diff
@@ -36,10 +36,8 @@ module ClassMethods
         #
         # @return [ String ] The object mongoized.
         def mongoize(object)
-          unless object.blank?
-            __numeric__(object).to_f rescue 0.0
-          else
-            nil
+          wrap_mongoize(object) do
+            __numeric__(object).to_f rescue nil
```
Neilshweky, line 40: I made the decision to make "" an uncastable value, and to raise instead of just returning nil, which was old functionality.

Neilshweky: Oleg and I agreed that "" should always return nil.

lib/mongoid/extensions/object.rb
```diff
@@ -239,6 +248,26 @@ def demongoize(object)
...
         end
+
+        # Mongoize the given value and return nil if an error is raised.
+        #
+        # @param [ Object ] object The object to mongoize.
+        #
+        # @return [ Object ] The object mongoized.
+        #
+        # @api private
+        def mongoize_safe(object)
```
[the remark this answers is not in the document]
Neilshweky: removed.

docs/reference/fields.txt
```diff
@@ -825,6 +825,35 @@ Mongoid also defines the ``id`` field aliased to ``_id``. The ``id``
...
 with systems that use the ``id`` field to store value different from ``_id``.
 
+.. _uncastable-values:
+
+Assigning Uncastable Values
+---------------------------
+
+In Mongoid 8, Mongoid has standardized the treatment of the assignment of
+"uncastable" values. A value is considered "uncastable" when it cannot be
+coerced to the type of the assigning field. For example:
```
p-mongo, line 835: I would just say "to the type of the field"

docs/reference/fields.txt
```diff
@@ -825,6 +825,35 @@ Mongoid also defines the ``id`` field aliased to ``_id``. The ``id``
...
+  end
+
+  User.new(name: [ "hello" ])
+
+Assigning an array to a field of type Integer doesn't work since an array can't
+be coerced to an Integer. To deal with this case, Mongoid has introduced the
+``validate_attribute_types`` flag.
+
+When the ``validate_attribute_types`` flag is turned on, the assignment of uncastable
+values to a field will cause an ``InvalidValue`` error to be raised.
```
p-mongo, line 852:
```suggestion
values to a field will cause a ``Mongoid::Errors::InvalidValue`` exception being raised.
```

docs/reference/fields.txt
```diff
@@ -976,6 +1005,14 @@ setter methods for fields of your custom type.
    venue = Venue.new(location: point) # This uses the Point#mongoize instance method.
    venue = Venue.new(location: [ 12, 24 ]) # This uses the Point.mongoize class method.
 
+.. note::
+
+  The ``mongoize`` method should raise an "InvalidValue" error on values that
```
p-mongo, line 1010:
```suggestion
  The ``mongoize`` method should raise a ``Mongoid::Errors::InvalidValue`` on values that
```

docs/reference/fields.txt
```diff
@@ -976,6 +1005,14 @@ setter methods for fields of your custom type.
...
+  are uncastable to your custom type. Mongoid will handle catching that error
```
p-mongo, line 1011: Suggest "Mongoid will handle that error and write `nil`"  since "catch" is an operation in Ruby that is different from "rescue".

docs/release-notes/mongoid-8.0.txt
```diff
@@ -47,6 +47,40 @@ changed in Mongoid 8.0:
 Please refer to :ref:`configuration option <configuration-options>` for
 the description and effects of each of these options.
 
+Storing Uncastable Types and the ``validate_attribute_types`` Flag
+------------------------------------------------------------------
+
+In Mongoid 8, Mongoid deals with "uncastable values" based on the values of the
+``validates_attribute_types`` flag. If the ``validates_attribute_types`` flag
+is set to true, an ``InvalidValue`` error is raise, and if it is set to false
```
p-mongo, line 55:
```suggestion
is set to true, a ``Mongoid:Errors::InvalidValue`` exception is raised, and if it is set to false
```

lib/mongoid/attributes.rb
```diff
@@ -157,10 +157,15 @@ def write_attribute(name, value)
...
         _assigning do
-          validate_attribute_value(field_name, value)
           localized = fields[field_name].try(:localized?)
           attributes_before_type_cast[name.to_s] = value
-          typed_value = typed_value_for(field_name, value)
+
+          typed_value = if Mongoid.validate_attribute_types
+            typed_value_for(field_name, value)
+          else
+            typed_value_for(field_name, value) rescue nil
```
p-mongo, line 166: Needs to rescue the specific exception

lib/mongoid/criteria/queryable/extensions/boolean.rb
```diff
@@ -20,7 +20,11 @@ module ClassMethods
...
+                end
```
p-mongo, line 27:
```
else
  # TODO return original value in evolve
```

lib/mongoid/extensions/object.rb
```diff
@@ -239,6 +239,15 @@ def demongoize(object)
         def mongoize(object)
           object.mongoize
         end
+
+        def _mongoid_wrap_mongoize(object)
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongoid/attributes.rb
Neilshweky: done

docs/reference/fields.txt
Neilshweky: done

docs/reference/fields.txt
Neilshweky: done

docs/reference/configuration.txt
```diff
@@ -434,6 +434,11 @@ for details on driver options.
       # (default: false)
       use_utc: false
 
+      # Raise an error on the assignment of a value that cannot be casted to
```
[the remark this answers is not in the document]
Neilshweky: done

docs/release-notes/mongoid-8.0.txt
```diff
@@ -47,6 +47,40 @@ changed in Mongoid 8.0:
...
+Some ``mongoize`` methods were also changed to perform consistently with rails
+and the other mongoize methods. The following is a table of the changes in
+functionality:
+
++--------------+------------------------+------------------------+-------------------+
+| Field Type   | Situation              | Previous Functionality | New Functionality |
++==============+========================+========================+===================+
+| Integer/Float| When a non-numeric     | return ``nil``         | return ``42``     |
+|              | string starts with a   |                        |                   |
+|              | number: "42bogus"      |                        |                   |
```
johnnyshields, line 68: Hmmm I disagree with this. Bogus chars should make the whole thing bogus.

Neilshweky: Hey Johnny, originally I thought so too, but this is rails behavior since it is the result of `to_i`. We've decided to go with the rails behavior for consistency.

johnnyshields:
@Neilshweky can you include https://github.com/mongodb/mongoid/pull/5220 in scope here (or handle it as a follow-up?)

Neilshweky:
> @Neilshweky please include https://github.com/mongodb/mongoid/pull/5220 in scope here (or handle it as a follow-up?)

After taking a quick look at that, it seems like this PR considers your changes and decides to go in a different direction. After discussion, oleg and I decided to I've consistent with rails behavior, and therefor that PR introduces incorrect behavior

docs/release-notes/mongoid-8.0.txt
johnnyshields: When it comes to working with potentially corrupted data, I don't think that "Doing what Rails happens to do" is a sensible barometer. There's no reason to change Mongoid's existing behavior in this case; we should error on the side of caution rather than **further corrupting corrupted data.** (125 --> "12h5" --> "12")

Before making this change specifically, I suggest you take a poll of actual users and ask what they think is reasonable.

johnnyshields: @dblock @yuki24 care to comment on this?

yuki24: I agree with @johnnyshields. Following what rails does could be a sensible default if (and only if) there is really no concern. It does seem like `validates_attribute_types` should make value checking more strict, so implicit casting may not be as strict as it should.

p-mongo: We created https://jira.mongodb.org/browse/MONGOID-5382 to implement a mechanism that would avoid potential data loss due to type/value conversions during attribute value assignment. The precedent (as set by Rails since, as far as I know, the beginning - circa 2004-2005) is to discard, partially or completely, incoming data when it doesn't neatly convert into the column type. Mongoid has this behavior right now and the changes we are making presently don't "further corrupt data". We do need to settle on one behavior though, which is either to discard the entire input value in cases like `42foo` or to follow the core language (`to_i` converts `42foo` to `42`). We decided to follow the behavior of core language.

Please add your feedback about how to implement stricter value conversions to https://jira.mongodb.org/browse/MONGOID-5382.

johnnyshields: @p-mongo in the spirit of minimizing change, I'd like to ask that we remove the `"42bogus" --> 42` change from this PR and add it only if we agree to do so after discussion of MONGOID-5382. Otherwise it feels like a "drive-by" change in this PR. It is the only change in this PR which coerces type; all other changes `return nil or raise InvalidError`.

johnnyshields:
@Neilshweky I'm a bit concerned about the performance cost of using rescue here on each attribute. Perhaps would be good to benchmark the performance of serializing/deserializing a doc with 100 attributes 10,000 times.

https://technicaldebt.com/the-cost-of-using-rubys-rescue-as-logic/

It might be more performant to do something like this:

```ruby
module Mongoid
   INVALID_VALUE = Object.new

   self.invalid_value
     INVALID_VALUE if Mongoid.validate_attribute_types
   end
end

def Array
  def mongoize(object)
     return if object.nil?
     if object.is_a?(::Array)
       evolve(object).collect{ |obj| obj.class.mongoize(obj) }
     else
       Mongoid.invalid_value
     end
   end
end

typed_value = typed_value_for(field_name, value)
raise Errors::InvalidValue if typed_value == Mongoid::INVALID_VALUE
```

Neilshweky:
> @Neilshweky I'm a bit concerned about the performance cost of using rescue here on each attribute. Perhaps would be good to benchmark the performance of serializing/deserializing a doc with 100 attributes 10,000 times.
> 
> https://technicaldebt.com/the-cost-of-using-rubys-rescue-as-logic/
> 
> It might be more performant to do something like this:
> 
> ```ruby
> module Mongoid
>   INVALID_VALUE = Object.new.freeze
> 
>   def self.value_for_invalid_attribute
>     INVALID_VALUE if Mongoid.validate_attribute_types
>   end
> end
> 
> def Array
>   def mongoize(object)
>      return if object.nil?
>      if object.is_a?(::Array)
>        evolve(object).collect{ |obj| obj.class.mongoize(obj) }
>      else
>        Mongoid.value_for_invalid_attribute
>      end
>    end
> end
> 
> # in lib/mongoid/attributes.rb
> typed_value = typed_value_for(field_name, value)
> raise Errors::InvalidValue if typed_value == Mongoid::INVALID_VALUE
> ```
> 
> Come to think of it, this could enable the interesting behavior where you could _read_ `Mongoid::INVALID_VALUE` into Ruby's memory, but the error is raised only when you attempt to _write_ it back to the DB. That would be really cool 😎

@p-mongo what do you think?

docs/release-notes/mongoid-8.0.txt
Neilshweky: I think I kind of agree, and I've changed this to not mongoize "42bogus" to 42

johnnyshields:
A refinement on the above:

```ruby
module Mongoid
  class UncastableValue
    attr_reader :raw_value

    def initialize(raw_value = nil)
      @raw_value = raw_value
    end
  end

  def self.value_for_uncastable_attribute(raw_value)
    UncastableValue.new(raw_value) if Mongoid.validate_attribute_types
  end
end

def Array # as an example
  def demongoize(object)
    return if object.nil?
    if object.is_a?(::Array)
      evolve(object).collect{ |obj| obj.class.demongoize(obj) }
    else
      Mongoid.value_for_uncastable_attribute(object)
    end
  end
end

# in lib/mongoid/attributes.rb
typed_value = typed_value_for(field_name, value)
if Mongoid.raise_invalid_error_on_demongoization && typed_value.is_a?(Mongoid::UncastableValue)
  raise Errors::InvalidValue
end
```

docs/release-notes/mongoid-8.0.txt
dblock: 💯 on not corrupting data

dblock:
I generally disagree that relying on exceptions to control flow is the way to go because as soon as the data submitted cannot be parsed the performance characteristics will vary dramatically (the exception has a cost). We're dealing with user data, so data X takes N time to process, while very similar data with some values that cannot be cast might take 2xN. That doesn't help me design stable systems.

johnnyshields:
@p-mongo
1. Re: performance, how about we benchmark the "before" and "after" state of this PR? That will put any speculation to rest either way. Seems reasonable to do before merging, doesn't it?
2. Even in absence of performance concerns, as @dblock notes, errors are not generally desirable for control flow in this case. Giving users a choice between (a) suppressing errors (`nil`), (b) a class which encapsulates the error state (`UncastableValue`) ,and (c) raising hard errors would be ultimately flexible, and as per the example code above, can be achieved in just a few lines.

p-mongo:
You can preprocess the user input using an algorithm of your choice with a tighter performance envelope, then feed the processed values to Mongoid. The implementation being proposed here provides easy to understand and easy to work with behavior for most applications, it does not prevent designing "stable systems" by users who choose to do so.

johnnyshields:
So we are now expecting users now to write specialized preprocessors if they want Mongoid to be stable? 😕

p-mongo:
Please do not confuse the issue @johnnyshields . The request was for a "stable system". This means the entire system.

dblock (commented):
I'll try to help this argument. First, @Neilshweky, this is good work! I think it should get merged one way or another, or some parts of it. @johnnyshields what are the specific changes that you propose be made in this PR to merge it, or are you saying "do not merge anything that raises errors"?

johnnyshields:
* Edited comment above.

lib/mongoid/config.rb
```diff
@@ -125,6 +125,13 @@ module Config
     # always return a Hash.
     option :legacy_attributes, default: false
 
+    # When this flag is true, Mongoid will validate the type of the value
+    # assigned to a field, such that it can be coerced to a value of the field
+    # type. If it is "uncastable," an InvalidValue error will be raised. If this
+    # feature flag is off, nil will be written to the attribute and the database
+    # on save.
+    option :validate_attribute_types, default: true
```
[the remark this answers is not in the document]
p-mongo: > For demongoize (reading from DB), the behavior needs to match the Mongoid 7.x behavior, we should not raise errors. 

There is a use case for failing demongoization. It is described in https://jira.mongodb.org/browse/MONGOID-5382.

> Otherwise, entire queries may fail due to just one bad field in one document.

This is something I will verify as part of the ongoing design work for https://jira.mongodb.org/browse/MONGOID-5332 but my impression is this statement is not correct - demongoization is only performed on field access and thus failing demongoization does not result in failing queries since demongoization happens after the query is already completed. If Mongoid does not already behave in this manner we are considering how to obtain this behavior, by default or via opt-in.

p-mongo:
We considered the performance impact of the changes proposed in this PR and we believe we can meet the requirements of https://jira.mongodb.org/browse/MONGOID-5332 (including tickets within the epic) without significantly impacting performance. Presently we are revising both the scope of work and the design.

Please add your use cases and requirements under appropriate tickets within https://jira.mongodb.org/browse/MONGOID-5332, or create new tickets if you have use cases and requirements related to mongoization/demongoization/evolution not presently addressed by Mongoid.

As  a result of feedback in this PR, we added maintaining existing Mongoid performance as one of the requirements.

PR #5321, opened 2022-06-16
MONGOID-5340 Add $set behavior change in Mongoid 7.1 to release notes


p-mongo:


docs/release-notes/mongoid-7.0.txt
```diff
@@ -38,3 +38,31 @@ supports Rails 5.1-6.0.
...
 
 New in version 7.0.5: Mongoid now officially supports Rails 6.0.
+
+
+``set`` Overwrites Complete Document
+------------------------------------
+
+**Breaking change:** In Mongoid 7.0.2 and higher, ``set`` writes
+the provided attributes as the complete document (or sub-document), replacing
+any other attributes stored in the database. This behavior matches that of
```
johnnyshields, line 48: This makes it sound like `set` affects the whole document, even things you aren't setting. How about something like this instead:

```
In Mongoid 7.0.2 and higher, ``set`` fully overwrites the targeted attributes
in the database, including any nested documents and nested attributes.
```

p-mongo: > To a casual reader, this may sound like set now affects the whole document, i.e. even things you aren't setting

Correct. The example below further clarifies the situation.

johnnyshields: Yes, the example is good 👍. However, the wording "complete document" in the intro sentence causes confusion.

Just sharing my experience as an actual user reading this.

p-mongo: This was done in https://github.com/mongodb/mongoid/commit/aa28ed9f4921f4cfa1579c08ae06ce84d5e474a0.

PR #5327, opened 2022-06-17
MONGOID-5221 Consistent demongoization behavior when database value cannot be cast to  application-declared field type

Neilshweky:


docs/reference/configuration.txt
```diff
@@ -434,6 +434,11 @@ for details on driver options.
       # (default: false)
       use_utc: false
 
+      # Raise an error on the assignment of a value that cannot be cast to
+      # the assigning field type. If this flag is off, no error will be raised
+      # and nil will be written.
+      validate_attribute_types: true
```
johnnyshields, line 440: See comment here: https://github.com/mongodb/mongoid/pull/5319#discussion_r902918901

PR #5329, opened 2022-06-18
MONGOID-4815 Add :conditions option example to reference documentation

p-mongo:


p-mongo:
https://jira.mongodb.org/browse/MONGOID-4815

docs/reference/validation.txt
```diff
@@ -23,3 +23,32 @@ Mongoid behaves slightly differently to Active Record when using ``#valid?``
...
+    validates_uniqueness_of :name, conditions: -> { where(:year.gte => 2000) }
+  end
+  
+  # OK
+  Band.create!(name: "Sun Project", year: 2000)
+  
+  # Fails validation because there is a band with the "Sun Project" name
+  # and year 2000 in the database, even though the model being created now
+  # does not have a year.
+  Band.create!(name: "Sun Project")
```
johnnyshields, line 54: I'm struggling to understand the real-world use case for `:conditions`; this example doesn't explain to me how it's actually useful in an app. In my 10+ years using Mongoid and making 100s of uniqueness validations, I've never used the `:conditions` option -- but I do use `:scope` very frequently.


Is this code correct?

```ruby
Band.create!(name: "Sun Project") # OK
Band.create!(name: "Sun Project") # OK
Band.create!(name: "Sun Project") # OK
Band.create!(name: "Sun Project", year: 2000) # OK
Band.create!(name: "Sun Project") # Fails!
```

If so, it seems like option is quite dangerous. I'd propose to deprecate it and if people really want it they can write a custom validator.

docs/reference/validation.txt
```diff
@@ -23,3 +23,32 @@ Mongoid behaves slightly differently to Active Record when using ``#valid?``
...
+
+
+``validates_uniqueness_of`` and ``:conditions``  Option
+=======================================================
+
+The ``:conditions`` option to ``validates_uniqueness_of`` can be used to
+provide additional conditions to add to the database query looking for
+identical documents. This option does not influence when the validation
+is executed because it is not considered when Mongoid retrieves the present
+value of the respective field from the model. Consider the following example:
```
johnnyshields, line 35: I've read this paragraph several times and I'm really struggling to understand it's meaning. I'd propose something like the following:

```
``validates_uniqueness_of`` takes a ``:conditions`` which adds a filter condition to the
query used to find existing documents in the database. This can be useful when < insert use case here >
```

johnnyshields:
I've commented on https://jira.mongodb.org/browse/MONGOID-4815. The gist is:
- Mongoid's behavior for `:conditions` is problematic
- It should be changed to match Rails (with feature flag to facilitate migration.) I highly doubt any user is relying on Mongoid's current behavior, but if they are, they make a custom validation when they upgrade.
- The documentation in this PR should reflect that the behavior is known to be problematic and will change in the future to match Rails.

p-mongo:
I responded in https://jira.mongodb.org/browse/MONGOID-4815.

PR #5330, opened 2022-06-19
MONGOID-5369 - Improvements to the Aggregation Pipeline doc page

johnnyshields:
This is a redux of my changes from https://github.com/mongodb/mongoid/pull/5308. It basically adds an intro section to the Aggregation Pipeline docs page, since its a standalone, along with a few small fixes.

I haven't changed any links and manually tested that all the links I could find work.

### DIFF
--- a/docs/reference/aggregation.txt
+++ b/docs/reference/aggregation.txt
@@ -12,8 +12,6 @@ Aggregation Pipeline
    :depth: 2
    :class: singlecol
 
-Aggregation Pipeline Overview
-=============================
 
 Mongoid exposes `MongoDB's aggregation pipeline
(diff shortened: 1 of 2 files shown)


docs/reference/aggregation.txt
```diff
@@ -76,8 +88,8 @@ having to send the list of document ids to Mongoid in the second query
 
 .. _aggregation-pipeline-builder-dsl:
 
-Builder DSL
------------
+Aggregation Pipeline Builder DSL
+================================
```
johnnyshields, line 92: This needs to be a H2 title, otherwise the table of contents gets messed up.

PR #5331, opened 2022-06-19
MONGOID-5390: Mongoid::Contextual::Memory#pluck should return nil values

johnnyshields:
Fixes MONGOID-5390

Conveniently, we can now use ActiveSupport's `Array#pluck` here, which was introduced in Rails 5.0.

The #pluck method was also missing a doc comment, so I added it.

In addition, the original method called `Array.wrap(fields)`. However, this had no useful effect. If the method was called with a first-arg Array like `pluck([:a, :b])`, then `Array.wrap(fields) #=> [[:a, :b]]`, which is non-sensical. So I've removed this call.

Lastly, I don't think we need to have a feature flag here. Refer to MONGOID-5390 JIRA ticket for details.

### DIFF
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -305,13 +305,13 @@ Mongoid 7 behavior:
   end
 
 
-`#pluck` on Embedded Criteria Returns Nil Values
+``#pluck`` on Embedded Criteria Returns Nil Values
 ------------------------------------------------
 
-Mongoid 8 fixes a bug where calling `#pluck` on a Mongoid::Criteria
-for embedded documents discards nil values. This behavior was inconsistent
-with the `#pluck` method in ActiveSupport, and also with how
-`#pluck` works when reading documents from the database.
+Mongoid 8 fixes a bug where calling ``#pluck`` on a Mongoid::Criteria
+for embedded documents discarded nil values. This behavior was
+inconsistent with both the ``#pluck`` method in ActiveSupport and
+with how ``#pluck`` works when reading documents from the database.
 
 The following example illustrates the bug:
 
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -305,7 +305,7 @@ Mongoid 7 behavior:
   end
 
 
-``#pluck`` on Embedded Criteria Returns Nil Values
+``#pluck`` on Embedded Criteria Returns ``nil`` Values
 ------------------------------------------------
 
 Mongoid 8 fixes a bug where calling ``#pluck`` on a Mongoid::Criteria
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -305,6 +305,50 @@ Mongoid 7 behavior:
   end
 
 
+`#pluck` on Embedded Criteria to Returns Nil Values
+---------------------------------------------------
+
+Mongoid 8 fixes a bug where calling `#pluck` on a Mongoid::Criteria
+for embedded documents discards nil values. This behavior was inconsistent
+with the `#pluck` method in ActiveSupport, and also with how
+`#pluck` works when reading documents from the database.
+
+The following example illustrates the bug:
+
+.. code-block:: ruby
+
+  class Address
+    include Mongoid::Document
+
+    embedded_in :mall
+
+    field :street, type: String
+  end
+
+  class Mall
+    include Mongoid::Document
+
+    embeds_many :addresses
+  end
+
+  mall = Mall.create!
+  mall.addresses.create!(street: "Elm Street")
+  mall.addresses.create!(street: nil)
+
+  # Pluck from database
+  Mall.all.pluck('addresses.street')
+    #=> Correct: Mongoid 7 and 8 both return [ ['Elm Street', nil] ]
+
+  # Pluck using ActiveSupport Array#pluck
+  mall.addresses.pluck(:street)
+    #=> Correct: Mongoid 7 and 8 both return ['Elm Street', nil]
+
+  # Pluck from embedded document criteria
+  mall.addresses.all.pluck(:street)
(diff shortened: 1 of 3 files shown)


docs/release-notes/mongoid-8.0.txt
```diff
@@ -305,6 +305,50 @@ Mongoid 7 behavior:
   end
 
 
+``#pluck`` on Embedded Criteria Returns Nil Values
```
p-mongo, line 308:
```suggestion
``#pluck`` on Embedded Criteria Returns ``nil`` Values
```

docs/release-notes/mongoid-8.0.txt
```diff
@@ -305,6 +305,50 @@ Mongoid 7 behavior:
...
+------------------------------------------------
+
+Mongoid 8 fixes a bug where calling ``#pluck`` on a Mongoid::Criteria
+for embedded documents discarded nil values. This behavior was
+inconsistent with both the ``#pluck`` method in ActiveSupport and
+with how ``#pluck`` works when reading documents from the database.
+
+The following example illustrates the bug:
```
p-mongo, line 316: The purpose of release notes is not to illustrate bugs but to describe new behavior. Please refer to other release notes and 1) adjust the language to focus on new behavior, and 2) adjust the code examples to give Mongoid 8 behavior first.

lib/mongoid/contextual/memory.rb
```diff
@@ -188,15 +188,16 @@ def limit(value)
         self
       end
 
+      # Pluck the field values in memory.
+      #
+      # @example Get the values in memory.
+      #   context.pluck(:name)
+      #
+      # @param [ Array<String, Symbol> ] *fields Field(s) to pluck.
```
p-mongo, line 196: We do not use the `@param ... *parameter` syntax currently and it looks confusing. Is each element of `fields` an array or is the entire `fields` an array?

Can you please reference where this syntax is defined, and if it doesn't have an established definition, either remove `*` or remove `Array<>` leaving `[ String | Symbol ]`.

spec/mongoid/contextual/memory_spec.rb
```diff
@@ -944,7 +948,7 @@
       end
 
       it "returns the values" do
-        expect(plucked).to eq([ "hobrecht", "friedel" ])
+        expect(plucked).to eq([ "hobrecht", "friedel", nil ])
```
p-mongo, line 951: Please separate the case of nil values to their own tests for clarity under a `context "when value in database is nil"` or similar.

p-mongo (changes requested):

Overall it looks good but there are several places where the documentation and tests can be made more clear, I hope that you can make the requested changes.

docs/release-notes/mongoid-8.0.txt
johnnyshields: This is done, please check again.

lib/mongoid/contextual/memory.rb
johnnyshields: Good point, I've gone with:

```
@param [ String | Symbol ] *fields Field(s) to pluck.
```

Since `*fields` is what appears in the method signature `def pluck(*fields)`. This is unambiguous that it's a splat. `Array` doesn't feel right because Arrays aren't actually passed in. I think we should adopt this convention for splats everywhere (I'm happy to raise a follow-up ticket to do so.)

johnnyshields: One more comment, I think we should clarify in other methods that `|` is used for the union of types, while `,` should be reserved for `Hash<Symbol, String>` i.e. separation of args within a generic type.

spec/mongoid/contextual/memory_spec.rb
johnnyshields: Done.

johnnyshields:
@p-mongo all requested changes done, ready for your review again.

lib/mongoid/contextual/memory.rb
```diff
@@ -188,15 +188,16 @@ def limit(value)
...
       def pluck(*fields)
-        fields = Array.wrap(fields)
-        documents.map do |doc|
-          if fields.size == 1
-            doc[fields.first]
-          else
-            fields.map { |n| doc[n] }.compact
-          end
-        end.compact
+        documents.pluck(*fields)
```
Neilshweky, line 200: I wonder if this wasn't implemented like this originally because we supported versions of ruby that didn't implement pluck... do you know when pluck was made available?

Neilshweky (approved):
Nice, would you mind verifying that pluck works in all ruby versions we support?

lib/mongoid/contextual/memory.rb
johnnyshields: Yes, correct. Enumerable#pluck was [introduced in ActiveSupport 5.0.0](https://github.com/rails/rails/issues/20339)

johnnyshields:
> would you mind verifying that pluck works in all ruby versions we support?

Yep, verified. The minimum we support is ActiveSupport 5.0.0 and coincidentally that is the same version in which Enumerable#pluck (with same behavior/args/etc) was introduced.

https://www.rubydoc.info/gems/activesupport/5.0.0/Enumerable:pluck

johnnyshields:
If all concerns are addressed let's please run Evergreen and merge?

Neilshweky (commented):
@johnnyshields if you're reworking the pluck method, it seems like when the changes went into allow for things like localizable fields and whatnot, it was not put into the memory method. Can you add tests to make sure that functionality is there? I would look at the pluck tests in criteria and copy some of those....

johnnyshields:
@Neilshweky order to support things like `pluck("descriptions.en", "emdedded.x")`, etc. in-memory, additional work will needed. These things aren't supported by `pluck` today. Adding them is beyond the scope of the original ticket, which is specifically intended to narrowly fix the `nil` case.

I've raised MONGOID-5272 to track this request, and I'm happy to work on it after my existing PRs are merged. Note that it applies to in-memory `pluck`, `distinct`, `map`, and possibly others.

p-mongo (approved):

PR #5332, opened 2022-06-19
MONGOID-5391 - Support Criteria#pluck_each method

johnnyshields:
Resolves MONGOID-5391

This PR extracts out some code into `Mongoid::Contextual::Mongo::PluckEnumerator`. This class is used for both #pluck and #pluck_each (#pluck is simply `pluck_each(fields).to_a`.) There has been no change to the logic of #pluck; all existing #pluck cases pass without modification.

### DIFF
--- a/lib/mongoid/contextual/mongo/pluck_enumerator.rb
+++ b/lib/mongoid/contextual/mongo/pluck_enumerator.rb
@@ -1,0 +1,157 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Contextual
+    class Mongo
+
+      # Utility class to add enumerable behavior for Criteria#pluck_each.
+      #
+      # @api private
+      class PluckEnumerator
+        include Enumerable
+
+        # Create the new PluckEnumerator.
+        #
+        # @api private
+        #
+        # @example Initialize a PluckEnumerator.
+        #   PluckEnumerator.new(klass, view, fields)
+        #
+        # @param [ Class ] klass The base of the binding.
+        # @param [ Mongo::Collection::View ] view The Mongo view context.
+        # @param [ String, Symbol ] *fields Field(s) to pluck,
+        #   which may include nested fields using dot-notation.
+        def initialize(klass, view, fields)
+          @klass = klass
+          @view = view
+          @fields = fields
+        end
+
+        # Iterate through plucked field value(s) from the database
+        # for the view context. Yields result values progressively as
+        # they are read from the database. The yielded results are
+        # normalized according to their Mongoid field types.
+        #
+        # @api private
+        #
+        # @example Iterate through the plucked values from the database.
+        #   context.pluck_each(:name) { |name| puts name }
+        #
+        # @param [ Proc ] block The block to call once for each plucked
+        #   result.
+        #
+        # @return [ Enumerator, PluckEnumerator ] The enumerator, or
+        #   self if a block was given.
+        def each(&block)
+          return to_enum unless block_given?
+
+          @view.projection(normalized_field_names.index_with(true)).each do |doc|
+            yield_result(doc, &block)
+          end
+
+          self
+        end
+
+        private
+
+        def database_field_names
+          @database_field_names ||= @fields.map {|f| @klass.database_field_name(f) }
+        end
+
+        def normalized_field_names
+          @normalized_field_names ||= if Mongoid.legacy_pluck_distinct
+            database_field_names
+          else
+            @fields.map {|f| @klass.cleanse_localized_field_names(f) }
+          end
+        end
+
+        def yield_result(doc)
+          values = database_field_names.map do |n|
+            if Mongoid.legacy_pluck_distinct
+              n.include?('.') ? doc[n.partition('.')[0]] : doc[n]
+            else
+              extract_value(doc, n)
+            end
+          end
+          yield(values.size == 1 ? values.first : values)
+        end
+
+        # Fetch the element from the given hash and demongoize it using the
+        # given field. If the obj is an array, map over it and call this method
+        # on all of its elements.
+        #
+        # @param [ Hash | Array<Hash> ] obj The hash or array of hashes to fetch from.
+        # @param [ String ] meth The key to fetch from the hash.
+        # @param [ Field ] field The field to use for demongoization.
+        #
+        # @return [ Object ] The demongoized value.
+        #
+        # @api private
+        def fetch_and_demongoize(obj, meth, field)
+          if obj.is_a?(Array)
+            obj.map { |doc| fetch_and_demongoize(doc, meth, field) }
+          else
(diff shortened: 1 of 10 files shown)


johnnyshields:
This feature works brilliantly. So fast! 🏎️

johnnyshields:
This is ready for review.

johnnyshields:
@Neilshweky please review this PR, think it will be useful for your work on `tally`.

lib/mongoid/contextual/mongo/pluck_enumerator.rb
```diff
@@ -0,0 +1,150 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Contextual
+    class Mongo
+
+      # Utility class to add enumerable behavior for
+      # Criteria#pluck_each. Should not be directly instantiated
+      # outside of Criteria#pluck_each.
```
Neilshweky, line 9: Delete last sentence I think. Marking it @api private implies this.

lib/mongoid/contextual/mongo/pluck_enumerator.rb
```diff
@@ -0,0 +1,150 @@
...
+        def database_field_names
+          @database_field_names ||= @fields.map {|f| @klass.database_field_name(f) }
+        end
+
+        def normalized_field_names
+          @normalized_field_names ||= if Mongoid.legacy_pluck_distinct
+                                        database_field_names
+                                      else
+                                        @fields.map {|f| @klass.cleanse_localized_field_names(f) }
+                                      end
```
Neilshweky, line 68: Your implementation of `normalized_field_names` differs slightly to the one I had previously. 

Can you add a test in which we pluck both a localized field, and it's _transaltions hash?

For example: 
```
Dictionary.pluck(:description, :description_translations)
```
and verify that this new implementation correctly returns 
```
[["t", {"en" => "t", "de" => "u" }]]
```
for current locale being `:en`

Neilshweky (commented):
Can you please answer my comments regarding the noramlzied_field_names and add a test?

lib/mongoid/contextual/mongo/pluck_enumerator.rb
johnnyshields: @Neilshweky this test is [already in this PR](https://github.com/mongodb/mongoid/pull/5332/files/91df1608b47256be1931a0b86afdb284e7f367f5#diff-56ee0ff139cc683178654f38c4899e8b2c716c60c79c8c0a681ed76fa997526aR2619) for both `#pluck` and `#pluck_each`, legacy and non-legacy cases (I copied all the tests you wrote for #pluck.)

Github [collapses the diff](https://github.com/mongodb/mongoid/pull/5332/files/91df1608b47256be1931a0b86afdb284e7f367f5#diff-56ee0ff139cc683178654f38c4899e8b2c716c60c79c8c0a681ed76fa997526a) for `criteria_spec.rb` b/c it is large so you may have missed it.

johnnyshields:
@Neilshweky responded.

lib/mongoid/contextual/mongo/pluck_enumerator.rb
Neilshweky: Okay great!

Neilshweky (approved):

p-mongo:
This is an optimization for a narrow use case that does not need to exist in Mongoid. To iterate result set in batches you should use the normal `.where` etc. calls that Mongoid provides.

johnnyshields:
@p-mongo `#pluck_each(*fields)` greatly reduces the number of object allocations when iterating over large data sets vs. the current alternative (`.only(*fields).each`). I am seeing it speed up my in-Ruby-memory data analysis jobs by ~10x or more. I'm already using this technique in production in dozens of reports in my app, so I don't consider it a narrow use case at all--many users will benefit.

Given the benefit, this refactor has a pretty low "complexity" cost; extracting out `PluckEnumerator` actually makes the code simpler (less code in the mixin, more code encapsulated in the `PluckEnumerator` class.

p-mongo:
To not construct Mongoid model instances when issuing queries from Mongoid we have https://jira.mongodb.org/browse/MONGOID-5411. If your object allocations are coming from somewhere else please elaborate.

johnnyshields:
@p-mongo I believe MONGOID-5411 and this PR serve similar, but slightly different purposes.

Suppose I have a model User with 100 different fields and some embeds (in the below examples "addresses" will be an embedded field.)

In MONGOID-5411, I'm assuming it adds a criteria modifier (lets call it `.as_hashes`) which causes the result to be returned as hashes rather than `Mongoid::Document` instances. Presumably this will work with `.only` and `.each`, so I can do:

```
User.where(name: "Ryan").only(:name, :rank, :addresses).as_hashes.each {|result| puts result }
#=> [ { name: "Ryan", rank: "Private", addresses: [{ street: "Elm Street", city: "Springwood" }] },
#     { name: "Ryan", rank: "Captain", addresses: [] },
#     { name: "Ryan", rank: "Major", addresses: [{ street: "Evergreen Terrace", city: "Springfield" }] } ]
```

(Please correct me if this is not how you're intending MONGOID-5411 to work)

#pluck_each, like #pluck, can actually do even fewer allocations, especially when working with embedded docs where you can use dot-notation path traversal:

```
User.where(name: "Ryan").pluck_each(:name, :rank, "addresses.street") {|result| puts result }
#=> [ "Ryan", "Private", ["Elm Street"] ] },
#     "Ryan", "Captain", [] },
#     "Ryan", "Major", ["Evergreen Terrace"] ]
```

[Array creation is about twice the speed of Hash creation](https://gist.github.com/johnnyshields/4ad359d099576cc2e6fc6914da3e159c) with lower memory footprint. `#pluck_each` is almost certainly to be lighter-weight than MONGOID-5411.

In summary, I think both #pluck_each and MONGOID-5411 are useful and *not mutually exclusive.* It's good for users to have a choose here, just like `pluck` vs. `only` today.

p-mongo:
1. `only` supports path traversal.
2. `pluck` takes the hashes as they come out of the driver and makes arrays out of them. It is more expensive than returning the arrays as they are.

johnnyshields:
@p-mongo that sounds reasonable. If [MONGOID-5411](https://jira.mongodb.org/browse/MONGOID-5411):
- supports `each` (batch iteration with GETMORE)
- supports `only` (trims docs) including dot-notation
- is faster/more memory efficient in cases above

Then there wouldn't be any need for this PR.

p-mongo:
Closing per the above comments. I extracted pluck-related changes to https://github.com/mongodb/mongoid/pull/5399.

johnnyshields:
Fine to close. I do think this is still useful to have (syntax is more terse than what is being proposed in MONGOID-5411) but can revisit later.

PR #5336, opened 2022-06-21
MONGOID-5222 always return nil on mongoize

Neilshweky:
This PR is for changing the mongoize behavior to not always raise, and rather always return nil on uncastable value. This PR also removes the proposed flag.

This leaves us, after the implementation of the strict fields ticket, with the following functionalities:
- return `nil` on uncastable
- raise on invalid type when `strict: true`

It does not contain the following functionalities:
- raise on uncastable
- return `nil` on invalid type when `strict: true`

@p-mongo and I have decided that:
a) the latter functionality doesn't make sense, as the point of setting the strict option is to avoid data loss
b) the former functionality is not desirable and inconsistent with rails 

@p-mongo is working on a longer explanation on the state of our implementation plan


docs/reference/fields.txt
```diff
@@ -976,6 +1004,12 @@ setter methods for fields of your custom type.
    venue = Venue.new(location: point) # This uses the Point#mongoize instance method.
    venue = Venue.new(location: [ 12, 24 ]) # This uses the Point.mongoize class method.
 
+.. note::
+
+  The ``mongoize`` method should return ``nil`` on values that are uncastable to
+  your custom type. See the secion on :ref:`Uncastable Values <uncastable-values>`
+  for more details.
```
Neilshweky, line 1011: Is this note needed?

lib/config/locales/en.yml
```diff
@@ -373,9 +373,9 @@ en:
           resolution: "Set the values from the parent, or redefine the association
             with only a single definition in the parent."
         invalid_value:
-          message: "Value of type %{value_class} cannot be written to a field of type %{field_class}"
-          summary: "Tried to set a value of type %{value_class} to a field of type %{field_class}"
-          resolution: "Verify if the value to be set correspond to field definition"
+          message: "The value %{value} cannot be written to a field of type %{field_class}."
+          summary: "The value %{value} could not be coerced to type %{field_class}."
+          resolution: "Verify that the value to be set corresponds to the field definition."
```
Neilshweky, line 378: This error is most likely no longer needed, and can be repurposed for raising when `strict: true`

johnnyshields: The error should be removed from this PR then. I'd like to ask that we think further around `strict: true` and consider other alternatives; we can add the error back when it is actually needed.

johnnyshields:
This PR looks good to merge as it is consistent with Mongoid 7.x and doesn't introduce any new options/behaviors. 👍 

> @p-mongo is working on a longer explanation on the state of our implementation plan

I propose that this plan should be written as an RFC, on which community members can comment, before we proceed with introducing any new options/behaviors.

lib/mongoid/extensions/range.rb
```diff
@@ -78,7 +77,7 @@ def __mongoize_hash__(object)
           hash.slice!('min', 'max', 'exclude_end')
           hash.compact!
           hash.transform_values!(&:mongoize)
-          hash
+          hash.blank? ? nil : hash
```
johnnyshields, line 80: This line should be reverted. What if I intentionally want to insert an empty hash to Mongoid?

lib/mongoid/criteria/queryable/extensions/boolean.rb
```diff
@@ -20,7 +20,8 @@ module ClassMethods
             # @return [ true, false ] The boolean value.
             def evolve(object)
               __evolve__(object) do |obj|
-                obj.to_s =~ (/\A(true|t|yes|y|on|1|1.0)\z/i) ? true : false
+                res = mongoize(object)
+                res.nil? ? object : res
```
johnnyshields, line 24: Nice.

lib/mongoid/extensions/hash.rb
```diff
@@ -217,10 +217,12 @@ module ClassMethods
         #
         # @param [ Object ] object The object to mongoize.
         #
-        # @return [ Hash ] The object mongoized.
+        # @return [ Hash | nil ] The object mongoized or nil.
         def mongoize(object)
           return if object.nil?
-          evolve(object.dup).transform_values!(&:mongoize)
+          if object.is_a?(Hash)
+            evolve(object.dup).transform_values!(&:mongoize)
```
johnnyshields, line 224: Don't we also need to mongoize the Hash keys?

lib/mongoid/extensions/array.rb
```diff
@@ -144,12 +144,11 @@ def __mongoize_fk__(association, object)
         #
         # @param [ Object ] object The object to mongoize.
         #
-        # @return [ Array ] The object mongoized.
+        # @return [ Array | nil ] The object mongoized or nil.
         def mongoize(object)
+          return if object.nil?
           if object.is_a?(::Array)
             evolve(object).collect{ |obj| obj.class.mongoize(obj) }
```
johnnyshields, line 151: How about Array like things here? Range, Set, even Hash?

lib/mongoid/extensions/set.rb
```diff
@@ -37,9 +37,12 @@ def demongoize(object)
         #
         # @param [ Set ] object The object to mongoize.
         #
-        # @return [ Array ] The object mongoized.
+        # @return [ Array | nil ] The object mongoized or nil.
         def mongoize(object)
-          object.to_a
+          return if object.nil?
+          case object
+          when Array, Set then ::Array.mongoize(object.to_a)
```
johnnyshields, line 44: Also Range and Hash here.

lib/mongoid/extensions/regexp.rb
```diff
@@ -12,12 +12,16 @@ module ClassMethods
...
+        # @param [ Object ] object The object to mongoize.
         #
-        # @return [ Regexp ] The object mongoized.
+        # @return [ Regexp | nil ] The object mongoized or nil.
         def mongoize(object)
-          return nil if object.nil?
-          ::Regexp.new(object)
+          return if object.nil?
+          case object
+          when String then ::Regexp.new(object)
```
johnnyshields, line 21: Does this potentially raise error?

lib/mongoid/extensions/range.rb
```diff
@@ -59,15 +59,14 @@ def demongoize(object)
...
-        # @return [ Hash ] The object mongoized.
+        # @return [ Hash | nil ] The object mongoized or nil.
         def mongoize(object)
+          return if object.nil?
           case object
-          when NilClass then nil
-          when String then object
           when Hash then __mongoize_hash__(object)
-          else __mongoize_range__(object)
+          when Range then __mongoize_range__(object)
```
johnnyshields, line 69: Arguably an Array or Set with two elements could be converted to range. Perhaps that's a stretch.

lib/mongoid/extensions/date.rb
```diff
@@ -60,7 +61,9 @@ def mongoize(object)
               else
                 time = object.__mongoize_time__
               end
-              ::Time.utc(time.year, time.month, time.day)
+              if time.acts_like?(:time)
+                ::Time.utc(time.year, time.month, time.day)
+              end
```
johnnyshields, line 66: this can go outside the rescue block

lib/mongoid/extensions/date.rb
```diff
@@ -50,8 +50,9 @@ def demongoize(object)
         #
         # @param [ Object ] object The object to mongoize.
         #
-        # @return [ Time ] The object mongoized.
+        # @return [ Time | nil ] The object mongoized or nil.
         def mongoize(object)
+          return if object.nil?
```
johnnyshields, line 55: `return if object.blank?`. You have a redundant `unless object.blank?` on the next line.

lib/mongoid/extensions/big_decimal.rb
```diff
@@ -67,21 +63,24 @@ def demongoize(object)
         # @param [ Object ] object The object to Mongoize
         #
         # @return [ String | BSON::Decimal128 | nil ] A String or Decimal128
-        #   representing the object or nil.
+        #   representing the object or nil. String if Mongoid.map_big_decimal_to_decimal128
+        #   is false, BSON::Decimal128 otherwise.
         def mongoize(object)
-          unless object.nil?
+          return if object.nil?
+          return if object.is_a?(String) && object.blank?
```
johnnyshields, line 70: probably fine to just do `return if object.blank?` here. Skipping blank Array/Hash is not a problem IMHO.

lib/mongoid/extensions/float.rb
```diff
@@ -34,12 +34,11 @@ module ClassMethods
...
         # @param [ Object ] object The object to mongoize.
         #
-        # @return [ String ] The object mongoized.
+        # @return [ Float | nil ] The object mongoized or nil.
         def mongoize(object)
-          unless object.blank?
-            __numeric__(object).to_f rescue 0.0
-          else
-            nil
+          return if object.nil?
```
johnnyshields, line 39: `return if object.blank?` is fine for all numeric types.

lib/mongoid/extensions/date.rb
```diff
@@ -60,7 +61,9 @@ def mongoize(object)
               else
                 time = object.__mongoize_time__
               end
-              ::Time.utc(time.year, time.month, time.day)
+              if time.acts_like?(:time)
```
johnnyshields, line 64: For `Date`, we should check if the driver can handle `Date` type, and `mongoize` can just return Date.

lib/mongoid/extensions/boolean.rb
```diff
@@ -11,9 +11,14 @@ class << self
       # @example Mongoize the object.
       #   Boolean.mongoize("123.11")
       #
-      # @return [ String ] The object mongoized.
+      # @return [ true | false | nil ] The object mongoized or nil.
       def mongoize(object)
-        evolve(object)
```
johnnyshields, line 16: General comment which does not need to be fixed in this PR. Sometimes, `mongoize` calls `evolve`, and other times `evolve` calls `mongoize`. We should think about putting both methods side-by-side in the same `extensions` file, rather than also having separate `criteria/queryable/extensions` files.

Neilshweky: I'm not sure about moving the evolves, but we do have a ticket to standardize the functionality of evolve methods, so that should help with the idea that evolve behavior is different across types...

johnnyshields: Cool, problem should be looked at holistically (after solving `mongoize` first, of course!)

lib/mongoid/extensions/hash.rb
p-mongo: Keys in BSON documents must be strings, see https://bsonspec.org/spec.html. Non-string keys are not going to round-trip.

johnnyshields: Perhaps `#mongoize` should call `.deep_stringify_keys!` here then?

lib/mongoid/extensions/regexp.rb
Neilshweky: Good catch

lib/mongoid/extensions/set.rb
Neilshweky: Good catch on the uniq

lib/mongoid/extensions/date.rb
Neilshweky: done

lib/mongoid/extensions/date.rb
Neilshweky: `__mongoize_time__` only returns a Time or `ActiveSupport::TimeWithZone`. I have added a test to demonstrate this.

lib/mongoid/extensions/float.rb
Neilshweky: It doesn't actually change functionality.... but sure

lib/mongoid/extensions/time.rb
```diff
@@ -63,17 +63,19 @@ def demongoize(object)
         #
         # @param [ Object ] object The object to mongoize.
         #
-        # @return [ Time ] The object mongoized.
+        # @return [ Time | nil ] The object mongoized or nil.
         def mongoize(object)
-          return nil if object.blank?
+          return if object.nil?
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongoid/extensions/big_decimal.rb
Neilshweky: sure

lib/config/locales/en.yml
Neilshweky: done

docs/release-notes/mongoid-8.0.txt
```diff
@@ -57,6 +57,45 @@ changed in Mongoid 8.0:
...
+| Symbol       | When a value that does | return ``nil``         | return ``nil``    |
+|              | not respond to         |                        |                   |
+|              | ``to_sym`` is          |                        |                   |
+|              | assigned: ``[]``       |                        |                   |
++--------------+------------------------+------------------------+-------------------+
+| Array/Set    | When a value that is   | return ``[ 1, 2, 3 ]`` | return ``nil``    |
+|              | not an array or set is | return ``[[ :x, 1 ]]`` | return ``nil``    |
+|              | assigned, but does     |                        |                   |
+|              | respond to ``to_a``:   |                        |                   |
+|              | ``1..3``, ``{ x: 1 }`` |                        |                   |
```
johnnyshields, line 87: If this was the legacy behavior for `Array/Set`, I really think we should keep this. It may potentially break apps otherwise.

p-mongo: Johnny, you wanted Mongoid to silently discard input data. For erroring on unconvertable values we have https://jira.mongodb.org/browse/MONGOID-5382.

Range and hash are not sensible inputs for an array field.

p-mongo, line 87: I would add in the second column something like

> While it doesn't usually make sense to convert Range and Hash instances to Arrays or Sets, instances of custom classes implementing to_a that were assigned to Array/Set would now be discarded. Please explicitly call ``to_a`` on the custom classes when assigning to Array/Set fields.

And we might need to go back on this change if we receive user feedback that it's problematic, this might be something to highlight in release notes for RC which would make sense to do.

johnnyshields: > Range and Hash are not sensible inputs for an array field

They may not be *common*, but they are *sensible*. Reasons to keep:
1. Range/Hash both define `.to_a`, which makes users assume that Mongoid will cast them. (Similar to `.to_i`, `.to_s`, `.to_sym`, etc.)
2. Range/Hash casting to Array is always loss-less.
3. Casting Ranges/Hashes to Array is fairly common in normal Ruby code `digits = (0..9).to_a`
4. It is supported today; for above reasons some users may rely on it. Converting to nil could cause data-loss (potentially severe issue) in existing apps, so *any removal **must** have a feature flag.*

> you wanted Mongoid to silently discard input data

We should *not* discard data if it can be casted in a lossless/unambiguous manner. Range/Hash-to-Array are lossless. Consider this example:

```ruby
class Person
  include Mongoid::Document

  field :name_ary, type: Array # [first, last]
  field :age, type: Integer
end

name_hash = { "Bob" => "Smith" } # oops!
person.name_ary = name_hash
person.save!

# Although name wasn't an Array, it's better to save the data rather than discarding it.
person.name_ary #=> ["Bob", "Smith"]

person.age = "Foobar"
person.save!

# "Foobar" is completely non-sensical as an Integer; we have to discard it.
person.age #=> nil
```

Now, please let me further clarify my position:
- Range/Hash-to-Array are "castable", so let's cast them.
- For bona-fide "uncastable" types, given a choice between (a) raising an error and (b) discarding data, we should discard data.
- However, this is a *false choice*. For "uncastable", the best thing to do would be (c) introduce a `Mongoid::UncastableValue` wrapper type. On *mongoization* (presistence), this should output a warning log line, and on demongoization (retrieval) it should return the value wrapped by `Mongoid::UncastableValue`, and allow user to call `Mongoid::UncastableValue#raw_value` to handle it.

johnnyshields: If we have to tell users to do this *or else they will silently lose data*, it's a clear indication that the change is problematic. At a minimum it needs a *feature flag*, but better not to do it at all IMHO.

Neilshweky: Excuse me, this documentation is incorrect. Our current functionality raises on any non-array value to an array field:
```
irb(main):001:1* class User
irb(main):002:1*   include Mongoid::Document
irb(main):003:1*   field :arr, type: Array
irb(main):004:0> end
=> #<Mongoid::Fields::Standard:0x00007f9138758b10 @name="arr", @options={:type=>Array, :klass=>User}, @label=nil, @default_val=nil>
irb(main):005:0> user = User.new
=> #<User _id: 62ca6c1f3282a42670df183c, arr: nil>
irb(main):006:0> user.arr = 1..3
Mongoid::Errors::InvalidValue ()
message:
  Value of type Range cannot be written to a field of type Array
summary:
  Tried to set a value of type Range to a field of type Array
resolution:
  Verify if the value to be set correspond to field definition
```
I would be okay leaving this in if we want don't want to change that but otherwise I think we're clear on the Range/Hash to_a situation.

johnnyshields: @Neilshweky well that changes things! In legacy code, it seems the `InvalidValue` error is *only* used for Array/Hash.

After some further consideration, **I'd propose to add back `InvalidValue` error behavior**, pending further thinking about `strict` etc. handling.

My reasoning is that, since I expect Range/Hash to be casted (they define `.to_a`), it's better to raise an error for now than silently drop the data. In particular I'm worried about users doing mass-assignment in Rails, where they may not be aware of the types which ActionController is giving them.

AFAIK none of the other changes in this PR increase the chance for "data loss" vs. the legacy state. For example, casting Boolean "Foobar" to `nil` (new) rather than `false` (legacy) is not anymore loss-y than the previous state.

spec/mongoid/extensions/date_class_mongoize_spec.rb
```diff
@@ -109,8 +109,12 @@
 
       context "when the string is an invalid time" do
 
+        let(:evolved) do
```
johnnyshields, line 112: This is a nit, but this variable should be `mongoized` not `evolved`

docs/reference/fields.txt
```diff
@@ -978,6 +1006,12 @@ setter methods for fields of your custom type.
    venue = Venue.new(location: point) # This uses the Point#mongoize instance method.
    venue = Venue.new(location: [ 12, 24 ]) # This uses the Point.mongoize class method.
 
+.. note::
+
+  The ``mongoize`` method should return ``nil`` on values that are uncastable to
+  your custom type. See the secion on :ref:`Uncastable Values <uncastable-values>`
```
johnnyshields, line 1012: Typo `secion` --> `section`

Neilshweky, line 1012:
```suggestion
  your custom type. See the section on :ref:`Uncastable Values <uncastable-values>`
```

docs/release-notes/mongoid-8.0.txt
Neilshweky: Im not completely opposed to raising for array and hash for consistency. @p-mongo what do you think

p-mongo: The current behavior in this PR (change from raising InvalidValue to discarding input and assigning nil) is OK with me. The raising on invalid input is covered by https://jira.mongodb.org/browse/MONGOID-5382.

@johnnyshields  I suggest you formulate your thoughts for raising on assignment to Array/Hash fields in a way that is aligned with the requirements/proposal in https://jira.mongodb.org/browse/MONGOID-5382 and add them there.

johnnyshields: Commented here: MONGOID-5222. The core reason *not* to do the change in this PR is that the logic in question is in `mongoid/attributes.rb` and is not part of any `mongoize` method, which makes it out-of-scope for the original ticket. 5222 specifically mentions `mongoize`.

(I believe the code being located outside `mongoize` is the reason both Neil and I were confused about the spec above.)

I'd like to think further about how exactly we handle MONGOID-5382 before commenting there, but we should defer any change to `mongoid/attributes.rb` until then.

spec/mongoid/extensions/date_class_mongoize_spec.rb
Neilshweky: done

docs/reference/fields.txt
```diff
@@ -827,6 +827,34 @@ Mongoid also defines the ``id`` field aliased to ``_id``. The ``id``
...
+  User.new(name: [ "hello" ])
+
+Assigning an array to a field of type Integer doesn't work since an array can't
+be coerced to an Integer. The assignment of uncastable values to a field will
+cause a ``nil`` to be written:
+
+.. code::
+
+  User.new(name: [ "hello" ])
+  # => #<User _id: 62b222d43282a47bf73e3264, name: nil>
```
p-mongo, line 856: Can you add after this that the original value is available in `attributes_before_type_cast`:

```
user.attributes_before_type_cast["name"]
# => ["hello"]
````

File: lib/mongoid/criteria/queryable/extensions/boolean.rb (hunk 4 above)
Comment (p-mongo) [inline lib/mongoid/criteria/queryable/extensions/boolean.rb:24]: Is there a behavior change here? If not I would prefer this hunk reverted here and done as part of evolution PR.

File: lib/mongoid/extensions/date.rb [hunk 21]
```diff
@@ -50,20 +50,21 @@ def demongoize(object)
...
-                # https://jira.mongodb.org/browse/MONGOID-4460
-                time = ::Time.parse(object)
-              else
-                time = object.__mongoize_time__
-              end
+          return if object.blank?
+          begin
+            if object.is_a?(String)
+              # https://jira.mongodb.org/browse/MONGOID-4460
+              time = ::Time.parse(object)
```
p-mongo, line 59: This should `rescue` something right?

lib/mongoid/extensions/hash.rb
```diff
@@ -217,10 +217,12 @@ module ClassMethods
         #
         # @param [ Object ] object The object to mongoize.
         #
-        # @return [ Hash ] The object mongoized.
+        # @return [ Hash | nil ] The object mongoized or nil.
         def mongoize(object)
           return if object.nil?
-          evolve(object.dup).transform_values!(&:mongoize)
+          if object.is_a?(Hash)
+            object.dup.transform_values!(&:mongoize)
```
p-mongo, line 224: Why not https://ruby-doc.org/core-3.1.2/Hash.html#method-i-transform_values without bang, omitting `dup`?

lib/mongoid/extensions/regexp.rb
```diff
@@ -12,12 +12,18 @@ module ClassMethods
...
+          when ::Regexp then object
+          when BSON::Regexp::Raw then object.compile
```
p-mongo, line 23: This is a lossy operation I'm afraid:

```

irb(main):014:0> a=BSON::Regexp::Raw.new('abc','m')
=> #<BSON::Regexp::Raw:0x00007fbd22b28e00 @options="m", @pattern="abc">
irb(main):015:0> a.compile.options
=> 0

```

The declared field type is a Ruby regexp though, not bson one.

What I think can be done is a warning here stating that BSON::Regexp::Raw conversion to Ruby Regexp is lossy, ideally one time per program execution.


Alternatively incoming value of BSON::Regexp::Raw could be  returned as is from `mongoize` and written exactly as is into attributes, with  `demongoize`  issuing the `compile` call to get the Ruby regexp, but conceptually this behavior doesn't align with what I would expect from a field declared as a Ruby regexp.

lib/mongoid/extensions/set.rb
```diff
@@ -37,9 +37,13 @@ def demongoize(object)
         #
         # @param [ Set ] object The object to mongoize.
         #
-        # @return [ Array ] The object mongoized.
+        # @return [ Array | nil ] The object mongoized or nil.
         def mongoize(object)
-          object.to_a
+          return if object.nil?
+          case object
+          when ::Set then ::Array.mongoize(object.to_a)
```
p-mongo, line 44: Shouldn't this line also have a `uniq` at the end in case the mongoizer maps multiple inputs to the same output?

spec/mongoid/extensions/set_spec.rb
```diff
@@ -31,5 +31,48 @@
...
+
+        it "mongoizes to a set" do
+          expect(mongoized).to be_a(Array)
+        end
+
+        it "converts the elements properly" do
+          expect(mongoized.first).to eq(Time.utc(2012, 1, 1, 0, 0, 0))
+        end
+      end
+    end
```
p-mongo, line 76: Is there test coverage in this file for deduplicating input array elements and for removing duplicates when mongoizers map multiple inputs to one output?

lib/mongoid/extensions/date.rb
```diff
@@ -50,20 +50,21 @@ def demongoize(object)
...
+            else
+              time = object.__mongoize_time__
+            end
+            if time.acts_like?(:time)
               ::Time.utc(time.year, time.month, time.day)
-            rescue ArgumentError
-              nil
             end
+          rescue ArgumentError
```
p-mongo, line 66: I would like this rescue to be only wrapped around the actual parse operations because it is very generic and could mask other problems.

spec/mongoid/extensions/date_spec.rb
```diff
@@ -76,4 +76,22 @@
       expect(date.mongoize).to eq(time)
     end
   end
+
+  describe ".mongoize" do
+    let(:date) do
+      Date.new(2010, 1, 1)
```
p-mongo, line 82: This file also needs tests for assigning valid and invalid stringified dates & times I think?

The stringified date case can be tricky because there is an implied time zone in which that date would be interpreted.

docs/reference/fields.txt
Neilshweky: Sure, I ended up adding that in the Mongoid-5221 pr but I can add it here

lib/mongoid/extensions/set.rb
Neilshweky: Yes

spec/support/models/catalog.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class Catalog
```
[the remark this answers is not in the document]
Neilshweky: I only found that after the fact but mine happens to have different fields anyway.

lib/mongoid/criteria/queryable/extensions/boolean.rb
johnnyshields: This logic could still be copied into mongoize for now.

spec/mongoid/extensions/date_spec.rb
johnnyshields: > The stringified date case can be tricky because there is an implied time zone in which that date would be interpreted.

AFAIK Ruby Date class ignores timezone:

```
> Date.parse("2022-01-01")
 => #<Date: 2022-01-01 ((2459581j,0s,0n),+0s,2299161j)>

> Date.parse("2022-01-01Z")
 => #<Date: 2022-01-01 ((2459581j,0s,0n),+0s,2299161j)>

> Date.parse("2022-01-01+0900")
 => #<Date: 2022-01-01 ((2459581j,0s,0n),+0s,2299161j)>
```

lib/mongoid/extensions/hash.rb
p-mongo: :+1:

lib/mongoid/attributes.rb
```diff
@@ -353,30 +352,6 @@ def unalias_attribute(name)
...
-    # being checked to be of an appropriate type (i.e. either Hash or Array,
-    # respectively, or nil).
-    #
-    # This method takes the name of the field as stored in the document
-    # in the database, not (necessarily) the Ruby method name used to read/write
-    # the said field.
-    #
-    # @param [ String, Symbol ] field_name The name of the field.
-    # @param [ Object ] value The value to be validated.
-    def validate_attribute_value(field_name, value)
```
johnnyshields, line 368: _(Other thread was closed)_ This is a drive-by-change in this PR:
- MONGOID-5222 references `mongoize` method specifically.
- This method `validate_attribute_value` does not call `mongoize` (and vice-versa.)

Suggest to revisit changes here as part of [MONGOID-5382](https://jira.mongodb.org/browse/MONGOID-5382). This PR is large already and there's no reason to rush this change in.

johnnyshields: @p-mongo My comment here is politely-worded, on-topic, and is a valid concern. I'd appreciate if the MongoDB team can leave it unminimized and write a brief response when time allows. Thank you!

spec/mongoid/extensions/date_spec.rb
p-mongo: Will be done in https://jira.mongodb.org/browse/MONGOID-5425.

p-mongo (approved):
Approved with everything we discussed.

lib/mongoid/extensions/set.rb
Neilshweky: done

spec/mongoid/extensions/set_spec.rb
Neilshweky: done

lib/mongoid/extensions/date.rb
Neilshweky: done

p-mongo (approved):

lib/mongoid/attributes.rb
p-mongo: @johnnyshields The reasoning for this change is as follows:

1. The overall epic being worked on is consistent type conversion of values (https://jira.mongodb.org/browse/MONGOID-5332).
2. For Mongoid 8.0, we decided to replace unconvertable input with nil for the value written into `attributes`. The original input is stored in `attributes_before_type_cast`.
3. This behavior applies to all field types, per the goal of the epic.
4. This PR implements the nil replacement behavior as just outlined.
5. I believe there is a valid use case for erroring (i.e. raising an exception) during type conversion instead of replacing with nil (https://jira.mongodb.org/browse/MONGOID-5382).
6. Erroring is not part of this PR and is not technically required for 8.0, since it can potentially be implemented as a compatible change in 8.1.

Your argument that the raise for non-Array/Hash incoming values for Array/Hash fields is out of scope of this PR because it's done in a method not called `mongoize` is nitpicking. Your opinion on this matter is noted, and your repeated posting of this opinion is off topic or if you prefer, spam.

If you wish to propose an adjustment to assignment behavior as explained in https://jira.mongodb.org/browse/MONGOID-5222, please leave a comment in that ticket. You need to write a *constructive* comment, meaning it needs to address the requirements in a plausibly better way than the existing implementation being proposed in this PR.

If you have any thoughts you haven't already expressed in regard to failing attribute assignment and/or attribute reads from database with exceptions, please add those to https://jira.mongodb.org/browse/MONGOID-5382, also in a constructive manner. Your current position there is that that ticket is not needed at all. Your opinion that instead of having an option for any field type to fail assignments of unconvertable input Mongoid should retain ad-hoc checks for some types scattered through the codebase is noted, we acknowledge that this is existing behavior of Mongoid and we are looking to improve it.

johnnyshields:
Thank you for this PR Neil, a lot of users are going to benefit 🎉

PR #5338, opened 2022-06-21
MONGOID-5381 Correct map_big_decimal_to_decimal128 feature flag default and write 8.0 release note

Neilshweky:
We decided to not include `map_big_decimal_to_decimal128` in a 7.x release, therefore the default of the feature flag should equal the 7.0 behavior. I have corrected the default in the code, and removed the reference to changing the feature flag in 8.0 (since it's first being introduced in 8.0). I also added a release note for the flag.

### DIFF
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -40,7 +40,6 @@ changed in Mongoid 8.0:
 - ``:legacy_attributes`` => true
 - ``:legacy_pluck_distinct`` => ``false``
 - ``:legacy_triple_equals`` => ``false``
-- ``:map_big_decimal_to_decimal128`` => ``true``
 - ``:object_id_as_json_oid`` => ``false``
 - ``:overwrite_chained_operators`` => ``false``
 
--- a/docs/release-notes/mongoid-8.0.txt
+++ b/docs/release-notes/mongoid-8.0.txt
@@ -475,12 +475,13 @@ it was stored and returned as a String.
 ``Decimal128``-backed ``BigDecimal`` fields
 -------------------------------------------
 
-Mongoid 8, introduces the ``map_big_decimal_to_decimal128`` feature flag, which
+Mongoid 8 introduces the ``map_big_decimal_to_decimal128`` feature flag, which
 allows values assigned to a field of type ``BigDecimal`` to be stored as type
-``Decimal128`` in the database. In Mongoid 7 and earlier, and in Mongoid 8 with
-this feature flag turned off, values assigned to fields of type ``BigDecimal``
-were stored in the database as Strings. In Mongoid 8, with this feature flag
-on, values assigned to fields of type ``BigDecimal`` are stored as type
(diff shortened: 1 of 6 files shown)


p-mongo:
I think the behavior in 8.0 should be the one we desire going forward, which from what I understand is the one currently in master?

Neilshweky:
> I think the behavior in 8.0 should be the one we desire going forward, which from what I understand is the one currently in master?

This will be the first time we are introducing the user to this feature flag, I thought protocol was to keep it with previous behavior... 

Or should we use the new behavior because we have a new major version?

Neilshweky:
@p-mongo this is rfal

docs/release-notes/mongoid-8.0.txt
```diff
@@ -472,3 +471,16 @@ it was stored and returned as a String.
   p registry.data
   # => Mongoid 7: "data!"
   # => Mongoid 8: <BSON::Binary:0x2600 type=generic data=0x6461746121...>
+
+``Decimal128``-backed ``BigDecimal`` fields
+-------------------------------------------
+
+Mongoid 8, introduces the ``map_big_decimal_to_decimal128`` feature flag, which
+allows values assigned to a field of type ``BigDecimal`` to be stored as type
+``Decimal128`` in the database. In Mongoid 7 and earlier, and in Mongoid 8 with
```
p-mongo, line 480:
```suggestion
``String`` in the database, for compatibility with Mongoid 7 and earlier. In Mongoid 8 by default and with
```

docs/release-notes/mongoid-8.0.txt
```diff
@@ -472,3 +471,16 @@ it was stored and returned as a String.
...
+this feature flag turned off, values assigned to fields of type ``BigDecimal``
+were stored in the database as Strings. In Mongoid 8, with this feature flag
```
p-mongo, line 482:
```suggestion
are stored in the database as ``BSON::Decimal128``.
```

docs/release-notes/mongoid-8.0.txt
```diff
@@ -472,3 +471,16 @@ it was stored and returned as a String.
...
+on, values assigned to fields of type ``BigDecimal`` are stored as type
```
p-mongo, line 483: And this needs to be adjusted also.

p-mongo (approved):
The prose needs to say that the feature flag is added for compatibility with 7.x behavior, not that the feature flag yields new behavior.

docs/release-notes/mongoid-8.0.txt
Neilshweky: done

docs/release-notes/mongoid-8.0.txt
Neilshweky: done

docs/release-notes/mongoid-8.0.txt
Neilshweky: done

PR #5341, opened 2022-06-22
MONGOID-5037 use $pull instead of $pullAll for when documents change before destroying all

Neilshweky:


johnnyshields:
> This change relies on the assumption that all embedded documents have _id fields, which I believe they do.

@Neilshweky please see this ticket https://jira.mongodb.org/browse/MONGOID-5025 and my comments on it.

johnnyshields:
What about doing something like `$pullAll => { "_id" => { "$exists" => true } }` ?

lib/mongoid/association/embedded/batchable.rb
```diff
@@ -50,20 +50,39 @@ def batch_clear(docs)
...
         #
         # @example Batch remove the documents.
         #   batchable.batch_remove([ doc_one, doc_two ])
         #
         # @param [ Array<Document> ] docs The docs to remove.
         # @param [ Symbol ] method Delete or destroy.
         def batch_remove(docs, method = :delete)
-          removals = pre_process_batch_remove(docs, method)
+          # If the _id is nil, we cannot use $pull and delete by searching for
+          # the id. Therefore we have to user pullAll with the documents
```
p-mongo, line 62:
```suggestion
          # the id. Therefore we have to user pullAll with the documents'
```

lib/mongoid/association/embedded/batchable.rb
```diff
@@ -50,20 +50,39 @@ def batch_clear(docs)
...
+          # attributes.
+          pulls, pull_alls = pre_process_batch_remove(docs, method).partition do |o|
+            o.key?("_id") && !o["_id"].nil?
```
p-mongo, line 65: The `key?` check here is redundant, no?

p-mongo (approved):

lib/mongoid/association/embedded/batchable.rb
Neilshweky, line 62:
```suggestion
          # the id. Therefore we have to use pullAll with the documents'
```

johnnyshields:
@Neilshweky this PR looks good but just so we have a record here, what's the final outcome if a write of a non-_id doc arrives at the same time `.destroy_all` is called?

This case will of course be exceedingly rare, and not happen at all in standard usage of Mongoid. If users do expect it, they can use a **transaction** to mitigate the impact (i.e. the read which detects _id would be guaranteed to be synchronized with the $pullAll or $pull). Perhaps we can add a note in the documentation to that effect?

PR #5344, opened 2022-06-24
MONGOID-5270 Implement .tally method for Mongoid#Criteria

Neilshweky:


johnnyshields:
@Neilshweky thanks for this.

One thought--potentially a follow-up PR is to accept a splat of args `tally(*fields)`, which similar to `pluck` would return results as an array (`[ { result1 }, { result2 } ]`). This would allow us to get multiple fields in a single database pass, saving on CPU.

lib/mongoid/contextual/mongo.rb
```diff
@@ -454,6 +454,73 @@ def pluck(*fields)
...
+      #
+      #   { _id: 1, array: [ { x: 1 }, { x: 2 }, { y: 3 } ] }
+      #
+      #   Model.tally("array.x")
+      #   # => { [ 1, 2 ] => 1 }
+      #
+      # @param [ String | Symbol ] field_name The field name.
+      #
+      # @return [ Hash ] The hash of counts.
+      def tally(field_name)
```
johnnyshields, line 494: We also need mirror implementations for `Contextual::Memory` and `Contextual::None`

lib/mongoid/contextual/mongo.rb
```diff
@@ -454,6 +454,73 @@ def pluck(*fields)
...
+          # demongoize to the same value. A good example of when this happens
+          # is when using localized fields. While the server query won't group
+          # together hashes that have other values in different languages, the
+          # demongoized value is just the translation in the current locale,
+          # which can be the same across multiple of those unequal hashes.
+          tallies[key] ||= 0
+          tallies[key] += doc["counts"]
+          tallies
+        end
+      end
```
johnnyshields, line 523: Need a blank line after this one.

lib/mongoid/contextual/mongo.rb
```diff
@@ -454,6 +454,73 @@ def pluck(*fields)
         end
       end
 
+      # Get a hash of counts for the values of a single field. For example,
```
johnnyshields, line 457: I would add `"using the MongoDB aggregation pipeline"` here to make it explicit that it's a pipeline calculation. Although its an implementation detail, I think it's important for users to know how is works under-the-hood to understand the performance implications of using it on large collections.

johnnyshields:
> Does this need a release note?

Definitely needs a release note! 🎉

lib/mongoid/contextual/mongo.rb
p-mongo: What would be the other option?

johnnyshields: For example, my (#pluck_each PR)[https://github.com/mongodb/mongoid/pull/5332], you could use that to do a relatively efficient tally in Ruby memory (using GETMORE commands, etc.)

spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -813,6 +813,372 @@
...
+
+    let(:criteria) { Band.where(origin: "tally") }
+
+    context "when tallying a string" do
+      let(:tally) do
+        criteria.tally(:name)
+      end
+
+      it "returns the correct hash" do
+        expect(tally).to eq("Depeche Mode" => 1, "New Order" => 1, "10,000 Maniacs" => 1)
```
johnnyshields, line 846: would be good to add a basic test for what happens if the key is missing here, i.e. `nil`. Would be good to include `nil` in the results, as ActiveSupport#tally does this today:

```
[1,2,2,nil,3].tally
#=> {1=>1, 2=>2, nil=>1, 3=>1}
```

lib/mongoid/contextual/mongo.rb
```diff
@@ -454,6 +454,74 @@ def pluck(*fields)
...
+      #   { [ 1, 2 ] => 2, [ 1, 3 ] => 1 }
+      #
+      # Note that if tallying an element in an array of hashes, and the key
+      # doesn't exist in some of the hashes, tally will not include those
+      # nil keys in the resulting hash:
+      #
+      #   { _id: 1, array: [ { x: 1 }, { x: 2 }, { y: 3 } ] }
+      #
+      #   Model.tally("array.x")
+      #   # => { [ 1, 2 ] => 1 }
```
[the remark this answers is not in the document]
Neilshweky: server behavior does not include `nil`, I believe duplicates would be there but I can add a test

spec/mongoid/contextual/mongo_spec.rb
Neilshweky: done

lib/mongoid/contextual/memory.rb
```diff
@@ -414,6 +422,59 @@ def prepare_remove(doc)
...
+            # If this is a localized field, and there are remaining, get the
+            # _translations hash so that we can get the specified translation in
+            # the remaining
+            if field&.localized?
+              document.send("#{segment}_translations")
+            end
+          end
+          res.nil? ? document.send(segment) : res
+        elsif document.is_a?(Hash)
+          document.with_indifferent_access.fetch(segment, nil)
```
p-mongo, line 464: This will repeatedly perform conversions of parts of (nested) hashes. When is this elsif branch taken that document is a Hash?

p-mongo: I'm pretty sure it would be better to try to fetch `segment`, then `segment.to_sym`, but I am quite curious now why there are two types of keys possible at all.

lib/mongoid/contextual/memory.rb
```diff
@@ -414,6 +422,59 @@ def prepare_remove(doc)
...
+        else
+          nil
+        end
+
+        return curr unless remaining
+
+        if curr.is_a?(Array)
+          # compact is used for consistency with server behavior.
+          curr.map { |d| retrieve_value(d, remaining) }.compact
```
p-mongo, line 473: If you move `compact` before `map` would that achieve the same result?

lib/mongoid/contextual/mongo.rb
```diff
@@ -704,11 +773,15 @@ def acknowledged_write?
       # @param [ Object ] The value for the given field name
       def extract_value(attrs, field_name)
         def fetch_and_demongoize(d, meth, klass)
```
p-mongo, line 775: Is this nested method definition required to be nested? https://stackoverflow.com/questions/39313535/ruby-doesnt-have-nested-methods-right-so-what-is-this-and-is-there-a-better

Doesn't look like it uses any variables not local to it, so it should be un-nested and gain a docstring?

docs/reference/queries.txt
```diff
@@ -1380,6 +1380,27 @@ Mongoid also has some helpful methods on criteria.
...
 
+   * - ``Criteria#tally``
+
+       *Get a mapping of values to counts for the provided field.*
+
+       *This method accepts the dot notation, thus permitting referencing
+       fields in embedded associations.*
+
+       *This method respects :ref:`field aliases <field-aliases>`,
+       including those defined in embedded documents.*
```
p-mongo, line 1391: And also expands storage field names, right?

p-mongo (commented):
Everything looks good except for the indifferent access conversion which I have a question on.

p-mongo (approved):
Approved with 1) two hash references instead of conversion of entire hash to indifferent access and 2) note added for 5410.

lib/mongoid/contextual/memory.rb
```diff
@@ -414,6 +422,59 @@ def prepare_remove(doc)
...
+      # Note that the result is in an array since accounts is an array. If it
+      # was nested in two arrays the result would be in a 2D array.
+      #
+      # @param [ Object ] document The object to traverse the field path.
+      # @param [ String ] field_path The dotted string that represents the path
+      #   to the value.
+      #
+      # @return [ Object | nil ] The value at the given field path or nil if it
+      #   doesn't exist.
+      def retrieve_value(document, field_path)
```
johnnyshields, line 446: This method name is quite generic, would suggest something more descriptive like `resolve_dot_notation_path` etc.

lib/mongoid/contextual/mongo.rb
johnnyshields: This is fixed in https://github.com/mongodb/mongoid/pull/5332

lib/mongoid/contextual/memory.rb
johnnyshields: > # compact is used for consistency with server behavior.

This may be true for `tally`, but it's not true for other methods like `pluck` which we've discussed applying this for. I would suggest that this method should return the nils, and the tally method specifically can strip them out in a second pass.

p-mongo: If retrieve_value demongoizes then it can in principle convert non-nil input to nil output, right? Then compact should stay at the end.

Neilshweky: I will handle this when I do this for pluck.

p-mongo (approved):

lib/mongoid/contextual/mongo.rb
```diff
@@ -703,21 +792,13 @@ def acknowledged_write?
...
-          if field = klass.fields[meth]
-            field.demongoize(res)
-          else
-            res.class.demongoize(res)
-          end
-        end
-
         i = 1
         num_meths = field_name.count('.') + 1
         k = klass
```
johnnyshields, line 797: @Neilshweky it seems that as a result of this PR, this local variable `k = klass` is no longer used (it is assigned also on line 837, but it's never read.)

PR #5345, opened 2022-06-24
MONGOID-3942 Test: Eager loading no longer broken when iterating criteria a second time

Neilshweky:


p-mongo (approved):
Can you retitle this please and make sure the merge commit gets the new title

PR #5346, opened 2022-06-25
MONGOID-5399: Contextual #map should take a splat of field args

johnnyshields:
Fixes the following issues:
- MONGOID-5399: Contextual #map should take a splat of field args
- MONGOID-5148: Add missing #map to Contextual::Memory and None


johnnyshields:
Closed due to MONGOID-5426

PR #5349, opened 2022-07-01
MONGOID-5311 Drop support for Rails < 5.2 and Ruby < 2.6

Neilshweky:


p-mongo:
JRuby 9.2 pretends to be Ruby 2.5 and doesn't work, these configurations need to be removed or changed to 9.3.

p-mongo:
Also it doesn't look like these versions have been deprecated, if so this needs to happen and be backported to at least 7.5 and if the commit cherry-picks to 7.4 and 7.3.

p-mongo (approved):
I approve this but the deprecation still needs to be added to docs for both master and 7.* stable.

docs/release-notes/mongoid-7.5.txt
```diff
@@ -18,14 +18,28 @@ please consult GitHub releases for detailed release notes and JIRA for
 the complete list of issues fixed in each release, including bug fixes.
 
 
+Ruby 2.5 Deprecated
+-------------------
+
+Mongoid 7.5 deprecates support for Ruby 2.5.
```
p-mongo, line 24: As well as JRuby 9.2 due to it presenting itself as Ruby 2.5.

p-mongo (approved):
Readme says Mongoid supports jruby 9.2, needs to be 9.3 with this PR.

The compat tables for JRuby also need to be updated to deprecate/remove 9.2 support as is done with 2.5.

docs/reference/compatibility.txt
```diff
@@ -91,13 +95,25 @@ is deprecated.
...
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - D
+     -
+     -
+     -
+     - |checkmark|
+     -
```
Neilshweky, line 115: @p-mongo we have a problem, if we deprecate JRuby-9.2 in 7.5, there will be no available version of JRuby, since we don't support JRuby-9.3 until 8.0

p-mongo:
We should operate the same with 9.3 as we do with 9.2, so probably we would need to add 9.3 as a supported version for 7.5.

PR #5351, opened 2022-07-06
MONGOID-5270: Follow-ups to tally PR

johnnyshields:
This contains a few things I missed in code review of #5344
1. Remove unused variable `k` in `Contextual::Mongo#extract_value`
2. Add missing code comment for `Contextual::Memory#tally`
3. Minor code comment fixes

### DIFF
--- a/lib/mongoid/contextual/memory.rb
+++ b/lib/mongoid/contextual/memory.rb
@@ -74,7 +74,7 @@ module Mongoid
       # @example Get the distinct values.
       #   context.distinct(:name)
       #
-      # @param [ String, Symbol ] field The name of the field.
+      # @param [ String | Symbol ] field The name of the field.
       #
       # @return [ Array<Object> ] The distinct values for the field.
       def distinct(field)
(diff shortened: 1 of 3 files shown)


Neilshweky (approved):
Nice

lib/mongoid/contextual/memory.rb
```diff
@@ -200,6 +200,14 @@ def pluck(*fields)
         documents.pluck(*fields)
       end
 
+      # Tally the field values in memory.
+      #
+      # @example Get the counts of values in memory.
+      #   context.tally(:name)
+      #
+      # @param [ String, Symbol ] field Field to tally.
```
p-mongo, line 208: `String | Symbol` per the convention you referred to earlier right?

p-mongo (approved):

lib/mongoid/contextual/memory.rb
johnnyshields: Yeah, ha I had a feeling you'd point that out. Actually Neil used `String, Symbol` in his PR so I just kept consistent with him. **Fixed now.**

I've raised https://jira.mongodb.org/browse/MONGOID-5415 to switch to pipe everywhere, as copy-pasting will likely lead to commas appearing more in the future.

PR #5354, opened 2022-07-06
MONGOID-4547 Implement #take / #take! for AR feature parity

Neilshweky:
A decision I made here is that `take` should return documents instead of a criteria. This draws a distinction between `take` and `limit`, as well as I think it would be a little strange if we returned a criteria on just `Person.take` (i.e. without params) since only one document is returned.

TODO:

- memory
- none
- release note
- crud documentation

### DIFF
--- a/docs/release-notes/mongoid-7.5.txt
+++ b/docs/release-notes/mongoid-7.5.txt
@@ -126,3 +126,35 @@ return a ``Hash`` when instantiating a new document:
   band = Band.first
   p band.attributes.class
   # => BSON::Document
+
+
+Implemented ``Criteria#take`` Method
+------------------------------------
+
+Mongoid 7.5 introduces the ``#take`` method which returns an unsorted document
+or list of documents from the database:
+
+.. code:: ruby
+
+  class Band
+    include Mongoid::Document
+  end
+
+  Band.create!
+  Band.create!
+
+  Band.take
+  # => #<Band _id: 62c835813282a4470c07d530, >
+  Band.take(2)
+  # => [ #<Band _id: 62c835813282a4470c07d530, >, #<Band _id: 62c835823282a4470c07d531, > ]
+
+If a parameter is given to ``#take``, a list is returned. If no parameter is
+given, a singular document is returned.
+
+Note that the ``#take`` method does not apply a sort to the view before retrieving
+the documents from the database, and therefore it may return different results
+than the ``#first`` and ``#last`` methods. This method is equivalent to calling
+``#first`` and ``#last`` with the ``{ id_sort: :none }`` option. This option
+has been deprecated in Mongoid 7.5 and it is recommended to use ``#take``
+instead going forward. Support for the ``:id_sort`` option will be dropped in
+Mongoid 8.
--- a/docs/release-notes/mongoid-7.5.txt
+++ b/docs/release-notes/mongoid-7.5.txt
@@ -131,8 +131,8 @@ return a ``Hash`` when instantiating a new document:
 Implemented ``Criteria#take/take!`` Method
 ------------------------------------------
 
-Mongoid 7.5 introduces the ``#take`` method which returns an unsorted document
-or list of documents from the database:
+Mongoid 7.5 introduces the ``#take`` method which returns a document
+or a set of documents from the database without ordering by ``_id``:
 
 .. code:: ruby
 
@@ -148,12 +148,11 @@ or list of documents from the database:
   Band.take(2)
   # => [ #<Band _id: 62c835813282a4470c07d530, >, #<Band _id: 62c835823282a4470c07d531, > ]
 
-If a parameter is given to ``#take``, a list is returned. If no parameter is
+If a parameter is given to ``#take``, an array of documents is returned. If no parameter is
 given, a singular document is returned.
 
-The ``#take!`` method does not take parameters but otherwise functions the same
-as calling ``#take``. The ``#take!`` method returns a single document, and
-raises an error if none are found.
+The ``#take!`` method functions the same as calling ``#take`` without arguments,
+but raises an error instead of returning nil if no documents are found.
 
 .. code:: ruby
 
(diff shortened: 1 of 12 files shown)


johnnyshields:
Agreed with decision to return documents. This mirrors what ActiveRecord does.

lib/mongoid/contextual/mongo.rb
```diff
@@ -400,6 +400,41 @@ def limit(value)
...
+      #
+      # @example Take 10 documents
+      #   context.take(10)
+      #
+      # @param [ Integer ] value The number of documents to return.
+      #
+      # @return [ Document | Array<Document> ] The list of documents, or one
+      #   document if no value was given.
+      def take(value = nil)
+        if value
```
Neilshweky, line 413: should I use the cache here?

lib/config/locales/en.yml
```diff
@@ -427,6 +427,11 @@ en:
             \_\_\_\_\_\_default:\n
             \_\_\_\_\_\_\_\_hosts:\n
             \_\_\_\_\_\_\_\_\_\_- localhost:27017\n\n"
+        no_documents_found:
+          message: "Could not find a document of class %{klass}."
+          summary: "Mongoid attempted to find a document from the class %{klass}
```
p-mongo, line 432:
```suggestion
          summary: "Mongoid attempted to find a document of the class %{klass}
```

lib/mongoid/contextual/memory.rb
```diff
@@ -164,6 +164,39 @@ def last
...
       end
 
+      # Take the given number of documents from the database.
+      #
+      # @example Take a document.
+      #   context.take
+      #
+      # @param [ Integer ] value The number of documents to take.
+      #
+      # @return [ Document ] The document.
```
p-mongo, line 174: or nil?

lib/config/locales/en.yml
```diff
@@ -427,6 +427,11 @@ en:
...
+            but none exist."
+          resolution: "Create a document of class %{klass}."
```
p-mongo, line 434:
```suggestion
          resolution: "Create a document of class %{klass} or use .take."
```

lib/mongoid/contextual/memory.rb
```diff
@@ -164,6 +164,39 @@ def last
...
+      def take(value = nil)
```
p-mongo, line 175:
```suggestion
      def take(limit = nil)
```

lib/mongoid/contextual/memory.rb
```diff
@@ -164,6 +164,39 @@ def last
...
+        if value
+          eager_load(documents.take(value))
+        else
+          eager_load([documents.first]).first
+        end
+      end
+
+      # Take the given number of documents from the database.
```
p-mongo, line 183: This method doesn't take arguments?

lib/mongoid/contextual/mongo.rb
```diff
@@ -400,6 +400,44 @@ def limit(value)
...
 
+      # Take the given number of documents from the database.
+      #
+      # @example Take 10 documents
+      #   context.take(10)
+      #
+      # @param [ Integer ] value The number of documents to return.
+      #
+      # @return [ Document | Array<Document> ] The list of documents, or one
+      #   document if no value was given.
```
p-mongo, line 411: or nil?

lib/mongoid/contextual/mongo.rb
```diff
@@ -400,6 +400,44 @@ def limit(value)
...
+      def take(value = nil)
```
p-mongo, line 412:
```suggestion
      def take(limit = nil)
```

lib/mongoid/contextual/none.rb
```diff
@@ -125,6 +125,28 @@ def initialize(criteria)
...
 
+      # Returns nil or empty array.
+      #
+      # @example Take a document in null context.
+      #   context.take
+      #
+      # @param [ Integer ] value The number of documents to take.
+      #
+      # @return [ [] | nil ] Empty array or nil.
+      def take(value = nil)
```
p-mongo, line 136:
```suggestion
      def take(limit = nil)
```

p-mongo:
I am confused how https://apidock.com/rails/v5.2.3/ActiveRecord/FinderMethods/take%21 says it's "the same as `take`" while not taking arguments that `take` takes?

lib/mongoid/contextual/memory.rb
Neilshweky: I don't think I can use `limit` as the parameter name since we already have a method called `limit` that's used in this method) and there will be collisions between the names

Neilshweky: Actually, looks like it works!

lib/mongoid/contextual/memory.rb
Neilshweky: it does not.

lib/mongoid/contextual/mongo.rb
Neilshweky: done

lib/mongoid/contextual/none.rb
Neilshweky: done

lib/mongoid/contextual/memory.rb
Neilshweky: done

Neilshweky:
> I am confused how https://apidock.com/rails/v5.2.3/ActiveRecord/FinderMethods/take%21 says it's "the same as `take`" while not taking arguments that `take` takes?

A little same, but I can confirm `take!` does not take arguments in AR. I did try it.

lib/mongoid/contextual/memory.rb
p-mongo: You can call the method via `self.limit` to disambiguate when not passing it any arguments, same to access an instance attribute (technically also a method) when there is a local variable of the same name.

lib/config/locales/en.yml
p-mongo: How about something like "or use a non-bang finder method"?

p-mongo: Or "or use a finder method that returns nil when no documents are found instead of raising an exception"

docs/release-notes/mongoid-7.5.txt
```diff
@@ -126,3 +126,44 @@ return a ``Hash`` when instantiating a new document:
...
+raises an error if none are found.
+
+.. code:: ruby
+
+  Band.take!
+  # => #<Band _id: 62c835813282a4470c07d530, >
+
+Note that the ``#take/take!`` methods do not apply a sort to the view before
+retrieving the documents from the database, and therefore it may return different
+results than the ``#first`` and ``#last`` methods. This method is equivalent to
```
p-mongo, line 165:
```suggestion
results than the ``#first`` and ``#last`` methods. ``#take`` is equivalent to
```

docs/release-notes/mongoid-7.5.txt
```diff
@@ -126,3 +126,44 @@ return a ``Hash`` when instantiating a new document:
...
+calling ``#first`` and ``#last`` with the ``{ id_sort: :none }`` option. This
```
p-mongo, line 166:
```suggestion
calling ``#first`` or ``#last`` with the ``{ id_sort: :none }`` option. This
```

p-mongo (approved):

lib/config/locales/en.yml
johnnyshields: `Create a document of class %{klass}, or use a finder method variant that does not raise an exception when no documents are found.`

(shouldn't assume return type in this error, though all are currently returning `nil`)

PR #5355, opened 2022-07-07
MONGOID-5107: Remove reference to mongoid.github.io

johnnyshields:
The https://mongoid.github.io/ looks broken CSS-wise and think these versions are deep, deep EOL, so should just remove.

(If you want to keep this, please close MONGOID-5107)


p-mongo:
The link is to http://mongoid.github.io/ which renders correctly.

johnnyshields:
This is rendering correctly?
image

PR #5358, opened 2022-07-07
MONGOID-4998 deprecate :id_sort and implement limit positional argument on #first/#last

Neilshweky:
PLEASE READ THIS before reviewing this PR. There's a lot going on here, and a lot still left to do.

First, this change is actually a few different changes wrapped in one (besides all of the #take stuff under it):
1) remove support for the `id_sort` option
2) add support for the `limit` option 
3) add support for passing options in all of our first/last methods, so that users don't get errors when they pass in options to different first/lasts. Note that our syntax differs from Ruby's. We use keyword args (well actually, a hash) whereas Ruby uses an optional arg to pass the limit. This is annoying, but I fear it would be too complicated to completely do away with the hash in favor of just a limit integer. 

Things still left to do:
- Release notes about changes to first/last
- Deprecate the id_none option in 7.5 and add a release note

Note that this PR should not be merged before that of MONGOID-4547, since this deprecation/removal of the id_none option only works once we have take. take provides the functionality that the id_none option did in first/last.

Future directives: 
- first! method
- last! method
- other take-like methods: `second`, `third` etc.

### DIFF
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1303,7 +1303,7 @@ Mongoid also has some helpful methods on criteria.
 
        *Finds a single document given the provided criteria. This automatically adds a sort on _id.
        Opt out of adding the id sort with the {id_sort: :none} option. This option is deprecated.
-       Get a list of documents by passing in a limit argument.*
+       please call ``take`` instead. Get a list of documents by passing in a limit argument.*
 
      -
         .. code-block:: ruby
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1302,7 +1302,8 @@ Mongoid also has some helpful methods on criteria.
    * - ``Criteria#first|last``
 
        *Finds a single document given the provided criteria. This automatically adds a sort on id.
-       Opt out of adding the id sort with the {id_sort: :none} option.*
+       Opt out of adding the id sort with the {id_sort: :none} option. This option is deprecated.
+       Get a list of documents by passing in a limit argument.*
 
      -
         .. code-block:: ruby
@@ -1310,6 +1311,7 @@ Mongoid also has some helpful methods on criteria.
           Band.first
           Band.where(:members.with_size => 3).first
           Band.where(:members.with_size => 3).last
+          Band.first(2)
 
    * - ``Criteria#first_or_create``
 
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1302,7 +1302,7 @@ Mongoid also has some helpful methods on criteria.
    * - ``Criteria#first|last``
 
        *Finds a single document given the provided criteria. This automatically adds a sort on _id.
-       Opt out of adding the id sort with the {id_sort: :none} option. This option is deprecated.
+       Opt out of adding the id sort with the {id_sort: :none} option. This option is deprecated -
        please call ``take`` instead. Get a list of documents by passing in a limit argument.*
 
      -
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1301,7 +1301,7 @@ Mongoid also has some helpful methods on criteria.
 
    * - ``Criteria#first|last``
 
-       *Finds a single document given the provided criteria. This automatically adds a sort on id.
+       *Finds a single document given the provided criteria. This automatically adds a sort on _id.
        Opt out of adding the id sort with the {id_sort: :none} option. This option is deprecated.
        Get a list of documents by passing in a limit argument.*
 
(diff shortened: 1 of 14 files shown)


docs/reference/queries.txt
```diff
@@ -1301,15 +1301,17 @@ Mongoid also has some helpful methods on criteria.
 
    * - ``Criteria#first|last``
 
-       *Finds a single document given the provided criteria. This automatically adds a sort on id.
-       Opt out of adding the id sort with the {id_sort: :none} option.*
+       *Finds a single document given the provided criteria. This automatically adds a sort on _id.
+       Opt out of adding the id sort with the {id_sort: :none} option. This option is deprecated.
+       Get a list of documents by passing in a limit argument.*
```
p-mongo, line 1306:
```suggestion
       please call ``take`` instead. Get a list of documents by passing in a limit argument.*
```

docs/reference/queries.txt
```diff
@@ -1301,15 +1301,17 @@ Mongoid also has some helpful methods on criteria.
 
    * - ``Criteria#first|last``
 
-       *Finds a single document given the provided criteria. This automatically adds a sort on id.
-       Opt out of adding the id sort with the {id_sort: :none} option.*
+       *Finds a single document given the provided criteria. This automatically adds a sort on _id.
+       Opt out of adding the id sort with the {id_sort: :none} option. This option is deprecated.
```
p-mongo, line 1305:
```suggestion
       Opt out of adding the id sort with the {id_sort: :none} option. This option is deprecated -
```

docs/release-notes/mongoid-7.5.txt
```diff
@@ -188,3 +188,30 @@ calling ``#first`` or ``#last`` with the ``{ id_sort: :none }`` option. This
 option has been deprecated in Mongoid 7.5 and it is recommended to use ``#take``
 instead going forward. Support for the ``:id_sort`` option will be dropped in
 Mongoid 8.
+
+
+Deprecate ``:id_sort`` Option and Support ``limit`` on ``#first/last``
+----------------------------------------------------------------------
+
+Mongoid 7.5 deprecates support for the ``:id_sort`` keyword argument for the
```
p-mongo, line 196:
```suggestion
Mongoid 7.5 deprecates the ``:id_sort`` keyword argument for the
```

docs/release-notes/mongoid-7.5.txt
```diff
@@ -188,3 +188,30 @@ calling ``#first`` or ``#last`` with the ``{ id_sort: :none }`` option. This
...
+``Criteria#first`` and ``Criteria#last`` methods. These methods have added
```
p-mongo, line 197:
```suggestion
``Criteria#first`` and ``Criteria#last`` methods. Please use ``Criteria#take`` to retrieve documents without sorting by id. ```

File: docs/release-notes/mongoid-7.5.txt [hunk 5]
```diff
@@ -188,3 +188,30 @@ calling ``#first`` or ``#last`` with the ``{ id_sort: :none }`` option. This
...
+support for adding a limit as a positional argument:
```
p-mongo, line 198:
```suggestion
The ``first`` and ``last`` methods now take the number of documents to return as a positional argument, mirroring the functionality of Ruby's ``Enumerable`` method and ActiveRecord's ``first`` and ``last`` methods. Both invocations (with limit as positional arguments and with ``:id_sort`` option) remain supported in Mongoid 7.x, but the ``:id_sort`` invocation will be removed in Mongoid 8.
```

lib/mongoid/contextual/memory.rb
```diff
@@ -116,9 +116,16 @@ def exists?
...
       #   context.first
       #
+      # @param [ Integer | Hash ] limit_or_opts The number of documents to
+      #   return, or a hash of options.
+      #
       # @return [ Document ] The first document.
-      def first(*args)
-        eager_load([documents.first]).first
+      def first(limit_or_opts = nil)
+        if !limit_or_opts || limit_or_opts.is_a?(Hash)
```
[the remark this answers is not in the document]
Neilshweky: done

lib/mongoid/contextual/none.rb
```diff
@@ -117,13 +117,35 @@ def initialize(criteria)
...
       # @example Get the last document in null context.
       #   context.last
       #
+      # @param [ Integer | Hash ] limit_or_opts The number of documents to
+      #   return, or a hash of options.
+      #
       # @return [ nil ] Always nil.
-      def last; nil; end
+      def last(limit_or_opts = nil)
+        if limit_or_opts && !limit_or_opts.is_a?(Hash)
```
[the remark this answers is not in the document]
Neilshweky: done

PR #5360, opened 2022-07-07
MONGOID-5320: Make MissingAttributeError more informative

johnnyshields:


Neilshweky (commented):
Definitely agree with Oleg's comment.

Also while you're doing this, do you think you can standardize the message somewhere? Maybe make a Mongoid error that subclasses the ActiveSupport::MissingAttribute error?

I think you would have to subclass it so that it won't be so backwards breaking, but I like the idea of having a more standard error than having to change the message in two places.

johnnyshields:
Thanks for comments, will do those. @Neilshweky great idea about subclass.

Neilshweky:
What's the status on this @johnnyshields?

johnnyshields:
Please give me another week, will try to get to it.

johnnyshields:
@Neilshweky I've investigated this a bit further, and it turns out that our usage of MissingAttribute is quite different than 

Mongoid raises MissingAttribute when using the `.only` method, like so:

```ruby
Band.only(:name).first.label
```

This is from the Rails docs:

```
  # Raised when an attribute is not defined.
  #
  #   class User < ActiveRecord::Base
  #     has_many :pets
  #   end
  #
  #   user = User.first
  #   user.pets.select(:id).first.user_id
  #   # => ActiveModel::MissingAttributeError: missing attribute: user_id
```

In this case, Mongoid simply raises a plain-old `MethodMissing` error.

I think we should consider renaming this error to `Mongoid::Errors::FieldNotLoaded` to better clarify its intention. We can still inherit from ActiveModel::MissingAttributeError for now, and perhaps deprecate that in the future.

johnnyshields:
Ready for review. In order to make this error behave like the others, I extracted out a mixin `ErrorComposable`

PR #5361, opened 2022-07-07
MONGOID-5407 Add phantom custom field type example


p-mongo:


johnnyshields:
How about `#mongoize` / `#demongoize` defined on object instance level? In this example, would Mongoid ever attempt to call `"black".mongoize` (i.e. instead of `ColorMapping.mongoize("black")`) and yield unexpected behavior?

Unless we have tested everything very thoroughly here and are confident the above is not a concern, I would suggest to hold off on this until [MONGOID-5409](https://jira.mongodb.org/browse/MONGOID-5409) is done. After MONGOID-5409, "Phantom Custom Field Types" will become the standard case (i.e. "Custom Field Adapters"), so if that ticket is coming soon it may also make sense to defer this.

p-mongo:
https://jira.mongodb.org/browse/MONGOID-5407

p-mongo:
@johnnyshields I'm afraid you misunderstood the purpose of this PR and the corresponding ticket.

The documentation added here documents existing behavior of Mongoid and is intended to help with figuring out what to do in the future, such as whether to implement 5409 and if so how.

docs/reference/fields.txt
```diff
@@ -994,6 +994,64 @@ when transforming objects for use in Mongoid query criteria.
...
+.. code-block:: ruby
+
+  class ColorMapping
+    
+    MAPPING = {
+      'black' => 0,
+      'white' => 1,
+    }.freeze
+    
+    INVERSE_MAPPING = MAPPING.invert
```
p-mongo, line 1018:
```suggestion
    INVERSE_MAPPING = MAPPING.invert.freeze
```

docs/reference/fields.txt
```diff
@@ -994,6 +994,64 @@ when transforming objects for use in Mongoid query criteria.
...
+    MAPPING = {
+      'black' => 0,
+      'white' => 1,
+    }.freeze
+    
+    INVERSE_MAPPING = MAPPING.invert.freeze
+
+    class << self
+
+      # Takes any possible object and converts it to how it would be
```
p-mongo, line 1022:
```suggestion
      # Takes application-scope value and converts it to how it would be
```

docs/reference/fields.txt
```diff
@@ -994,6 +994,64 @@ when transforming objects for use in Mongoid query criteria.
...
+      # stored in the database. Converts invalid values to nil.
+      def mongoize(object)
+        MAPPING[object]
+      end
+
+      # Get the object as it was stored in the database, and instantiate
+      # this custom class from it. Converts invalid values to nil.
```
p-mongo, line 1029:
```suggestion
      # value. Converts invalid values to nil.
```

johnnyshields:
@p-mongo it's fine, and I'm all for the direction this is headed. My concern is simply that if this documentation is released in Mongoid 8.0, then users will expect this functionality to be "ready for prime time" without caveats/edge-case bugs. If not sure, better not to release. That's all.

johnnyshields:
@p-mongo please refer to my original comment:

> How about #mongoize / #demongoize defined on object instance level? In this example, would Mongoid ever inadvertently attempt to call "black".mongoize (i.e. instead of ColorMapping.mongoize("black")) and yield unexpected behavior?

I am asking specifically "have you tested this case"? If you have not, the given the complexity of Mongoid, you and I should assume there are edge cases there. The responsible thing to do would be to test this case and others thoroughly before releasing documentation to users.

p-mongo:
This PR includes execution results of the pattern being demonstrated. If you want to do something else please create a  ticket for your use case.

johnnyshields:
Yes, you are testing one usage pattern narrowly, and that's a _great_ first step. However, I do not believe that based on this single "happy-path" test case alone that you should advertise to Mongoid users that this usage pattern is "ready for prime time".

The prudent thing to do would be to wait until detailed Rspec tests are added as part of [MONGOID-5409](https://jira.mongodb.org/browse/MONGOID-5409) before releasing this documentation to users. If MONGOID-5409 is certain to come in the next release (8.0), then no issue here.

Anyway, I've said my piece. It's your call.

Neilshweky (approved):

PR #5362, opened 2022-07-07
MONGOID-4998 Drop support for the :id_sort option and only accept  positional argument

Neilshweky:


p-mongo (approved):
This needs a release note that `id_sort` is removed right?

Neilshweky:
> This needs a release note that `id_sort` is removed right?

done

PR #5366, opened 2022-07-09
MONGOID-5415: Code docs: denote type unions with pipe

johnnyshields:
This PR covers only the union comma --> pipe to pipe change. Because this PR is large, splats will be done in a follow-up.

In addition, I've written up the following draft of conventions to use. Comments welcome (please provide a suggested alternate way)

---

#### Draft: Ruby Code Doc Conventions

Use pipe character to denote a union of allowed types:

```ruby
  # @param [ Symbol | String ] name Either a Symbol or a String.
```

Hash: Use comma to denote key and value types:

```ruby
  # @param [ Hash<Symbol, Integer> ] hash A Hash whose keys are Symbols, and whose values are Integers.
```

Array: Use pipe to denote a union of allowed types:

```ruby
  # @param [ Array<Symbol | String> ] array An Array whose members must be either Symbols or Strings
```

Array: Use comma to denote the types of each position in a tuple:

```ruby
  # @return [ Array<Symbol, Integer, [ true | false ]> ] A 3-member Array whose first element is a Symbol,
  #   second element is an Integer, and third element is a boolean value.
```

For clarity, use brackets to denote nested unions when commas are also used:

```ruby
  # @param [ Hash<Symbol, [ true | false ]> ] hash A Hash whose keys are Symbols, and whose values are boolean values.
```

Specific values may be denoted in the type using Ruby syntax.

```ruby
  # @param [ :before | :after ] timing One of the Symbol values :before or :after.
```

Use "true", "false", and "nil" rather than "TrueClass", "FalseClass", and "NilClass". Do not use "Boolean" as a type as it does not exist in Ruby.

```ruby
  # GOOD:
  # @param [ true | false | nil ] bool A boolean or nil value.

  # BAD:
  # @param [ TrueClass | FalseClass | NilClass ] bool A boolean or nil value.
  # @param [ Boolean ] bool A boolean value.
```

Return value "self" is allowed.

```ruby
  # @return [ self ] Returns the object self.
```

Splats: Use * in the parameter name to denote a splat:

```ruby
  # @param [ String ] *items The list of items name(s) as Strings.
  # ...
  def buy_groceries(*items)
```

Splats: Use comma to denote positionality in a splat:

```ruby
  # @param [ Symbol, Hash ] *args A list of names, followed by a hash of options.
  # ...
  def say_hello(*args)
```

Named keyword arguments: following YARD convention, use `@param` as normal for named keyword arguments.

```ruby
  # @param [String] query The search string
  # @param [Boolean] exact_match Whether to do an exact match
  # @param [Integer] results_per_page Number of results
  def search(query, exact_match: false, results_per_page: 10)
```

Double-splats: Use ** in the parameter name to denote a double-splat. Note that type does not need declared on the double-splat element (it is implicly <Symbol, Object>). Define the types with @option macro below:

```ruby
  # @param **kwargs The optional keyword argument(s).
  #
  # @option **kwargs [ String | Array<String> ] :items The items(s) as Strings to include.
  # @option **kwargs [ Integer ] :limit An Integer denoting the limit.
  # ...
  def buy_groceries(**kwargs)
```

Long comments: Use double-space indent when wrapping lines.

```ruby
  # @return [ Symbol ] This sentence is so long that is must be double-space
  #   idented on the second, third, etc. lines.
```

### DIFF
--- a/lib/mongoid/association/many.rb
+++ b/lib/mongoid/association/many.rb
@@ -17,7 +17,7 @@ module Mongoid
       # @example Is the association empty??
       #   person.addresses.blank?
       #
-      # @return [ true, false ] If the association is empty or not.
+      # @return [ true | false ] If the association is empty or not.
       def blank?
         !any?
       end
@@ -133,7 +133,7 @@ module Mongoid
       # @param [ Symbol ] name The method name.
       # @param [ true | false ] include_private Whether to include private methods.
       #
-      # @return [ true, false ] If the proxy responds to the method.
+      # @return [ true | false ] If the proxy responds to the method.
       def respond_to?(name, include_private = false)
         [].respond_to?(name, include_private) ||
           klass.respond_to?(name, include_private) || super
--- a/lib/mongoid/association/many.rb
+++ b/lib/mongoid/association/many.rb
@@ -131,7 +131,7 @@ module Mongoid
       #   relation.respond_to?(:name)
       #
       # @param [ Symbol ] name The method name.
-      # @param [ true, false ] include_private Whether to include private methods.
+      # @param [ true | false ] include_private Whether to include private methods.
       #
       # @return [ true, false ] If the proxy responds to the method.
       def respond_to?(name, include_private = false)
--- a/lib/mongoid/association/many.rb
+++ b/lib/mongoid/association/many.rb
@@ -158,9 +158,9 @@ module Mongoid
       #
       # @param [ Hash ] options The options to pass.
       #
-      # @option options [ Symbol ] :include What associations to include
-      # @option options [ Symbol ] :only Limit the fields to only these.
-      # @option options [ Symbol ] :except Dont include these fields.
+      # @option options [ Symbol | Array<Symbol> ] :include Which association(s) to include.
+      # @option options [ Symbol | Array<Symbol> ] :only Limit the field(s) to only these.
+      # @option options [ Symbol | Array<Symbol> ] :except Do not include these field(s).
       #
       # @return [ Hash ] The documents, ready to be serialized.
       def serializable_hash(options = {})
--- a/lib/mongoid/association/nested/nested_buildable.rb
+++ b/lib/mongoid/association/nested/nested_buildable.rb
@@ -12,7 +12,7 @@ module Mongoid
         # @example Do we allow a destroy?
         #   builder.allow_destroy?
         #
-        # @return [ true, false ] True if the allow destroy option was set.
+        # @return [ true | false ] True if the allow destroy option was set.
         def allow_destroy?
           options[:allow_destroy] || false
         end
@@ -25,7 +25,7 @@ module Mongoid
         # @param [ Document ] document The parent document of the association
         # @param [ Hash ] attrs The attributes to check for rejection.
         #
-        # @return [ true, false ] True and call proc or method if rejectable, false if not.
+        # @return [ true | false ] True and call proc or method if rejectable, false if not.
         def reject?(document, attrs)
           case callback = options[:reject_if]
             when Symbol
@@ -43,7 +43,7 @@ module Mongoid
         # @example Is this update only?
         #   builder.update_only?
         #
-        # @return [ true, false ] True if the update_only option was set.
+        # @return [ true | false ] True if the update_only option was set.
         def update_only?
           options[:update_only] || false
         end
--- a/lib/mongoid/association/nested/nested_buildable.rb
+++ b/lib/mongoid/association/nested/nested_buildable.rb
@@ -56,7 +56,7 @@ module Mongoid
         # @param [ Class ] klass The class we're trying to convert for.
         # @param [ String ] id The id, usually coming from the form.
         #
-        # @return [ BSON::ObjectId, String, Object ] The converted id.
+        # @return [ BSON::ObjectId | String | Object ] The converted id.
         def convert_id(klass, id)
           klass.using_object_ids? ? BSON::ObjectId.mongoize(id) : id
         end
--- a/lib/mongoid/attributes/nested.rb
+++ b/lib/mongoid/attributes/nested.rb
@@ -33,7 +33,7 @@ module Mongoid
         #     accepts_nested_attributes_for :addresses, :game, :posts
         #   end
         #
-        # @param [ Array<Symbol>, Hash ] args A list of association names, followed
+        # @param [ Array<Symbol> | Hash ] args A list of association names, followed
         #   by a hash of options.
         #
         # @option *args [ true, false ] :allow_destroy Can deletion occur?
--- a/lib/mongoid/attributes/nested.rb
+++ b/lib/mongoid/attributes/nested.rb
@@ -36,12 +36,12 @@ module Mongoid
         # @param [ Array<Symbol> | Hash ] args A list of association names, followed
         #   by a hash of options.
         #
-        # @option *args [ true, false ] :allow_destroy Can deletion occur?
+        # @option *args [ true | false ] :allow_destroy Can deletion occur?
         # @option *args [ Proc, Symbol ] :reject_if Block or symbol pointing
         #   to a class method to reject documents with.
         # @option *args [ Integer ] :limit The max number to create.
-        # @option *args [ true, false ] :update_only Only update existing docs.
-        # @options *args [ true, false ] :autosave Whether autosave should be enabled on the
+        # @option *args [ true | false ] :update_only Only update existing docs.
+        # @option *args [ true | false ] :autosave Whether autosave should be enabled on the
         #   association. Note that since the default is true, setting autosave to nil will still
         #   enable it.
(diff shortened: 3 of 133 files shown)


lib/mongoid/association/many.rb
```diff
@@ -158,9 +158,9 @@ def scoped
       #
       # @param [ Hash ] options The options to pass.
       #
-      # @option options [ Symbol ] :include What associations to include
-      # @option options [ Symbol ] :only Limit the fields to only these.
-      # @option options [ Symbol ] :except Dont include these fields.
+      # @option options [ Symbol | Array<Symbol> ] :include Which association(s) to include.
+      # @option options [ Symbol | Array<Symbol> ] :only Limit the field(s) to only these.
+      # @option options [ Symbol | Array<Symbol> ] :except Do not include these field(s).
```
johnnyshields, line 163: This was incorrect type.

lib/mongoid/attributes/nested.rb
```diff
@@ -33,15 +33,15 @@ module ClassMethods
...
-        # @option *args [ true, false ] :allow_destroy Can deletion occur?
-        # @option *args [ Proc, Symbol ] :reject_if Block or symbol pointing
+        # @option *args [ true | false ] :allow_destroy Can deletion occur?
+        # @option *args [ Proc | Symbol ] :reject_if Block or symbol pointing
         #   to a class method to reject documents with.
         # @option *args [ Integer ] :limit The max number to create.
-        # @option *args [ true, false ] :update_only Only update existing docs.
-        # @options *args [ true, false ] :autosave Whether autosave should be enabled on the
+        # @option *args [ true | false ] :update_only Only update existing docs.
+        # @option *args [ true | false ] :autosave Whether autosave should be enabled on the
```
johnnyshields, line 44: FYI this said `@options` instead of `@option`. I assume its a typo.

lib/mongoid/criteria/queryable/selectable.rb
```diff
@@ -635,7 +635,7 @@ def or(*criteria)
         # @example Same as previous example, also deprecated.
         #   selectable.any_of([{field: 1}], [{field: 2}])
         #
-        # @param [ Hash | Criteria | Array<Hash | Criteria>, ... ] criteria
+        # @param [ Hash | Criteria | Array<Hash | Criteria>] *criteria
```
johnnyshields, line 638: Did this splat change in this PR. This is only place where `, ... ]` was used.

lib/mongoid/matcher.rb
```diff
@@ -24,17 +24,14 @@ module Matcher
...
     # @param [ String ] key The key path to extract.
     #
-    # @return [ Array<true | false, Object | Array, true | false> ]
-    #   Whether the value existed in the document, the extracted value
-    #   and the array expansion flag.
+    # @return [ Array<Object | Array, true | false> ]
+    #   A two-element array. The first element is the value retrieved, or an
+    #   array of values. The second element is a boolean flag indicating
+    #   whether an array was expanded at any point during the key traversal
+    #   (because the respective document field was an array).
```
johnnyshields, line 34: This was quite funky one. I think I have it correct here. Note I move the return comment to below the @return macro.

johnnyshields:
Ready for review.

lib/mongoid/criteria/queryable/selectable.rb
johnnyshields: Hmmm, let's keep this actually. Will revert.

p-mongo:
Did you review every change this PR makes for correctness?

johnnyshields:
> Did you review every change this PR makes for correctness?

Yes, I checked it 3 times, but I am human, so I appreciate it someone else can scan it once and make sure I didn't make any mistakes.

johnnyshields:
OK, make that 4 times! I think we're good.

lib/mongoid/association/nested/nested_buildable.rb
```diff
@@ -56,7 +56,7 @@ def update_only?
         # @param [ Class ] klass The class we're trying to convert for.
         # @param [ String ] id The id, usually coming from the form.
         #
-        # @return [ BSON::ObjectId, String, Object ] The converted id.
+        # @return [ BSON::ObjectId | String | Object ] The converted id.
```
Neilshweky, line 59: Shouldn't this just be Object?

lib/mongoid/contextual/memory.rb
```diff
@@ -240,7 +240,7 @@ def limit(value)
       #
       # @param [ String | Symbol ] *fields Field(s) to pluck.
       #
-      # @return [ Array ] The array of plucked values.
+      # @return [ Array<Object> | Array<Array<Object>> ] The plucked values.
```
Neilshweky, line 243: I don't agree with this type annotation... Could you explain why you need the double nested array? Because theoretically we can have n-nested array return values from this method. I think were better off keeping it as just an array.

lib/mongoid/contextual/mongo.rb
```diff
@@ -453,9 +453,9 @@ def map_reduce(map, reduce)
       # @note This method will return the raw db values - it performs no custom
       #   serialization.
       #
-      # @param [ String, Symbol, Array ] fields Fields to pluck.
+      # @param [ String | Symbol ] *fields Field(s) to pluck.
       #
-      # @return [ Array<Object, Array> ] The plucked values.
+      # @return [ Array<Object> | Array<Array<Object>> ] The plucked values.
```
Neilshweky, line 458: See my comment on memory#pluck

lib/mongoid/contextual/mongo.rb
```diff
@@ -453,9 +453,9 @@ def map_reduce(map, reduce)
       # @note This method will return the raw db values - it performs no custom
       #   serialization.
       #
-      # @param [ String, Symbol, Array ] fields Fields to pluck.
+      # @param [ String | Symbol ] *fields Field(s) to pluck.
```
Neilshweky, line 456: Is this how we do * fields?

lib/mongoid/contextual/none.rb
```diff
@@ -88,7 +88,7 @@ def exists?; false; end
       # @example Get the values for null context.
       #   context.pluck(:name)
       #
-      # @param [ String | Symbol ] *_fields Field or fields to pluck.
+      # @param [ String | Symbol ] *_fields Field(s) to pluck.
```
Neilshweky, line 91: Looks like we're inconsistent about how we manage * fields

lib/mongoid/extensions/regexp.rb
```diff
@@ -12,7 +12,7 @@ module ClassMethods
         # @example Mongoize the object.
         #   Regexp.mongoize(/\A[abc]/)
         #
-        # @param [ Object ] object The object to mongoize.
+        # @param [ Regexp | String ] object The object to mongoize.
```
Neilshweky, line 15: This method actually accepts arbitrary input as of MONGOID-5222

lib/mongoid/association/nested/nested_buildable.rb
Neilshweky: I'm okay with leaving this as it is if you prefer

lib/mongoid/contextual/mongo.rb
johnnyshields: See my answer 👍

lib/mongoid/contextual/none.rb
johnnyshields: See above comment. Note in diff the `*` isn't changing here, I just changed the text (for consistency)

lib/mongoid/association/nested/nested_buildable.rb
johnnyshields: It could be Object (as obviously Object is a superset of the others) but I think its reasonable to provide color on the "usually" expected type(s) here. So would suggest to leave this as I have it.

lib/mongoid/extensions/regexp.rb
johnnyshields: Good catch, I've fixed this.

Neilshweky (approved):
LGTM @p-mongo just waiting on your LGTM

lib/mongoid/contextual/memory.rb
```diff
@@ -318,7 +318,7 @@ def sort(values)
       #
       # @param [ Hash ] attributes The new attributes for the document.
       #
-      # @return [ nil, false ] False if no attributes were provided.
+      # @return [ nil | false ] False if no attributes were provided.
```
p-mongo, line 321: This looks suspicious.

johnnyshields: Please change if needed, I just simply following what was in the docs before.

lib/mongoid/matcher.rb
p-mongo: I think I got rid of that flag ultimately and didn't update the docstring.

p-mongo (approved):

PR #5367, opened 2022-07-10
MONGOID-5422 - Configuration DSL should not require an argument to its block (Rails parity)

johnnyshields:
This PR removes the need for `|config|` arg in `Mongoid.configure`. It makes it more terse and consistent with `Rails.application.configure`. The change is fully backwards compatible.

This syntax is now possible:

```ruby
  Mongoid.configure do
    # can call methods by themselves:
    connect_to("mongoid_test")

    # `config` is defined internally, so this also works here:
    config.connect_to("mongoid_test")

    # must use `config.variable =` when assigning variables
    config.preload_models = true
```

Previous syntax (still supported):

```ruby
  Mongoid.configure do |config|
    config.connect_to("mongoid_test")
    config.preload_models = true
```

To do this magic, like Rails, it uses `instance_exec`. For variable assignment, in the block, you need to use `self.variable = "foo"` (otherwise its a local variable.) To get around this, we introduce a method `config` that returns self, which is also what Rails does:

```ruby
Rails.application.configure do
  config.cache_classes = true
```

Note this only affects assigning with `=`, so other methods work fine (such as `connect_to` in example above. (I now fully grok why Puma's config DSL uses only methods and not variable assignment.) Most IDEs will warn about local variable assignment in the block.


spec/mongoid_spec.rb
```diff
@@ -13,22 +13,51 @@
...
     end
 
-    context "when a block is supplied" do
+    context "when a block is given" do
 
-      before do
+      after do
         Mongoid.configure do |config|
-          config.preload_models = true
+          config.preload_models = false
```
p-mongo, line 20: This should use `config_override` to return the value to the previous setting.

p-mongo (commented):
If you can adjust the tests to correctly restore config options everything else looks good.

spec/mongoid_spec.rb
johnnyshields: Done

johnnyshields:
Done > using config_override now

p-mongo (approved):
The release notes addition will conflict with https://github.com/mongodb/mongoid/pull/5396.

johnnyshields:
@Neilshweky the behavior here is the same as Rails:

```ruby
Mongoid.configure do
  self.preload_models = true # can do this
  config.preload_models = true # can also do this, because there is now a magic `config` method which returns self

  preload_models = true # cannot do this, because its a local variable assignment  
end
```

johnnyshields:
@p-mongo release note conflict resolved; ready to merge.

docs/release-notes/mongoid-8.1.txt
```diff
@@ -18,6 +18,31 @@ please consult GitHub releases for detailed release notes and JIRA for
 the complete list of issues fixed in each release, including bug fixes.
 
 
+Configuration DSL No Longer Requires an Argument to its Block
+-------------------------------------------------------------
+
+In Mongoid 8.1, it is now possible to use ``Mongoid.configure`` without
```
p-mongo, line 24: Please do not add version numbers to release notes like this, it makes backporting changes more difficult and the release notes already specify the version they are for.

p-mongo, line 24:
```suggestion
It is now possible to use ``Mongoid.configure`` without
```

johnnyshields:
Thanks @p-mongo for getting this one to the finish line!

PR #5368, opened 2022-07-10
Proof-of-concept: Mongoid::RawValue

johnnyshields:
Proof-of-Concept/Draft

This PR introduces a new type Mongoid::RawValue which affects `mongoize, demongoize, evolve`. It does three things:

- For `mongoize`, it is used in `strict` behavior (MONGOID-5382) to facilitate either (a) raise an error, (b) log a warning and return nil, or (c) return nil without a warning. Note new global config 
- For `demongoize` based on global config it is either (a) returned to user wrapped in RawValue or (b) returns in the underlying value to user. (No ticket for this yet)
- For `evolve`, it facilitates pass-thru of raw values (MONGOID-5408). You can now do:
   ```ruby
   User.where(name: Mongoid::RawValue(42)) # queries for the Integer 42
   ```


Neilshweky:
I haven't thought about this PR yet but just from an initial look.... Wouldn't it be better to have Mongoid::RawValue have its own evolve method which just returns the inner value? Instead of hard coding checks everywhere?

johnnyshields:
> Wouldn't it be better to have Mongoid::RawValue have its own evolve method which just returns the inner value?

Yeah that thought crossed my mind too. Hmm... 🤔

p-mongo:
@johnnyshields Can you please either rebase this PR on master or close it if no longer relevant.

I also get the impression that this is a "solution in search of a problem", so to speak, namely that you are trying to get your idea of `mongoize` returning variant values into Mongoid. This is not what either of the tickets you referenced is asking for (https://jira.mongodb.org/browse/MONGOID-5382, https://jira.mongodb.org/browse/MONGOID-5408). Please ensure that you are starting from the requirements and are working on a solution toward the requirements.

johnnyshields:
I'm glad to rebase this, please give me a few days.

johnnyshields:
PR is rebased.

My next step will be to split this into two PRs which clearly show how it solves each problem of [MONGOID-5382](https://jira.mongodb.org/browse/MONGOID-5382) and [MONGOID-5408](https://jira.mongodb.org/browse/MONGOID-5408).

johnnyshields:
I have extracted out https://github.com/mongodb/mongoid/pull/5448 to address MONGOID-5408 only.

johnnyshields:
@jamis can you please give me some feedback on the spec written above and the draft code in this PR? If I invest my time to complete this, will the MongoDB team merge it?

PR #5374, opened 2022-07-12
MONGOID-5419 Add warnings to Mongoid::Warnings module

Neilshweky:


spec/mongoid/indexable_spec.rb
```diff
@@ -423,6 +423,18 @@ def self.hereditary?
       end
     end
 
+    context "when providing a geo haystack index with a bucket_size" do
+
+      let(:message) do
+        'The geoHaystack option is deprecated.'
```
p-mongo, line 429: I think it should be "geoHaystack index type" and not "option". https://www.mongodb.com/docs/manual/core/geohaystack/ just calls it an "index" which I suppose also is valid if you prefer that.

spec/mongoid/warnings_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      it "logs the warning only once" do
+        expect_any_instance_of(Logger).to receive(:warn).once.with(message)
+        Mongoid::Warnings.send("warn_#{id}")
+        Mongoid::Warnings.send("warn_#{id}")
+      end
+    end
+  end
+end
```
p-mongo, line 35: Add a newline to the end of file please, otherwise it gets added by git automatically when rebasing/applying patches and causes conflicts.

p-mongo (approved):

PR #5379, opened 2022-07-15
MONGOID-5429 Require Driver 2.18 for Mongoid 8.0

Neilshweky:


gemfiles/bson_min.gemfile
```diff
@@ -4,7 +4,7 @@ gemspec path: '..'
 gem 'bson', '4.14.0'
 # This configuration doesn't require a specific driver version. When bson-4.14.0
 # was released, current driver version was 2.17.0
-gem 'mongo', '2.17.0'
+gem 'mongo', '2.18.beta1'
```
Neilshweky, line 7: We'll have to switch this when we release 2.18

lib/mongoid/query_cache.rb
```diff
@@ -20,11 +14,7 @@ class << self
       # @return [ Hash ] The hash of cached queries.
       # @api private
       def cache_table
```
p-mongo, line 16: If this method is not used by Mongoid I think we should get rid of it completely.

spec/integration/dots_and_dollars_spec.rb
```diff
@@ -4,6 +4,7 @@
 
 describe "Dots and Dollars" do
   min_driver_version '2.18'
```
p-mongo, line 6: This line can be removed right?

gemfiles/driver_oldstable.gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem 'mongo', git: "https://github.com/mongodb/mongo-ruby-driver", branch: '2.15-stable'
+gem 'mongo', git: "https://github.com/mongodb/mongo-ruby-driver", branch: 'master'
```
p-mongo, line 3: These would point to 2.18-stable once that branch exists.

p-mongo (approved):

lib/mongoid/query_cache.rb
Neilshweky: done

spec/integration/dots_and_dollars_spec.rb
Neilshweky: done

gemfiles/driver_oldstable.gemfile
Neilshweky: done

PR #5381, opened 2022-07-15
MONGOID-5428 Make feature flag connected to Mongo.broken_view_options

Neilshweky:
This PR won't work until 2.18 is released. Driver master works! That's a good sign.

### DIFF
--- a/lib/mongoid/config/options.rb
+++ b/lib/mongoid/config/options.rb
@@ -30,13 +30,13 @@ module Mongoid
 
         class_eval do
           # log_level accessor is defined specially below
-          unless name.to_sym == :log_level
+          unless name == :log_level
             define_method(name) do
               settings[name]
             end
           end
 
-          if name.to_sym == :broken_view_options
+          if name == :broken_view_options
             Mongo.send("#{name}=", options[:default])
             define_method("#{name}=") do |value|
               Mongo.send("#{name}=", value)
--- a/lib/mongoid/config/options.rb
+++ b/lib/mongoid/config/options.rb
@@ -95,7 +95,10 @@ module Mongoid
           Mongo.send(name)
         end
 
+        # This causes the driver option to be set to Mongoid's default
+        # when Mongoid is loaded.
         Mongo.send("#{name}=", options[:default])
+
         define_method("#{name}=") do |value|
           Mongo.send("#{name}=", value)
           settings[name] = value
--- a/lib/mongoid/config/options.rb
+++ b/lib/mongoid/config/options.rb
@@ -36,8 +36,16 @@ module Mongoid
             end
           end
 
-          define_method("#{name}=") do |value|
-            settings[name] = value
+          if name.to_sym == :broken_view_options
+            Mongo.send("#{name}=", options[:default])
+            define_method("#{name}=") do |value|
+              Mongo.send("#{name}=", value)
+              settings[name] = value
+            end
+          else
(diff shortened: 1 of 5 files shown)


lib/mongoid/config/options.rb
```diff
@@ -36,8 +36,16 @@ def option(name, options = {})
             end
           end
 
-          define_method("#{name}=") do |value|
-            settings[name] = value
+          if name.to_sym == :broken_view_options
```
p-mongo, line 39: I would do something like `name = name.to_sym` one time in this method right in the beginning.

p-mongo: Also, the docstring says name can only be a symbol, so either this conversion is unnecessary or the docstring should be amended to specify string as allowed input.

lib/mongoid/config/options.rb
```diff
@@ -36,8 +36,16 @@ def option(name, options = {})
...
+            Mongo.send("#{name}=", options[:default])
+            define_method("#{name}=") do |value|
+              Mongo.send("#{name}=", value)
+              settings[name] = value
```
p-mongo, line 43: If possible I wouldn't have a setting in Mongoid at all, this way the two can't get out of sync.

spec/mongoid/config_spec.rb
```diff
@@ -233,6 +233,12 @@
       end
     end
 
+    around do |ex|
+      temp = Mongoid.send(option)
+      ex.run
+      Mongoid.send("#{option}=", temp)
```
p-mongo, line 239: I would write a macro similar to the one we have for Mongoid options.

Neilshweky: That would be almost the same as config_override. The difference in this case is that the Mongoid options are being modified in other ways within the tests (using loaded configs), and I just want to put it back to what it was before the test.

lib/mongoid/config/options.rb
Neilshweky: Well, I want the default to be true in Mongoid and false in mongo-ruby-driver.... how else would you do that?

lib/mongoid/config/options.rb
Neilshweky: Probably the former.

Neilshweky: done

Neilshweky:
@p-mongo I made some changes: 

I have fully delegated the `Mongoid.broken_view_options` flag to the `Mongo.broken_view_options flag`, so that when one changes so does the other. The only problem is that when modifying the `Mongo.broken_view_options`, the `Mongoid::Config.settings` does not include the new value (and vice versa). Which could be viewed as problematic. I'm not really sure how to navigate this correctly.

p-mongo (approved):

p-mongo:
> The only problem is that when modifying the Mongo.broken_view_options, the Mongoid::Config.settings does not include the new value (and vice versa).

In the `settings` method augment the internal hash with the value from driver.

p-mongo:
Actually the settings are mutable so that won't work quite so easily.

Neilshweky:
> Actually the settings are mutable so that won't work quite so easily.

right

Neilshweky:
I thought about excluding it from the settings entirely, but that caused validation errors... 

I have documented this oddity if you want to keep current behavior (updating the Mongo flag doesn't update Mongoid settings)

PR #5382, opened 2022-07-16
MONGOID-4500 Mongoid::Contextual::Aggregable::Memory should ignore nil values.

johnnyshields:
This change makes the in-memory behavior closer to MongoDB's native operators.

This PR is a pared-back version of https://github.com/mongodb/mongoid/pull/5038 which contains only nil changes and not type conversion. I hope this can be merged.

### DIFF
--- a/spec/mongoid/contextual/aggregable/memory_table.yml
+++ b/spec/mongoid/contextual/aggregable/memory_table.yml
@@ -53,15 +53,15 @@ sets:
     values:
       - 233
       - '-974.332'
-      - ~
+      - ~ # ignored
       - 532.45642
       - '-5675'
       - 348.434
-      - 'Foobar'
+      - 'Foobar' # ignored
       - -3394
       - -493
-      - ''
-      - '0 1 2 3'
+      - '' # ignored
+      - '0 1 2 3' # ignored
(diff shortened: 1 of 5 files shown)


spec/mongoid/contextual/aggregable/memory_table.yml
```diff
@@ -0,0 +1,88 @@
...
+      - ''
+      - '0 1 2 3'
+      - !ruby/object:BigDecimal 18:0.53245642e3
+      - -5675
+      - '0.434'
+      - !ruby/object:BigDecimal 18:-0.97676e5
+      - 0.434
+      - 7455.96
+      - -0.0544
+      - '   '
```
p-mongo, line 72: Everything in this PR looks good except for this list. It is unclear which values are used in the calculation and which are not. For example, the numeric strings. Can you please annotate the values that aren't used with comments?

johnnyshields: Done. I've added `# ignored` to all unused values.

Note that strings like `"1.234"` *are* used because of the type coercion logic (i.e. the logic recently updated in MONGOID-5222).

p-mongo (approved):

PR #5383, opened 2022-07-18
MONGOID-5414 deprecate criteria cache functionality

Neilshweky:
MONGOID-5414

This PR keeps the cache and cached? methods but removes all of their functionality. Do you think it would be better to remove these methods outright and put a warning in 7.5?

The failing distros are driver min, which will stop failing once we merge MONGOID-5429


p-mongo (approved):
8.0 is a good time to remove those methods, otherwise we would need to keep them until 9.0.

p-mongo:
If you can add the deprecation warning in 7.x that would be great.

Neilshweky:
I have raised #5388 for this purpose @p-mongo

p-mongo (approved):

PR #5386, opened 2022-07-18
MONGOID-5426 remove support for field argument to Mongo#map

Neilshweky:


spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -2959,7 +2959,9 @@
     context "when passed the symbol field name" do
 
       it "performs mapping" do
```
p-mongo, line 2961: It doesn't anymore right?

p-mongo (approved):

PR #5387, opened 2022-07-18
MONGOID-5425 Verify string mongoization into Date, Time, DateTime with time zones

Neilshweky:
MONGOID-5425
Some things to note:
- Time always mongoizes to utc. It seems like we only store dates in the db as UTC
- I didn't write expansive DateTime tests because DateTime just calls the Time.(de)mongoize methods
- I didn't write Date tests, because as far as I can tell Dates aren't affected by time zones: 
```
irb(main):005:0> Date.parse("2022-07-11 14:03:42 +1500")
=> Mon, 11 Jul 2022
irb(main):006:0> Date.parse("2022-07-11 14:03:42 -1500")
=> Mon, 11 Jul 2022
```

### DIFF
--- a/spec/mongoid/extensions/time_spec.rb
+++ b/spec/mongoid/extensions/time_spec.rb
@@ -301,7 +301,7 @@ describe Mongoid::Extensions::Time do
               Time.now.utc_offset
             end
 
-            let(:expected_time) { Time.parse("2010-11-19 05:00:00 +0000").in_time_zone }
+            let(:expected_time) { Time.parse("2010-11-19 00:00:00 +0000") - Time.parse(string).utc_offset }
 
             it 'test operates in multiple time zones' do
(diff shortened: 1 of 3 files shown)


spec/mongoid/extensions/time_spec.rb
```diff
@@ -166,113 +166,316 @@
...
+            let(:string) do
+              # JST is +0900
+              "2010-11-19 00:24:49.123457 +1100"
+            end
+
+            let(:mongoized) do
+              Time.demongoize(string)
+            end
+
+            let(:expected_time) { Time.parse("2010-11-18 13:24:49.123457 +0000").in_time_zone }
```
p-mongo, line 186: This would be more clear if the time zone was given explicitly in the construction call.

Neilshweky:
@p-mongo  is this good to merge?

p-mongo (approved):
Yes go ahead.

PR #5389, opened 2022-07-19
MONGOID-5421: Mongoid::Contextual::Aggregable::Memory should do type coercion for String and Decimal128

johnnyshields:
Replaces https://github.com/mongodb/mongoid/pull/5038

Not really sure this is needed at this point but putting here for review.


johnnyshields:
I'm going to close this and will re-open if I encounter a use case in the real-world.

PR #5390, opened 2022-07-19
MONGOID-5414 Remove caching on Mongoid::Contextual::Mongo#length

Neilshweky:
Originally this method was cached because, before the view options were fixed to be passed into the count method, it was impossible to pass options into length. Now that it is possible to pass options into length, the following case would cause stale results to be returned:
```
crit = Band.all
crit.limit(5).length 
# => 5
crit.limit(3).length
# => 5
```
This is incorrect, and so I removed the caching from this method. 

The caching would get delegated to the query cache, but it seems like no caching is done on agg pipelines, which is what appears to be performed underneath the call to count.

### DIFF
--- a/spec/mongoid/contextual/mongo_spec.rb
+++ b/spec/mongoid/contextual/mongo_spec.rb
@@ -2587,12 +2587,7 @@ describe Mongoid::Contextual::Mongo do
         end
 
         context "when calling more than once with different limits" do
-          around do |ex|
-            temp = Mongo.broken_view_options
-            Mongo.broken_view_options = false
-            ex.run
-            Mongo.broken_view_options = temp
-          end
+          driver_config_override :broken_view_options, false
 
           it "does not cache the value" do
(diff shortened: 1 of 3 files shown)


spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -2586,10 +2586,17 @@
           expect(context.send(method)).to eq(2)
         end
 
-        context "when calling more than once" do
-          it "returns the cached value for subsequent calls" do
-            expect(context.view).to receive(:count_documents).once.and_return(2)
-            2.times { expect(context.send(method)).to eq(2) }
+        context "when calling more than once with different limits" do
+          around do |ex|
+            temp = Mongo.broken_view_options
```
Neilshweky, line 2591: use config_override once the MONGOID-5428 is merged

p-mongo (approved):

p-mongo (approved):

PR #5392, opened 2022-07-19
MONGOID-5428 add driver_options to Mongoid config

Neilshweky:


lib/mongoid/config.rb
```diff
@@ -216,6 +216,7 @@ def load_configuration(settings)
       configuration = settings.with_indifferent_access
       self.options = configuration[:options]
       self.clients = configuration[:clients]
+      Mongo.options = configuration[:driver_options] || {}
```
Neilshweky, line 219: One of these things is not like the other... here's why: https://github.com/mongodb/mongo-ruby-driver/pull/2515#discussion_r883942106

p-mongo: I think this should iterate the contents of `configuration[:driver_options]` and call `Mongo.send("#{option}=", value)`.

p-mongo: Which also would stop discarding options that are already set for driver and don't exist in Mongoid config.yml. In fact there should be a test for this.

p-mongo: If an option value is nil here would you write nil to driver's option or do nothing?

Neilshweky: That's what the drivers `options=` method does: 
``` 
options.each_pair do |option, value|
  Validators::Option.validate(option)
  send("#{option}=", value)
end

Comment (Neilshweky) [reply lib/mongoid/config.rb:219]: I have added that test

Comment (Neilshweky) [reply lib/mongoid/config.rb:219]: > If an option value is nil here would you write nil to driver's option or do nothing?

If the option is unset, nothing, if the option is set to `nil`, set the driver's option

File: spec/config/mongoid.yml [hunk 2]
```diff
@@ -40,6 +40,8 @@ test:
     app_name: 'testing'
     background_indexing: false
     join_contexts: false
+  driver_options:
+    broken_view_options: false
```
p-mongo, line 44: Can you please add a comment here that `validate_update_replace` shouldn't be given, because if someone gave it here as false it would make the test not do anything.

lib/mongoid/config.rb
p-mongo: Can you add a test for an option with nil value please.

Neilshweky: as there are only two options available in Mongo, I cannot. One is already being used for regular assignment, and one is being used to show that it does not get overridden.

spec/config/mongoid.yml
Neilshweky: done

lib/mongoid/config.rb
p-mongo: Add another config file?

Neilshweky: oh lol yeah good point

Neilshweky: done

p-mongo (approved):

spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -2582,39 +2582,28 @@
...
+            expect(context.send(method)).to eq(1)
           end
         end
 
-        context "when the results have been iterated over" do
-
-          before do
-            context.entries
-          end
+        context "when calling more than once with different limits" do
```
Neilshweky, line 2593:
```suggestion
        context "when broken_view_options is true" do
```

spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -2582,39 +2582,28 @@
           described_class.new(criteria)
         end
 
-        it "returns the number of documents that match" do
-          expect(context.send(method)).to eq(2)
-        end
-
         context "when calling more than once with different limits" do
```
Neilshweky, line 2585:
```suggestion
        context "when broken_view_options is false" do
```

PR #5397, opened 2022-07-20
MONGOID-4979 Add :touch option to model save method

Neilshweky:


p-mongo:
:+1: Looks good so far.

Neilshweky:
@p-mongo  this is rfal

johnnyshields:
Is it correct to say this has a slightly different meaning than `timeless`? i.e.:
- `timeless` applies only to parent
- `touch: false` applies to both parent and any modified embedded childs?

lib/mongoid/persistable/updatable.rb
```diff
@@ -171,10 +172,10 @@ def update_document(options = {})
       # cleared in the before_update callback.
       #
       # @param [ Hash ] options The options.
-      def process_touch_option(options)
+      def process_touch_option(options, children)
```
p-mongo, line 175: `children` needs to be added to docstring?

p-mongo: Could you add at least the `touch` option also to the docstring?

lib/mongoid/persistable/updatable.rb
```diff
@@ -160,6 +165,19 @@ def update_document(options = {})
           end
         end
       end
+
+      # If there is a touch option and it is false, call the timeless method so
+      # that the updated_at attribute is not updated. Call the timeless method
```
p-mongo, line 170: Call*s*? Unclear whether this says this method will call `timeless` or the caller of this method needs to do that.

p-mongo (approved):

johnnyshields:
Needs release note and update to docs in persistence section. Should explain difference re: timeless if any.

lib/mongoid/persistable/updatable.rb
Neilshweky: done

lib/mongoid/persistable/updatable.rb
Neilshweky: done

p-mongo (approved):

docs/release-notes/mongoid-8.1.txt
```diff
@@ -42,3 +42,18 @@ with a bang (!) raise an error:
...
+Support for the ``:touch`` option has been added to the ``#save`` and ``#save!``
+methods. When this option is false, the ``updated_at`` field on the saved
+document and all of it's embedded documents will not be updated with the
+current time. When this option is true or unset, the ``updated_at`` field will
+be updated with the current time.
+
+If the document being saved is a new document (i.e. it has not yet been
+persisted to the database), then the ``:touch`` option will be ignored, and the
+``updated_at`` (and ``created_at``) fields will be updated with the current
+time.
```
johnnyshields, line 59: Please add docs to Persistence (?) page, in addition to Release Notes. Ideally it should explain the difference with `timeless` method, if any.

PR #5398, opened 2022-07-21
MONGOID-5435 Fix ArgumentError when block has positional and keyword arguments

kyoshidajp:
https://jira.mongodb.org/browse/MONGOID-5435


Neilshweky (approved):

Neilshweky:
@kyoshidajp This seems to be failing on >Ruby-3.0... Do you think you could fix those failures? Thanks!

johnnyshields:
@Neilshweky failure looks like its on Ruby 2.6.

kyoshidajp:
For Ruby 2.6, I switched to use ruby2_keywords. It will be tested successfully.

PR #5400, opened 2022-07-21
Add splats to docs (WIP)

johnnyshields:


johnnyshields:
Ready for review

lib/mongoid/association/embedded/embeds_many/proxy.rb
```diff
@@ -235,8 +235,8 @@ def exists?
           # @example Finds the first matching document using a block.
           #   person.addresses.find { |addr| addr.state == 'CA' }
           #
-          # @param [ Array<Object> ] args Various arguments.
-          # @param [ Proc ] block Optional block to pass.
+          # @param [ Object... ] *args Various arguments.
+          # @yield &block Optional block to pass.
```
p-mongo, line 239: Why is this a `@yield`?

p-mongo (commented):
`@yield` should document parameters yielded to the block, it is not meant for documenting block as a parameter to the method as far as I know.

lib/mongoid/association/embedded/embeds_many/proxy.rb
johnnyshields: In my view the current way `@param [ Proc ] block` is not correct, because (a) it's not a standard arg, and (b) its not a Proc object.

Here's what [Yard officially says](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#yield):

```
# For a block {|a,b,c| ... }
# @yield [a, b, c] Gives 3 random numbers to the block
def provide3values(&block)
  yield(42, 42, 42)
end
```

It seems they don't add document the `&block` parameter name, however, I think there is no harm in adding it to `@yield`. I would propose the following syntax:

```
# @yield [a, b, c] &block Gives 3 random numbers to the block
def provide3values(&block)
  yield(42, 42, 42)
end

# @yield [a, b, c] Gives 3 random numbers to the block
def provide3values
  yield(42, 42, 42)
end

# @yield [a, b, c] &block Gives 3 random numbers to the block
def provide3values(&block)
  block.call(42, 42, 42)
end
```

p-mongo:
Please take the `@yield` changes out of this PR, you can remove them and commit that as a separate commit then make a new branch and revert this commit to get back to the present diff and make a follow-up PR in which we can discuss the yields.

lib/mongoid/association/referenced/has_many/proxy.rb
```diff
@@ -199,7 +199,7 @@ def exists?
           # @note This will keep matching documents in memory for iteration
           #   later.
           #
-          # @param [ Object | Array<Object> ] *args The ids.
+          # @param [ [ BSON::ObjectId | String ]... ] *args The ids.
```
p-mongo, line 202: You can add the `...` here but the types were correct as written, please put them back the way they were.

johnnyshields:
Done -- splats only in this PR now.

lib/mongoid/criteria.rb
```diff
@@ -88,7 +88,7 @@ def ==(other)
     # @example Tries to find a document whose _id is the stringification of the provided Proc, typically failing.
     #   enumerator = criteria.find(-> { "Default Band" })
     #
-    # @param [ Object | Array<Object> ] *args The ids.
+    # @param [ Object... ] *args The id(s).
```
p-mongo, line 91: And here.

lib/mongoid/criteria/findable.rb
```diff
@@ -32,7 +32,7 @@ def execute_or_raise(ids, multi)
       # @example Find by multiple ids.
       #   criteria.find([ BSON::ObjectId.new, BSON::ObjectId.new ])
       #
-      # @param [ Object | Array<Object> ] *args The ids to search for.
+      # @param [ [ Object | Array<Object> ]... ] *args The id(s) to find.
```
p-mongo, line 35: This is the correct specification for the other two finds.

lib/mongoid/criteria/queryable/selectable.rb
```diff
@@ -69,8 +69,9 @@ def all(*criteria)
         # @example Add the criterion.
         #   selectable.and({ field: value }, { other: value })
         #
-        # @param [ Array<Hash | Criteria> ] criteria Multiple key/value pair
-        #   matches or Criteria objects that all must match to return results.
+        # @param [ [ Hash | Criteria ]... ] *criteria
```
p-mongo, line 72: This should be `Hash | Criteria | Array<Hash | Criteria>`

lib/mongoid/criteria/queryable/selectable.rb
```diff
@@ -516,7 +517,7 @@ def nin(condition)
         # @example Add the $nor selection.
         #   selectable.nor(field: 1, field: 2)
         #
-        # @param [ Array<Hash | Criteria> ] criteria Multiple key/value pair
+        # @param [ [ Hash | Criteria ]... ] *criteria Multiple key/value pair
```
p-mongo, line 520: This should retain `Array<>` to indicate the acceptance of arrays.

lib/mongoid/findable.rb
```diff
@@ -149,7 +149,7 @@ def exists?
     # @note Each argument can be an individual id, an array of ids or
     #   a nested array. Each array will be flattened.
     #
-    # @param [ Object | Array<Object> ] *args The _id value(s) to find.
+    # @param [ Object... ] *args The id(s) to find.
```
p-mongo, line 152: This is wrong.

p-mongo (changes requested):

lib/mongoid/association/referenced/has_many/proxy.rb
johnnyshields: Fixed

lib/mongoid/criteria.rb
johnnyshields: Fixed

lib/mongoid/criteria/queryable/selectable.rb
johnnyshields: Noted

lib/mongoid/criteria/queryable/selectable.rb
johnnyshields: Fixed

lib/mongoid/findable.rb
johnnyshields: Fixed

lib/mongoid/criteria/findable.rb
johnnyshields: Noted

johnnyshields:
All requested fixes are done, please give final review.

p-mongo (approved):

johnnyshields:
Thank you 🙇‍♂️

PR #5403, opened 2022-07-22
MONGOID-4030 Typed Array Fields

Neilshweky:
This was fun, but I'm not sure it's feasible. Would love your thoughts @p-mongo


johnnyshields:
This is an interesting POC and certainly a useful feature, however, I suggest to hold off until MONGOID-5409 is implemented as its likely the fundamental structure will change--probably in a way that will make this ticket easier/cleaner to implement.

Neilshweky:
I think you're right that followup work for this ticket is typed sets and typed hashes. That makes me wonder if this solution if feasible, and perhaps we should wait for some redesign of the mongoization/demongoization process. This would only be possible in a major version so this would have to wait until then. 

This was a cool solution for what it was, though.

p-mongo:
I think the implementation in this PR is very reasonable. The current behavior of mongoize/demongoize of (always) replacing invalid input with nil I think becomes more problematic when dealing with arrays as in the scope of this PR. Another interesting use case is if a user wanted to declare a field to be an array of Foos but produce validation errors when elements aren't Foos, but given that in general this type of operation isn't very well supported by Mongoid presently perhaps asking for it here is not appropriate either.

docs/reference/fields.txt
```diff
@@ -447,6 +447,81 @@ matches strings containing "hello" before a newline, besides strings ending in
...
+
+Mongoid supports declaring a typed array field on models. This field will
+enforce the type of the elements in the array by ``mongoizing`` each element
+inserted into the array. To make a typed array field:
+
+.. code:: ruby
+
+  class Band
+    include Mongoid::Document
+    field :members, type: Array(String)
```
p-mongo, line 464: The `Array(String)` syntax does not extend to Set or Hash. If you are intending to support type restrictions for the other container types  I think it is worth figuring out a syntax that would work for all of them.

Neilshweky:
This is real @p-mongo

docs/reference/fields.txt
```diff
@@ -447,6 +447,81 @@ matches strings containing "hello" before a newline, besides strings ending in
...
+  end
+
+The members field on the Band class will have class ``Mongoid::StringArray``.
```
p-mongo, line 467: One consequence of ``Mongoid::StringArray`` is that the class is reused for multiple models, in other words if an application wanted to customize their typed array for a particular model this wouldn't work well. Do you think this is a valid use case?

lib/mongoid/fields.rb
```diff
@@ -835,18 +835,61 @@ def retrieve_and_validate_type(name, type)
...
+                when self then object
+                when Array, Set
+                  new(object.to_a)
+                end
+              end
+              alias :demongoize :mongoize
+            end
+          end
+          array_class.const_set("Type", type)
+          Mongoid.const_set(const_string, array_class)
```
p-mongo, line 890: The entire body of this method needs to be under a mutex so that multiple threads don't try to define the same class at the same time.

docs/reference/fields.txt
Neilshweky: I'm not necessarily considering supporting type restrictions for other container types right now. 

It seems silly to make a Mongoid::StringIntegerHash class for every Type x Type combination. 

Do you have any recommendations on what an alternate syntax could be?

docs/reference/fields.txt
Neilshweky: Can you explain a specific use case for a customization? 

I feel like at some point we have to choose a functionality to provide out of the box, and if the user wants something else they can make a custom class.

p-mongo:
Correct me if I'm wrong:

- TypedArray defines demongoize as an alias of mongoize
- Meaning, during demongoization, on each individual element the method being called is mongoize, not demongoize

... which looks wrong.

I see there is a test for TypedArray(Range) but I don't understand it.

spec/mongoid/fields_spec.rb
```diff
@@ -486,6 +502,288 @@
...
+        it "demongoizes to a hash" do
+          expect(band.range_array).to eq([ { "min" => 1, "max" => 3 } ])
+        end
+
+        it "is stored as a hash in the database" do
+          expect(from_db.attributes["range_array"]).to eq([ { "min" => 1, "max" => 3 } ])
+        end
+
+        it "demongoizes to a hash from the database" do
+          expect(from_db.range_array).to eq([ { "min" => 1, "max" => 3 } ])
```
p-mongo, line 770: Shouldn't this yield Range instances?

p-mongo:
`evolve` doesn't need to be defined for typed arrays?

Neilshweky:
@p-mongo i went back and forth on this, that's why this range test is there, since it used to yield range objects like you suggested.

There was thought to that change, however, and it has to do with MONGOID-2951.

In order for the following case to be true:

```
Band.field :members, type: Array
band.members = []
band.members << 2
band.members
# => [ 2 ]
```
Then the demongoizer cannot perform any modifications on the mongoize values, since, the value returned by the demongoizer will be different than the one stored in the attributes hash. If we did this, any modification of the value returned by the demongoizer would not affect the value stored in the attributes hash, and actually in the above example, we would get the empty hash even though we pushed a value.

Because of this, container types have to return their mongoized values in the demongoizer, so that when they are modified, they also modify the value in the attributes hash. This leads to when getting regular array fields with mongoizable values:
```
band.genres = [1..3]
band.genres 
# => [{ "min" => 1, "max" => 3 }]
```
Because of this, I changed the typed array demongoizer to return the mongoized value to be consistent with regular array, and to avoid issues caused by MONGOID-2951

Neilshweky:
Note: change .*Array demongoize to default

lib/mongoid/fields.rb
Neilshweky: That's a good idea, however I think it would be better to name it `TypedArrayClassFactory` as what it produces is not TypedArrays, rather typed array classes.

Neilshweky: done

Neilshweky:
There is an existing problem with this PR that has to do with MONGOID-2951. I don't think it's worth holding off on this until then, as the problem already exists wrt Sets. I think we should move forward with this and document the weirdness, and hopefully fix it when we can. I have added a test that shows the oddity.

@p-mongo this is rfal.

Neilshweky:
> I can approve typed arrays behaving the same way as normal arrays, meaning the typed array itself permits mutation but the mutation won't be persisted by Mongoid due to 2951. Reads of Array(Range) should in my opinion return an array of range instances.

Regular arrays return ranges as hashes. You want regular arrays and typed arrays to behave differently in this case?

p-mongo:
That is untyped field behavior, incidentally one that I typed up in https://github.com/mongodb/mongoid/pull/5412. For typed arrays yes I think an array of ranges should return ranges as elements when reading.

Neilshweky:
> That is untyped field behavior, incidentally one that I typed up in #5412. For typed arrays yes I think an array of ranges should return ranges as elements when reading.

So the reason I balk at doing that is, if we do it that way we have to transform all of the elements in the array to and from mongoized form every time it's assigned, which, besides for the runtime issues, exacerbates the problems of MONGOID-2951. 

Currently, the way we're doing mongoize/demongoize is that when an array is assigned, it's converted to a TypedArray (or rather, a new one is created) and stored in the attributes hash, then on every demongoization, it is returned without any conversion, which avoids the issues from MONGOID-2951.

Now, the problems from MONGOID-2951 arise when we retrieve a document from the database and it's sitting in the attributes hash as a *regular* array. Now, on demongoization we have to convert the array to a typed array, which causes the MONGOID-2951 problems.

Now, if we were to demongoize all of the elements in the array every time (and return an array of ranges, for example), we would get the problems of MONGOID-2951 in every case, and not just when retrieving from the database.

If you are okay with that, I will make the change.

lib/mongoid/extensions/typed_array_class_factory.rb
```diff
@@ -0,0 +1,80 @@
...
+              array_class.const_set("Type", type)
+              Mongoid.const_set(const_string, array_class)
+            end
+          end
+        end
+
+        private
+
+        # @api private
+        LOCK = Mutex.new
```
johnnyshields, line 36: Declaring a `CONST` after `private` has no effect. Suggest to follow what seems to be Mongoid's code convention elsewhere to declare consts at/near top of class.

docs/reference/fields.txt
johnnyshields: Using square brackets `[ ]` instead of parenthesis `( )` happens to work:

```ruby
Array[String]  #=> [String]
Set[String]    #=> #<Set: {String}>
Hash[String, Integer]  #=> {String=>Integer}
```

I also think we should consider if there can be an equivalent Symbol syntax for each of these.

```ruby
:string_array
:string_set
:string_to_integer_hash
```

where `*_array`, `*_set`, and `*_to_*_hash` are privileged symbols

johnnyshields:
Neil, food for thought: as per your above comment it seems you are introducing `TypedArray` class as a solution to MONGOID-2951 specific to the scope of this PR. But since it doesn't solve the case of vanilla `Array`, *as a thought experiment*, should Mongoid:
- implement a more generic solution to MONGOID-2951 that would make `TypedArray` unnecessary here (use vanilla Array instead)
- do the opposite extreme and consider using ObjectArray (TypedArray with base type Object) as a base for all vanilla Arrays, i.e. `field type: Array` (non-typed)

Neilshweky:
I think you misunderstood the above comments... This PR does not fix 2951 it is affected by the problems caused by 2951

lib/mongoid/extensions/typed_array_class_factory.rb
p-mongo: No, this is a private constant and it doesn't need to be at the top.

Neilshweky:
> > That is untyped field behavior, incidentally one that I typed up in #5412. For typed arrays yes I think an array of ranges should return ranges as elements when reading.
> 
> So the reason I balk at doing that is, if we do it that way we have to transform all of the elements in the array to and from mongoized form every time it's assigned, which, besides for the runtime issues, exacerbates the problems of [MONGOID-2951](https://jira.mongodb.org/browse/MONGOID-2951).
> 
> Currently, the way we're doing mongoize/demongoize is that when an array is assigned, it's converted to a TypedArray (or rather, a new one is created) and stored in the attributes hash, then on every demongoization, it is returned without any conversion, which avoids the issues from [MONGOID-2951](https://jira.mongodb.org/browse/MONGOID-2951).
> 
> Now, the problems from [MONGOID-2951](https://jira.mongodb.org/browse/MONGOID-2951) arise when we retrieve a document from the database and it's sitting in the attributes hash as a _regular_ array. Now, on demongoization we have to convert the array to a typed array, which causes the [MONGOID-2951](https://jira.mongodb.org/browse/MONGOID-2951) problems.
> 
> Now, if we were to demongoize all of the elements in the array every time (and return an array of ranges, for example), we would get the problems of [MONGOID-2951](https://jira.mongodb.org/browse/MONGOID-2951) in every case, and not just when retrieving from the database.
> 
> If you are okay with that, I will make the change.

Waiting for your reply w.r.t this @p-mongo

spec/mongoid/extensions/typed_array_class_factory_spec.rb
```diff
@@ -0,0 +1,359 @@
...
+
+        it "has the correct type" do
+          expect(evolved.class).to eq(Array)
+        end
+      end
+
+      context "when passing in a bogus value" do
+        let(:arg) { "bogus" }
+
+        it "returns nil" do
```
Neilshweky, line 208:
```suggestion
        it "returns the original value" do
```

spec/mongoid/extensions/typed_array_class_factory_spec.rb
```diff
@@ -0,0 +1,359 @@
...
+        end
+
+        let!(:band) { Band.create!(range_array: [ 1..3 ]) }
+        let(:from_db) { Band.first }
+
+        it "mongoizes to a hash" do
+          expect(band.attributes["range_array"]).to eq([ { "min" => 1, "max" => 3 } ])
+        end
+
+        it "demongoizes to a hash" do
```
Neilshweky, line 304:
```suggestion
        it "demongoizes to a range" do
```

spec/mongoid/extensions/typed_array_class_factory_spec.rb
```diff
@@ -0,0 +1,359 @@
...
+          expect(band.range_array).to eq([ 1..3 ])
+        end
+
+        it "is stored as a hash in the database" do
+          expect(from_db.attributes["range_array"]).to eq([ { "min" => 1, "max" => 3 } ])
+        end
+
+        it "demongoizes to a hash from the database" do
```
Neilshweky, line 312:
```suggestion
        it "demongoizes to a range from the database" do
```

Neilshweky:
@p-mongo I have changed typed array values to contain demongoized values as we discussed. This is rfal.

docs/reference/fields.txt
```diff
@@ -471,6 +471,93 @@ matches strings containing "hello" before a newline, besides strings ending in
...
```
p-mongo, line 491: I would prefer the class nested under Fields, e.g. `Mongoid::Fields::StringArray` or in a dedicated module like `Mongoid::Fields::TypedArray::StringArray`.

docs/reference/fields.txt
```diff
@@ -471,6 +471,93 @@ matches strings containing "hello" before a newline, besides strings ending in
...
+``mongoized`` to Strings. When retrieving the array from the members field, the
+array is ``demongoized`` into a ``Mongoid::StringArray`` and all of its
+elements are ``demongoized`` as well. When subsequently appending to this array,
+the appended elements will also be ``demongoized`` to Strings.
+
+When a typed array field is declared, Mongoid will create a custom class for
+the type: ``Mongoid::<Type>Array``. This class overrides the ``mongoize``,
+``demongoize`` and ``evolve`` methods to convert the elements in the array to
+the array's inner type. This class inherits from the ``Mongoid::TypedArray``
+class.
```
p-mongo, line 502: Do you intend users to modify these classes? If not I think this needs to be explicitly discouraged here.

docs/reference/fields.txt
```diff
@@ -471,6 +471,93 @@ matches strings containing "hello" before a newline, besides strings ending in
...
+- ``unshift``
+- ``prepend``
+- ``insert``
+- ``fill``
+- ``replace``
+- ``concat``
+
+The ``Mongoid::TypedArray`` class also differs from ``Array`` in that its
+constructor takes in a type as well as the normal ``Array#new`` parameters.
+A user should never need to directly instantiate a ``Mongoid::TypedArray``
```
p-mongo, line 522:
```suggestion
Do not directly instantiate a ``Mongoid::TypedArray``
```

lib/config/locales/en.yml
```diff
@@ -205,6 +205,13 @@ en:
           resolution: "Please provide a valid type value for the field.
             Refer to:
             https://docs.mongodb.com/mongoid/current/reference/fields/#using-symbols-or-strings-instead-of-classes"
+        invalid_field_type_array:
+          message: "Invalid typed array field type for field '%{field}' on model
+            '%{klass}'."
+          summary: "Model '%{klass}' defines a field '%{field}' with an invalid
+            typed value array value. Attempted to pass %{types} types, when only
+            one is permitted."
+          resolution: "Please only pass in one type for a typed array."
```
p-mongo, line 214:
```suggestion
          resolution: "Please pass in a one-element array as the type for a typed array."
```

lib/mongoid/extensions/typed_array.rb
```diff
@@ -0,0 +1,239 @@
+# frozen_string_literal: true
+
+module Mongoid
+  class TypedArray < ::Array
```
p-mongo, line 4: This isn't an extension, TypedArray a type defined by Mongoid. It should be in `lib/mongoid/typed_array.rb`.

lib/mongoid/extensions/typed_array_class_factory.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Extensions
+
+    # A factory for manufacturing Mongoid::<Type>Array classes.
+    module TypedArrayClassFactory
```
p-mongo, line 7: This module should be `@api private` also methinks.

p-mongo (commented):
Looking at all of the demongoization code in typed array class, I am thinking that the entire feature as presently implemented should be implementable via a phantom field type, where the field type class will implement mongoize and demongoize converting plain Arrays to other plain Arrays with different values. The resulting arrays will be not persistable when mutated but this is also the behavior of the PR as it currently stands.

I might be missing the benefit of the demongoize methods right now, and the other reason why I think they should potentially be retained is for aligning this change with storing values in declared field type. But if XArray is the (effective) declared field type, mutating methods on it should be calling mongoize, not demongoize, which is something that confuses me in the current diff.

docs/reference/fields.txt
Neilshweky: mmm alright, but for the record, I think `Mongoid::StringArray` is super clean and nice looking.

docs/reference/fields.txt
Neilshweky: what do you mean by modify?

docs/reference/fields.txt
Neilshweky: I actually prefer to use my language, unless you really prefer yours.

Neilshweky:
> Looking at all of the demongoization code in typed array class, I am thinking that the entire feature as presently implemented should be implementable via a phantom field type, where the field type class will implement mongoize and demongoize converting plain Arrays to other plain Arrays with different values. The resulting arrays will be not persistable when mutated but this is also the behavior of the PR as it currently stands.
> 
> I might be missing the benefit of the demongoize methods right now, and the other reason why I think they should potentially be retained is for aligning this change with storing values in declared field type. But if XArray is the (effective) declared field type, mutating methods on it should be calling mongoize, not demongoize, which is something that confuses me in the current diff.

@p-mongo I don't understand this comment. Are you saying that the implementation of this feature as written in this PR is suboptiomal, and you're suggesting a new implementation?

lib/config/locales/en.yml
```diff
@@ -246,13 +246,21 @@ en:
...
+            one is permitted."
+          resolution: "Please pass in a one-element array as the type for a typed array."
         invalid_global_executor_concurrency:
           message: "Invalid global_executor_concurrency option."
           summary: "You set global_executor_concurrency while async_query_executor
             option is not set to :global_thread_pool. The global_executor_concurrency is
             allowed only for the global thread pool executor."
           resolution: "Set global_executor_concurrency option to :global_thread_pool
             or remove global_executor_concurrency option."
+            https://docs.mongodb.com/mongoid/current/reference/fields/#using-symbols-or-strings-instead-of-classes"
```
Neilshweky, line 263:
```suggestion
```

lib/mongoid/extensions/typed_array_class_factory.rb
Neilshweky: done

johnnyshields:
Are all the issues in this PR related to object mutation ([MONGOID-2951](https://jira.mongodb.org/browse/MONGOID-2951)) now resolved?

Neilshweky:
> Are all the issues in this PR related to object mutation ([MONGOID-2951](https://jira.mongodb.org/browse/MONGOID-2951)) now resolved?

Nope, that's going to be have to solved independently of this ticket...

docs/reference/fields.txt
```diff
@@ -471,6 +471,93 @@ matches strings containing "hello" before a newline, besides strings ending in
...
+Here's an example of using typed array fields:
+
+.. code:: ruby
+
+  > band = Band.new(members: [ 1, 2, 3 ])
+  # => #<Band _id: 62e1c39e3282a40b321f8e8e, members: ["1", "2", "3"]>
+  > string_array = band.members
+  # => ["1", "2", "3"]
+  > string_array.class
+  # => Mongoid::StringArray
```
johnnyshields, line 536: Should consider renaming this `Mongoid::ArrayOfString` rather than `StringArray`

I think if we look at other potential types I think feels better/less ambiguous (`SetOfString`, `HashOfStringToInteger`)

PR #5405, opened 2022-07-25
MONGOID-5293 Add Rails-style defaults & config.load_defaults

Neilshweky:
MONGOID-5293

### DIFF
--- a/lib/mongoid/config/defaults.rb
+++ b/lib/mongoid/config/defaults.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Config
+
+    # Encapsulates logic for loading defaults.
+    module Defaults
+
+      def load_defaults
+
+      end
+    end
+  end
+end
--- a/lib/mongoid/config/defaults.rb
+++ b/lib/mongoid/config/defaults.rb
@@ -26,32 +26,32 @@ module Mongoid
         case version.to_s
         when "7.3"
           # flags introduced in 7.4 - old functionality
-          Mongoid.broken_aggregables = true
-          Mongoid.broken_alias_handling = true
-          Mongoid.broken_and = true
-          Mongoid.broken_scoping = true
-          Mongoid.broken_updates = true
-          Mongoid.compare_time_by_ms = false
-          Mongoid.legacy_pluck_distinct = true
-          Mongoid.legacy_triple_equals = true
-          Mongoid.object_id_as_json_oid = true
+          self.broken_aggregables = true
+          self.broken_alias_handling = true
+          self.broken_and = true
+          self.broken_scoping = true
+          self.broken_updates = true
+          self.compare_time_by_ms = false
+          self.legacy_pluck_distinct = true
+          self.legacy_triple_equals = true
+          self.object_id_as_json_oid = true
 
           load_defaults "7.4"
         when "7.4"
           # flags introduced in 7.5 - old functionality
-          Mongoid.overwrite_chained_operators = true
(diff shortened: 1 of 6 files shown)


lib/mongoid/config/defaults.rb
```diff
@@ -0,0 +1,59 @@
...
+        # Note that for 7.x, since all of the feature flag defaults have been
+        # flipped to the new functionality, all of the settings for those
+        # versions are to give old functionality. Because of this, it is
+        # possible to recurse to later version to get all of the options to
+        # turn off. Note that this won't be true when adding feature flags to
+        # 8.1, since the default will be the old functionality until the next
+        # major version is released. More likely, the recursion will have to go
+        # in the other direction (towards earlier versions).
+
+        case version.to_s
```
p-mongo, line 26: This should explicitly round to 1 decimal place before the `to_s` call for good measure.

Or maybe more than one but there should be some sort of rounding to a certain precision performed.

lib/mongoid/config/defaults.rb
```diff
@@ -0,0 +1,59 @@
...
+        when "7.5"
+          # flags introduced in 8.0 - old functionality
+          Mongoid.legacy_attributes = true
+          Mongoid.map_big_decimal_to_decimal128 = false
+        when "8.0"
+          # All flag defaults currently reflect 8.0 behavior.
+        when "8.1"
+          # All flag defaults currently reflect 8.1 behavior.
+        else
+          raise ArgumentError, "Unknown version: #{version.to_s}"
```
p-mongo, line 54: `to_s` is not needed when the value is already inside a string being interpolated.

lib/mongoid/config/defaults.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        case version.to_s
+        when "7.3"
+          # flags introduced in 7.4 - old functionality
+          Mongoid.broken_aggregables = true
+          Mongoid.broken_alias_handling = true
+          Mongoid.broken_and = true
+          Mongoid.broken_scoping = true
+          Mongoid.broken_updates = true
+          Mongoid.compare_time_by_ms = false
```
p-mongo, line 34: Is `self` here the `Mongoid` module?

p-mongo (commented):
Can you please add the user documentation for this feature?

lib/mongoid/config/defaults.rb
Neilshweky: I followed Rails' lead on this one... The parameter can also be a string and it feels awkward to convert to a number, round, and convert back to a string.

lib/mongoid/config/defaults.rb
Neilshweky: done

lib/mongoid/config/defaults.rb
Neilshweky: yes, done.

lib/mongoid/config/defaults.rb
p-mongo: If that is how Rails does it it can't be wrong.

p-mongo (approved):
I approve what is here but this needs user documentation on https://www.mongodb.com/docs/mongoid/master/reference/configuration/ giving  a usage example and ideally listing the option bundles that are defined.

p-mongo (approved):
:+1: Very nice.

johnnyshields:
Neil, great work on this, it makes configuration much, much simpler 🎉  I've raised https://jira.mongodb.org/browse/MONGOID-5439 as a follow-up.

PR #5407, opened 2022-07-26
MONGOID-5264 clean up and comment the fields file

Neilshweky:


lib/mongoid/fields.rb
```diff
@@ -757,38 +779,72 @@ def generated_methods
...
+      # Returns the type of the field if the type was not in the TYPE_MAPPINGS
+      # hash.
+      #
+      # @param [ Symbol | Class ] type The type of the field.
+      #
+      # @return [ Class ] The type of the field.
+      #
+      # @api private
+      def unmapped_type(type)
+        if "Boolean" == type.to_s
```
p-mongo, line 844: I would like to get rid of this special case at some point but it's not in scope of the this PR.

p-mongo (approved):

lib/mongoid/fields.rb
Neilshweky: Im actually going to expand this method in the typed array PR lol

PR #5408, opened 2022-07-26
MONGOID-5264 add :present option to localized fields

Neilshweky:


lib/mongoid/fields/localized.rb
```diff
@@ -29,6 +29,16 @@ def localized?
         true
       end
 
+      # Is the localized field enforcing values to be present?
+      #
+      # @example Is the field localized?
+      #   field.present?
+      #
+      # @return [ true | false ] If the field enforces present.
+      def present?
```
johnnyshields, line 38: `present?` already is defined on Object by ActiveSupport and should be treated as a reserved method name. I propose to rename this to `remove_blank_localized_values?` or something similarly specific, since its also defined on Fields::Standard.

lib/mongoid/fields/localized.rb
```diff
@@ -29,6 +29,16 @@ def localized?
         true
       end
 
+      # Is the localized field enforcing values to be present?
+      #
+      # @example Is the field localized?
+      #   field.present?
```
johnnyshields, line 35: @example needs to be updated here, and also in `standard.rb`

docs/reference/fields.txt
```diff
@@ -1396,6 +1396,42 @@ You can get and set all the translations at once by using the corresponding ``_t
      { "en" => "Marvelous!", "de" => "Wunderbar!" }
 
 
+.. _present-fields:
+
+Present Fields
```
johnnyshields, line 1401: "Present Fields" as a header doesn't really sound clear to me as a reader. How about something like: "Auto-Remove Blank Translations"

lib/mongoid/fields/standard.rb
```diff
@@ -100,7 +100,7 @@ def localized?
       # Is the localized field enforcing values to be present?
       #
       # @example Is the field localized?
```
johnnyshields, line 102: `Is the field localized?` is still not the right text for this example.

docs/reference/fields.txt
```diff
@@ -1396,6 +1396,42 @@ You can get and set all the translations at once by using the corresponding ``_t
...
+
+   product.description_translations
+   # { "en" => "Marvelous!", "de" => "Fantastisch!" }
+
+   product.description = ""
+   product.description_translations
+   # { "en" => "Marvelous!" }
+
+When the empty string is written for the ``:de`` locale, the ``"de"`` key is
+removed from the ``_translations`` hash instead of writing an empty string.
```
p-mongo, line 1432:
```suggestion
removed from the ``_translations`` hash instead of writing the empty string.
```

lib/mongoid/attributes.rb
```diff
@@ -177,8 +177,14 @@ def write_attribute(name, value)
             attribute_will_change!(field_name)
           end
           if localized
-            attributes[field_name] ||= {}
-            attributes[field_name].merge!(typed_value)
+            present = fields[field_name].try(:localize_present?)
+            loc_key, loc_val = typed_value.to_a.flatten
```
p-mongo, line 181: If this does what I think it is attempting to do, I suggest `typed_value.first` instead of `typed_value.to_a.flatten`. If I am mistaken what is the purpose of `flatten` here?

lib/mongoid/attributes.rb
```diff
@@ -177,8 +177,14 @@ def write_attribute(name, value)
...
+            if present && loc_val.blank?
+              attributes[field_name].delete(loc_key)
```
p-mongo, line 183: You need the `||= {}` defaulting here otherwise empty string write as the first value fails:

```
serene2% mongoid=master rails c
Loading development environment (Rails 6.0.5.1)
irb(main):001:0>    class Product
     include Mongoid::Document
     field :description, localize: :present
   end

   I18n.default_locale = :en
   product = Product.new
   product.description = ''

/home/w/apps/mongoid/lib/mongoid/attributes.rb:183:in `block in write_attribute': undefined method `delete' for nil:NilClass (NoMethodError)

              attributes[field_name].delete(loc_key)
```

And this should be added as a test.

p-mongo: Or only call `delete` if `attributes[field_name]` is not nil.

p-mongo (approved):

p-mongo:
I think there should also be  a test which removes all non-blank values for documenting that Mongoid would write the empty hash to the database in that case. Also a test which writes  a single blank value and asserts that the outcome is empty hash written (as opposed to nothing/nil).

lib/mongoid/attributes.rb
Neilshweky: TIL first is a method on hashes. Much nicer. Thanks.

Neilshweky: done

lib/mongoid/attributes.rb
Neilshweky: done

lib/mongoid/attributes.rb
```diff
@@ -177,8 +177,14 @@ def write_attribute(name, value)
             attribute_will_change!(field_name)
           end
           if localized
-            attributes[field_name] ||= {}
-            attributes[field_name].merge!(typed_value)
+            present = fields[field_name].try(:localize_present?)
+            loc_key, loc_val = typed_value.first
+            if present && loc_val.blank?
+              attributes[field_name].delete(loc_key) unless attributes[field_name].nil?
```
p-mongo, line 183:
```suggestion
              attributes[field_name]&.delete(loc_key)
```

p-mongo (approved):

PR #5409, opened 2022-07-27
MONGOID-5417 Fix memory leak when call 'with'

comandeo:
https://jira.mongodb.org/browse/MONGOID-5417

### DIFF
--- a/lib/mongoid/persistence_context.rb
+++ b/lib/mongoid/persistence_context.rb
@@ -246,8 +246,8 @@ module Mongoid
       #
       # @api private
       def get_context(object)
-        Thread.current[:context] ||= {}
-        Thread.current[:context][object.object_id]
+        Thread.current["mongoid:persistence_context"] ||= {}
+        Thread.current["mongoid:persistence_context"][object.object_id]
       end
 
       # Store persistence context for a given object in the thread local
@@ -258,8 +258,12 @@ module Mongoid
       #
       # @api private
       def store_context(object, context)
-        Thread.current[:context] ||= {}
-        Thread.current[:context][object.object_id] = context
+        Thread.current["mongoid:persistence_context"] ||= {}
+        if context.nil?
+          Thread.current["mongoid:persistence_context"].delete(object.object_id)
+        else
+          Thread.current["mongoid:persistence_context"][object.object_id] = context
+        end
       end
     end
   end
--- a/lib/mongoid/persistence_context.rb
+++ b/lib/mongoid/persistence_context.rb
@@ -189,8 +189,7 @@ module Mongoid
       #
       # @return [ Mongoid::PersistenceContext ] The persistence context for the object.
       def set(object, options_or_context)
-        key = "[mongoid][#{object.object_id}]:context"
-        existing_context = Thread.current[key]
+        existing_context = get_context(object)
         existing_options = if existing_context
           existing_context.options
         else
@@ -201,7 +200,7 @@ module Mongoid
         end
         new_options = existing_options.merge(options_or_context)
         context = PersistenceContext.new(object, new_options)
-        Thread.current[key] = context
+        store_context(object, context)
       end
 
       # Get the persistence context for a particular class or model instance.
@@ -213,7 +212,7 @@ module Mongoid
       #
       # @return [ Mongoid::PersistenceContext ] The persistence context for the object.
       def get(object)
-        Thread.current["[mongoid][#{object.object_id}]:context"]
+        get_context(object)
       end
 
       # Clear the persistence context for a particular class or model instance.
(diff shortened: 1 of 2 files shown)


lib/mongoid/persistence_context.rb
```diff
@@ -232,7 +231,35 @@ def clear(object, cluster = nil, original_context = nil)
...
+      #
+      # @param [ Object ] object Object to get the persistance context for.
+      #
+      # @return [ Mongoid::PersistenceContext | nil ] The persistence context
+      #   for the object if previously stored, otherwise nil.
+      #
+      # @api private
+      def get_context(object)
+        Thread.current[:context] ||= {}
+        Thread.current[:context][object.object_id]
```
johnnyshields, line 250: The key `:context` is too generic and may be used by other gems.

I recommend key `"mongoid:persistence_context"`

lib/mongoid/persistence_context.rb
```diff
@@ -232,7 +231,35 @@ def clear(object, cluster = nil, original_context = nil)
...
+      # Store persistence context for a given object in the thread local
+      #   storage.
+      #
+      # @param [ Object ] object Object to store the persistance context for.
+      # @param [ Mongoid::PersistenceContext ] context Context to store
+      #
+      # @api private
+      def store_context(object, context)
+        Thread.current[:context] ||= {}
+        Thread.current[:context][object.object_id] = context
```
johnnyshields, line 262: It's not clear to me how this fixes the memory leak, since the context hash itself is never reset. It will keep growing and growing as more objects accumulate.

lib/mongoid/persistence_context.rb
comandeo: Good point, thanks. Fixed.

lib/mongoid/persistence_context.rb
comandeo: The problem was that a new symbol was created every time we store the context, and never released. If we use `object_id` as a key, no new object is created.
However, you are right that we can do better here.

lib/mongoid/persistence_context.rb
```diff
@@ -232,7 +231,39 @@ def clear(object, cluster = nil, original_context = nil)
...
+      #
+      # @param [ Object ] object Object to get the persistance context for.
+      #
+      # @return [ Mongoid::PersistenceContext | nil ] The persistence context
+      #   for the object if previously stored, otherwise nil.
+      #
+      # @api private
+      def get_context(object)
+        Thread.current["mongoid:persistence_context"] ||= {}
+        Thread.current["mongoid:persistence_context"][object.object_id]
```
Neilshweky, line 250: Is there a reason why you're changing how the contexts are stored?

lib/mongoid/persistence_context.rb
```diff
@@ -232,7 +231,39 @@ def clear(object, cluster = nil, original_context = nil)
...
+      #   storage.
+      #
+      # @param [ Object ] object Object to store the persistance context for.
+      # @param [ Mongoid::PersistenceContext ] context Context to store
+      #
+      # @api private
+      def store_context(object, context)
+        Thread.current["mongoid:persistence_context"] ||= {}
+        if context.nil?
+          Thread.current["mongoid:persistence_context"].delete(object.object_id)
```
Neilshweky, line 263: When would the context ever be nil?

Neilshweky (commented):
This is MONGOID-5417 not 5414 btw

lib/mongoid/persistence_context.rb
Neilshweky: I saw your other comment. Cool

lib/mongoid/persistence_context.rb
```diff
@@ -232,7 +231,39 @@ def clear(object, cluster = nil, original_context = nil)
...
+      #   storage.
+      #
+      # @param [ Object ] object Object to get the persistance context for.
+      #
+      # @return [ Mongoid::PersistenceContext | nil ] The persistence context
+      #   for the object if previously stored, otherwise nil.
+      #
+      # @api private
+      def get_context(object)
+        Thread.current["mongoid:persistence_context"] ||= {}
```
p-mongo, line 249: Let's stick with `[mongoid]` prefix which is used everywhere except for https://jira.mongodb.org/browse/MONGOID-5441.

lib/mongoid/persistence_context.rb
```diff
@@ -232,7 +231,39 @@ def clear(object, cluster = nil, original_context = nil)
...
+
+      # Store persistence context for a given object in the thread local
+      #   storage.
+      #
+      # @param [ Object ] object Object to store the persistance context for.
+      # @param [ Mongoid::PersistenceContext ] context Context to store
+      #
+      # @api private
+      def store_context(object, context)
+        Thread.current["mongoid:persistence_context"] ||= {}
```
p-mongo, line 261: This line can be moved into the `else` path and line 263 changed to

```
          Thread.current["mongoid:persistence_context"]&.delete(object.object_id)
``

lib/mongoid/persistence_context.rb
p-mongo: This method also restores context to the previous one, and if all contexts are unwound there wouldn't be a previous one eventually.

p-mongo:
Can the submitter's test be added to this PR? It should call .with on an object to warm up the symbols then call .with on another object, GC and verify symbol count hasn't grown.

lib/mongoid/persistence_context.rb
comandeo: Done

lib/mongoid/persistence_context.rb
comandeo: Done

comandeo:
> Can the submitter's test be added to this PR? It should call .with on an object to warm up the symbols then call .with on another object, GC and verify symbol count hasn't grown.

I added the test; however, I am afraid it may become flaky. We have multiple threads running in the driver, what if a symbol will be created there meanwhile?

johnnyshields:
This is quite hacky, but you could try wrapping in the tests in `mutex.synchronize` from the appropriate mutex lock(s) from the driver.

p-mongo:
The driver has solo test infrastructure, something like that ought to work here well enough. I can look into it next week.

comandeo:
@p-mongo Are you still going to look into it? I have no objection merging with this test. If is notice it is flaky, we can remove (or improve) it later.

p-mongo:
Yes, 2 out of 3 evergreen failures for this PR are in the counting test.

comandeo:
@p-mongo It turned out that the test fails on jruby only. I think it would be OK if we just skip this test on jruby, coverage on MRI should be enough.

p-mongo (approved):
Yes if MRI is always green that's sufficient.

PR #5410, opened 2022-07-28
MONGOID-5415: Code docs for blocks should use @yield

johnnyshields:
Includes an update to the docs page, including a few unrelated items to docs like `@raise` and `@todo` which are both used currently in the codebase.

### DIFF
--- a/docs/contributing/code-documentation.txt
+++ b/docs/contributing/code-documentation.txt
@@ -35,7 +35,7 @@ Structure
     class CardboardBox
 
 - **Methods:** All method definitions should be preceded by a documentation comment.
-  Use ``@param`` and ``@return`` to specify input(s) and output respectively.
+  Use ``@param``, ``@yield``, and ``@return`` to specify inputs and output.
   For further details, refer to
   :ref:`Type Declaration <code-documentation-type-declaration>` below.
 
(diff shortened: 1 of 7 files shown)


johnnyshields:
@p-mongo this is ready for review.

johnnyshields:
Ready for final review

docs/contributing/code-documentation.txt
```diff
@@ -286,3 +295,58 @@ Type Declaration
     # @option **kwargs [ String | Array<String> ] :items The items(s) as Strings to include.
     # @option **kwargs [ Integer ] :limit An Integer denoting the limit.
     def buy_groceries(**kwargs)
+
+- **Blocks:** Use ``@yield`` to specify when the method yields to a block.
```
p-mongo, line 299: Can you reference where in the Ruby stdlib this behavior is defined please?

docs/contributing/code-documentation.txt
```diff
@@ -286,3 +295,58 @@ Type Declaration
...
+
+    # @yield [ Symbol, Symbol, Symbol ] Evaluates the guess of who did the crime.
+    #   Must take the person, location, and weapon used. Must return true or false.
+    def whodunit
+      yield(:mustard, :ballroom, :candlestick)
+    end
+
+- **Blocks:** If the method explicitly specifies a block argument, include the
+  argument name in ``@yield`` preceded by an ampersand ``&``. Also use the ``@yield``
+  macro when the method uses ``block.call`` rather than ``yield`` internally.
```
p-mongo, line 311: I don't see the purpose of this paragraph. `@yield` documents that the method yields, it doesn't matter that the block is taken as an explicit argument.

docs/contributing/code-documentation.txt
johnnyshields: `yield` is defined as `rb_yield` in [vm_eval.c](https://github.com/ruby/ruby/blob/master/vm_eval.c#L1357). Here's the (doc in Japanese)[https://docs.ruby-lang.org/ja/latest/function/rb_yield.html]. I can't find the English equivalent.

docs/contributing/code-documentation.txt
johnnyshields: @p-mongo If the discussion here is meant to be whether the behavior of a `&block` arg should be specified by `@param` or `@yield`, in [YARD itself](https://github.com/lsegal/yard/commit/50384a2bdcad6ffc90c9ecfd5877ab02bb9ab9b2) there is ambiguity and it seems that both ways are supported.

For my tastes, it seems more useful to always use `@yield`, `@yieldparam`, `@yieldreturn` to describe the shape of *how a passed-in block behaves*, and not get hung up over whether (a) the block is explicitly declares as a param `&block` and (b) the method is uses `yield` and `block.call` internally because in both (a) and (b) the end result is functionally equivalent.

Again: **this is a matter of preference.** My opinion here is that the code docs should focus on *semantic behavior*, and not a literal statement of `yield` vs. `block.call` appears in the code. In my view, if you want to know that, you can just read the code 😄 

If you'd like to do things differently, its your call, so just let me know what to do.

docs/contributing/code-documentation.txt
p-mongo: I meant the `@yield` clause. Looking at this again, the yield parameters should either go in `@yieldparam` etc. or they should be in `@yield` but both (or arbitrary choosing) doesn't make sense to me.

lib/mongoid/association/embedded/embeds_many/proxy.rb
```diff
@@ -236,7 +236,7 @@ def exists?
           #   person.addresses.find { |addr| addr.state == 'CA' }
           #
           # @param [ Object... ] *args Various arguments.
-          # @param [ Proc ] block Optional block to pass.
+          # @yield &block Optional block to pass.
```
p-mongo, line 239: For example, here, the block is a parameter and should be documented as `@param`. There may additionally be a `@yield` clause stating the method yields to the passed block, with `@yieldparam` listing the parameters yielded, if any.

docs/contributing/code-documentation.txt
```diff
@@ -48,33 +48,30 @@ Structure
     # @return [ Tiger ] The transmogrified result.
     def transmogrify(person)
 
+- **Errors:** Use ``@raise`` to explain errors specific to the method.
+
+  .. code-block:: ruby
+
+    # @raise [ Errors::Validations ] If validation failed.
+    def validate!
```
p-mongo, line 56: Can you please put everything that is not related to yielding to another PR.

johnnyshields:
Will do all the changes mentioned, please give me a few days. Thanks!

docs/contributing/code-documentation.txt
johnnyshields: Extracted here: https://github.com/mongodb/mongoid/pull/5429

johnnyshields:
@p-mongo this is done as you've specified above (I think). If you'd like to make any changes, please commit them directly.

I personally prefer doing `@yield &block` to `@param &block + @yield` as the former feels more terse, but it's your call. The important thing is to make a standard and stick to it.

p-mongo (approved):

Neilshweky (approved):

comandeo (approved):

PR #5413, opened 2022-07-28
MONGOID-5441 prefix Thread.current[:session]

Neilshweky:


lib/mongoid/threaded.rb
```diff
@@ -322,7 +322,7 @@ def validations_for(klass)
     #
     # @param [ Mongo::Session ] session The session to save.
     def set_session(session)
-      Thread.current[:session] = session
+      Thread.current["[mongoid]:session"] = session
```
johnnyshields, line 325: should be a const marked `@api private` ?

p-mongo (approved):

comandeo (approved):

PR #5416, opened 2022-07-29
MONGOID-5334 ensure localized demongoization works with symbol keys

Neilshweky:
Only use case would be if the user modified the translations hash directly, but there were already some places that checked for both string and symbol keys so I just made it work in fallbacks too.

### DIFF
--- a/lib/mongoid/fields/localized.rb
+++ b/lib/mongoid/fields/localized.rb
@@ -14,9 +14,8 @@ module Mongoid
       #
       # @return [ Object ] The value for the current locale.
       def demongoize(object)
-        if object
-          type.demongoize(lookup(object))
-        end
+        return if object.nil?
+        type.demongoize(lookup(object))
       end
 
       # Is the field localized or not?
@@ -86,7 +85,10 @@ module Mongoid
         end
         return value unless value.nil?
         if fallbacks? && ::I18n.respond_to?(:fallbacks)
-          object[::I18n.fallbacks[locale].map(&:to_s).find{ |loc| object.has_key?(loc) }]
+          fallback_key = ::I18n.fallbacks[locale].find do |loc|
+            object.has_key?(loc.to_s) || object.has_key?(loc)
+          end
+          object[fallback_key.to_s] || object[fallback_key]
         end
       end
     end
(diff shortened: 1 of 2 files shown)


johnnyshields:
Nice one. I vaguely remember encountering this case in my app.

lib/mongoid/fields/localized.rb
```diff
@@ -86,7 +85,10 @@ def lookup(object)
         end
         return value unless value.nil?
         if fallbacks? && ::I18n.respond_to?(:fallbacks)
-          object[::I18n.fallbacks[locale].map(&:to_s).find{ |loc| object.has_key?(loc) }]
+          fallback_key = ::I18n.fallbacks[locale].find do |loc|
+            object.has_key?(loc.to_s) || object.has_key?(loc)
+          end
+          object[fallback_key.to_s] || object[fallback_key]
```
p-mongo, line 91: What is the reason for stringified key being used in preference to the actual key?

Neilshweky: I'm not sure, but if I had to guess, I would think it's because rails does it this way. Just like they stringify the keys of hashes... I wouldn't be opposed to changing this to be a BSON document, but I also think this is fine as it is.

p-mongo: If you don't have a strong reason can this be

```
object[fallback_key] || object[fallback_key.to_s]
```

p-mongo (approved):
Approved with the requested key consulted first.

lib/mongoid/fields/localized.rb
Neilshweky: I rather not, and I'll tell you why. When adding values to the localized fields, they are added to the translations hash using string keys. The symbol keys would only be in the translations hash if users put them there themselves. It seems better to prioritize string keys for that reason, and that is what's done a few lines above this one (existing code).

p-mongo: i18n keys are symbols, the localized field value hashes' keys are strings, therefore it makes some sense to check first for the stringified key and then use the original key.

p-mongo (approved):

PR #5417, opened 2022-07-29
MONGOID-5328 (de)mongoize ActionController::Paramters

Neilshweky:


lib/mongoid/extensions/action_controller_parameters.rb
```diff
@@ -0,0 +1,38 @@
...
+        # Turn the object from the Ruby type into a Mongo-friendly type.
+        #
+        # @example Mongoize the object.
+        #   ActionController::Parameters.mongoize([ 1, 2, 3 ])
+        #
+        # @param [ Object ] object The object to mongoize.
+        #
+        # @return [ Hash ] The object mongoized.
+        def mongoize(object)
+          ::Hash.mongoize(object)
```
johnnyshields, line 28: should probably make a note somewhere that this intentionally uses `.to_h` and not `.to_unsafe_h`.

PR #5423, opened 2022-08-01
swap delete & destroy


p-mongo:


p-mongo:
No content changes in this PR, just moving existing language around for hopefully better flow.

docs/reference/crud.txt
```diff
@@ -11,6 +11,188 @@ CRUD Operations
...
+``read_attribute`` and ``write_attribute`` methods can be used to read and
+write the values directly into the attributes hash:
+
+.. code-block:: ruby
+
+  class Person
+    include Mongoid::Document
+
+    def first_name
+      read_attribute(:fn)
```
Neilshweky, line 67: why is this read_attribute(:fn) I would think the purpose of using read_attribute is when you override the getter of the same name? 

So if you had:
```
class Person 
  include Mongoid::Document
  field :first_name, type: String

  def first_name
    fn = read_attribute(:first_name)
    # some modifications to fn
  end    
end
```
This way you could get the value for :first_name without calling self.first_name and causing an infinite loop.

Neilshweky: Otherwise, if the field is called fn, just call `self.fn`

docs/reference/crud.txt
```diff
@@ -228,56 +410,56 @@ operations with examples.
           person.touch
           person.touch(:audited_at)
 
-   * - ``Model#delete``
+   * - ``Model#destroy``
```
Neilshweky, line 413: what was the purpose of switching the order of delete and destroy here? 

I think it was put in this order for the sake of alphabetical sorting...

Neilshweky: same for delete_all, destroy_all

docs/reference/fields.txt
```diff
@@ -576,93 +578,341 @@ you have three options:
...
+        when Hash then Point.new(object[:x], object[:y]).mongoize
+        else object
+        end
+      end
 
-    field :state, type: String, default: 'created'
+      # Get the object as it was stored in the database, and instantiate
+      # this custom class from it.
+      def demongoize(object)
+        Point.new(object[0], object[1])
```
Neilshweky, line 631: Do we care that this code is not representative with how we have implemented our demongoize methods? I would check that the input is an array or else return nil...

docs/reference/fields.txt
```diff
@@ -576,93 +578,341 @@ you have three options:
...
 
-  Default values that are not ``Proc`` instances are evaluated at class load
-  time, meaning the following two definitions are not equivalent:
+  The ``mongoize`` and ``demongoize`` methods should return ``nil`` on values
+  that are uncastable to your custom type. See the section on :ref:`Uncastable
+  Values <uncastable-values>` for more details.
 
-  .. code-block:: ruby
+Lastly, the class method ``evolve`` is similar to ``mongoize``, however it is used
+when transforming objects for use in Mongoid query criteria.
```
Neilshweky, line 674: You should note that the difference between evolve and mongoize is that it doesn't return nil on uncastable values.

docs/reference/fields.txt
```diff
@@ -576,93 +578,341 @@ you have three options:
...
+  profile.save!
+
+
+.. _custom-field-options:
+
+Custom Field Options
+====================
+
+You may define custom options for the ``field`` macro function
+which extend its behavior at the your time model classes are loaded.
```
Neilshweky, line 746:
```suggestion
which extend its behavior at the time your model classes are loaded.
```

docs/reference/fields.txt
```diff
@@ -576,93 +578,341 @@ you have three options:
...
+
+As an example, we will define a ``:required`` option which will add a presence
+validator for the field. First, declare the new field option in an initializer,
+specifiying its handler function as a block:
+
+.. code-block:: ruby
+
+  # in /config/initializers/mongoid_custom_fields.rb
+
+  Mongoid::Fields.option :required do |model, field, value|
```
Neilshweky, line 756: `required` is already a defined option on fields.

Neilshweky: It's used here: https://github.com/mongodb/mongoid/blob/fe9d44baae7f9fb8c9ba561e3cb33561bdfe02ce/lib/mongoid/association/referenced/belongs_to.rb#L208

docs/reference/fields.txt
```diff
@@ -576,93 +578,341 @@ you have three options:
...
+
+It is possible to replace the ``_id`` field definition to change the type
+of the ``_id`` values or have different default values:
+
+.. code-block:: ruby
+
+  class Band
+    include Mongoid::Document
+    field :name, type: String
+    field :_id, type: String, default: ->{ name }
```
Neilshweky, line 830: Do we want to note that associations are not available inside the default proc?

i.e. the following won't work:
```
class Band
     include Mongoid::Document
     field :name, type: String
     field :_id, type: String, default: ->{ label.name }
     
     embeds_one :label
end
```

docs/reference/fields.txt
```diff
@@ -576,93 +578,341 @@ you have three options:
...
+
+.. note::
+
+  Default values that are not ``Proc`` instances are evaluated at class load
+  time, meaning the following two definitions are not equivalent:
+
+  .. code-block:: ruby
+
+    field :submitted_at, type: Time, default: Time.now
+    field :submitted_at, type: Time, default: ->{ Time.now }
```
Neilshweky, line 903: Perhaps now would be the right time to mention that using associations in procs won't work, rather than above.

docs/reference/crud.txt
p-mongo: This example shows how to implement storage field names (which an application wouldn't do because this functionality is provided by Mongoid itself). The read is intentionally of another attribute than what the model exposes.

docs/reference/crud.txt
p-mongo: I personally find a "logical" order easier to read and follow. With that said, I can see how what is "logical" can vary from one person to the next.

In this case, `destroy` is an operation which should be used by default to remove documents, and `delete` should be explicitly chosen over `destroy` for performance reasons after considering whether there would be orphan objects left, whether any callbacks would be skipped, etc. Thus `destroy` comes first and `delete` comes second in my mind.

docs/reference/fields.txt
p-mongo: Yes, I filed https://jira.mongodb.org/browse/MONGOID-5460 for this.

docs/reference/fields.txt
p-mongo: Added this to https://jira.mongodb.org/browse/MONGOID-5460.

docs/reference/fields.txt
p-mongo: Filed https://jira.mongodb.org/browse/MONGOID-5461 for this.

docs/reference/fields.txt
p-mongo: https://jira.mongodb.org/browse/MONGOID-3267 would imply to me that it is in fact possible?

PR #5425, opened 2022-08-01
MONGOID-5293 put legacy_attributes in 7.5 defaults

Neilshweky:
Once we've back ported `legacy_attributes` to 7.5, the defaults for 7.4 should have `legacy_attributes` true, and 7.5 should have `legacy_attributes` false (as per the current default).

### DIFF
--- a/lib/mongoid/config/defaults.rb
+++ b/lib/mongoid/config/defaults.rb
@@ -39,8 +39,8 @@ module Mongoid
           load_defaults "7.4"
         when "7.4"
           # flags introduced in 7.5 - old functionality
-          self.overwrite_chained_operators = true
           self.legacy_attributes = true
+          self.overwrite_chained_operators = true
 
           load_defaults "7.5"
         when "7.5"
--- a/lib/mongoid/config/defaults.rb
+++ b/lib/mongoid/config/defaults.rb
@@ -40,11 +40,11 @@ module Mongoid
         when "7.4"
           # flags introduced in 7.5 - old functionality
           self.overwrite_chained_operators = true
+          self.legacy_attributes = true
 
           load_defaults "7.5"
         when "7.5"
           # flags introduced in 8.0 - old functionality
-          self.legacy_attributes = true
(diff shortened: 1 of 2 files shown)


lib/mongoid/config/defaults.rb
```diff
@@ -39,12 +39,12 @@ def load_defaults(version)
           load_defaults "7.4"
         when "7.4"
           # flags introduced in 7.5 - old functionality
+          self.legacy_attributes = true
           self.overwrite_chained_operators = true
 
           load_defaults "7.5"
         when "7.5"
           # flags introduced in 8.0 - old functionality
-          self.legacy_attributes = true
```
p-mongo, line 47: I think each branch here should set all relevant flags, in case any of them were changed previously from Mongoid's defaults, so that when an application asks for e.g. 7.5 defaults, all flags that are relevant to 7.5 are set to their default values for 7.5 regardless of existing state.

spec/mongoid/config/defaults_spec.rb
```diff
@@ -40,26 +40,26 @@
...
     shared_examples "turns on 7.5 flags" do
       it "turns on the 7.5 flags" do
+        expect(Mongoid.legacy_attributes).to be false
         expect(Mongoid.overwrite_chained_operators).to be false
       end
     end
 
     shared_examples "turns off 8.0 flags" do
       it "turns off the 8.0 flags" do
-        expect(Mongoid.legacy_attributes).to be true
```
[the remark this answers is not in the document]
Neilshweky: Each of the tests includes examples on/off from 7.4/7.5/8.0

lib/mongoid/config/defaults.rb
Neilshweky: I don't think that's necessary as the flags you're talking about are never going to get flipped back to old functionality. They are about to get removed, actually.

Neilshweky:
Note, this has to be back ported to 8.1

p-mongo (approved):

PR #5427, opened 2022-08-01
MONGOID-5438 Implement local override for i18n parameters

Neilshweky:


Neilshweky:
@johnnyshields I didn't want to reset the I18n configs for all examples so I created a new macro that does it for the specific examples.... Let me know what you think.

spec/support/macros.rb
```diff
@@ -77,5 +77,19 @@ def query_cache_enabled
         end
       end
     end
+
+    def with_default_i18n_configs
+      around do |ex|
+        I18n.locale = :en
```
Neilshweky, line 83: I'm not sure if it's necessary to do this before and after, but I felt like it was a more airtight solution. I also tried making a method that does this, but I wasn't able to access it from within the around block.

Neilshweky:
TIL that around blocks always have higher precedence over before blocks in RSpec. Even if the around block is embedded in a context! Take the following example: 
```
describe "rspec stuff" do

  before do
    puts "before"
  end

  around do |ex|
    puts "before around"
    ex.run
    puts "after around"
  end

  context "context" do

    before do
      puts "inner before"
    end

    around do |ex|
      puts "inner before around"
      ex.run
      puts "inner after around"
    end


    it "passes" do
      expect(true).to be true
    end
  end
end
```
yields: 
```
before around
inner before around
before
inner before
inner after around
after around
```

johnnyshields:
> I didn't want to reset the I18n configs for all examples so I created a new macro that does it for the specific examples.... Let me know what you think.

In my opinion the best practice is to always reset at the base level as I had in my original PR. A few reasons:
1. We already have a precedent of resetting timezone and a few other globals. There is nothing that makes locale special from other global state.
2. Humans writing specs are unaware, lazy and forgetful: they will unlikely to know about or use the `with_default_i18n_configs` macro unless we remind them--and we ourselves may even forget in six months time!

p-mongo (approved):
The macro name feels a bit awkward to me but I am not sure what to rename it to right now, maybe it'll become more obvious in the future.

johnnyshields:
For my knowledge, is there a reason a macro which needs to be added each time is preferable to a global reset here?

p-mongo:
It is possible to reset everything for every single test. This 1) makes the test suite run longer, 2) removes the indication of what state, or other environment aspects, each test depends on. I personally find such indication valuable.

johnnyshields:
> makes the test suite run longer

Doubt it would be more than 1s in total across suite. Resetting in-Ruby variables is negligible compared to latency of DB actions done on `before(:each)`, such as dropping all collections.

> removes the indication of what state... each test depends on. I personally find such indication valuable.

It's a matter of preference I guess. I raised these I18n PRs originally because various Mongoid developers had not used the test helpers consistently. It's a virtual certainty that `with_default_i18n_configs` will become similarly forgotten/ignored over time. Newbie Mongoid developers will burn human cycles debugging why their locale-related specs are strangely failing, then fix it by adding an ad-hoc `before { I18n.locale = :en }`... the experienced devs will ask them to change it to `with_default_i18n_configs` in the code review, creating additional toil to fix it. And so the cycle repeats... ☸️

Anyway, as you wish...

p-mongo:
Need to figure out how to reset fallbacks.

Neilshweky:
> Need to figure out how to reset fallbacks.

@p-mongo I think I figured out how to reset the fallbacks.

spec/support/macros.rb
```diff
@@ -47,11 +44,23 @@ def with_i18n_fallbacks
       require_fallbacks
 
       around do |example|
-        old_backend = I18n.backend
-        I18n.backend = I18nBackendWithFallbacks.new
+        include I18n::Backend::Fallbacks
+        if I18n.respond_to?(:temp_fallbacks)
+          class << I18n
```
Neilshweky, line 49: This is a little hacky. Turns out. once you include the `I18n::Backend::Fallbacks` module anywhere, the fallbacks methods are defined on I18n, and reassigning the old backend doesn't change that. This is because the `fallbacks` and `fallbacks=` methods are defined outside of the fallbacks module onto I18n itself.

This changed makes it such that the two fallbacks methods are undefined after each test, and redefined later.

p-mongo: These methods actually are defined when the backend is loaded (from filesystem), it doesn't even need to be instantiated:

```
serene2% irb -ri18n
irb(main):001:0> I18n.fallbacks
(irb):1:in `<main>': undefined method `fallbacks' for I18n:Module (NoMethodError)
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
	from /home/w/.rbenv/versions/3.1/bin/irb:25:in `load'
	from /home/w/.rbenv/versions/3.1/bin/irb:25:in `<main>'
irb(main):002:0> require 'i18n/backend/fallbacks'
=> true
irb(main):003:0> I18n.fallbacks
=> {}
```

The presence of `fallbacks` method doesn't mean fallbacks actually are used:
```
irb(main):001:0> I18n.enforce_available_locales = false
=> false
irb(main):002:0> I18n.fallbacks
(irb):2:in `<main>': undefined method `fallbacks' for I18n:Module (NoMethodError)
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
	from /home/w/.rbenv/versions/3.1/bin/irb:25:in `load'
	from /home/w/.rbenv/versions/3.1/bin/irb:25:in `<main>'
irb(main):003:0> I18n.backend.store_translations('en',{'a'=>'a-en'})
=> {:a=>"a-en"}
irb(main):004:0> I18n.backend.store_translations('ru',{'b'=>'b-ru'})
=> {:b=>"b-ru"}
irb(main):005:0> I18n.t('b')
=> "translation missing: en.b"
irb(main):006:0> require 'i18n/backend/fallbacks'
=> true
irb(main):007:0> I18n.fallbacks
=> {}
irb(main):008:0> I18n.fallbacks[:en]=[:ru]
=> [:ru]
irb(main):009:0> I18n.t('b')
=> "translation missing: en.b"
irb(main):010:0> I18n.backend.class.included_modules.include?(I18n::Backend::Fallbacks)
=> false
irb(main):012:0> I18n.locale='en'
=> "en"
irb(main):013:0> I18n.t('b')
=> "translation missing: en.b"
irb(main):014:0> I18n::Backend::Simple.include(I18n::Backend::Fallbacks)
=> I18n::Backend::Simple
irb(main):015:0> I18n.t('b')
=> "b-ru"
```

The fallbacks logic doesn't identify whether it is active in a sensible way. The best condition for checking whether fallbacks are actually enabled is `I18n.backend.class.included_modules.include?(I18n::Backend::Fallbacks)` as far as I can tell, quite ugly.

What we can do is always test fallbacks in the test configs with the fallbacks enabled, as is the current setup, and add an assertion at the end of a test run that `fallbacks` isn't defined if the test config doesn't have fallbacks set. With this we just need to make sure we don't require any fallback code accidentally and we won't need to go into i18n guts.

p-mongo:
We decided to:

- run the entire test suite in evergreen in the fallback configurations (instead of individual files)
- add macros to skip tests when fallbacks are enabled and disabled (the disabled one already sort of exists)
- add an `after_suite` hook to spec helper to verify that when fallbacks aren't enabled in the test configuration, `I18n.fallbacks` is not defined at the end of the test run (to guard against unintentional fallback loading)
- add back setting of locales in the tests themselves
- adjust tests as needed to not fail when fallbacks are enabled and the entire test suite is run (assuming there is no meaningful behavior change when fallbacks are on vs off)

johnnyshields:
Can you explain how this decision reflects best practices established by Rails, I18n gem, and elsewhere? I'd like to see an example of one other gem that runs its entire test suite of the cases of fallbacks on/off.

p-mongo:
Once the fallbacks backend is required it mutates global state of i18n. There is no public/sensible API to check for fallbacks being loaded, as per my earlier comment. The options are do what I just said or insert i18n's module names into Mongoid for checking whether fallbacks are being used.

johnnyshields:
> Once the fallbacks backend is required it mutates global state of i18n

Why not mutate it back to the original state? For example using [Module#remove_const](https://apidock.com/ruby/Module/remove_const). In Ruby, all things are possible.

p-mongo:
Because then we have to update our code if i18n decide to change theirs, and in the meantime we could have either test failures upon a release of new i18n versions or our tests won't behave as we intend, potentially silently. The proposed solution doesn't have these issues.

johnnyshields:
> Because then we have to update our code if i18n decide to change theirs

This is true of any dependency, be it Rails, I18n, Ruby Driver, etc. I18n gem code seems pretty mature/stable, but perhaps RSpec Mocks [allow/expect](https://relishapp.com/rspec/rspec-mocks/docs/configuring-responses) and [hide_const](https://www.rubydoc.info/github/rspec/rspec-mocks/RSpec%2FMocks%2FExampleMethods:hide_const) can also be useful.

> potentially silently

If tests cover the negative cases, i.e. "If fallbacks disabled then localized fields do not fallback", etc. then silent failures would be unlikely.

Anyway, I would do it differently, but it's your call.

spec/spec_helper.rb
```diff
@@ -123,6 +123,9 @@ class Query
 end
 
 I18n.config.enforce_available_locales = false
+if %w(yes true 1).include?((ENV['TEST_I18N_FALLBACKS'] || '').downcase)
+  include I18n::Backend::Fallbacks
```
p-mongo, line 127: This includes into the global scope no? If you want `I18n.fallbacks` defined at this point this should just `require` the backend.

Neilshweky:
TIL that not only does ActiveSupport 7 require fallbacks by default, it depends on the functionality provided by fallbacks. Therefore, if we were to undefine the `fallbacks` method when the `TEST_I18N_FALLBACKS` flag is off, we would get a bunch of other, unrelated, failures.

p-mongo:
Let's not test without fallbacks on AS 7.

p-mongo (approved):

p-mongo:
Evergreen is green!

PR #5431, opened 2022-08-02
MONGOID-5456 cast castable values for integer/float/big decimal

Neilshweky:


spec/mongoid/extensions/big_decimal_spec.rb
```diff
@@ -273,6 +273,17 @@
         end
       end
 
+      context "when the value is castable" do
+
+        let(:value) do
+          2.hours
+        end
+
```
p-mongo, line 281: Could you add a `before` here please asserting the type of `value` to be the `Duration`?

p-mongo (approved):

spec/mongoid/extensions/big_decimal_spec.rb
Neilshweky: done

p-mongo (approved):

PR #5432, opened 2022-08-03
MONGOID-4403 Support Dirty Tracking changed from/to

Neilshweky:


docs/release-notes/mongoid-8.1.txt
```diff
@@ -116,3 +116,30 @@ from the ``_translations`` hash:
 
 See the section on :ref:`Localize :present Field Option <present-fields>` for
 more details on how these are used.
+
+
+Added ``:to`` and ``:from`` options to ``attribute_changed?``
+-------------------------------------------------------------
+
+Mongoid 8.1 adds the ``:to`` and ``:from`` options on the ``attribute_changed?``
+method. These options can be used to assert that the attribute has been changed
```
p-mongo, line 125:
```suggestion
method. These options can be used to inquire whether the attribute has been changed
```

docs/release-notes/mongoid-8.1.txt
```diff
@@ -116,3 +116,30 @@ from the ``_translations`` hash:
...
+to or from certain values:
```
p-mongo, line 126:
```suggestion
to or from a certain value:
```

docs/release-notes/mongoid-8.1.txt
```diff
@@ -116,3 +116,30 @@ from the ``_translations`` hash:
...
+  person.name = "Ohtani"
+
+  person.name_changed?
+  # => true
+  person.name_changed?(from: "Trout")
+  # => true
+  person.name_changed?(to: "Ohtani")
+  # => true
+  person.name_changed?(from: "Trout", to: "Ohtani")
+  # => true
```
p-mongo, line 145: One example here where the return value is false would be nice.

lib/mongoid/changeable.rb
```diff
@@ -161,12 +161,24 @@ def attribute_change(attr)
     #   model.attribute_changed?("name")
     #
     # @param [ String ] attr The name of the attribute.
+    # @param **kwargs The optional keyword arguments.
+    #
+    # @option **kwargs [ Object ] :from The object the attribute was changed from.
+    # @option **kwargs [ Object ] :to The object the attribute was changed to.
```
p-mongo, line 167: Did we agree on `**kwargs` here in the documentation conventions PR?

p-mongo (approved):

lib/mongoid/changeable.rb
Neilshweky: I think we did. I checked back there when I decided on the naming.

docs/release-notes/mongoid-8.1.txt
Neilshweky: done

lib/mongoid/changeable.rb
p-mongo: We didn't do this before and I think omitting `**kwargs` wouldn't add any ambiguity, if we don't need that extra word I'd rather not have it.

p-mongo (approved):

lib/mongoid/changeable.rb
Neilshweky: It's been documented here: https://github.com/mongodb/mongoid/blob/49ff14f4a2187ef61abd81ecbc5a0fcd043f2401/docs/contributing/code-documentation.txt#L285

PR #5435, opened 2022-08-05
Drop **kwargs from @option


p-mongo:


johnnyshields:
The intention here was that a param `**kwargs` is actually defined on the method signature, e.g. `def my_method(**kwargs)`

Hence the syntax here would be congruent with other cases in use today:

```
    # @option opts [ Class ] :type The class of the field.
    # @option *args [ true | false ] :allow_destroy Can deletion occur?
    # @option **kwargs [ Object ] :from The object the attribute was changed from.
```

PR #5442, opened 2022-08-10
MONGOID-4698 update warning when overriding criteria methods in scope

Neilshweky:


Neilshweky:
The resulting warning: 
```
irb(main):003:0> Band.send(:check_scope_name, "text_search")
W, [2022-08-10T16:47:22.508027 #91436]  WARN -- : Creating scope :text_search which collides with Band.text_search. Calls to this scope will delegate to Band.text_search and will ignore the declared scope.
```

lib/mongoid/scopable.rb
```diff
@@ -249,12 +249,11 @@ def check_scope_name(name)
...
             raise Errors::ScopeOverwrite.new(self.name, name)
           else
-            if Mongoid.logger
-              Mongoid.logger.warn(
-                "Creating scope :#{name}. " +
-                "Overwriting existing method #{self.name}.#{name}."
-              )
-            end
+            Mongoid.logger.warn(
+              "Creating scope :#{name} which collides with #{self.name}.#{name}. " +
```
p-mongo, line 253:
```suggestion
              "Creating scope :#{name} which conflicts with #{self.name}.#{name}. " +
```

lib/mongoid/scopable.rb
```diff
@@ -249,12 +249,11 @@ def check_scope_name(name)
...
+              "Calls to this scope will delegate to #{self.name}.#{name} and " +
```
p-mongo, line 254:
```suggestion
              "Calls to `#{name}' will delegate to #{self.name}.#{name} and " +
```

p-mongo (approved):
The "calls to the scope" part is confusing to me because the call isn't actually made to the scope.

johnnyshields:
I see this warning when I declare a scope on a base class, then add a mixin which overrides it. For example:

```ruby
module MyMixin
  extend ActiveSupport::Concern

  included do
    scope :foo, -> { none }
  end
end

class Band
  include Mongoid::Document
  include MyMixin

  scope :foo, -> { where(name: "Pixies") }
  scope :text_search, -> { where(name: "The Cure") }
end

# Creating scope :foo. Overwriting existing method Band.foo.
Band.foo #=> #<Mongoid::Criteria selector: {"name"=>"Pixies"}
Band.text_search #=> #<Mongoid::Criteria selector: {"name"=>"The Cure"}
```

The error method you have in this PR seems to indicate that the old method won't be overwritten, when in fact it is...

Perhaps the logic should check for methods which are defined on Mongoid::Criteria and make a hard rule of not allowing to override them with `:scope`?

johnnyshields:
`Calls to ``#{name}' will delegate to #{self.name}.#{name} and will ignore the declared scope.`

Perhaps I'm not understanding something--won't the calls go to the (newly) declared scope? This is the behavior on 7.5:

```ruby
class Band
  include Mongoid::Document
  scope :text_search, -> { where(name: "The Cure") }
end
# WARN -- : Creating scope :text_search. Overwriting existing method Band.text_search.

Band.text_search #=> #<Mongoid::Criteria selector: {"name"=>"The Cure"}
```

Neilshweky:
@johnnyshields you're right that it will go to the newly declared scope when it's not being called on the criteria. However, if you call it on a criteria like: 
```
Band.all.text_search
```
 it will be delegated to the criteria method, not the scope.

PR #5447, opened 2022-08-16
MONGOID-4629 Non-block form of persistence context #with

Neilshweky:
The following warning comes up when calling storage_options on criteria: 
```
Mongoid::PersistenceContext#storage_options at /Users/neil.shweky/.rbenv/versions/2.7.5/lib/ruby/2.7.0/forwardable.rb:154 forwarding to private method Mongoid::Criteria#storage_options
```
I think it has to do with this line: 
```
def_delegators :@object, :storage_options
```
apparently you can't delegate to a private method anymore after MRI-2.4


Neilshweky:
The novelty of this PR is twofold: 
- Criteria can now accept persistence contexts
- Criteria, documents, and classes, can now use #with without a block:
  - Calling #with without a block on a criteria or a document returns that criteria or document
  - Calling #with without a block on a class returns a criteria with the new persistence context

Neilshweky:
Note: Docs/release notes will come in a separate PR

lib/mongoid/criteria/options.rb
```diff
@@ -6,18 +6,30 @@ class Criteria
...
 
-      private
-
+      # Returns the current persistence context or a new one constructed with
+      # self.
+      #
+      # @return [ Mongoid::PersistenceContext ] the persistence context.
       def persistence_context
-        klass.persistence_context
+        PersistenceContext.get(self) || klass.persistence_context
```
Neilshweky, line 14: this change allows persistence contexts to be stored _per criteria_.

spec/mongoid/contextual/mongo/persistence_context_spec.rb
```diff
@@ -0,0 +1,703 @@
...
+            end
+          end
+        end
+      end
+    end
+  end
+
+  # this might be a difficult method to implement because we'll have to keep around
+  # the original cluster and configuration.
+  xdescribe "#clear_persistence_context!" do
```
Neilshweky, line 662: Do we need a public method like this?

spec/mongoid/contextual/mongo/persistence_context_spec.rb
```diff
@@ -0,0 +1,703 @@
...
+
+          it "has the correct write preference" do
+            expect(criteria.destroy_all).to eq(0)
+          end
+        end
+      end
+
+      # TODO: cant figure out if this is relevant or not. havent been able to
+      # test effectively
+      xcontext "when calling #with on a document" do
```
Neilshweky, line 627: Any ideas on how to test this? this is for testing client options (like read preference or write concern) when working with a single document....

spec/mongoid/contextual/mongo/persistence_context_spec.rb
Neilshweky: done.

docs/release-notes/mongoid-8.1.txt
```diff
@@ -145,3 +145,40 @@ to or from a certain value:
...
+
+.. code-block:: ruby
+
+  # The #with method on Model classes returns a criteria.
+  criteria = Band.with(database: "music-non-stop")
+  # => #<Mongoid::Criteria
+         selector: {}
+         options:  {}
+         class:    Band
+         embedded: false>
```
johnnyshields, line 165: perhaps it makes sense to change the Criteria#inspect method output to include the persistence context (if set)?

lib/mongoid/clients/options.rb
```diff
@@ -21,10 +21,30 @@ module Options
...
+            clear_persistence_context(original_cluster, original_context)
+          end
+        else
+          # If #with is called on a Criteria we want to return a new criteria,
+          # but if it's called on a document, we want to return the same document
+          # since duping a document changes its _id.
+          if is_a?(Criteria)
+            crit = dup
+            crit.instance_variable_set("@original_cluster", original_cluster)
+            crit.instance_variable_set("@original_context", original_context)
```
johnnyshields, line 38: what happens if I chain multiple non-block calls to `#with` (`Band.with(...).with(...).all`)? Does `@original_cluster` etc get set to the second-to-last call to `with`, or does it remember the state at the beginning of the Criteria?

Neilshweky: Interesting, the first `with` would probably be ignored tbh. I think this would be a separate feature however.

spec/mongoid/contextual/mongo/persistence_context_spec.rb
johnnyshields: How about `document.with(read: { mode: :primary }).reload`, ensure reload does a query to primary?

docs/reference/persistence-configuration.txt
```diff
@@ -177,22 +188,52 @@ failing subsequent updates:
...
+  # documents will also return the same document.
+  band = Band.new
+  band.with(client: :tertiary)
+  band.save!
+  band.clear_persistance_context!
+
+  # The #with method on criteria returns a new criteria.
+  criteria = Band.all.with(collection: "artists")
+  band = criteria.first
+  criteria.clear_persistance_context!
```
p-mongo, line 227: I suggest putting this line under an `ensure` block for proper cleanup. As written exceptions during result processing will leak clients.

docs/reference/persistence-configuration.txt
```diff
@@ -177,22 +188,52 @@ failing subsequent updates:
...
+         embedded: false>
+
+  criteria.first
+  criteria.clear_persistance_context!
+
+  # The #with method on documents modifies the document object itself since
+  # calling #dup on it will change that document's _id. The #with method on
+  # documents will also return the same document.
+  band = Band.new
+  band.with(client: :tertiary)
```
p-mongo, line 220: This line mutates `band` permanently? I understand how this came to be (`band` would be mutated in the block previously) but this permanent mutation feels too iffy to me.

comandeo: This looks weird to me, too. But I think this is an inevitable side effect of the feature.

comandeo (approved):

docs/reference/persistence-configuration.txt
johnnyshields: This feature should not cause any permanent mutation of globals. That will be a disaster, not to mention not thread-safe, etc. It should return a new criteria object where the modification happens specific to that criteria.

Neilshweky: Note that calling clear_persistence_context does put it back to its original context still...

p-mongo: For the model instances which are mutated by `with`, my suggestion was to define `with!` and in the `with` implementation require a block. For model classes retain the current behavior of `with` only which returns a new scope.

docs/reference/persistence-configuration.txt
Neilshweky: how's this?

johnnyshields: Should be `persistence` with an "e" as the more common spelling.

docs/reference/persistence-configuration.txt
johnnyshields: I think `with!` is too dangerous to be practical in real-world applications. It would be better to declare any such options in model itself (declarative > imperitive)

Neilshweky: done

docs/reference/persistence-configuration.txt
```diff
@@ -177,22 +188,70 @@ failing subsequent updates:
...
+    criteria.first
+  ensure
+    criteria.clear_persistance_context!
+  end
+
+  begin
+    # Since a document instance cannot be duped (doing so changes its _id), in
+    # order to change its persistence context, the document itself has to be
+    # modified. Use the #with! method instead, which modifies and returns the
+    # document instance itself.
```
johnnyshields, line 231: This explanation doesn't make a whole lot of sense to a normal user. Why would I need to dupe a document instance in order to change it's persistence context? I can change many other properties of the document without duping it...

docs/release-notes/mongoid-8.1.txt
```diff
@@ -145,3 +145,45 @@ to or from a certain value:
...
+Mongoid 8.1 adds support for a non-block form for the ``#with`` method and for
+a new ``with!`` method. The ``#with`` method is used for changing the
+``persistence context`` of a Model class, a document, and, starting in Mongoid
+8.1, a criteria.
+
+Note that, ``with`` cannot be called on document instances without a block, since
+in order to change the persistence context of a document instance, the document
+itself must be modified. This is where the ``#with!`` method comes in: The
+``#with!`` method modifies its caller, and returns it with the given
+persistence context. This method can be used on document and criteria instances.
```
johnnyshields, line 162: My two cents: it doesn't make a lot of sense to have a block form of `.with` on a document instance. AFAIK, `band.with { ... }` is 100% equivalent to `band.class.with { ... }` (i.e. `Band.with { ... }`)

If my understanding above is correct, I would advocate to get rid of the instance block syntax, as it redundant/confusing, and instead allow only the non-block form on the instance `band.with()` -- which is what you currently have as the `.with!` method.

docs/reference/persistence-configuration.txt
```diff
@@ -177,22 +188,70 @@ failing subsequent updates:
...
+    criteria = Band.all.with(collection: "artists")
+    band = criteria.first
+  ensure
+    criteria.clear_persistance_context!
+  end
+
+As of Mongoid 8.1, the ``with`` method can be called with or without a block.
+If it is called with a block, the temporary persistence context will exist only
+for the duration of the block. If the ``with`` method is called without a block,
+the ``clear_persistance_context!`` method must be called on the criteria or
```
johnnyshields, line 250: Should be `clear_persistence_context` with "e"

docs/release-notes/mongoid-8.1.txt
Neilshweky: I would be okay with deprecating document instance with, what do you think @p-mongo ?

docs/reference/persistence-configuration.txt
Neilshweky:
```suggestion
the ``clear_persistence_context!`` method must be called on the criteria or
```

Neilshweky: good catch.

johnnyshields: Think you still have the same mistake in several more places. Please scan the codebase for it.

johnnyshields:
How does Document#with! handle relations? For example, support Band has_many Albums. What does this do?

```ruby
band = Band.first
band.with(read: { mode: :secondary }).albums
```

If there is complexity here we may want to consider splitting the changes to Criteria and to Document into two separate PRs...

johnnyshields:
I've been thinking about this more. For Document instances, it would really make more sense to have the persistence options be kwargs on the specific persistence actions.

For example:

```
band = Band.first
band.save!(with: { collection: "artists" })
band.set(name: "Steely Dan", with: { collection: "artists" })
band.reload(with: { read: { mode: :secondary } })
```

This would:
1. Allow users to achieve their core use-case here, which is to influence specific document operations.
2. Avoid introducing mutable state in a potentially confusing/knock-on effect way.
3. Avoid the `with / with!` dilemma.
4. Avoid ambiguity between class-level `Band.with(...) { ... }` and instance/operation-level overrides (operation-specific overrides would take precedence.)
5. Avoid confusion about whether the instances returned by a criteria should inherit the criteria's `with` directives (the answer would be "no")

You may want to consider splitting the changes to Criteria and to Document into two separate PRs, since Criteria seems to be a straight-shot but Document there may be more thinking to do...

p-mongo:
@johnnyshields Can you please add use cases for all new behavior that you are either proposing or reviewing. Upon review of MONGOID-4629 I realized that I misstated the original use case, which is saving the instances back to where they were loaded from, and this use case is not at all addressed by changing the persistence context on an instance. We are currently considering whether changing persistence context on an instance makes sense to do at all. So, if you have a use case for this, please add it here or to the ticket.

johnnyshields:
@p-mongo re-stated use cases here: https://jira.mongodb.org/browse/MONGOID-4629

Neilshweky:
> How does Document#with! handle relations? For example, support Band has_many Albums. What does this do?
> 
> ```ruby
> band = Band.first
> band.with(read: { mode: :secondary }).albums
> ```

Instance level with returns the document itself, so it should return the albums using secondary read. However, as of last commit, we have prohibited using with (in favor of with!) on instances

docs/release-notes/mongoid-8.1.txt
p-mongo: We decided that if instance-level #with didn't exst, we would not implement it now because it is in practice probably attainable via `band.class.with { ... }`, but since the functionality is already implemented, and it is not currently causing issues, let's leave it as is.

p-mongo:
@johnnyshields for your comment with the three examples:

```
band = Band.first
band.save!(with: { collection: "artists" })
band.set(name: "Steely Dan", with: { collection: "artists" })
```

I think in practice this is highly unlikely to do anything useful. The above operations are updates, if routed to a different collection the underlying document is unlikely to be present, thus the update would do nothing (other than indicate on the client side that the update happened, even though no data was written to the database, which could be considered potential data loss).

```
band.reload(with: { read: { mode: :secondary } })
```

This is a legitimate use case but it should already be attainable via the class-level with:

```
Band.with(read: {mode: :secondary}) { band.reload }
```

lib/mongoid/clients/options.rb
```diff
@@ -39,16 +89,28 @@ def mongo_client
...
 
+      # Was a persistence context set?
+      #
+      # @return [ true | false ] true if a persistence context was set,
+      #   false otherwise.
       def persistence_context?
         !!(PersistenceContext.get(self) || PersistenceContext.get(self.class))
       end
 
+      def clear_persistence_context!
```
Neilshweky, line 110: TODO: add comment here.

docs/reference/persistence-configuration.txt
```diff
@@ -177,22 +188,70 @@ failing subsequent updates:
...
+    criteria = Band.with(database: "music-non-stop")
+    # => #<Mongoid::Criteria
+          selector: {}
+          options:  {}
+          class:    Band
+          embedded: false>
+
+    criteria.first
+  ensure
+    criteria.clear_persistance_context!
```
Neilshweky, line 224: typo

Neilshweky:
Closing this for now, it should be revisited at a later time.

PR #5448, opened 2022-08-18
MONGOID-5408 .evolve should support wrapper class

johnnyshields:
This is now possible:

```ruby
# queries for the Integer 42, even though :name is type String.
User.where(name: Mongoid::RawValue(42))

# queries for the String "42", even though :age is type Integer.
User.where(age: Mongoid::RawValue(42))
```


johnnyshields:
@p-mongo this is ready for initial review.

p-mongo:

Can you explain why you are evolving raw values? The ticket says to skip evolution when querying, this is opposite behavior.

johnnyshields:
@p-mongo we still need to evolve Ruby classes into Mongo driver-compatible ones, e.g. BigDecimal into BSON::Decimal128. Otherwise, the driver won't be able to handle it AFAIK. The difference is the evolution ignores any field type definition, and instead evolves purely using the value's class.

johnnyshields:
Raised: MONGOID-5468

johnnyshields:
@p-mongo on this PR I'll remove the evolution logic for RawValues. Can be done as a separate ticket follow-up if needed pending discussion in [MONGOID-5468](https://jira.mongodb.org/browse/MONGOID-5468)

lib/mongoid/criteria/queryable/selector.rb
```diff
@@ -150,6 +150,8 @@ def evolve_multi(specs)
         #
         # @return [ Object ] The serialized object.
         def evolve(serializer, value)
+          return value.raw_value if value.is_a?(Mongoid::RawValue)
```
johnnyshields, line 153: This is the key line.

Note that under the current structure, we cannot define `Mongoid::RawValue#evolve`. The reason is that this method `Mongoid::Criteria::Queryable::Selector#evolve(serializer, value)` calls the class-specific `#evolve` methods based on the field-specific `serializer`, which defeats the goal of `RawValue` which is to *bypass the field-specific serialization*.

In the future, I think this line should be `raw_value.mongoize` as I have noted in MONGOID-5468.

johnnyshields:
@p-mongo this is ready for another review.

lib/mongoid/extensions/raw_value.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def initialize(raw_value)
+      @raw_value = raw_value
+    end
+
+    # Delegate all missing methods to the raw value.
+    #
+    # @param [ String, Symbol ] method_name The name of the method.
+    # @param [ Array ] args The arguments passed to the method.
+    ruby2_keywords def method_missing(method_name, *args, &block)
```
p-mongo, line 21: This implementation is not necessary. The topic of proxies is a complex one, for example proxies generally need to derive from BasicObject. Please remove `method_missing` and `respond_to_missing?` from here and if you have a use case which requires RawValue to be a proxy please file a ticket describing it.

What might be useful is a `to_s` or `inspect` override which would call the raw value's `to_s` or `inspect` and prefix the output with `RawValue:`.

spec/integration/criteria/raw_value_spec.rb
```diff
@@ -0,0 +1,397 @@
...
+  let!(:band2) { Band.create!(name: '2', likes: 1, rating: 1.0, sales: BigDecimal('100'), decibels: 30..90, founded: today, updated_at: now_utc + 1.days) }
+  let!(:band3) { Band.create!(name: '3', likes: 1, rating: 2.2, sales: BigDecimal('220'), decibels: 40..100, founded: today + 1.days, updated_at: now_utc + 2.days) }
+  let!(:band4) { Band.create!(name: '3', likes: 2, rating: 3.1, sales: BigDecimal('310'), decibels: 50..120, founded: today + 1.days, updated_at: now_utc + 3.days) }
+  let!(:band5) { Band.create!(name: '4', likes: 3, rating: 3.1, sales: BigDecimal('310'), decibels: 60..150, founded: today + 2.days, updated_at: now_utc + 3.days, labels: labels) }
+
+  context 'Mongoid::RawValue<String> criteria' do
+
+    context 'Integer field' do
+      it 'does not match objects' do
+        expect(Band.where(likes: Mongoid::RawValue('1')).to_a).to eq []
```
p-mongo, line 24: The tests should assert that RawValue works, i.e. that it returns query results. You can use `Band.collection.insert_many` to populate the database with values that bypass Mongoid field logic. Do you have positive assertions on this field in other tests? It is unclear which tests that would be if so.

spec/mongoid/criteria/queryable/selector_spec.rb
```diff
@@ -403,6 +403,32 @@ def localized?
...
+
+          context "and raw_value is a complex type" do
+            config_override :map_big_decimal_to_decimal128, true
+
+            before do
+              selector.send(method, "key", Mongoid::RawValue([{ foo: 1, "Bar" => [/baz/, BigDecimal('2'), 4..5] }, 3]))
+            end
+
+            it "returns the raw_value" do
+              expect(selector["key"]).to eq([{ foo: 1, "Bar" => [/baz/, BigDecimal('2'), 4..5] }, 3])
```
p-mongo, line 428: Can you also add a test which uses a type that isn't handled by bson-ruby and that is not transformed by Mongoid for RawValue logic, e.g. either a Range or `Object.new` for example.

lib/mongoid/criteria/queryable/selector.rb
p-mongo: This looks fine, I don't see any issues here right now.

lib/mongoid/extensions/raw_value.rb
johnnyshields: Done.

spec/mongoid/criteria/queryable/selector_spec.rb
johnnyshields: Done.

spec/integration/criteria/raw_value_spec.rb
johnnyshields: Done. I've used the `Band.collection.insert_one` approach you advocated.

johnnyshields:
@p-mongo this is ready for final review.

johnnyshields:
@p-mongo this is now unblocked and ready for final review.

johnnyshields:
@Neilshweky @comandeo this is ready for final review. Please remove the "blocked" tag by the way.

docs/reference/queries.txt
```diff
@@ -245,6 +232,33 @@ Since `id` and `_id` fields are aliases, either one can be used for queries:
   #   embedded: false>
 
 
+Embedded Documents
```
johnnyshields, line 235: I just copy-pasted the "Embedded Documents" section after "Fields" here w/out edits to it, as I felt it made more sense order-wise vis-a-vis additions to Fields section in this PR.

johnnyshields:
@alexbevi please unmark this PR as "Blocked". It is ready for review and feedback.

lib/mongoid/criteria/queryable/selector.rb
jamis, line 153: Hey @johnnyshields -- I hate to keep drawing this PR out, but I wanted to ask if there was a specific reason for using a guard here, rather than adding `when Mongoid::RawValue` as a clause in the `case` statement below.

lib/mongoid/extensions/raw_value.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Wrapper class used when a value cannot be casted in evolve method.
+module Mongoid
+  def RawValue(*args)
```
jamis, line 5: I know it's pretty self-evident, but a comment to document this method would be clearer, I think.

jamis (changes requested):
This looks good, @johnnyshields! Just a couple of minor comments on things I wanted to clarify, but once those are answered I'm happy to approve this.

lib/mongoid/extensions/raw_value.rb
johnnyshields: Done!

lib/mongoid/criteria/queryable/selector.rb
johnnyshields: Changed it. Good catch.

johnnyshields:
@jamis thanks for your review! Comments addressed.

jamis (approved):

comandeo (approved):

lib/mongoid/extensions/raw_value.rb
```diff
@@ -0,0 +1,32 @@
...
+# Wrapper class used when a value cannot be casted in evolve method.
+module Mongoid
+
+  # Instantiates a new Mongoid::RawValue object. Used as a syntax shortcut.
+  #
+  # @example Create a Mongoid::RawValue object.
+  #   Mongoid::RawValue("Beagle")
+  #
+  # @return [ Mongoid::RawValue ] The object.
+  def RawValue(*args)
```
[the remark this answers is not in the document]
johnnyshields: I agree on this point, I have some more PRs coming for this area so will do it in a follow-up. Thanks!

PR #5449, opened 2022-08-18
MONGOID-5226 Allow setting "store_in collection" on document subclass

Neilshweky:


lib/mongoid/clients/storage_options.rb
```diff
@@ -7,9 +7,10 @@ module StorageOptions
 
       included do
 
-        cattr_accessor :storage_options, instance_writer: false do
-          storage_options_defaults
-        end
+        # NOTE: use class_attribute here so that when the child modifies its
+        # storage options, the parent is not also modified. Using cattr_accessor
+        # causes both to be modified
+        class_attribute :storage_options, instance_writer: false, default: storage_options_defaults
```
Neilshweky, line 13: use class_attribute here so that when the child modifies its storage options, the parent is not also modified. Using cattr_accessor causes both to be modified.

lib/mongoid/clients/storage_options.rb
```diff
@@ -49,7 +50,7 @@ module ClassMethods
         # @return [ Class ] The model class.
         def store_in(options)
           Validators::Storage.validate(self, options)
-          storage_options.merge!(options)
+          self.storage_options = self.storage_options.merge(options)
```
Neilshweky, line 53: We need to assign here so when the child is modified the parent is also not modified. If you merge!, the parent's storage_options will be modified

docs/reference/inheritance.txt
```diff
@@ -253,3 +253,50 @@ either normal setting or through the build and create methods on the association
...
    rect = Rectangle.new(width: 100, height: 200)
    firefox.shapes
+
+
+.. _inheritance-persistence-context:
+
+Persistence Contexts
+====================
+
+Mongoid allows the persistence context of subclasses to be changed from the
```
p-mongo, line 263:
```suggestion
Mongoid allows the persistence context of a subclass to be changed from the
```

docs/reference/inheritance.txt
```diff
@@ -253,3 +253,50 @@ either normal setting or through the build and create methods on the association
...
+persistence context of its parent. This means that, using the ``store_in``
+method, we can store the documents of the subclasses in different collections
+(as well as different databases, clients) than their parent:
```
p-mongo, line 266:
```suggestion
(as well as different databases, clients) than their parents:
```

lib/mongoid/clients/storage_options.rb
p-mongo: :+1:

lib/mongoid/errors/invalid_storage_parent.rb
```diff
@@ -1,25 +0,0 @@
-# frozen_string_literal: true
-
-module Mongoid
-  module Errors
-
-    # Raised when calling store_in in a sub-class of Mongoid::Document
-    class InvalidStorageParent < MongoidError
```
p-mongo, line 7: The exception class needs to be retained if this change is going into 8.1. It should be marked deprecated and won't be used by Mongoid.

p-mongo (approved):
I would like the documentation to also state explicitly that if the collection is overridden on a child, the child documents will no longer be found when the parent class is queried via `where` etc. (i.e. the inheritance effectively will no longer work the way it does by default).

comandeo (approved):

PR #5450, opened 2022-08-18
MONGOID-4698 Correct warning to point out that klass.band doesnt conflict

Neilshweky:


johnnyshields:
Hmm.... users are going to assume that since `Band.my_scope` typically delegates to a criteria that this warning is going to apply there as well. You could add `"... However, calls to #{klass}.scope will use the overridden method."`

I really think we should raise a ticket to improve the fundamental behavior here.

p-mongo (approved):

p-mongo:
@johnnyshields  https://jira.mongodb.org/browse/MONGOID-4698?focusedCommentId=4775096&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4775096

PR #5451, opened 2022-08-19
MONGOID-5150 Use primary read preference for validates_uniqueness_of 

Neilshweky:
https://jira.mongodb.org/browse/MONGOID-5150

### DIFF
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1487,6 +1487,15 @@ Mongoid also has some helpful methods on criteria.
           Band.all.pluck(:name, :likes)
             #=> [ ["Daft Punk", 342], ["Aphex Twin", 98], ["Ween", 227] ]
 
+   * - ``Criteria#read``
+
+       *Sets the read preference for the criteria.*
+
+     -
+        .. code-block:: ruby
+
+          Band.all.read(mode: :primary)
+
    * - ``Criteria#second``
 
        *Get the second document for the given criteria.*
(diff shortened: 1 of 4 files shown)


p-mongo:
I cannot think of any objections to this particular change but shouldn't all validations then use the primary read preference?

Neilshweky:
> I cannot think of any objections to this particular change but shouldn't all validations then use the primary read preference?
That would make sense honestly

Neilshweky:
@p-mongo what other validations do we have that touch the database? Here are the contents of the `lib/mongoid/validatable/` directory:
- associated.rb - checks that the associated document is valid
- format.rb - checks the field value matches a regex
- length.rb - checks the length of a specific field value
- localizable.rb - Adds localization support to validations
- macros.rb - defines macros
- presence.rb - checks that the value exists for a field
- queryable.rb - Wrap the validation inside the an execution block that alert's the client not to clear its persistence options.
- uniqueness.rb - Hits the db

Neilshweky:
> I was thinking of the custom validator methods (`validate :foo`) which would result in database calls. But if the only affected Mongoid validator is uniqueness maybe the current solution is sufficient.

If users make their own validators, should we be enforcing primary reads? The default read preference is primary, so if a user elects to use secondary reads, they are most likely doing it purposely. I vote that this solution is solution.

p-mongo (approved):

p-mongo:
I copied the notes here in the ticket and approve this PR as it currently is.

johnnyshields:
AFAIK the only validator which reads from the DB is `validates_uniqueness`. I would not want Mongoid to force custom validators (i.e. user-defined) to read from primary; the user can do that explicitly within the validator.

comandeo (approved):

PR #5452, opened 2022-08-22
MONGOID-5370 Add collection options support

comandeo:
This PR solves https://jira.mongodb.org/browse/MONGOID-5370 by providing a mechanism to specify any collection options for a Mongoid document. It means that a user still have to use the driver's documentation to find out how to set up a time series collection or a capped collection. It also means that Mongoid won't validate options that user provided, and will rely on the driver's errors.

On the other hand, with this PR it is not necessary to use the driver directly, all the configuration can be done using `store_in` macro and a rake task. So, I believe this is still an improvement.


johnnyshields:
Should look here if there are more options that can be supported: https://www.mongodb.com/docs/v6.0/reference/method/db.createCollection/#mongodb-method-db.createCollection

p-mongo (commented):
I typed this up yesterday and I think I will need to do another pass but I'll submit this for now.

docs/reference/collection-configuration.txt
```diff
@@ -10,51 +10,88 @@ Collection Configuration
    :depth: 2
    :class: singlecol
 
+Configuring a Document Collection
+=================================
 
-Capped Collections
-------------------
+You can specify collection options for documents using ``store_in`` macro.
```
p-mongo, line 16:
```suggestion
You can specify collection options for documents using the ``store_in`` macro.
```

docs/reference/collection-configuration.txt
```diff
@@ -10,51 +10,88 @@ Collection Configuration
...
+This macro accepts ``:collection_options`` argument, which can contain any collection
+options that is supported by the driver.
```
p-mongo, line 18:
```suggestion
options that are supported by the driver.
```

docs/reference/collection-configuration.txt
```diff
@@ -10,51 +10,88 @@ Collection Configuration
...
+
+.. note::
+
+  In order to apply the options, the collection must be explicitly created up-front.
+  This should be done using  :ref:`Collection Management Rake Task<collection-management-task>`.
+
+Please refer to `the driver collections page
+<https://mongodb.com/docs/ruby-driver/current/reference/collection-tasks/>`_
+for the more information about collection options.
```
p-mongo, line 27: For indexes there is a documented way to create them by directly executing the respective method without going through a rake task: https://www.mongodb.com/docs/mongoid/master/reference/indexes/#index-management-rake-tasks

I use this all the time and I think this pattern should be continued for the collection creation as well.

docs/reference/collection-configuration.txt
```diff
@@ -10,51 +10,88 @@ Collection Configuration
...
+
+.. note::
 
-Mongoid does not provide a mechanism for creating capped collections on the fly - you
-will need to create these yourself one time up front either with the driver or via the
-Mongo console.
+  Collection options depend on the driver version and MongoDB server version.
+  It is possible that some options, like time series collectios, are not available
+  on older server versions. Mongoid does not validate collection options.
```
p-mongo, line 33: The "Mongoid does not validate" bit could potentially be interpreted as that users are on their own for validation, but this isn't actually the case - either the driver or the server will validate the options. I suggest either adding something like "but you may receive an error from the driver or from the server for using options that are not supported",  replacing the last sentence ("Mongoid does not validate") with this bit or removing the "Mongoid does not validate" bit without replacement.

docs/reference/collection-configuration.txt
```diff
@@ -10,51 +10,88 @@ Collection Configuration
...
+
+.. _collection-management-task:
 
-  client["name", :collation => { :locale => 'fr'}].create
+Collection Management Rake Task
+===============================
 
-To create a collection with a default collation from the Mongo console:
+If you specify collection options for a document, then the corresponding collection
+must be explicitly created. To do so, use the provided
```
p-mongo, line 92:
```suggestion
must be explicitly created prior to use. To do so, use the provided
```

lib/config/locales/en.yml
```diff
@@ -31,6 +31,17 @@ en:
             since the document did not actually get saved."
           resolution: "Double check all before callbacks to make sure they are
             not unintentionally returning false."
+        create_collection:
+          message: "Cannot create collection %{collection_name}
+            with options %{collection_options}. The following error was raised:
+            %{server_error}."
```
p-mongo, line 37: This should contain both the exception class and the exception message following the `"#{e.class}: #{e}"` pattern  we normally use.

lib/mongoid/collection_configurable.rb
```diff
@@ -0,0 +1,30 @@
...
+        if coll_options = collection.database.list_collections(filter: { name: collection_name.to_s }).first
+          logger.info(
+            "MONGOID: Collection '#{collection_name}' already exist " +
+            "in database '#{database_name}' with options '#{coll_options}'."
+          )
+        else
+          begin
+            collection.database[collection_name].create(storage_options[:collection_options])
+          rescue Mongo::Error::OperationFailure => e
+            raise Errors::CreateCollection.new(
```
p-mongo, line 20: Suggest `Errors::InvalidCreateCollection` or `Errors::CreateCollectionFailed` or similar, `CreateCollection` alone isn't an error.

lib/mongoid/collection_configurable.rb
```diff
@@ -0,0 +1,30 @@
...
+
+module Mongoid
+
+  # Encapsulates behavior around defining collections.
+  module CollectionConfigurable
+    extend ActiveSupport::Concern
+
+    module ClassMethods
+      def create_collection
+        if coll_options = collection.database.list_collections(filter: { name: collection_name.to_s }).first
```
p-mongo, line 11: The fact that this method (and thus rake task) doesn't create collections that don't have collection options specified needs to be documented. I think it's better to create all collections always? @Neilshweky

lib/mongoid/collection_configurable.rb
```diff
@@ -0,0 +1,30 @@
...
+          logger.info(
+            "MONGOID: Collection '#{collection_name}' already exist " +
+            "in database '#{database_name}' with options '#{coll_options}'."
+          )
```
p-mongo, line 15: I also think I would prefer this method to 1) retrieve collection options and 2) raise if they are different from current specification instead of succeeding in that case but leaving db in a state different from what one might expect. Thoughts?

docs/reference/collection-configuration.txt
comandeo: Done

lib/mongoid/tasks/database.rb
```diff
@@ -5,6 +5,18 @@ module Tasks
...
       extend self
 
+      def create_collections(models = ::Mongoid.models)
+        models.each do |model|
+          next if model.storage_options.fetch(:collection_options, {}).empty?
+          if !model.embedded? || model.cyclic?
+            model.create_collection
+            logger.info("MONGOID: Created collection for #{model}:")
+          else
+            logger.info("MONGOID: collection options ignored on: #{model}, please define in the root model.")
```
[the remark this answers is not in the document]
comandeo: `create_indexes` method below does the same. I prefer consistency, it would be strange if two similar pieces of code have different structure. If we want to move the warning to `store_in`, the same should be done for `create_indexes`, and probably other methods here.

lib/mongoid/collection_configurable.rb
comandeo: I like the idea of creating all collections regardless of `store_in` macro, it seems to be more intuitive.

lib/mongoid/collection_configurable.rb
comandeo: This would be ideal; however, implementing this adds some knowledge about the driver and collection options to Mongoid. When we create time series collection, we pass options using `time_series` key. When we get collection information using `listCollections`, the key is `timeseries`. Same with `expire_after` and `expireAfterSeconds`, `clustered_index` and `clusteredIndex`, and so on - https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/collection.rb#L68

Further, if you do not specify, for example, `granularity` for a time series collection, it will be defaulted to seconds. However, what if a collection was created with with granularity = hours, but the corresponding Mongoid model omits the granularity option? Should we also validate defaults (=> adding even more knowledge about driver/server implementation details to the Mongoid)? Without such validation we cannot guarantee the state of db.

The above can be done in Mongoid, but we need to agree on it. I believe we usually avoid adding such low level details.

johnnyshields: I would propose that we should only check for conflicts in cases where values have been explicitly set by the user in Mongoid on the `store_in` macro, i.e. not validate vs Mongoid's defaults.

lib/mongoid/errors/create_collection.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      # Instantiate the create collection error.
+      #
+      # @param [ String ] collection_name The name of the collection that
+      #   Mongoid failed to create.
+      # @param [ Hash ] collection_options The options that were used when
+      #   tried to create the collection.
+      # @param [ String ] server_error The error raised by the server when
+      #   tried to create the collection.
+      def initialize(collection_name, collection_options, server_error)
```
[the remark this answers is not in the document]
comandeo: Done

docs/reference/collection-configuration.txt
comandeo: Done

docs/reference/collection-configuration.txt
comandeo: Done

lib/config/locales/en.yml
comandeo: I implemented this in the exception constructor, I believe this is what you suggested.

lib/mongoid/collection_configurable.rb
comandeo: @johnnyshields In the proposed implementation if we add validations to the ` db:mongoid:create_collections` task, it can be run as often as possible. It does not modify existing collections; so, it would be possible to execute it as a deployment step.

p-mongo: I see, retrieving collection options and munching them is not something I think belongs in this PR.

I tested and collection cannot be created twice with identical options:
```
irb(main):006:0> a['foo1',capped:true,size:2].create
=> #<Mongo::Operation::Result:0x2660 documents=[{"ok"=>1.0}]>
irb(main):007:0> a['foo1',capped:true,size:2].create
/home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/result.rb:364:in `raise_operation_failure': [48:NamespaceExists]: Collection already exists. NS: admin.foo1 (on localhost:14400) (Mongo::Error::OperationFailure)
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/result.rb:330:in `validate!'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/response_handling.rb:36:in `block (3 levels) in validate_result'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/response_handling.rb:108:in `add_server_diagnostics'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/response_handling.rb:35:in `block (2 levels) in validate_result'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/response_handling.rb:54:in `add_error_labels'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/response_handling.rb:34:in `block in validate_result'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/response_handling.rb:94:in `unpin_maybe'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/response_handling.rb:33:in `validate_result'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/executable.rb:81:in `block in execute'
	from <internal:kernel>:90:in `tap'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/executable.rb:80:in `execute'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/polymorphic_operation.rb:50:in `execute_with_connection'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/polymorphic_operation.rb:36:in `block in execute'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/server/connection_pool.rb:667:in `with_connection'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/server.rb:448:in `with_connection'
	from /home/w/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/mongo-2.18.1/lib/mongo/operation/shared/polymorphic_operation.rb:35:in `execute'
	... 9 levels...
irb(main):008:0> 
```

How about:

`Model.create_collection` creates the collection if it does not exist
`Model.create_collection!` attempts to create the collection every time, will fail if collection exists

+ have two rake tasks, maybe `mongoid:create_collections` and `mongoid:create_collections:force` or something like that.

@Neilshweky

lib/mongoid/tasks/database.rb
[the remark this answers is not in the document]
p-mongo: Yep that is fine.

lib/mongoid/collection_configurable.rb
johnnyshields: Should bear in mind users are lazy creatures, and they are not going to want to run a Rake task to create collections. Today everyone is used having it happen automatically when the first model instance is persisted at runtime. The Rake task is useful as a diagnostic tool perhaps, but should not be the golden path.

Perhaps the first time a given collection is accessed in a process, Mongoid could soft-validate that the collection's options match what is actually in the DB, and output a warning log line saying "either revert the options or make a new collection" (AFAIK collections can't be reconfigured.)

spec/mongoid/collection_configurable_spec.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+class CollectionConfigurableCapped
```
Neilshweky, line 5: I think these should be defined in a `before(:all)` block, and call `remove_const` on them in an `after(:all)`

spec/mongoid/collection_configurable_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        ).not_to be_empty
+      end
+    end
+
+    context 'with supported options' do
+      let(:subject) do
+        CollectionConfigurableCapped
+      end
+
+      it 'creates the collection' do
```
Neilshweky, line 52: maybe add an `it` that calls `collstats` and makes sure the options were actually set?

spec/mongoid/collection_configurable_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+
+      before do
+        expect_any_instance_of(Mongo::Collection)
+          .to receive(:create)
+          .with({ capped: true, size: 10000 })
+          .and_raise(Mongo::Error::OperationFailure)
+      end
+
+      it 'raises an error' do
```
Neilshweky, line 73: why does this raise an error?

spec/mongoid/tasks/models.rb
```diff
@@ -0,0 +1,13 @@
+module DatabaseSpec
```
Neilshweky, line 1: maybe name this file `database_spec_models.rb` for consistency with the rest of the codebase?

spec/mongoid/tasks/models.rb
```diff
@@ -0,0 +1,13 @@
...
+  class Measurement
+    include Mongoid::Document
+
+    field :timestamp, type: Time
+    field :temperature, type: Integer
+
+    store_in collection: "measurement",
+      collection_options: {
+        capped: true, size: 10000
+      }
```
Neilshweky, line 11: Is there a test for time series options?

spec/mongoid/tasks/database_spec.rb
```diff
@@ -18,6 +19,30 @@
...
+      let(:models) do
+        [Person]
+      end
+
+      it 'creates the collection' do
+        expect(Person).to receive(:create_collection).once
+        Mongoid::Tasks::Database.create_collections(models)
+      end
+    end
+  end
```
Neilshweky, line 44: Add tests when called on embedded (cyclic and non cyclic) collections.

lib/mongoid/collection_configurable.rb
p-mongo: `mongoid:create_collections:force` in the Rails sense would be useful for development. The `mongoid:create_collections:validate` I think is impractical - it would always fail in production and in development on a non-empty database, for the empty database in development `mongoid:create_collections:force` would work just as well. The `mongoid:create_collections` with warning behavior I think could be useful in production and would be the equivalent to Rails' migrations, alternatively it could perhaps be renamed to `mongoid:create_collections:missing` and not warn about the existing ones.

For indexes we have precedent of checking indexes in db vs indexes declared in models, this is also non-trivial and from what I remember doesn't quite work 100% of the time.

lib/mongoid/collection_configurable.rb
comandeo: I updated the code after I got @p-mongo comment, now it always creates collections.

docs/reference/collection-configuration.txt
```diff
@@ -10,51 +10,96 @@ Collection Configuration
...
+    $ rake db:mongoid:create_collections
+
+The create collections command also works for just one model by running
+in Rails console:
+
+.. code-block:: ruby
 
-  db.createCollection("name", { collation: { locale: 'fr' } });
+    # Create collection for Model
+    Model.create_collections
```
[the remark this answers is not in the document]
comandeo: Method name is singular, good catch.

`Model.collection.create` does not consider `collection_options` set by `store_in` macro. The intention is to promote using `store_in` macro for collection options, instead of using driver.

lib/mongoid/collection_configurable.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Mongoid
+
+  # Encapsulates behavior around defining collections.
+  module CollectionConfigurable
+    extend ActiveSupport::Concern
+
+    module ClassMethods
+      # Create collection for the Mongoid document this module is included into.
```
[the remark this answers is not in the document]
comandeo: Updated according to the suggestion.

spec/mongoid/collection_configurable_spec.rb
comandeo: Not sure if we should have the driver functionality here, but some additional coverage won't hurt. Added the requested.

spec/mongoid/collection_configurable_spec.rb
comandeo: Because of the we the `expect_any_instance_of` above. It the driver raises an error, Mongoid should, too.
However, I can see why this test might seem unclear, I reworked it.

spec/mongoid/tasks/models.rb
comandeo: Good point, thanks. Updated.

spec/mongoid/tasks/models.rb
comandeo: No, I do not think we should test all the possible options in Mongoid. I use capped collection options because they are accepted by most server versions.

lib/mongoid/tasks/database.rb
```diff
@@ -11,10 +11,13 @@ module Database
       # @param [ Array<Mongoid::Document> ] models. Array of document classes for
       #   which collections should be created. Defaulted to all document classes
       #   in the application.
-      def create_collections(models = ::Mongoid.models)
+      # @param [ true | false ] force If true, the method will drop existing
+      #   collections before creating new ones. If false, the method will create
+      #   only new collection (that do not exist in the database).
+      def create_collections(models = ::Mongoid.models, force: false)
```
johnnyshields, line 17: I would rename this variable to `:drop_existing_collections`. The name force is not 100% that it will drop your collections, and given the potential severity of a mistake here, we should be 100% clear.

lib/mongoid/tasks/database.rake
```diff
@@ -40,5 +39,12 @@ namespace :db do
     task :purge => :environment do
       ::Mongoid.purge!
     end
+
+    namespace :create_collections do
+      desc "Drop and create collections for Mongoid models"
+      task :force => [:environment, :load_models] do
```
johnnyshields, line 45: Same here, `task :drop_existing`

lib/mongoid/tasks/database.rb
comandeo: A similar method in ActiveRecord they use[force](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-create_table), hence `force` here. I would leave it as it is now.

However, I agree that `drop_existing_collections` might be more explicit. @p-mongo @Neilshweky what do you think?

p-mongo: The behavior is clearly described and matches Rails behavior. I don't see the need to change it.

p-mongo (approved):

lib/mongoid/tasks/database.rb
johnnyshields: If Rails is using `force` then OK.

image

p-mongo: The point of the cartoon is exactly opposite of what you are requesting @johnnyshields .

johnnyshields: I just think about this cartoon every time I hear dropping tables.

PR #5455, opened 2022-09-07
MONGOID-5474 add readonly! and raise on save/update

Neilshweky:
Flip feature flag on 8.1: #5456

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -127,6 +127,16 @@ module Mongoid
     # always return a Hash.
     option :legacy_attributes, default: false
 
+    # When this flag is true, a document is only readonly if it has been
+    # projected using #only or #without, and readonly documents will only not
+    # be able to be deleted/destroyed. When this flag is false, a document will
+    # become readonly only once the #readonly! method is called, and an error
+    # will be raised on attempting to save or update such documents, instead
+    # of just on delete.
+    # When this feature flag is turned on, the readonly state will be reset on
+    # reload, but when it is turned off, it won't be.
+    option :legacy_readonly, default: true
+
     # Returns the Config singleton, for use in the configure DSL.
     #
     # @return [ self ] The Config singleton.
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -135,7 +135,7 @@ module Mongoid
     # of just on delete.
     # When this feature flag is turned on, the readonly state will be reset on
     # reload, but when it is turned off, it won't be.
-    option :legacy_readonly, default: true
+    option :legacy_readonly, default: false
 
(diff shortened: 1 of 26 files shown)


spec/mongoid/config_spec.rb
```diff
@@ -349,6 +349,13 @@
     it_behaves_like "a config option"
   end
 
+  context 'when setting the legacy_attributes option in the config' do
+    let(:option) { :legacy_readonly }
+    let(:default) { true }
```
Neilshweky, line 354:
```suggestion
    let(:default) { false }
```

lib/mongoid/config.rb
```diff
@@ -127,6 +127,16 @@ module Config
     # always return a Hash.
     option :legacy_attributes, default: false
 
+    # When this flag is true, a document is only readonly if it has been
+    # projected using #only or #without, and readonly documents will only not
```
p-mongo, line 131:
```suggestion
    # projected using #only or #without, and readonly documents will not
```

lib/mongoid/config.rb
```diff
@@ -127,6 +127,16 @@ module Config
...
+    # be able to be deleted/destroyed. When this flag is false, a document will
```
p-mongo, line 132:
```suggestion
    # be deletable nor destroyable. When this flag is false, a document will
```

lib/mongoid/stateful.rb
```diff
@@ -95,14 +95,37 @@ def pushable?
...
+    # Flags the document as readonly. Will cause a ReadonlyDocument error to be
+    # raised if the document is attempted to be saved, updated or destroyed.
+    #
+    # @example Flag the document as readonly.
+    #   document.readonly!
+    #
+    # @return [ true | false ] true if the document was successfully marked
+    #   readonly, false otherwise.
+    #
+    # @since 4.0.0
```
p-mongo, line 107: I don't think this is right?

p-mongo (approved):

lib/mongoid/stateful.rb
Neilshweky: Oops lol

lib/mongoid/stateful.rb
```diff
@@ -95,14 +95,37 @@ def pushable?
...
 
+    # Flags the document as readonly. Will cause a ReadonlyDocument error to be
+    # raised if the document is attempted to be saved, updated or destroyed.
+    #
+    # @example Flag the document as readonly.
+    #   document.readonly!
+    #
+    # @return [ true | false ] true if the document was successfully marked
+    #   readonly, false otherwise.
+    #
```
Neilshweky, line 106:
```suggestion
```

PR #5457, opened 2022-09-08
MONGOID-5474/MONGOID-5478 add docs and release notes

Neilshweky:


p-mongo (approved):
The convention we have for release notes is to show new behavior first. I would also apply it to normal documentation, in this case explaining the behavior with legacy_readonly false and then with it true.

p-mongo (approved):

Neilshweky:
@p-mongo  I switched the order for the flag docs (so new functionality is first), and added the use case outlined in MONGOID-5477

p-mongo (approved):

PR #5458, opened 2022-09-08
MONGOID-4960 Create an exception subclass for calling estimated_count on models with default scopes

Neilshweky:


lib/config/locales/en.yml
```diff
@@ -181,8 +181,12 @@ en:
...
           summary: "Estimated count is strictly a collection-level operation and cannot be called
             on a filtered criteria."
-          resolution: "Try calling estimated_count directly on the class: %{class_name}.estimated_count.\n\n
-            \_If the class defines a default scope, use unscoped: %{class_name}.unscoped.estimated_count."
+          resolution: "Try calling estimated_count directly on the class: %{class_name}.estimated_count."
+        invalid_estimated_count_scoping:
+          message: "Invalid criteria for estimated_count."
+          summary: "Estimated count is strictly a collection-level operation and cannot be called
+            on a model that uses a default scope."
+          resolution: "Try calling estimated_count using unscoped: %{class_name}.unscoped.estimated_count."
```
p-mongo, line 189: I would also suggest alternatively using plain `count` which can be scoped.

p-mongo (approved):

lib/config/locales/en.yml
Neilshweky: done

p-mongo (approved):

PR #5460, opened 2022-09-09
MONGOID-5476 update 7.4 installation version

Neilshweky:


p-mongo (approved):
Reason why it's 7.4.0 is this way we don't need to update the version number each time we release a patch release. 7.4.3 would be more correct but adds some work at release tmie.

PR #5466, opened 2022-09-13
MONGOID-5100 allow selector arguments for .exists?

Neilshweky:
`#exists?` now takes a number of argument types:
- `Hash`: A hash of conditions.
- `Object`: An _id to search for.
- ``false``/``nil``: Always returns false.

I've decided to follow Rails' lead on `nil`/`false`... automatically returning `false` on `nil` makes sense, since in actuality what you're doing is checking for a `nil` or `false` _id. If that returned true, it would make no sense.

### DIFF
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1248,6 +1248,8 @@ Mongoid also has some helpful methods on criteria.
 
           Band.exists?
           Band.where(name: "Photek").exists?
+          Band.exists?(name: "Photek")
+          Band.exists?(BSON::ObjectId('6320d96a3282a48cfce9e72c'))
 
    * - ``Criteria#fifth``
 
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1243,6 +1243,13 @@ Mongoid also has some helpful methods on criteria.
        *Determine if any matching documents exist. Will return true if there
        are 1 or more.*
 
+       ``#exists?`` *now takes a number of argument types:*
+
+       - ``Hash``: *A hash of conditions.*
+       - ``BSON::ObjectId``: *An _id to search for.*
+       - ``String``: *An _id corresponding to the given string to search for.*
+       - ``false``: *Always returns false.*
+
      -
         .. code-block:: ruby
 
@@ -1250,6 +1257,8 @@ Mongoid also has some helpful methods on criteria.
           Band.where(name: "Photek").exists?
           Band.exists?(name: "Photek")
           Band.exists?(BSON::ObjectId('6320d96a3282a48cfce9e72c'))
+          Band.exists?('6320d96a3282a48cfce9e72c')
+          Band.exists?(false)
 
    * - ``Criteria#fifth``
 
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1246,8 +1246,7 @@ Mongoid also has some helpful methods on criteria.
        ``#exists?`` *now takes a number of argument types:*
 
        - ``Hash``: *A hash of conditions.*
-       - ``BSON::ObjectId``: *An _id to search for.*
-       - ``String``: *An _id corresponding to the given string to search for.*
+       - ``Object``: *An _id to search for.*
        - ``false``: *Always returns false.*
 
      -
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1247,7 +1247,6 @@ Mongoid also has some helpful methods on criteria.
 
        - ``Hash``: *A hash of conditions.*
        - ``Object``: *An _id to search for.*
-       - ``false``: *Always returns false.*
 
      -
         .. code-block:: ruby
@@ -1257,7 +1256,6 @@ Mongoid also has some helpful methods on criteria.
           Band.exists?(name: "Photek")
           Band.exists?(BSON::ObjectId('6320d96a3282a48cfce9e72c'))
           Band.exists?('6320d96a3282a48cfce9e72c')
-          Band.exists?(false)
 
    * - ``Criteria#fifth``
 
--- a/docs/reference/queries.txt
+++ b/docs/reference/queries.txt
@@ -1247,6 +1247,7 @@ Mongoid also has some helpful methods on criteria.
 
        - ``Hash``: *A hash of conditions.*
        - ``Object``: *An _id to search for.*
+       - ``false``/``nil``: *Always returns false.*
 
      -
         .. code-block:: ruby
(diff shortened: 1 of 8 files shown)


lib/mongoid/contextual/mongo.rb
```diff
@@ -161,9 +161,16 @@ def each(&block)
...
+      #   search for, or a hash of conditions.
+      #
       # @return [ true | false ] If the count is more than zero.
-      def exists?
-        !!(view.projection(_id: 1).limit(1).first)
+      def exists?(id_or_conditions = nil)
+        case id_or_conditions
+        when nil then !!(view.projection(_id: 1).limit(1).first)
+        when Hash then Mongo.new(criteria.where(id_or_conditions)).exists?
+        else Mongo.new(criteria.where(_id: id_or_conditions)).exists?
```
Neilshweky, line 172: I don't love explicitly creating a new Contextual object, but this is the only way to verify that a `Contextual::Mongo` object is used for the later `exists?` and not a `Memory`... 

I believe there can be a situation where calling exists? on a criteria in this method or the Memory method will result in the opposite contextual object being created and used. Explicitly calling the constructors here, instead of allowing create_context to do it, prevents that.

comandeo (approved):
[ActiveRecord `exists?`](https://api.rubyonrails.org/v7.0.3.1/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F) also accepts `false` as an argument, and always returns `false` in this case. Should we implement this as well?

Neilshweky:
@comandeo  good call, I just added that

docs/reference/queries.txt
```diff
@@ -1243,11 +1243,22 @@ Mongoid also has some helpful methods on criteria.
        *Determine if any matching documents exist. Will return true if there
        are 1 or more.*
 
+       ``#exists?`` *now takes a number of argument types:*
+
+       - ``Hash``: *A hash of conditions.*
+       - ``BSON::ObjectId``: *An _id to search for.*
+       - ``String``: *An _id corresponding to the given string to search for.*
+       - ``false``: *Always returns false.*
```
p-mongo, line 1251: Could you explain where the last two usages come from? The `false` one I don't see in https://apidock.com/rails/ActiveRecord/Base/exists%3f/class at all. The string one, in my estimation, comes from AR using integers for ids and providing the convenience shortcut of accepting a string from browser form submission which would be cast to integer.

In MongoDB _id can be of any type, so the second line should be `Object: an _id to search for`. String input I suppose could be handy but I'm not seeing where in this diff this is actually handled, if string input is handled via ObjectId's mongoizer then this documentation here should say that any input which is handled by the defined _id type will be appropriately cast.

Neilshweky: I originally didn't find the false either... I only saw it in the link that @comandeo sent: https://api.rubyonrails.org/v7.0.3.1/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F

Yes, you're right I will change that about the object id.

comandeo: @p-mongo apidock.com/rails seems to be broken. It shows the documentation for the _Ruby on Rails latest stable (**v5.2.3**)_ However, it is still the very first when one google a rails method :facepalm:

p-mongo: I see, that documentation doesn't explain the use case of `false` but since Rails cannot be wrong I approve us copying it.

Neilshweky: sounds good to me!

lib/mongoid/contextual/memory.rb
```diff
@@ -110,9 +110,24 @@ def each
...
+      # @param [ Hash | Object | false ] id_or_conditions an _id to
+      #   search for, a hash of conditions, or false.
+      #
       # @return [ true | false ] If the count is more than zero.
-      def exists?
-        any?
+      #   Always false if passed false.
+      def exists?(id_or_conditions = nil)
+        case id_or_conditions
+        when nil then any?
```
p-mongo, line 126: Per https://github.com/rails/rails/commit/d48dd18bb2a3d0c46708a9ee217909783b997cb2 I think this should return false (and this case added to docs).

p-mongo (commented):
I am trying to work out what the `false` input is supposed to do, haven't located it yet. Besides the comment I left about `nil` input there is also https://github.com/rails/rails/commit/debb3a178f9. You'll note that in that commit the signature defaults the parameter to `false`, maybe this is where `false` comes from. In current rails master the parameter is defaulted to `:none` symbol.

lib/mongoid/contextual/memory.rb
Neilshweky: I don't know if I agree... Then we would have to change the default parameter?

lib/mongoid/contextual/mongo.rb
```diff
@@ -158,12 +158,27 @@ def each(&block)
...
+      #   search for, a hash of conditions, or false.
+      #
       # @return [ true | false ] If the count is more than zero.
-      def exists?
-        !!(view.projection(_id: 1).limit(1).first)
+      #   Always false if passed false.
+      def exists?(id_or_conditions = nil)
+        case id_or_conditions
+        when nil then !!(view.projection(_id: 1).limit(1).first)
+        when false then false
```
johnnyshields, line 178: Whats the reason `false` returns false? Is this something from Rails?

Neilshweky: Yeah check out the discussion above.... Considering removing this

johnnyshields:
I agree Mongoid should diverge from Rails in the handling of `false`. Would suggest the value `false` should simply be disallowed to avoid confusion.

johnnyshields:
I think I maybe understand why Rails is using `:none`. Consider this code:

```
my_id = Pet.where(color: :red).first&._id
Pet.exists?(my_id)
```

If `my_id` is `nil`, then the second expression will return true. This is probably the case they are trying to prevent.

Neilshweky:
@comandeo this is rfal.

lib/mongoid/contextual/mongo.rb
```diff
@@ -168,14 +168,12 @@ def each(&block)
...
       #
       # @param [ Hash | Object | false ] id_or_conditions an _id to
-      #   search for, a hash of conditions, or false.
+      #   search for or a hash of conditions.
       #
       # @return [ true | false ] If the count is more than zero.
-      #   Always false if passed false.
       def exists?(id_or_conditions = nil)
         case id_or_conditions
         when nil then !!(view.projection(_id: 1).limit(1).first)
```
johnnyshields, line 176: @Neilshweky did you see my comment re: the reasons for Rails behavior? I really think we should change this to use :none rather than nil as default value, and make `nil` return `false`.

Neilshweky: hmmm point taken... I think it might be smart to take Rails' lead on this one... 

use :none for default and then automatically return false on "falsey" value

comandeo (approved):

PR #5467, opened 2022-09-14
MONGOID-5164 Use None criteria instead of $in with empty list in HABTM association

Neilshweky:
There is no behavior change here, as no query was being executed anyway (as per this [line](https://github.com/mongodb/mongoid/blob/master/lib/mongoid/association/referenced/has_many/enumerable.rb#L484))... I wouldn't mind closing this as won't do, or if we want to do this because it's cleaner. 

MONGOID-5164

### DIFF
--- a/lib/mongoid/association/referenced/has_and_belongs_to_many.rb
+++ b/lib/mongoid/association/referenced/has_and_belongs_to_many.rb
@@ -264,7 +264,12 @@ module Mongoid
         def query_criteria(id_list)
           crit = relation_class.criteria
           crit = crit.apply_scope(scope)
-          crit = crit.all_of(primary_key => {"$in" => id_list || []})
+          crit = if id_list
+            crit.all_of(primary_key => { "$in" => id_list })
+          else
+            crit.all.to_a
+            crit.none
+          end
           with_ordering(crit)
         end
       end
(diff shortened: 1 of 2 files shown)


lib/mongoid/association/referenced/has_and_belongs_to_many.rb
```diff
@@ -264,7 +264,11 @@ def with_ordering(criteria)
         def query_criteria(id_list)
           crit = relation_class.criteria
           crit = crit.apply_scope(scope)
-          crit = crit.all_of(primary_key => {"$in" => id_list || []})
+          crit = if id_list
+            crit.all_of(primary_key => { "$in" => id_list })
+          else
+            crit.none
```
p-mongo, line 270: Could you also omit `apply_scope` call for this branch?

p-mongo (approved):

lib/mongoid/association/referenced/has_and_belongs_to_many.rb
Neilshweky: sure

Neilshweky: done

comandeo (approved):

PR #5471, opened 2022-09-18
MONGOID-5488 - Bugfix: Date#__evolve_time__ should use the configured zone instead of Time.local (i.e. respecting Mongoid.use_activesupport_time_zone)

johnnyshields:
Currently `Date#__evolve_time__` uses "local" (not "UTC") which is not affected by Mongoid configuration.

This PR changes it to correct respect the `Mongoid.use_activesupport_time_zone` setting.

Refer to MONGOID-5488

### DIFF
--- a/docs/release-notes/mongoid-9.0.txt
+++ b/docs/release-notes/mongoid-9.0.txt
@@ -43,7 +43,7 @@ while ``ActiveModel::MissingAttributeError`` does not.
 
 
 Use configured time zone to typecast Date to Time in queries
-------------------------------------------------------------------
+-------------------------------------------------------------
 
 When querying for a Time field using a Date value, Mongoid now correctly
 considers the ``Mongoid.use_activesupport_time_zone`` to perform type
--- a/docs/release-notes/mongoid-9.0.txt
+++ b/docs/release-notes/mongoid-9.0.txt
@@ -46,8 +46,8 @@ Use configured time zone to typecast Date to Time in queries
 -------------------------------------------------------------
 
 When querying for a Time field using a Date value, Mongoid now correctly
-considers the ``Mongoid.use_activesupport_time_zone`` to perform type
-conversion.
+considers the ``Mongoid.use_activesupport_time_zone`` configuration option
+to perform type conversion.
 
 .. code-block:: ruby
 
--- a/docs/release-notes/mongoid-9.0.txt
+++ b/docs/release-notes/mongoid-9.0.txt
@@ -40,3 +40,36 @@ any Mongoid-specific usages of this class, and change them to
 ``Mongoid::Errors::AttributeNotLoaded``. Note additionally that
 ``AttributeNotLoaded`` inherits from ``Mongoid::Errors::MongoidError``,
 while ``ActiveModel::MissingAttributeError`` does not.
+
+
+Use configured Rails time zone to typecast Date to Time in queries
+------------------------------------------------------------------
+
+When querying for a Time field using a Date value, Mongoid now correctly
+considers the ``Mongoid.use_activesupport_time_zone`` to perform type
+conversion.
+
+.. code-block:: ruby
+
+  Mongoid.use_activesupport_time_zone = true
+
+  class Magazine
+    include Mongoid::Document
+
+    field :published_at, type: Time
+  end
+
+  Time.zone = 'Asia/Tokyo'
+
+  Magazine.gte(published_at: Date.parse('2022-09-26'))
+  #=> will return all results on or after Sept 26th, 2022
+  #   at 0:00 in Asia/Tokyo time zone.
+
+In prior Mongoid versions, the above code would ignore the
+``Mongoid.use_activesupport_time_zone`` setting and behave as if
+it were false, i.e. always using the system time zone to perform
+the type conversion.
+
+Note that in prior Mongoid versions, typecasting Date to Time during
+persistence operations was already correctly using the
+``Mongoid.use_activesupport_time_zone`` setting.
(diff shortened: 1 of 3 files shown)


spec/mongoid/criteria/queryable/extensions/date_spec.rb
```diff
@@ -25,20 +25,54 @@
...
+      end
+
+      it 'is a Time' do
+        expect(evolved.class).to eq(Time)
+      end
+
+      it 'is equal to expected time' do
+        expect(expected_time).to be_a(Time)
+        expect(evolved).to eq(expected_time)
+      end
```
johnnyshields, line 75: These specs were copied nearly verbatim from `Date#__mongoize_time__`.

lib/mongoid/criteria/queryable/extensions/date.rb
```diff
@@ -25,7 +25,7 @@ def __evolve_date__
           #
           # @return [ Time ] The date as a local time.
           def __evolve_time__
-            ::Time.local(year, month, day)
+            ::Time.configured.local(year, month, day).to_time
```
johnnyshields, line 28: I've added .to_time here in order to preserve the return time. However, it's not really necessary; AS::TimeWithZone works equivalently. `__mongoize_time__` will return either Time or AS::TimeWithZone based on Time.configured. Please let me know if you'd like me to remove this `to_time` to make it consistent with `__mongoize_time__`.

lib/mongoid/criteria/queryable/extensions/date.rb
```diff
@@ -23,9 +23,9 @@ def __evolve_date__
           # @example Evolve the date.
           #   date.__evolve_time__
           #
-          # @return [ Time ] The date as a local time.
+          # @return [ Time | ActiveSupport::TimeWithZone ] The date as a local time.
           def __evolve_time__
-            ::Time.local(year, month, day)
+            ::Time.configured.local(year, month, day)
```
johnnyshields, line 28: This code is now consistent with the `__mongoize_time__` behavior. There is no issue making this return `ActiveSupport::TimeWithZone` as the Ruby driver has no problem with it. However, we could suffix this with `.to_time` to avoid changing return type, but I think it's better to just be consistent with Mongoize.

p-mongo (changes requested):

lib/mongoid/criteria/queryable/extensions/date.rb
johnnyshields: AFAIK this code is correct as-is. Note `Time.configured` is a Mongoid-defined method which handles the branching for `use_activesupport_time_zone`.

It is not necessary to call `use_utc` here, as it only affects how time values are demongoized. From the Mongoid docs:

> If true, times stored in MongoDB will be returned in UTC.... The use_utc setting does not affect how times are parsed - parsing is always done in local time when the input being parsed does not include time zone information.

johnnyshields:
@p-mongo please see my response above. Your comment is marked "changes requested", please kindly either provide further specific instructions or else mark "resolved".

p-mongo (approved):

comandeo:
Is it possible that this change breaks existing code? If yes, we should at least mention this in the changelog.

johnnyshields:
@comandeo I think it's rare, but it is certainly *possible*, so I've added the changelog. Good call.

docs/release-notes/mongoid-9.0.txt
```diff
@@ -40,3 +40,36 @@ any Mongoid-specific usages of this class, and change them to
 ``Mongoid::Errors::AttributeNotLoaded``. Note additionally that
 ``AttributeNotLoaded`` inherits from ``Mongoid::Errors::MongoidError``,
 while ``ActiveModel::MissingAttributeError`` does not.
+
+
+Use configured Rails time zone to typecast Date to Time in queries
```
p-mongo, line 45:
```suggestion
Use configured time zone to typecast Date to Time in queries
```

docs/release-notes/mongoid-9.0.txt
```diff
@@ -40,3 +40,36 @@ any Mongoid-specific usages of this class, and change them to
...
+------------------------------------------------------------------
```
p-mongo, line 46:
```suggestion
-------------------------------------------------------------
```

johnnyshields:
@comandeo @Neilshweky please kindly approve, thanks!

docs/release-notes/mongoid-9.0.txt
```diff
@@ -40,3 +40,36 @@ any Mongoid-specific usages of this class, and change them to
 ``Mongoid::Errors::AttributeNotLoaded``. Note additionally that
 ``AttributeNotLoaded`` inherits from ``Mongoid::Errors::MongoidError``,
 while ``ActiveModel::MissingAttributeError`` does not.
+
+
+Use configured time zone to typecast Date to Time in queries
+-------------------------------------------------------------
+
+When querying for a Time field using a Date value, Mongoid now correctly
+considers the ``Mongoid.use_activesupport_time_zone`` to perform type
```
Neilshweky, line 49:
```suggestion
considers the ``Mongoid.use_activesupport_time_zone`` configuration option to perform type
```
Or something like this... Make sure the line lengths add up correctly.

Neilshweky (changes requested):

docs/release-notes/mongoid-9.0.txt
johnnyshields: Done

johnnyshields:
Comment addressed about. This is ready for merge.

comandeo (approved):

johnnyshields:
@Neilshweky thanks for merging the other PR. This one is waiting on your approval as well.

Neilshweky (approved):

johnnyshields:
thank you!

PR #5472, opened 2022-09-21
MONGOID-5082 Create contributing guidelines

Neilshweky:


docs/contributing/contributing-guidelines.txt
```diff
@@ -0,0 +1,35 @@
...
+#. Create a new branch in that fork.
+#. Make your changes.
+#. Ensure that the proposed changes have adequate test coverage.
+#. Raise a PR against Mongoid master. If these changes correspond to a specific
+   Jira ticket, title the PR: "MONGOID-<ticket #> Description of Changes".
+#. The Mongoid team will review the PR and make comments/suggest changes.
+#. Once all of the changes and fixes are made, and the Mongoid team determine
+   the PR fit for merging, we will merge the PR into master and determine
+   whether it needs to be backported.
+#. Backports to 7.3-present are done if the change is a bug fix, is not backwards
```
johnnyshields, line 31: It would be ideal if this could be expressed on a relative basis, e.g. "Backports to the current and previous major version are done..." so that it doesn't need to be updated on every release.

If the policy is taking effect from version 7.3, you could prefix it with "Beginning from version 7.3..."

alexbevi: I like this suggestion, but it's worth being specific for the time being. @Neilshweky I'd recommend changing the following bullet to 

```
#. Backports to previous stable versions are done if the change is a bug fix, is not backwards breaking, and if the commit is applicable to the corresponding stable branch. Presently backport candidates would include versions 7.3-8.0.
```

Neilshweky: done!

alexbevi (approved):
lgtm

p-mongo:
Ideally before writing code users should submit tickets via Jira. This is not strictly required, but I suggest recommending this workflow so that users who read these guidelines have an opportunity to receive feedback on their proposals before starting on implementation.

I would make this the first bullet point ahead of forking Mongiod, something like:

> If you wish to propose an enhancement to Mongoid, please create a Jira ticket describing the enhancement and what it would enable you to achieve in your application that is not already possible. If you believe Mongoid is not behaving correctly, please create a Jira ticket describing how you use Mongoid, what the existing behavior is that you consider incorrect or problematic, and what your desired behavior is.

Neilshweky:
> Ideally before writing code users should submit tickets via Jira. This is not strictly required, but I suggest recommending this workflow so that users who read these guidelines have an opportunity to receive feedback on their proposals before starting on implementation.
> 
> I would make this the first bullet point ahead of forking Mongiod, something like:
> 
> > If you wish to propose an enhancement to Mongoid, please create a Jira ticket describing the enhancement and what it would enable you to achieve in your application that is not already possible. If you believe Mongoid is not behaving correctly, please create a Jira ticket describing how you use Mongoid, what the existing behavior is that you consider incorrect or problematic, and what your desired behavior is.

@p-mongo  I made this the first paragraph and kept the list the same.

docs/contributing/contributing-guidelines.txt
johnnyshields: A second thought here, this probably belongs in a completely separate policy like "Security & Patching Policy". Typically backports aren't covered in Contributing Guidelines (unless you are asking the contributors to raise the backport PRs which some repos do, GRPC comes to mind.)

comandeo (approved):

PR #5474, opened 2022-09-23
MONGOID-5227 Add test and adjust docs for dotted attribute assignment in default scope

Neilshweky:


docs/reference/queries.txt
```diff
@@ -1900,23 +1898,9 @@ not possible:
     default_scope ->{ where('tags.foo' => 'bar') }
```
p-mongo, line 1898: This should have an example showing what the result of `Band.new` is, would it have the `tags.foo` as the attribute name?

docs/reference/queries.txt
```diff
@@ -1886,9 +1886,7 @@ in the default scope, the value in the default scope takes precedence:
 
   Band.new # => #<Band _id: 5c3f74ddce4ef3791abbb088, name: nil, active: false>
 
-Because a default scope initializes fields in new models as just described,
-defining a default scope with a dotted key and a simple literal value is
-not possible:
+As of Mongoid 8, a default scope can be defined with a dotted key:
```
johnnyshields, line 1889: When I first read this, I thought it was referring to dot/dollar key names. I'd recommend to be more explicit like the following:

```
As of Mongoid 8, a default scope can be defined using dot-notation for path traversal:
```

docs/reference/queries.txt
Neilshweky: I did not realize this would be a thing... I added docs and tests that demonstrate this.

p-mongo (approved):
Much appreciated.

docs/reference/queries.txt
```diff
@@ -1887,8 +1887,8 @@ in the default scope, the value in the default scope takes precedence:
   Band.new # => #<Band _id: 5c3f74ddce4ef3791abbb088, name: nil, active: false>
 
 Because a default scope initializes fields in new models as just described,
-defining a default scope with a dotted key and a simple literal value is
-not possible:
+defining a default scope with a dotted key and a simple literal value, while
+possible, is not recommended.
```
p-mongo, line 1891:
```suggestion
possible, is not recommended:
```

PR #5477, opened 2022-09-28
MONGOID-5312 Document attributes method in reference

Neilshweky:


docs/reference/crud.txt
```diff
@@ -37,6 +37,31 @@ operations with examples.
    * - Operation
      - Example
 
+   * - ``Model#attributes``
+
+       *Returns a hash of the document's attributes with its values in
+       mongoized form (i.e. the way it's stored in the db).*
```
p-mongo, line 43:
```suggestion
       mongoized form (i.e. the way they are stored in the db).*
```

docs/reference/crud.txt
```diff
@@ -37,6 +37,31 @@ operations with examples.
...
+
+       *In Mongoid 8.0 by default, and Mongoid 7.5 with the*
+       ``legacy_attributes`` *flag set to false, the object returned from this
+       method will have class* ``Hash``, *and have string keys. In Mongoid 7.5
+       by default, and in Mongoid 8.0 with the* ``legacy_attributes`` *flag set
+       to true, the object returned from this method will be a*
+       ``BSON::Document`` *(and thus, indifferently accessable) if the document
```
p-mongo, line 50:
```suggestion
       ``BSON::Document`` *(and thus, indifferently accessible) if the document
```

docs/reference/crud.txt
```diff
@@ -37,6 +37,31 @@ operations with examples.
...
+       has been retrieved from the database, and a* ``Hash`` *with string keys
+       if it is a new document.*
+
+       *If an embedded association is empty, its key will not show up in the
+       returned hash.*
```
p-mongo, line 55: Could you add before this sentence, in this paragraph, something like

> The `attributes` hash also contains the attributes of all embedded documents.

If I call `attributes` on a document which is itself embedded and it has another document embedded in it do I get the 2nd and 3rd level (counting from ultimate parent) attributes back?

Neilshweky:
Is that better @p-mongo ?

p-mongo (approved):

comandeo (approved):

PR #5486, opened 2022-09-30
MONGOID-5487 - evolve/mongoize Time/DateTime to BigDecimal (since it can cast to Integer/Float)

johnnyshields:
These conversions should be supported, since they work for Integer and Float.

- [ ] TODO: specs


Neilshweky:
Closing this PR as we've decided to not do this.

johnnyshields:
I've given some comments in the ticket MONGOID-5487 on why I think we should do this from a standardization perspective. No big issue to close, but perhaps we can revisit when we cleanup evolve/mongoize further.

PR #5492, opened 2022-10-03
MONGOID-5363 Change #upsert to perform updating upsert rather than replacing upsert

Neilshweky:

PR #5493, opened 2022-10-03
MONGOID-5363 switch default in 8.1

Neilshweky:
TODO: 

- Rebase on 8.1-stable once https://github.com/mongodb/mongoid/pull/5492 is merged 
and backported.
- Change the base of this PR to 8.1-stable


johnnyshields:
Great, this is very good to add the flipped defaults to the 9.0 changelog.

If there are going to be multiple defaults flipped in 9.0, it may be easier/more terse just to list them as bullet points under the heading "Flipped feature flag defaults" with a one sentence explanation of each.

Neilshweky:
@johnnyshields this isn't a feature flag, but an option default within a method.

johnnyshields:
Ah, sorry didn't read diff 🤦‍♂️. I am correct to understand that this is non-breaking vs 8.0 because https://github.com/mongodb/mongoid/pull/5492 flipped the actual behavior, so "flipping the default" of this method preserves the 8.0 behavior/

lib/mongoid/persistable/upsertable.rb
```diff
@@ -12,13 +12,13 @@ module Upsertable
       #
       # If the replace option is true, unspecified attributes will be dropped,
       # and if it is false, unspecified attributes will be maintained. The
-      # replace option defaults to false in Mongoid 9.
+      # replace option defaults to true in Mongoid 8.1 and earlier.
```
comandeo, line 15: Does it worth mentioning that the default will be changed to `false` in 9.0, since we know that?

comandeo (approved):

lib/mongoid/persistable/upsertable.rb
Neilshweky: good call

Neilshweky:
> Ah, sorry didn't read diff 🤦‍♂️. I am correct to understand that this is non-breaking vs 8.0 because #5492 flipped the actual behavior, so "flipping the default" of this method preserves the 8.0 behavior?

Correct, I did it kind of backwards since you have to work from master backwards. This is a non-breaking change.

PR #5494, opened 2022-10-03
MONGOID-5331 Document hash assignment to associations

Neilshweky:
Turns out this only works for embedded associations (but not `embedded_in` until this PR), but they cannot possible work for referenced associations, since any non document assigned to a referenced association is interpreted as a queryable item for retrieving the base document. 

I don't think we need to add this functionality for referenced association, and we know that referenced and embedded associations are not the same.

### DIFF
--- a/docs/reference/associations.txt
+++ b/docs/reference/associations.txt
@@ -906,6 +906,7 @@ example:
 This works for ``embeds_one``, ``embeds_many``, and ``embedded_in`` associations.
 Note that you cannot assign hashes to referenced associations.
 
+
 Common Behavior
 ===============
 
--- a/docs/reference/associations.txt
+++ b/docs/reference/associations.txt
@@ -875,12 +875,14 @@ documents that exist in the application:
   band.tours.destroy_all
 
 
+.. _hash-assignment:
+
 Hash Assignment
 ---------------
 
 Embedded associations allow the user to assign a ``Hash`` instead of a document
-to an association. On assignment, this hash is coerced into the class of the
-document for the association that it's being assigned to. Take the following
+to an association. On assignment, this hash is coerced into a document of the
+class of the association that it's being assigned to. Take the following
 example:
 
 .. code:: ruby
--- a/docs/reference/associations.txt
+++ b/docs/reference/associations.txt
@@ -875,6 +875,35 @@ documents that exist in the application:
   band.tours.destroy_all
 
 
+Hash Assignment
+---------------
+
+Embedded associations allow the user to assign a ``Hash`` instead of a document
+to an association. On assignment, this hash is coerced into the class of the
+document for the association that it's being assigned to. Take the following
+example:
+
+.. code:: ruby
+
+  class Band
+    include Mongoid::Document
+    embeds_many :albums
+  end
+
+  class Album
+    include Mongoid::Document
+    field :name, type: String
+    embedded_in :band
+  end
+
+  band = Band.create!
+  band.albums = [ { name: "Narrow Stairs" }, { name: "Transatlanticism" } ]
+  p band.albums
+  # => [ #<Album _id: 633c71e93282a4357bb608e5, name: "Narrow Stairs">, #<Album _id: 633c71e93282a4357bb608e6, name: "Transatlanticism"> ]
+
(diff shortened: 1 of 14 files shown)


spec/mongoid/association/embedded/embedded_in/proxy_spec.rb
```diff
@@ -616,4 +616,31 @@
...
+    let(:building_address) { BuildingAddress.new }
+
+    before do
+      building_address.building = { name: "Chrysler" }
+      building_address.building = { name: "Empire State" }
+    end
+
+    it "creates the objects correctly" do
+      expect(building_address.building).to be_a(Building)
+      expect(building_address.building.name).to eq("Empire Station")
```
Neilshweky, line 643:
```suggestion
      expect(building_address.building.name).to eq("Empire State")
```

spec/mongoid/association/embedded/embedded_in/proxy_spec.rb
```diff
@@ -616,4 +616,31 @@
       end
     end
   end
+
+  context "when assigning a hash" do
```
johnnyshields, line 620: Suggest to add (1) aliased and (2) localized fields to all specs.

(I know it's a pain but they always tend to break :)

johnnyshields:
@Neilshweky this should be tested for both embeds_one and embeds_many. AFAIK it seems that only the embeds_one case is covered.

johnnyshields:
Re: other types of relations (referenced, embedded_in, etc.) you are correct that hash assignment does not need to be supported, I don't think anyone would reasonably expect it to be.

Neilshweky:
> @Neilshweky this PR is definitely a convenience to have.
> 
> Test cases should cover:

I think I added tests for embeds_one/embeds_many/embedded_in (which is also implemented in this PR)
> * embeds_one and embeds_many. AFAIK it seems that only the embeds_one case is covered.

These should be covered by the `Factory\.build` method tests, these tests are just to verify that it actually makes it there. 
> * aliased fields
> * localized fields
> * whether or not _id field is automatically added.

This is not in the scope of this PR, we can have a discussion about that separately.
> In addition, we might want to think about ActionController::Parameters here in a separate PR. In the controller code, there's a good chance someone will unwittingly assign `params` (or a subsection of `params`) which not a Hash but actually ActionController::Parameters. This was the kind of case that I was intending to address in my PR that was closed FYI.

docs/reference/associations.txt
```diff
@@ -875,6 +875,37 @@ documents that exist in the application:
...
+  end
+
+  band = Band.create!
+  band.albums = [ { name: "Narrow Stairs" }, { name: "Transatlanticism" } ]
+  p band.albums
+  # => [ #<Album _id: 633c71e93282a4357bb608e5, name: "Narrow Stairs">, #<Album _id: 633c71e93282a4357bb608e6, name: "Transatlanticism"> ]
+
+This works for ``embeds_one``, ``embeds_many``, and ``embedded_in`` associations.
+Note that you cannot assign hashes to referenced associations.
+
```
Neilshweky, line 908:
```suggestion


```

spec/mongoid/association/embedded/embedded_in/proxy_spec.rb
```diff
@@ -616,4 +616,31 @@
...
+      building_address.building = { name: "Chrysler" }
+    end
+
+    it "creates the objects correctly" do
+      expect(building_address.building).to be_a(Building)
+      expect(building_address.building.name).to eq("Chrysler")
+    end
+  end
+
+  context "when assigning an association with a hash" do
```
comandeo, line 633: I am not sure if the context description reflects its content. The difference between the test above and this one is that above we assign hash once, and here twice. I do not see that we assign "an association with a hash" here.

comandeo (approved):

spec/mongoid/association/embedded/embedded_in/proxy_spec.rb
Neilshweky: ah right this should replacing

```suggestion
  context "when replacing an association with a hash" do
```

PR #5495, opened 2022-10-04
MONGOID-5016 Change the default of :touch option to true on embedded associations

Neilshweky:


lib/mongoid/touchable.rb
```diff
@@ -133,10 +133,10 @@ def define_touchable!(association)
     # @return [ Symbol ] The method name.
     def define_relation_touch_method(name, association)
       relation_classes = if association.polymorphic?
-                           association.send(:inverse_association_classes)
-                         else
-                           [ association.relation_class ]
-                         end
+        association.send(:inverse_association_classes)
```
Neilshweky, line 136: Now, all embedded associations that don't specify touch false, go through this code path... This is significant because below, the `#relation_class` method is called, which causes the associated class to be loaded (and thus if it has errors, those will be raised)... This segues pretty well into the Resolving Association Class Names epic, and I think that it would give us the opportunity to have a discussion about what we want to do about that. 

Since, now, most embedded associations will be loaded almost immediately (unless touch is set to false), we may want to just do this by default?

Neilshweky: Additionally we can change the way touch works, and instead of adding an after_save callback that does the touching, we can just do it within the actual save. This is a prerequisite to MONGOID-5392

johnnyshields: > after_save callback that does the touching, we can just do it within the actual save

Yep, absolutely this is what we should be doing.

johnnyshields: One random thought is we should have `touch` (non-persistent) and `touch!` immediately persistent. Could be revisited in the persistence deferral epic.

johnnyshields: Hmmm... @Neilshweky just thinking aloud. Is it correct that, versus the Mongoid 8.x behavior, this change will now make it so calling `save` on an embedded document will now do an extra persistence op in the after_save to update it's parent's timestamp (unless we implement MONGOID-5392)

If that's the case, its a minor performance regression... if MONGOID-5392 is straightforward should probably do it in Mongoid 9.0 as well.

Neilshweky:
This cannot be done until we rework touch or implement the Resolve Association Names epic. Closing for now.

johnnyshields:
@Neilshweky what's the plan for Mongoid 9.0 then? My previous PR fixes bugs that ended up changing behavior, i.e. from a user perspective.

PR #5485, opened 2022-09-30
Fix two bugs in BigDecimal mongoization

johnnyshields:
1. demongoize - has redundant elsif object.numeric?
2. mongoize - missing to_s in return condition when map_big_decimal_to_decimal128 is false

PR #5496, opened 2022-10-04
MONGOID-5515 Add release notes for smaller bug fixes/improvements for 9.0 features until now

Neilshweky:
depends on #5485


johnnyshields:
Neil this is great, very thorough. Should consider to add the JIRA ticket numbers on the big items as well.

Neilshweky:
> Neil this is great, very thorough. Should consider to add the JIRA ticket numbers on the big items as well.

I feel like it's only necessary to add the jira ticket number when I don't have the opportunity to explain in detail, like with these smaller explanations.

comandeo (approved):

PR #5500, opened 2022-10-24
MONGOID-5522: Update README

alexbevi:


alexbevi:
Based on https://www.markdownguide.org/basic-syntax/#paragraph-best-practices I updated the PR to remove the forward-slash and use two spaces instead (to denote a line break)

comandeo (approved):

PR #5504, opened 2022-11-08
MONGOID-5505 Skip jruby failing tests

comandeo:


Neilshweky (approved):
What's the purpose of this PR? Why did you get rid of the reference to MONGOID-5505?

PR #5517, opened 2022-12-20
RUBY-5530 Add transaction method to Mongoid module

comandeo:


spec/mongoid/clients/transactions_spec.rb
```diff
@@ -735,4 +735,71 @@
...
+          begin
+            Mongoid.transaction do
+              Person.create!
+              Account.create!
+            end
+          rescue => ex
+            e = ex
+          end
+          e
+        end
```
jamis, line 787: What would you think of pushing this logic into an auxiliary method, so that the test's focus isn't obscured by the mechanics of capturing the exception? Something like:

```ruby
let!(:error) do
  capture_exception do
    Mongoid.transaction do
       # ...
    end
  end
end
```

jamis (approved):
Thanks for sharing the PR with me! This is very helpful, to see the expected comment style for internal methods, as well as a focused example of new tests.

spec/mongoid/clients/transactions_spec.rb
comandeo: A very good hint, thank you! Updated the test accordingly.

PR #5519, opened 2023-01-06
MONGOID-5531 Add transaction callbacks

comandeo:


lib/mongoid/clients/sessions.rb
```diff
@@ -82,27 +80,52 @@ def transaction(options = {}, session_options: {})
...
+        # after_rollback callbacks on modified documents.
+        #
+        # @param [ Mongo::Session ] session Session on which
+        #   a transaction is started.
+        def abort_transaction(session)
+          session.abort_transaction
+          Threaded.clear_modified_documents(session).each do |doc|
+            doc.run_after_callbacks(:rollback)
+          end
+        end
```
johnnyshields, line 128: I understand that these `after_commit` and `after_rollback` callbacks are how Rails ActiveRecord works, but I'm struggling to understand the real-world use case for them.

It seems to be a more meaning for thing to have would be `after_persistence` which would be equivalent to `after_save` OR `after_commit`, depending on whether the persistence operation is happening as part of a transaction. AFAIK 

Note rails also has:

```
after_create_commit
after_save_commit
after_destroy_commit


I can't envision a real-world use case for `after_commit` by itself.

File: lib/mongoid/interceptable.rb [hunk 2]
```diff
@@ -43,6 +43,8 @@ module Interceptable
       # @api private
       define_model_callbacks :persist_parent
 
+      define_model_callbacks :commit, :rollback, only: :after
```
johnnyshields, line 46: Note Rails also has [these methods](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html#method-i-after_save_commit):


```
after_create_commit
after_update_commit
after_save_commit
after_destroy_commit
```

comandeo: @johnnyshields Good catch, thank you! I think we forgot about them when planing this featue. I create a ticket to implement them in score of this epic - https://jira.mongodb.org/browse/MONGOID-5540

lib/mongoid/clients/sessions.rb
comandeo: Rails tutorial provides [an example](https://guides.rubyonrails.org/active_record_callbacks.html#transaction-callbacks) how these callbacks can be used. I think I've done something similar earlier, so I believe that there are use cases at least for `after_commit`.

As to your proposal for `after_persistence` callback - I am not sure if I like the idea of a callback that behaves differently with or without transaction. However, If you believe there is a need for one - let's discuss it in a dedicated ticket.

johnnyshields: I don't think the example is very good. What if you destroy the `Picture` object without using a transaction?

comandeo: After discussion with @jamis I believe I understand the concern here. 

In Rails every `save` is implicitly wrapped in a transaction; therefore, `after_commit` is called after a successful `save` **even without an explicit outer transaction block**. In Mongoid, however, we do not have implicit transactions for `save`, so, `after_commit` **won't be called on "regular" saves**. Hence the inconsistent behaviour that was mentioned.

Thank you @johnnyshields for raising this topic early! We will come up with a better solution.

johnnyshields: Cool, glad my meaning was clear.

If ActiveRecord indeed triggers `after_commit` after every persistence operation (since all AR persistence is wrapped in a transaction), then I think Mongoid should do the same.

jamis (approved):
Looks good! I think paring back the scope (re: after_rollback) really helped keep things cleaner here.

PR #5521, opened 2023-01-13
MONGOID-5442 implement missing update operators

jamis:
This PR adds support for the following update operators:

* `$min`
* `$max`
* `$mul`
* `$setOnInsert` (via `#upsert`).

Note that specifying`set_on_insert` with `replace: true` in `#upsert` will raise an exception.


lib/mongoid/contextual/atomic.rb
```diff
@@ -163,11 +175,97 @@ def unset(*args)
...
+      #   of being named for the MongoDB operation they perform.
+      #
+      # @example Set "views" to be no more than 100.
+      #   context.update_min(views: 100)
+      #
+      # @param [ Hash ] fields The fields with the maximum value that each
+      #   may be set to.
+      #
+      # @return [ nil ] Nil.
+      def update_min(fields)
```
johnnyshields, line 218: Using the word `update_` here is confusing vis-a-vis `update_attributes` in Mongoid, which invokes model callbacks. As far as Mongoid is concerned, we should reserve the word `update` exclusively for actions which do callbacks.

(Yes the driver has `update_many`, but Mongoid users will not ever be using these driver methods directly.)

I would propose something like `atomic_min` or `set_to_min` as an alternative (the latter being congruent with the existing atomic `set` method.)

lib/mongoid/persistable/datable.rb
```diff
@@ -0,0 +1,59 @@
...
+      #   document.current_date(updated: :date)
+      #
+      # @example Set multiple fields to the current date, in different formats:
+      #   document.current_date(:updated, { founded: true, touched: :timestamp })
+      #
+      # @param [ Array<Array | Hash | Symbol | String> ] specs The fields to
+      #   set to the current date/time.
+      #
+      # @return [ Document ] The document.
+      def current_date(*specs)
```
johnnyshields, line 45: This is going to be problematic because it is possible that a Document would already have a field named `current_date`. Here `set_to_current_date` would be better.

lib/mongoid/contextual/atomic.rb
```diff
@@ -163,11 +175,97 @@ def unset(*args)
...
+      #   context.current_date(completed: true)
+      #
+      # @example Set multiple fields to the current date.
+      #   context.current_date(:completed, last_viewed: :timestamp)
+      #
+      # @param [ Array<Hash | Symbol | String> ] fields The fields with the
+      #   corresponding date format to use (true, :timestamp, or :date)
+      #
+      # @return [ nil ] Nil.
+      def current_date(*fields)
```
johnnyshields, line 195: See comment below; `set_to_current_date` proposed.

lib/mongoid/persistable/maxable.rb
```diff
@@ -0,0 +1,35 @@
...
+      # value, or a given value.
+      #
+      # @example Set a field to be no less than 100.
+      #   document.max(field: 100)
+      #
+      # @param [ Hash<Symbol | String, Comparable> ] fields The fields to
+      #   set, with corresponding minimum values.
+      #
+      # @return [ Document ] The document.
+      def max(fields)
```
johnnyshields, line 21: The naming here should be consistent with whatever the atomic operator is named. Proposed to be `set_to_ceil/floor`

lib/mongoid/persistable/maxable.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Persistable
+
+    # Defines behavior for setting a field (or fields) to the larger of
+    # either it's current value, or a given value.
+    module Maxable
```
johnnyshields, line 8: General comment but I don't see how having one module per method makes the code more readable. Probably all these Persistable atomic-like methods should be grouped into a single module.

lib/mongoid/persistable/datable.rb
```diff
@@ -0,0 +1,59 @@
...
+      #   to translate.
+      #
+      # @return [ true | Hash ] The type description indicated by the argument.
+      #
+      # @api private
+      def self.translate_date_field_spec(type)
+        case type
+        # `current_date(:foo)` == `current_date(foo: true)`
+        when nil then true
+        when "timestamp", :timestamp, "date", :date
```
johnnyshields, line 23: Consider allowing :time as an alias for timestamp here.

lib/mongoid/persistable/datable.rb
```diff
@@ -0,0 +1,59 @@
...
+          { "$type" => type.to_s }
+        else type
```
johnnyshields, line 25: Shouldn't this be `else true`?

lib/mongoid/contextual/atomic.rb
jamis: Good point about the `update` prefix; I'll switch to `set`, which I agree is clearer. I agree that the `min`/`max` naming is confusing, but I think it is important that the canonical method mirror the naming of the underlying MongoDB operation. That said, I'm totally on board with having an alias or two to make the intention clearer.

lib/mongoid/persistable/maxable.rb
jamis: I actually disagree here; looking at the `Contextual::Atomic` module, I've considered breaking it up so that single module isn't so monolithic. (I won't, though---other priorities!) A compromise could be to group similar methods (e.g. `min` and `max`) into modules. Something to consider for future work.

lib/mongoid/persistable/datable.rb
jamis: Good eye, but no, this is intentional. The idea is that we can allow programmers to pass either `true` explicitly, or a Hash object representing the payload to send to the `$currentDate` command, e.g.:

```ruby
  model.current_date(stamp: { "$type" => "date" })
```

This is primarily to grant access to the underlying API, where (e.g.) `stamp: :date` is shorthand for `stamp: { "$type" => "date" }`. It is conceivable that a future version of MongoDB could expose new parameters here, and this would allow access to those even before Mongoid itself is updated.

lib/mongoid/persistable/datable.rb
jamis: > how does MongoDB know the timezone in order to determine the date?

Excellent point. The answer is "it doesn't". It will use (I believe) whatever time and date are configured on the server. In a world where time zones matter, it seems more useful to explicitly set the timestamp to an appropriate value, rather than letting the server do it.

I'll discuss with the team; I still think it's important to have functions from the underlying API accessible, but maybe this one could use *less* visibility.

lib/mongoid/contextual/atomic.rb
johnnyshields: usually I'm not a big fan of aliases but in this particular case `min/max` are going to be impossible for humans to member which is which here.

lib/mongoid/persistable/maxable.rb
johnnyshields: It's a matter of taste. At my company, we only use mixins where there is a bona-fide need to add the same method(s) to two or more classes. When I see a mixin like `Maxable` or `Datable`, firstly the names are very silly, but a bigger problem is that it's not obvious to me what other classes/modules besides `Persistable` might be including `Maxable` and `Datable`, and what knock-on effects refactoring them might cause.

lib/mongoid/persistable/datable.rb
johnnyshields: how about `when Hash` then instead? I can see people doing something like `model.current_date(stamp: true)`, `model.current_date(stamp: 1)`, etc.

lib/mongoid/contextual/atomic.rb
```diff
@@ -52,6 +52,18 @@ def inc(incs)
...
 
+      # Perform an atomic $mul operation on the matching documents.
+      #
+      # @example Perform the atomic multiplication.
+      #   context.set_mul(likes: 10)
+      #
+      # @param [ Hash ] factors The operations.
+      #
+      # @return [ nil ] Nil.
+      def set_mul(factors)
```
johnnyshields, line 63: `mul` I think is fine to be as-is (not `set_mul`). It's analogous to `inc` and is a unique word (unlikely to be used as a field by a model)

lib/mongoid/contextual/atomic.rb
johnnyshields: should remove this if now out-of-scope for the PR

lib/mongoid/persistable/datable.rb
jamis: We opted to remove `#current_date` altogether. A better implementation for Mongoid would be a "#set_timestamp" method, which could take into account both the type of the field, and any configured time zone. That would be out of scope for this change, though.

lib/mongoid/contextual/atomic.rb
```diff
@@ -163,11 +175,99 @@ def unset(*args)
...
+      #   of being named for the MongoDB operation they perform.
+      #
+      # @example Set "views" to be no less than 100.
+      #   context.set_max(views: 100)
+      #
+      # @param [ Hash ] fields The fields with the minimum value that each
+      #   may be set to.
+      #
+      # @return [ nil ] Nil.
+      def set_max(fields)
```
johnnyshields, line 241: should be *fields (also set_min)

lib/mongoid/contextual/atomic.rb
```diff
@@ -163,10 +175,62 @@ def unset(*args)
...
+        view.update_many("$max" => collect_operations(fields))
+      end
+      alias :clamp_lower_bound :set_max
```
johnnyshields, line 222: In a separate PR we should add a `#clamp` method which combines $max and $min in to one atomic operation:

https://ruby-doc.org/core-3.1.1/Comparable.html#method-i-clamp

To be congruent with Ruby, args can be Array<min, max> OR a Range in the form `field: [-10, 30]` or `field: -10..30`. Ruby supports one-sided ranges for this as well.

comandeo (approved):
:+1:

PR #5522, opened 2023-01-15
Fix for rails runtime logging

dekelna:
Simple fix to show the runtime in ms as expected.

When runtime is actually 0.16633599996566772 (seconds)
Before:
INF | Completed 200 OK in 4987ms (Views: 0.2ms | MongoDB: 0.2ms | Allocations: 220793)
After:
INF | Completed 200 OK in 4987ms (Views: 0.2ms | MongoDB: 166.3ms | Allocations: 220793)

### DIFF
--- a/lib/mongoid/railties/controller_runtime.rb
+++ b/lib/mongoid/railties/controller_runtime.rb
@@ -43,7 +43,7 @@ module Mongoid
           def log_process_action(payload)
             messages = super
             mongoid_runtime = payload[:mongoid_runtime]
-            messages << ("MongoDB: %.1fms" % mongoid_runtime.to_f) if mongoid_runtime
+            messages << ("MongoDB: %.1fms" % (mongoid_runtime * 1000.0)) if mongoid_runtime
             messages
           end
 


lib/mongoid/railties/controller_runtime.rb
```diff
@@ -43,7 +43,7 @@ module ClassMethods
           def log_process_action(payload)
             messages = super
             mongoid_runtime = payload[:mongoid_runtime]
-            messages << ("MongoDB: %.1fms" % mongoid_runtime.to_f) if mongoid_runtime
+            messages << ("MongoDB: %.1fms" % (mongoid_runtime * 1000.0)) if mongoid_runtime
```
johnnyshields, line 46: Maybe it would be better to multiply by 1000 on this line:

```
          self.mongoid_runtime = mongo_rt_before_render + mongo_rt_after_render
          runtime - mongo_rt_after_render
```

That way the variable `mongoid_runtime` in the payload would be in milliseconds.

We should take a peek with what ActiveRecord does for this and try to be consistent.

johnnyshields:
@jamis I think this was already fixed in https://jira.mongodb.org/browse/MONGOID-4811

@dekelna you may be using an outdated version of Mongoid--can you check on Mongoid [8.0.1](https://jira.mongodb.org/issues/?jql=project+%3D+MONGOID+AND+fixVersion+%3D+8.0.1) or later and see if the issue persisted.

alexbevi:
@johnnyshields thanks for calling this out. I'm going to close this PR as [MONGOID-4811](https://jira.mongodb.org/browse/MONGOID-4811) addressed this issue already.

@dekelna if you're seeing this issue in a version of Mongoid > 8.0.1 please let us know.

PR #5524, opened 2023-01-18
MONGOID-5453 Add .none_of query method

jamis:
Jira: https://jira.mongodb.org/browse/MONGOID-5453

This adds a new query method, `#none_of`, which works as the opposite of `#any_of`. It will wrap all of the criteria it is given in a `$nor` operator.


comandeo (changes requested):
Looks good to me!

Please add this method to queries documentation (`docs/reference/queries.txt`), and also mention this in 8.1 release notes (`docs/release-notes/mongoid-8.1txt`)

PR #5526, opened 2023-01-20
MONGOID-5547 create index needs to load models

jamis:
The `:load_models` task for the rake tasks was a no-op, presumably because back in 2014 some work was done to decouple the tasks from Rails, but the model loading code was never completely decoupled.

This PR implements model loading for non-Rails apps, adding a `model_paths` option that will use the Rails configuration if present, and will fall back to some sane defaults otherwise (`./app/models` and `./lib/models`).

In order to make sure this works correctly under environments with and without Rails, I also added a new helper for testing, called `FeatureSandbox`, which can quarantine a code block in a test to ensure that the load paths, loaded features, and constants are all reset back to their original value when the block finishes.

ref: https://jira.mongodb.org/browse/MONGOID-5547

### DIFF
--- a/lib/mongoid.rb
+++ b/lib/mongoid.rb
@@ -21,6 +21,7 @@ require "mongoid/version"
 require "mongoid/deprecable"
 require "mongoid/config"
 require "mongoid/persistence_context"
+require "mongoid/loadable"
 require "mongoid/loggable"
 require "mongoid/clients"
 require "mongoid/document"
@@ -41,6 +42,7 @@ I18n.load_path << File.join(File.dirname(__FILE__), "config", "locales", "en.yml
 module Mongoid
   extend Forwardable
   extend Loggable
+  extend Loadable
   extend self
   extend Clients::Sessions::ClassMethods
 
@@ -70,7 +72,7 @@ module Mongoid
   #
   #     config.preload_models = true
   #
-  # @return [ Config ] The configuration object.
+  # @return [ Config ] The xuration object.
   def configure(&block)
     return Config unless block_given?
 
--- a/lib/mongoid/loadable.rb
+++ b/lib/mongoid/loadable.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+module Mongoid
+
+  # Defines how Mongoid can autoload all defined models.
+  module Loadable
+    # Search a list of model paths to get every model and require it, so
+    # that indexing and inheritance work in both development and production
+    # with the same results.
+    #
+    # @example Load all the application models from default model paths.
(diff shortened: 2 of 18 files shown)


johnnyshields:
Nice, thank you. This issue was always an annoyance for me.

lib/mongoid.rb
```diff
@@ -70,7 +72,7 @@ module Mongoid
   #
   #     config.preload_models = true
   #
-  # @return [ Config ] The configuration object.
+  # @return [ Config ] The xuration object.
```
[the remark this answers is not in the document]
jamis: Wups! Yes, I must have fat-fingered that somehow. Thanks for catching it!

lib/mongoid/loadable.rb
```diff
@@ -0,0 +1,77 @@
...
+    # `require` will be used.
+    #
+    # @example Load the model.
+    #   Mongoid.load_model("/mongoid/behavior")
+    #
+    # @param [ String ] file The base filename.
+    #
+    # @api private
+    def load_model(file)
+      if defined? require_dependency
```
comandeo, line 45:
```suggestion
      if defined?(require_dependency)
```

lib/mongoid/loadable.rb
```diff
@@ -0,0 +1,77 @@
...
+    #
+    # Note that these paths are the *roots* of the directory hierarchies where
+    # the models are located; it is not necessary to indicate every subdirectory,
+    # as long as these root paths are located in `$LOAD_PATH`.
+    #
+    # @return [ Array<String> ] the array of model paths
+    def model_paths
+      @model_paths ||= defined?(Rails) ?
+        Rails.application.config.paths["app/models"].expanded :
+        %w( ./app/models ./lib/models )
```
comandeo, line 65: I would extract this array to a constant. Not because it is necessary, but for better visibility of the default model paths.

comandeo (changes requested):
Great job! I particularly like the feature sandbox idea, it will be also useful in the future.

I left a couple of minor comments, not very important. However, we should update the documentation accordingly, particularly here - https://www.mongodb.com/docs/mongoid/master/reference/indexes/#using-rake-tasks-with-non-rails-applications

comandeo (approved):

jamis:
I know Friday is the day for digging into test failures, but these app_spec failures are maddening. For some reason, the Gemfile.lock is requiring concurrent_ruby 1.2.0, and Mongoid requires *less than* 1.2.0. I can't duplicate the issue locally, though. I'm going to spend a little more time this morning trying to puzzle this out.

PR #5528, opened 2023-01-20
MONGOID-5546 First/Last failures when empty hash passed as criteria

jamis:
`Mongo#first` and `Mongo#last` are trying to interpret the argument as either nil, an integer, or a Hash. When checking to see if the argument might be a Hash with the `:id_sort` key, the code uses `Hash#fetch`, which raises an exception if the requested key does not exist. This meant that passing an empty hash to these methods caused problems.

This PR gives those functions explicit `limit` (nil or Integer) and `opts` (Hash) values, extracted from the argument, which makes it easier to test for the `:id_sort` key without requiring it to be present.

### DIFF
--- a/lib/mongoid/contextual/mongo.rb
+++ b/lib/mongoid/contextual/mongo.rb
@@ -260,16 +260,16 @@ module Mongoid
       #
       # @return [ Document ] The first document.
       def first(limit_or_opts = nil)
-        limit = limit_or_opts unless limit_or_opts.is_a?(Hash)
+        limit, opts = extract_limit_and_opts(limit_or_opts)
         if cached? && cache_loaded?
           return limit ? documents.first(limit) : documents.first
         end
         try_numbered_cache(:first, limit) do
-          if limit_or_opts.try(:key?, :id_sort)
+          if opts.key?(:id_sort)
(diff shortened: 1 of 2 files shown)


lib/mongoid/contextual/mongo.rb
```diff
@@ -376,12 +376,12 @@ def initialize(criteria)
       #
       # @return [ Document ] The last document.
       def last(limit_or_opts = nil)
-        limit = limit_or_opts unless limit_or_opts.is_a?(Hash)
+        limit, opts = extract_limit_and_opts(limit_or_opts)
         if cached? && cache_loaded?
           return limit ? documents.last(limit) : documents.last
         end
         res = try_numbered_cache(:last, limit) do
```
Neilshweky, line 383: I thought we got rid of this numbered cache in favor of the regular QueryCache? Do any of the other methods use it?

Neilshweky (approved):
Nice! Glad to see you're improving mongoid!

lib/mongoid/contextual/mongo.rb
jamis: This is 7.5-stable---confusing, I know! This is fixing a regression. It doesn't affect 8.

comandeo (approved):

PR #5529, opened 2023-01-23
MONGOID-5439 Generate `config/initializers/mongoid.rb`

jamis:
Update the Mongoid rails generator to add config/initializers/mongoid.rb. The mongoid.yml file has been changed to be prepopulated with comments describing all existing `Mongoid::Config` options. Because the initializer can also be used for specifying configuration options, I've added a comment there indicating that the preferred location for configuration is the mongoid.yml file.


lib/rails/generators/mongoid/config/templates/mongoid.yml
```diff
@@ -117,62 +117,6 @@ development:
...
-
-    # Return stored times as UTC. See the time zone section below for
-    # further information. Most applications should not use this option.
-    # (default: false)
-    # use_utc: false
-
-    # (Deprecated) In MongoDB 4.0 and earlier, set whether to create
-    # indexes in the background by default. (default: false)
-    # background_indexing: false
-
```
johnnyshields, line 175: Careful here. Most of these options are "matters user of preference" which are not deprecated. Perhaps they should be moved to `mongoid.rb` instead, rather than being set in `mongoid.yml`.

The ones that can be removed are:
- `use_activesupport_time_zone`
- `background_indexing`

alexbevi:
Can we ensure https://github.com/mongodb/mongoid/blob/master/docs/reference/configuration.txt is updated to reflect how this will change configuration behavior?

lib/rails/generators/mongoid/config/config_generator.rb
```diff
@@ -25,9 +25,50 @@ def app_name
...
+      #
+      # @param [ Integer ] indent How many spaces each comment ought to
+      #   be re-indented.
+      # @param [ true | false ] include_deprecated Whether deprecated options
+      #   should be included in the list.
+      #
+      # @return [ Array<Hash<desc: [String], name: [String], default: [String],
+      #   deprecated: [nil | Integer]>> ] the array of hashes representing
+      #   each defined option, in alphabetical order by name.
+      def mongoid_config_options(indent: 2, include_deprecated: false)
```
comandeo, line 48: I like everything else this PR, but the idea of parsing source code with a regex to generate another source code scares me. It seems very fragile.
Is there any other reasonable way to do so?

johnnyshields: Strongly agreed. Perhaps the code generation could be extracted to a Rake task.

jamis: In defense of the current approach, it doesn't generate _code_, it generates _comments_. Thus, in the worst case, it generates nonsense that doesn't actually break anything. It is unlikely to do so, though, unless we drastically change the format of `mongoid/config.rb`, in which case, the corresponding test (in app_spec.rb) would likely also break and alert us to that fact.

However, here other options we could consider:

- We could **nix the automation altogether** and just duplicate the documentation. This is what is being done in the released versions currently, but it is fragile in another signficant way: it requires the documentation to be maintained in two places (mongoid/config.rb AND the generator template file), and it is very easy to forget about the template file when adding or updating configuration options. I do not recommend this approach.
- We could **use a rake task** to parse `mongoid/config.rb` ahead of time, scraping the documentation and building the template. This does nothing to address the fragility concern, though, because it's doing the exact same work, merely adding a new build step (which is itself prone to be forgotten).
- We could **store the documentation in a separate location**, as @johnnyshields suggested, and then use a rake task to generate `mongoid/config.rb` by combining the doc file with a separate code template. I don't like this approach either, though, because it adds a new build step that must be remembered every time the code template changes, or the docs change. Further, what happens if a developer forgets that `mongoid/config.rb` is autogenerated and edits it directly? A huge warning comment at the top of the file might help (e.g. "THIS IS AUTOGENERATED"), with instructions on where to go to find the template and the docs file, but I think that just adds a lot of complexity on top of something that really doesn't need to be complicated.
- We could **make the comments themselves executable**, like the `desc` directive in Rake tasks. Then, `Mongoid::Config` could be queried directly for all options and their descriptions. This sounds wonderful...except that it breaks our existing doc generator pipeline (which expects rdoc-style comments).
- We could **extract the `mongoid_config_options` method** to a new module, `Mongoid::Config::Introspection`. This would allow us to write more rigorous tests around the scraper, and would hopefully build more confidence in it.

I prefer the last option (`Mongoid::Config::Introspection`) since it better isolates the functionality, gives us an opportunity to write focused tests for it, and makes it clearer that there is a relationship between the format of the comments in `Mongoid::Config`, and the code that scrapes them.

johnnyshields: @jamis how about we stick with the existing manual approach for this PR, then split this refactor out into a new ticket/PR? I think you're absolutely right to consolidate the options (esp. so we can deprecate them easier), it just feels beyond the narrow scope of this particular change.

jamis: Good point, @johnnyshields -- the PR has evolved and I think you're right, the consolidation of the options in the yaml file is out of scope now. I'll split that out and treat it separately.

comandeo (approved):

PR #5531, opened 2023-01-25
Fix failing app_spec tests in evergreen

jamis:
When `rails new` is run during the app specs, it sets up a new Rails app with a default dependency on concurrent-ruby version 1.2.0. However, this version of concurrent-ruby is too new for Mongoid, which requires ">= 1.0.5, < 1.2". So when this test is run in evergreen, it fails because it can't resolve the dependency.

To fix this, I ensured the new Gemfile has an explicit dependency on the correct version of concurrent-ruby.

### DIFF
--- a/spec/integration/app_spec.rb
+++ b/spec/integration/app_spec.rb
@@ -272,7 +272,7 @@ describe 'Mongoid application tests' do
     gemfile_lines << "gem 'mongoid', path: '#{File.expand_path(BASE)}'\n"
     if rails_version
       gemfile_lines.delete_if do |line|
-        line =~ /rails/ || line =~ /concurrent_ruby/
+        line =~ /rails/ || line =~ /concurrent-ruby/
       end
       if rails_version == 'master'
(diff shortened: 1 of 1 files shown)


comandeo (commented):
There is something really strange happening indeed. It does not look like that there is an actual problem; when I create a new rails app, and add Mongoid as a dependency, everything works.

However, I believe that we should rather relax Mongiod's constraints for `concurrent-ruby`. I do not remember any reason for requiring a version below 1.2.

jamis:
Oh, ha, I just noticed that concurrent-ruby-1.2.0 was just released on the 23rd -- right at the beginning of the week, when I noticed these failing errors! So, yeah, maybe the right answer is to bump Mongoid's dependency. I'll check and make sure there's no issues with compatibility.

jamis:
PR for bumping concurrent-ruby is here: https://github.com/mongodb/mongoid/pull/5533 -- if that works out, we can close this one in favor of that.

jamis:
Closing this PR in favor of #5533.

PR #5533, opened 2023-01-26
Allow concurrent-ruby versions >= 1.2

jamis:
Mongoid was previously restricting concurrent-ruby versions to be < 1.2. This caused problems recently in evergreen with the release of concurrent-ruby 1.2.0. There do not appear to be any incompatibilities in 1.2.0, so it should be okay to be more permissive with which concurrent-ruby versions we support.


comandeo (approved):
:+1:

PR #5534, opened 2023-01-30
MONGOID-5504 Clear changes after :touch

comandeo:
While working on https://jira.mongodb.org/browse/MONGOID-5504, I realized that `touch` method leaves the object in the changes state:
```ruby
band = Band.create!
band.changed? # => false
band.update(name: 'John')
band.changed? # => false
band.touch
band.changed? # => true
band.changes # => {"updated_at"=>[2023-01-30 13:12:57.477191135 UTC, 2023-01-30 13:13:11.482975646 UTC]}
```

This leads to the problem described in https://jira.mongodb.org/browse/MONGOID-5504.

This behavior seems to be as designed - https://github.com/mongodb/mongoid/blob/master/spec/mongoid/touchable_spec.rb#L258. However, I believe that this behavior should be changed to be consistent with other updating operations.

### DIFF
--- a/docs/release-notes/mongoid-9.0.txt
+++ b/docs/release-notes/mongoid-9.0.txt
@@ -17,6 +17,30 @@ The complete list of releases is available `on GitHub
 please consult GitHub releases for detailed release notes and JIRA for
 the complete list of issues fixed in each release, including bug fixes.
 
+``touch`` method now clears changed state
+-----------------------------------------
+
+In Mongoid 8.x and older ``touch`` method leaves models in the changed state:
+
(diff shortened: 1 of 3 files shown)


jamis (approved):
This looks good.  It seems like the original behavior was from bd11b9145e, back in 2013, and appears to have been done so that callbacks can see the "changed" state. I think you've addressed that here, though (by calling `_clear_touch_updates` in the `ensure` block, after the callbacks run), so I suspect we're good. 👍

johnnyshields:
I agree with this change 👍

PR #5536, opened 2023-02-01
MONGOID-5539: Fix example in aggregation reference documentation

alexbevi:


jamis (approved):
Yeah, this looks good. It does make me wonder, though, if the `#find` method shouldn't be smart enough to try calling `#to_a` on any argument it doesn't otherwise understand.

johnnyshields:
> if the #find method shouldn't be smart enough to try calling #to_a on any argument it doesn't otherwise understand

I don't think it should cast to array. The typical use case of `#find` is to pass in a scalar id value. I used `#find` 1000s of times in my app and don't think I've every once used it with an array of IDs.

jamis:
It's not the primary use case, true. But the fact that it is a use case that is mentioned in the documentation, and which others have tripped over, suggests it might be worth smoothing the rough corners here. Certainly not a high priority, though.

comandeo (approved):

PR #5537, opened 2023-02-02
Use .where instead of .and in .any_of / .none_of

johnnyshields:
Since `.and` is a chaining operator, and `.any_of/.none_of` are explicitly non-chaining, it feels safer to use `.where` operator (non-chaining) instead of and.

There is probably no actual behavior difference here, but it's more for future-proofing.


johnnyshields:
🤦 `.and` is unfortunately needed because of the non-sensical way operator changing works since Mongoid 7.1. Closing this PR.

PR #5539, opened 2023-02-03
MONGOID-5555 Avoid duplicating option documentation

jamis:
The `mongoid.yml` template previously contained the same content as the comments in `Mongoid::Config`. This PR introduces an introspection mechanism that scrapes the `Mongoid::Config` module and extracts the comments for each of the options. The `mongoid.yml` template then uses that information to build the option list that users can choose from.

### DIFF
--- a/lib/mongoid/config/introspection.rb
+++ b/lib/mongoid/config/introspection.rb
@@ -1,0 +1,150 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Config
+
+    # This module provides a way to inspect not only the defined configuration
+    # settings and their defaults (which are available via
+    # `Mongoid::Config.settings`), but also the documentation about them. It
+    # does this by scraping the `mongoid/config.rb` file with a regular
+    # expression to match comments with options.
+    #
+    # @api private
+    module Introspection
+      extend self
+
+      # A helper class to represent an individual option, its name, its
+      # default value, and the comment that documents it.
+      class Option
+        # The name of this option.
+        #
+        # @return [ String ] The name of the option
+        attr_reader :name
+
+        # The default value of this option.
+        #
+        # @return [ Object ] The default value of the option, typically a
+        #   String, Symbol, nil, true, or false.
+        attr_reader :default
+
+        # The comment that describes this option, as scraped from
+        # mongoid/config.rb.
+        #
+        # @return [ String ] The (possibly multi-line) comment. Each line is
+        #   prefixed with the Ruby comment character ("#").
+        attr_reader :comment
+
+        # Instantiate an option from an array of Regex captures.
+        #
+        # @param [ Array<String> ] captures The array with the Regex captures
+        #   to use to instantiate the option. The element at index 1 must be
+        #   the comment, at index 2 must be the name, and at index 3 must be
+        #   the default value.
+        #
+        # @return [ Option ] The newly instantiated Option object.
+        def self.from_captures(captures)
+          new(captures[2], captures[3], captures[1])
+        end
+
+        # Create a new Option instance with the given name, default value,
+        # and comment.
+        #
+        # @param [ String ] name The option's name.
+        # @param [ String ] default The option's default value, as a String
+        #   representing the actual Ruby value.
+        # @param [ String ] comment The multi-line comment describing the
+        #   option.
+        def initialize(name, default, comment)
+          @name, @default, @comment = name, default, unindent(comment.strip)
+        end
+
+        # Indent the comment by the requested amount, optionally indenting the
+        # first line, as well.
+        #
+        # param [ Integer ] indent The number of spaces to indent each line
+        #   (Default: 2)
+        # param [ true | false ] indent_first_line Whether or not to indent
+        #   the first line of the comment (Default: false)
+        #
+        # @return [ String ] the reformatted comment
+        def indented_comment(indent: 2, indent_first_line: false)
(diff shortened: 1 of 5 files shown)


spec/integration/app_spec.rb
```diff
@@ -129,6 +129,15 @@ def start_app(cmd, port, timeout)
...
           config_text.should =~ /mongoid_test_config_test/
+
+          # deprecated options should not be included
+          config_text.should_not =~ /background_indexing/
+
+          # make sure the different option types are emitted
+          config_text.should =~ /# app_name: nil/
+          config_text.should =~ /# discriminator_key: "_type"/
+          config_text.should =~ /# join_contexts: false/
+          config_text.should =~ /# log_level: :info/
```
johnnyshields, line 140: I would be better to switch to expectation syntax everywhere:

```
expect(config_text).to match /.../
```

Note that .match? is faster than =~. (Mongoid would really benefit from Rubocop.)

johnnyshields:
My two cents: Purely in your tests (*not* in the /lib app code), I suggest to generate a copy of `mongoid.yml` as a test fixture, and assert that the generated one matches.

Of course, this test would break whenever the config code is modified, but the upside is that the config which is *actually released to users* will never break/change without the MongoDB team catching it first.

I think this strict testing approach is preferable to doing a few one off `line =~ /one line/` assertions. Otherwise, you'll be dealing with user tickets when things do inevitably break. Pick your poison 🧪

spec/integration/app_spec.rb
jamis: In this context, I'm not worried about optimizing for speed (I can't imagine `.match?` is fast enough here to make much difference), but it would be good to be consistent about which syntax we use in the tests. I'll update these to use expectation syntax.

jamis:
> My two cents: Purely in your tests (_not_ in the /lib app code), I suggest to generate a copy of `mongoid.yml` as a test fixture, then assert that the generated `mongoid.yml` matches this fixture exactly.

Thanks for the suggestion, @johnnyshields. I get what you're saying, and it's a good point. However, in practice, I've found matching file contents verbatim leads to challenges when trying to figure out what changed. Also, some tests explicitly check that the file does *not* contain a thing (e.g. checking that deprecated options are omitted), and just matching file contents loses the explicit intention behind those tests.

You're right, ultimately, that we need to pick our poison; for now, I'm opting to stick with the poison we've got.

johnnyshields:
If this test is not in place, would suggest that the team add a manual check of the generators on each release to ensure the generated file is correct.

> However, in practice, I've found matching file contents verbatim leads to challenges when trying to figure out what changed.

RSpec give you a line-by-line diff that's pretty easy to read.

spec/mongoid/config/introspection_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+
+    Mongoid::Config.defaults.each do |name, default_value|
+      context "for the `#{name}` option" do
+        let(:option) { all_options.detect { |opt| opt.name == name.to_s } }
+        let(:live_option) { options.detect { |opt| opt.name == name.to_s } }
+
+        it "should be parsed by the introspection scraper" do
+          expect(option).not_to be_nil
+          expect(eval(option.default)).to be == default_value
```
johnnyshields, line 44: `eval` even in tests should be avoided. Is there anyway this could be done with a regexp?

spec/integration/app_spec.rb
```diff
@@ -127,8 +127,17 @@ def start_app(cmd, port, timeout)
           File.exist?(mongoid_config_file).should be true
 
           config_text = File.read(mongoid_config_file)
-          config_text.should =~ /mongoid_test_config_development/
-          config_text.should =~ /mongoid_test_config_test/
+          expect(config_text).to match /mongoid_test_config_development/
+          expect(config_text).to match /mongoid_test_config_text/
```
johnnyshields, line 131: Would suggest to add a verbatim assertion here of the format of one full option that is not likely to change, e.g. 

```
text = <<~OPT
    # Mark belongs_to associations as required by default, so that saving a
    # model with a missing belongs_to association will trigger a validation
    # error. (default: true)
    # belongs_to_required_by_default: true
OPT
expect(config_text).to include(text)
```

jamis: That seems very reasonable. I'll take a closer look.

spec/mongoid/config/introspection_spec.rb
jamis: Yeah, that's fair. I think I can turn it around and compare `option.default` to to `default_value.inspect`. I'll experiment.

lib/mongoid/config/introspection.rb
```diff
@@ -0,0 +1,150 @@
...
+        # Create a new Option instance with the given name, default value,
+        # and comment.
+        #
+        # @param [ String ] name The option's name.
+        # @param [ String ] default The option's default value, as a String
+        #   representing the actual Ruby value.
+        # @param [ String ] comment The multi-line comment describing the
+        #   option.
+        def initialize(name, default, comment)
+          @name, @default, @comment = name, default, unindent(comment.strip)
```
comandeo, line 58: `unindent` method calls '#strip' on its argument. Should we do it here as well?

comandeo (approved):

lib/mongoid/config/introspection.rb
jamis: Good catch!

PR #5541, opened 2023-02-08
[READY FOR REVIEW] MONGOID-5556: #tally should support splatting array results

johnnyshields:
This uses the `$unwind` aggregation operator under the hood.

It turned out pretty nicely, it works for embedded and localized fields too.

Using this code I was able to run an array tally on a collection with 100,000,000 docs in 5 minutes 😎

### DIFF
--- a/lib/mongoid/contextual/memory.rb
+++ b/lib/mongoid/contextual/memory.rb
@@ -285,13 +285,23 @@ module Mongoid
       #   context.tally(:name)
       #
       # @param [ String | Symbol ] field Field to tally.
+      # @param [ Boolean ] :splat_arrays Whether to tally array
+      #   member values individually. Default false.
       #
       # @return [ Hash ] The hash of counts.
-      def tally(field)
-        return documents.each_with_object({}) do |d, acc|
+      def tally(field, splat_arrays: false)
+        documents.each_with_object({}) do |d, acc|
           v = retrieve_value_at_path(d, field)
-          acc[v] ||= 0
-          acc[v] += 1
+
+          if splat_arrays && v.is_a?(Array)
+            v.each do |vv|
+              acc[vv] ||= 0
+              acc[vv] += 1
+            end
+          else
+            acc[v] ||= 0
+            acc[v] += 1
+          end
         end
       end
 
--- a/lib/mongoid/contextual/memory.rb
+++ b/lib/mongoid/contextual/memory.rb
@@ -290,17 +290,17 @@ module Mongoid
       #
       # @return [ Hash ] The hash of counts.
       def tally(field, splat_arrays: false)
-        documents.each_with_object({}) do |d, acc|
-          v = retrieve_value_at_path(d, field)
+        documents.each_with_object({}) do |doc, tallies|
+          key = retrieve_value_at_path(doc, field)
 
-          if splat_arrays && v.is_a?(Array)
-            v.each do |vv|
-              acc[vv] ||= 0
-              acc[vv] += 1
+          if splat_arrays && value.is_a?(Array)
+            key.each do |array_value|
+              tallies[array_value] ||= 0
+              tallies[array_value] += 1
             end
           else
-            acc[v] ||= 0
-            acc[v] += 1
+            tallies[key] ||= 0
+            tallies[key] += 1
           end
         end
       end
--- a/lib/mongoid/contextual/memory.rb
+++ b/lib/mongoid/contextual/memory.rb
@@ -293,7 +293,7 @@ module Mongoid
         documents.each_with_object({}) do |doc, tallies|
           key = retrieve_value_at_path(doc, field)
 
-          if splat_arrays && value.is_a?(Array)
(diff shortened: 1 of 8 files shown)


lib/mongoid/contextual/memory.rb
```diff
@@ -285,13 +285,23 @@ def pick(*fields)
...
+      # @param [ Boolean ] :splat_arrays Whether to tally array
+      #   member values individually. Default false.
       #
       # @return [ Hash ] The hash of counts.
-      def tally(field)
-        return documents.each_with_object({}) do |d, acc|
-          v = retrieve_value_at_path(d, field)
-          acc[v] ||= 0
-          acc[v] += 1
+      def tally(field, splat_arrays: false)
```
johnnyshields, line 292: I'm open to calling this parameter `unwind`, which matches the MongoDB name, but I think splat_arrays may be more intuitive for Rubyists.

lib/mongoid/contextual/mongo.rb
```diff
@@ -465,23 +465,28 @@ def take!
...
+
+        # Must add a $project stage when using $unwind with nested fields
+        projected = 'p' if splat_arrays && (is_translation || name.include?('.'))
 
         fld = klass.traverse_association_tree(name)
-        pipeline = [ { "$group" => { _id: "$#{name}", counts: { "$sum": 1 } } } ]
-        pipeline.unshift("$match" => view.filter) unless view.filter.blank?
+        pipeline = []
+        pipeline << { "$match" => view.filter } if view.filter.present?
+        pipeline << { "$project" => { "#{projected}" => "$#{name}" } } if projected
```
alexbevi, line 482: Having a `$project` stage in the middle of a pipeline can negatively impact performance.

I would be extremely wary of introducing a change like this

alexbevi: See https://www.mongodb.com/docs/rapid/core/aggregation-pipeline-optimization/#-project-stage-placement

johnnyshields: @alexbevi if you don't use `$project`, the `$unwind` on nested fields will yield a null result. I have raised https://jira.mongodb.org/browse/SERVER-73888 for this because I believe it is a bug; other operators such as `$group` handle nested fields just fine.

as you can see in the code, the `$project` is only used in the case it needed:

- the `:splat_arrays` arg is true (note it is false by default) *AND*
- the field is a nested field (either `.` is present, or it is translation)

```ruby
        # Must add a $project stage when using $unwind with nested fields
        projected = 'p' if splat_arrays && (is_translation || name.include?('.'))
```

Note that the code in my PR does *not* apply the `$project` stage under default conditions, so there is no behavior change for any existing usage of this method.

Re: performance, fortunately I have a large production database we can try it and see what happens😄. Using it on a collection with 109 million+ records:

```
# non-nested array<string> field, does *not* use $project
Customer.tally('tags', splat_arrays: true)
#=> 481 seconds
#=> 761774 tally keys, total tallied value count 8031404

# nested field (array<document> -> string), uses $project
Customer.tally('addresses.city', splat_arrays: true)
#=> 592 seconds
#=> 65362 tally keys, total tallied value count 3284490
```

CPU, memory, disk util etc. were all equivalent & healthy on both throughout. Not apples-to-apples obviously, but clearly, nothing catastrophic happens when adding `$project` where it is needed due.
image

TLDR;
image

johnnyshields: One final note, the link you posted: https://www.mongodb.com/docs/rapid/core/aggregation-pipeline-optimization/#-project-stage-placement does not say putting a `$project` in the beginning will *hurt* performance, it says it will *not improve* performance. As per above, the reason we are using `$project` is nothing related to performance, it is required as per [SERVER-73888](https://jira.mongodb.org/browse/SERVER-73888).

alexbevi:
Closing this PR. We can revisit array tally functionality once SERVER-59713 is available as this would provide a huge performance boost over a client-side implementation.

johnnyshields:
@alexbevi The question here should not be about SERVER-59713, but rather: "does adding a $project stage in front of an $unwind _on the current MongoDB versions_ cause any performance regression?"

My benchmark data above clearly indicates that it does not, at least not any sort of hockey-stick or O(n^2) explosion, even when working with a 100 million document collection.

I feel the data I have provided is enough for you as PM to make a reasonable assessment that this is safe to go ahead with, but if you have concern, why not ask someone at MongoDB who is knowledgeable about the aggregation pipeline about the effects of putting a $project before an $unwind?

johnnyshields:
I'd also like to remind that the $project workaround is**ONLY applied for nested fields**, i.e. its a workaround for a corner-case. We're losing the forest for the trees here.

Would you be willing to merge this if I remove that workaround (i.e. drop support for nested fields), and then we add the workaround once additional homework is done?

johnnyshields:
Merged into [Mongoid Ultra](https://github.com/tablecheck/mongoid-ultra) :guitar:

PR #5544, opened 2023-02-16
MONGOID-5562 Remove BSON::Document#transform_keys hack

johnnyshields:
This hack can be removed as Mongoid requires Rails 5.1+ and the issue was fixed in Mongoid 5.0.

See: https://github.com/rails/rails/pull/24517

Should be fine to merge if tests pass.


jamis (approved):
Good catch, @johnnyshields. It always feels good to delete code!

PR #5547, opened 2023-02-16
MONGOID-5563 Code Docs: Correctly use Mongoid::Association::Relatable instead of Mongoid::Association

johnnyshields:
**This PR is code-doc changes only, it does not change any logic**

`Mongoid::Association` is poorly named. It is actually a **mixin** included in the `Mongoid::Document` class, which makes a document capable of being associated. (`Mongoid::Associable` would be a far better name, as the general practice is for mixins to end in `-able` suffix.)

The association metadata classes like `HasOne`, `EmbedsMany`, etc. do **not** inherit from `Mongoid::Association`. Instead, these `include Mongoid::Association::Relatable` mixin as their common base.

Therefore, where docs currently reference type `Association` (i.e. metadata) it is **incorrect**, we should replace it with `Mongoid::Association::Relatable`.

### DIFF
--- a/lib/mongoid/association/accessors.rb
+++ b/lib/mongoid/association/accessors.rb
@@ -17,7 +17,7 @@ module Mongoid
       #
       # @param [ String | Symbol ] name The name of the association.
       # @param [ Hash | BSON::ObjectId ] object The id or attributes to use.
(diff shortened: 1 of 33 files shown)


jamis (approved):
Good clarification!

PR #5551, opened 2023-02-18
[READY FOR REVIEW] MONGOID-5567 Remove "dead wood" Mongoid::Association::Referenced::NestedAttributes::One class

johnnyshields:
The class `Mongoid::Association::Referenced::NestedAttributes::One` is not used anywhere in the code, and moreover, attempting to load it will result in a `LoadError` because its supposed parent class `NestedBuilder` does not exist. Hence no deprecation needs to be done.

It looks like this class was forgotten to be removed in a refactor done ages ago.


johnnyshields:
Will be done in https://github.com/mongodb/mongoid/pull/5566

PR #5552, opened 2023-02-18
[READY FOR REVIEW] MONGOID-5571 Fix options arg in #destroy! being ignored + specs

johnnyshields:
The `#destroy!` (bang) method takes `options = {}` as an arg, but does not pass them to `#destroy` (non-bang). This looks unintentional.

The options here look like they are intended to be used internally, but they do exist on the public interface, so they should be documented and tested.

This PR adds spec coverage for `#delete`, `#destroy`, and `#destroy!` `:persist` and `:suppress` options.

### DIFF
--- a/spec/mongoid/persistable/destroyable_spec.rb
+++ b/spec/mongoid/persistable/destroyable_spec.rb
@@ -51,9 +51,7 @@ describe Mongoid::Persistable::Destroyable do
       end
 
       it 'deletes the matching document from the database' do
-        lambda do
-          person.reload
-        end.should raise_error(Mongoid::Errors::DocumentNotFound, /Document\(s\) not found for class Person with id\(s\)/)
+        expect { person.reload }.to raise_error(Mongoid::Errors::DocumentNotFound, /Document\(s\) not found for class Person with id\(s\)/)
       end
     end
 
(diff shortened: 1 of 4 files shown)


spec/mongoid/persistable/destroyable_spec.rb
```diff
@@ -297,6 +363,149 @@
...
+      let(:unsaved_person) { Person.new(id: person.id) }
+
+      before do
+        unsaved_person.destroy!
+      end
+
+      it 'deletes the matching document from the database' do
+        lambda do
+          person.reload
+        end.should raise_error(Mongoid::Errors::DocumentNotFound, /Document\(s\) not found for class Person with id\(s\)/)
```
jamis, line 377: For new code, we're preferring `expect { ... }.to raise_error` over the `should` form.

jamis (changes requested):
Good catch, @johnnyshields. If you can address the one issue I noted with the use of `should raise_error` in one of the tests, I think this will be fine to merge.

johnnyshields:
Fixed. Will raise another PR to remove all shoulds from all specs everywhere.

spec/mongoid/persistable/destroyable_spec.rb
johnnyshields: Fixed. Will raise another PR to remove all shoulds from all specs everywhere.

jamis (approved):

PR #5554, opened 2023-02-22
MONGOID-5016 default embedded touch to true

jamis:
This PR changes the default for the `embedded_in` relation from `touch: false` to `touch: true`, meaning that changing a subdocument will result in the parent being touched.

This resulted in a surprising number of broken tests, and most surprisingly, revealed a couple of bugs, as well, both of which have been addressed in this PR.

## Bug No. 1: The Memoized `_descendants` List

The first bug was a result of the `_descendants` list being memoized when first referenced. Given a system of models, and a series of operations like the following:

```ruby
class Person
  include Mongoid::Document
  include Mongoid::Timestamps

  field :name, type: String
  embeds_one :hometown
end

class Hometown
  include Mongoid::Document
  include Mongoid::Timestamps

  field :name, type: String
  embedded_in :person
end

bob = Person.create(name: "Bob")
home = bob.create_hometown(name: "Chicago")

bob.hometown = nil # <-- should destroy the hometown
bob.save!

p bob.hometown # <-- but it wasn't destroyed!
```
In the above sequence, with the default (and implicit) `touch: true` on the `embedded_in` relation, the hometown wouldn't actually be destroyed--calling `save!` would "resurrect" it. This was because of the `_descendants` list being initialized and memoized during the `create_hometown` call; when `#save!` was called later, that memoized list still had the newly created Hometown instance in it, which was marked as a new record as part of the substitution process, and it was therefore be _readded_ to the database!

The solution was to make sure to reset the memoized `_descendants` list after setting the hometown attribute to `nil`.

## Bug No. 2: `save` with `touch: false` doesn't override the implicit `touch: true`

When saving a subdocument, and explicitly specifying `save(touch: false)`, the parent document was being touched. Consider the following model system and operations:

```ruby
class Book
  include Mongoid::Document
  include Mongoid::Timestamps

  field :title, type: String

  # omitting `cascade_callbacks` revealed a related bug, where the touch callback was also being called; I'm
  # considering both scenarios under the same heading...
  embeds_many :covers, cascade_callbacks: true
end

class Cover
  include Mongoid::Document
  include Mongoid::Timestamps

  field :color, type: String
  embedded_in :book 
end

book = Book.create(covers: [ Cover.new(color: "blue") ])

sleep 1 # wait long enough for time-of-update to be different from time-of-create
cover = book.covers.first
cover.color = "green"

book.save!(touch: false) # should NOT push the touch up to the parent...

# ...but it does!
book.reload
puts "created at: #{book.created_at}"
puts "updated at: #{book.updated_at}"
```
The issue here is that the `timeless` scope was being reset as soon as the `updated_at` field was updated, and the touch callbacks were occurring *after* that. The solution was to add a small system for suppressing touch callbacks within a block, and then ensuring the entire update happens within such a block.

### DIFF
--- a/docs/release-notes/mongoid-9.0.txt
+++ b/docs/release-notes/mongoid-9.0.txt
@@ -191,6 +191,24 @@ persistence operation per parent document, even when using multiple
 levels of nested embedded documents.
 
 
+``embedded_in`` associations now default to ``touch: true``
+-----------------------------------------------------------
+
+Updating an embedded subdocument will now automatically touch the parent,
+unless you explicitly set ``touch: false`` on the relation:
+
+.. code-block:: ruby
+
+  class Address
+    include Mongoid::Document
+    include Mongoid::Timestamps
+
+    embedded_in :mall, touch: false
+  end
+
+For all other associations, the default remains ``touch: false``.
+
+
 Flipped default for ``:replace`` option in ``#upsert``
 ------------------------------------------------------
 
--- a/lib/mongoid/association/accessors.rb
+++ b/lib/mongoid/association/accessors.rb
@@ -341,12 +341,11 @@ module Mongoid
           klass.re_define_method("#{name}=") do |object|
             without_autobuild do
               if value = get_relation(name, association, object)
-                if value.respond_to?(:substitute)
-                  set_relation(name, value.substitute(object.substitutable))
-                else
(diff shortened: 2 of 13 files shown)


johnnyshields:
@jamis this is brilliant work! 🎉

comandeo (approved):
Great job! 👏

PR #5555, opened 2023-02-23
MONGOID-5565 Remove Ruby 2.6 from GH Actions

comandeo:
For some reason ruby 2.6 and jruby 9.3 configurations of github actions install an ancient version of actionpack (1.4). This situation is not reproducible neither on evergreen or locally. Therefore, it must be a GH actions issue.

Since we have coverage for these ruby versions on evergreen, it seems safe to remove them from GH actions.

This PR also reduces the amount of GH actions tasks and introduces mongodb 6.0.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,16 +49,6 @@ jobs:
           gemfile: Gemfile
           experimental: false
         - mongodb: '6.0'
-          ruby: jruby-9.4
-          topology: replica_set
-          os: ubuntu-20.04
-          task: test
-          driver: current
-          rails:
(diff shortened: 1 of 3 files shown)


jamis (approved):

johnnyshields:
How about dropping support for Ruby 2.6 (and Rails 5) altogether? Ruby 2.6 nearly 1 year in EOL, and Mongoid 7/8 are stable enough that users on older Ruby/Rails versions should have no problem using them.

https://endoflife.date/ruby

johnnyshields:
Please see #5565 for the correct fix.

PR #5556, opened 2023-02-23
MONGOID-5565 - Fix Github Actions CI for Ruby 2.6 and JRuby 9.3

johnnyshields:
This PR reverts the changes of https://github.com/mongodb/mongoid/pull/5555 and adds a proper fix. As @comandeo noted, the issue was than an old ActionPack version is installed, so the fix is:
- Use latest Bundler & Rubygems
- Correctly constrain the Gemfiles with `gem 'actionpack', '>= 5.1'`

Let's keep Github Actions in sync with Evergreen, so that community contributor PRs will be tested visibly against all supported Ruby versions. Remember, community members cannot see Evergreen run details, and Evergreen does not run automatically [because it insecure](EVG-16785).

If you run into trouble and would like me to help fix Github CI if it breaks in the future, please just ask before removing test coverage.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,9 +18,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - mongodb: '6.0'
+        - mongodb: '5.0'
           ruby: ruby-3.1
-          topology: replica_set
+          topology: server
           os: ubuntu-20.04
           task: test
           driver: current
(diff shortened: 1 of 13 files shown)


johnnyshields:
The JRuby failure above is due to timeout and not due to the ActionPack issue:
image

The timeout issue should be investigated separately.

alcaeus:
> Let's keep Github Actions in sync with Evergreen, so that community contributor PRs will be tested visibly against all supported Ruby versions.

I don't believe this is necessary. While we do want to have a number of CI builds available on GItHub Actions, I believe it is wasteful to run so many builds on two CI platforms testing the same thing. It wastes GitHub Actions resources that should be available for the many non-corporate-sponsored open source projects, it wastes a considerable amount of compute time and thus energy, and it wastes maintenance time by having the team maintain two different CI environments which each may break due to unrelated issues.

The set of builds that we run on GitHub actions should be considered a smoke test, able to spot obvious problems on the popular platforms. Everything beyond that is something that Evergreen will test and where the team will have to ensure that the build is stable - similar to how you are asking for us to investigate the JRuby failure separately. With that in mind, I'm closing this PR as we're covering Ruby 2.6 on Evergreen and will be dropping it in the near future anyways - the engineering time spent on fixing this build can be better spent on other things that provide value to the community.

johnnyshields:
>I believe it is wasteful to run so many builds on two CI platforms testing the same thing.

Andreas, I and others in the community *cannot see builds** on Evergreen. This is an open source project and should be treated as such. I will have no qualms if Evergreen results can be (a) auto-run and (b) publicly visible. Until that time, please keep full coverage of all builds on the public CI.

> we're covering Ruby 2.6 on Evergreen and will be dropping it in the near future anyways

Fine to drop Ruby 2.6 from CI once it's no longer supported.

> similar to how you are asking for us to investigate the JRuby failure separately

As stated above I am happy to contribute on the CI if the team lacks the expertise or bandwidth, please let me know.

PR #5553, opened 2023-02-22
MONGOID-5572 Rspec use expectation syntax

johnnyshields:
Fixes MONGOID-5572

Done using Transpec gem

PR #5564, opened 2023-03-03
MONGOID-5549 Fix app spec failures on JRuby

comandeo:


spec/integration/app_spec.rb
```diff
@@ -90,7 +90,12 @@ def start_app(cmd, port, timeout)
     end
 
     # Exit should be either success or SIGTERM
-    [0, 15, 128 + 15].should include(status)
+    allowed_statuses = [0, 15, 128 + 15]
+    if RUBY_PLATFORM == 'java'
+      # Puma on JRuby exits with status 1 when it receives a TERM signal.
+      allowed_statuses << 1
```
jamis, line 96: Oh, sneaky! Good catch. 👍

jamis (approved):

spec/integration/app_spec.rb
```diff
@@ -90,7 +90,12 @@ def start_app(cmd, port, timeout)
...
+    end
+    allowed_statuses.should include(status)
```
johnnyshields, line 98: Please remember to use RSpec expect syntax. It would be great if you can merge PR https://github.com/mongodb/mongoid/pull/5553 which does the switch everywhere.

PR #5489, opened 2022-10-01
MONGOID-5509: Deprecate all feature flags which were introduced in Mongoid 7.x

johnnyshields:
Also includes https://github.com/mongodb/mongoid/pull/5488 as there will be a conflict. Please merge that one first.

PR #5572, opened 2023-03-05
MONGOID-5509 Fix deprecation logic broken on Ruby <= 2.7 on 8.1-stable branch

johnnyshields:
This PR fixes deprecation warning logic to work correctly on Ruby 2.7 and earlier. PR #5489 didn't support it correctly for Ruby <= 2.7, and as a result Ruby 2.7/2.6 specs are failing on `8.1-stable` branch.

As you can see in the PR, there are differences in Ruby 2.7 and earlier handle `Module#prepend` on modules. I've intentionally done if/else for the different Ruby versions so it will be easy to cleanup in the future.

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -412,10 +412,25 @@ module Mongoid
                     object_id_as_json_oid
                     overwrite_chained_operators ]
 
-      OPTIONS.each do |option|
-        define_method(:"#{option}=") do |value|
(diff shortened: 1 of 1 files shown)


johnnyshields:
@jamis can you merge in this one? My "Deprecation" feature is actually broken on Ruby 2.6 and 2.7.

jamis:
Good catch, @johnnyshields -- thanks for the quick fix!

PR #5575, opened 2023-03-14
MONGOID-5585 Declare FLE schema in Mongoid document

comandeo:
This is the first PR in the series that adds support of CSFLE to Mongoid.

In this PR we implement only CSFLE schema generation from Mongoid documents. For more information about the schema syntax see https://www.mongodb.com/docs/manual/core/csfle/reference/encryption-schemas/

Schemas generated in the tests are for now manually tested with the Ruby driver. In the upcoming PRs we will implement configuring client(s) with the schema; it will allow us to add proper testing.

### DIFF
--- a/lib/mongoid/composable.rb
+++ b/lib/mongoid/composable.rb
@@ -2,6 +2,7 @@
 
 require "mongoid/changeable"
 require "mongoid/collection_configurable"
+require "mongoid/encryptable"
 require "mongoid/findable"
 require "mongoid/indexable"
 require "mongoid/inspectable"
@@ -59,6 +60,7 @@ module Mongoid
     include Interceptable
     include Copyable
     include Equality
+    include Encryptable
 
     MODULES = [
       Atomic,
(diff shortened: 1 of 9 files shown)


jamis (approved):

comandeo:
@johnnyshields thank you for your feedback!

> 1. It should be possible to add `:encrypt` to an existing field w/ existing data and have it encrypt just the records which are saved. Ideally one query should be able to return both encrypted and non-encrypted records. This would make migration/adoption of the feature a lot easier.

This is a very good point. We discussed this multiple times when preparing the epic, and we still do not have an agreement on how to implement this properly. But this is definitely on our radar.

> 2. It would be good to have a Mongoid global flag to disable encryption, even if `:encrypt` is defined on fields. This is useful for local and staging environments where you'd like to skip encryption, e.g. for portability reasons.

Created https://jira.mongodb.org/browse/MONGOID-5587 that includes your suggestion.

PR #5576, opened 2023-03-15
MONGOID-5582 Correctly interpret shard keys in embedded documents

jamis:
If a shard key referenced a field in an embedded document, Mongoid was incorrectly retrieving the value of that field when building the shard key selector. This PR properly parses and evaluates such shard keys.

See: https://jira.mongodb.org/browse/MONGOID-5582

### DIFF
--- a/lib/mongoid/shardable.rb
+++ b/lib/mongoid/shardable.rb
@@ -53,12 +53,10 @@ module Mongoid
     # @return [ Hash ] The shard key selector.
     #
     # @api private
-    def shard_key_selector
-      selector = {}
-      shard_key_fields.each do |field|
-        selector[field.to_s] = send(field)
+    def shard_key_selector(prefer_persisted: false)
+      shard_key_fields.each_with_object({}) do |field, selector|
+        selector[field.to_s] = shard_key_field_value(field.to_s, prefer_persisted: prefer_persisted)
       end
-      selector
     end
 
     # Returns the selector that would match the existing version of this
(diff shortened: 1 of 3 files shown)


lib/mongoid/shardable.rb
```diff
@@ -53,12 +53,10 @@ def shard_key_fields
     # @return [ Hash ] The shard key selector.
     #
     # @api private
-    def shard_key_selector
-      selector = {}
-      shard_key_fields.each do |field|
-        selector[field.to_s] = send(field)
+    def shard_key_selector(prefer_persisted: false)
```
johnnyshields, line 56: Maybe `use_persisted` instead of `prefer_persisted`? A preference to me would indicate a fallback situation: "use A if available, otherwise use B." But in this case, it setting the option to `true` looks like it deterministically always uses A instead of B.

spec/mongoid/shardable_spec.rb
```diff
@@ -41,163 +41,258 @@
...
           }
         end
 
         it 'uses foreign key as shard key in shard key fields' do
-          SmDriver.shard_key_fields.should == %i(age agency_id)
+          expect(SmDriver.shard_key_fields).to be == %i(age agency_id)
         end
       end
+
+      context 'when passed subdocument fields' do
```
comandeo, line 78: We probably want to remove this.

lib/mongoid/shardable.rb
comandeo, line 56: I suggest updating the documentation for this method, I think currently id does not describe correctly what the method is doing.

comandeo (approved):

lib/mongoid/shardable.rb
jamis: I guess, in my mind, `prefer_persisted` *does* indicate the fallback situation, more than `use_persisted` does. When `true`, it will *prefer* the persisted version of the field; when that isn't possible, it will fall back to the current version.

comandeo (approved):

PR #5580, opened 2023-03-16
MONGOID-5577 Restore previous behavior of #only/#without marking documents readonly

jamis:
Previous to MONGOID-5474, specifying a projection with `#only` or `#without` would implicitly mark the resulting documents readonly. When the explicit `readonly!` state was added, this implicit readonly state was lost. This PR makes sure documents produced with a projection are explicitly marked readonly.

See https://jira.mongodb.org/browse/MONGOID-5577

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ DONATE
 SPECS.txt
 .yardoc/*
 doc/*
+data/*
 yard-docs
 *.swp
 *.un~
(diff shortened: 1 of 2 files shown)


johnnyshields:
Test failures look related to this PR.

jamis:
Per discussion at https://jira.mongodb.org/browse/MONGOID-5577, it appears that Mongoid is working as designed and no modification is necessary in this regard. I'm therefore closing this PR.

PR #5583, opened 2023-03-23
MONGOID-5586 Make sure nested `with_scope` calls restore the previous scope

jamis:
It looks like this issue was indeed introduced in 7.3, as a result of refactoring `Mongoid::Fields::IDS` to `Mongoid::Fields#id_fields`. This refactoring relied on `Criteria#method_missing` to delegate the call to the `klass`, but that delegation involved wrapping the invocation in `with_scope`. The problem here is that from 7.3 until 8.0, there was a bug in `with_scope` that caused nested `with_scope` calls to misbehave, resetting the scope when the innermost invocation exited.

The fix is straightforward: essentially to port the basics of the bugfix from [MONGOID-5186](https://jira.mongodb.org/browse/MONGOID-5186) to 7.3-stable, 7.4-stable, and 7.5-stable.

This PR handles the fix for the 7.3-stable branch.

ref: https://jira.mongodb.org/browse/MONGOID-5586

### DIFF
--- a/lib/mongoid/scopable.rb
+++ b/lib/mongoid/scopable.rb
@@ -226,11 +226,12 @@ module Mongoid
       #
       # @since 1.0.0
       def with_scope(criteria)
+        previous = Threaded.current_scope(self)
         Threaded.set_current_scope(criteria, self)
         begin
           yield criteria
         ensure
-          Threaded.set_current_scope(nil, self)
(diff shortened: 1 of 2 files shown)


jamis:
Wups, ignore this PR. It looks like the issue was "fixed" in 7.4 and 7.5 with a feature flag, but was never ported further back to 7.3.

PR #5584, opened 2023-03-24
MONGOID-5520 allow Hash to act like a sort direction

jamis:
When attempting to specify `score: { "$meta" => "textScore" }` as a sort direction for a text search query, Mongoid raises an exception because it does not know how to convert a Hash instance into a sort direction.

ref: https://jira.mongodb.org/browse/MONGOID-5520


johnnyshields:
The feature request here is fine, but implemention-wise we should be getting rid of monkey patches in Mongoid, not adding them.

It would be much better to refactor the other methods so that `to_direction` is either removed or deprecated from Numeric, etc.

alexbevi:
I like the translator approach

PR #5588, opened 2023-03-28
MONGOID-5590 Update sharding docs re: keys with "." characters in them

jamis:
MONGOID-5582 updated Mongoid so that shard keys with "." in them are treated as delimited fields identifying attributes of embedded subdocuments. This precludes the possibility of using shard keys that literally contain the "." character. Since we've decided this design decision is acceptable, this PR updates the sharding documentation for Mongoid to mention this caveat.

Once accepted, I'll backport this PR to 7.5-stable, 8.0-stable, and 8.1-stable (which were also updated for MONGOID-5582).

ref: https://jira.mongodb.org/browse/MONGOID-5590

### DIFF
--- a/docs/reference/sharding.txt
+++ b/docs/reference/sharding.txt
@@ -83,6 +83,21 @@ configured in the association as the field name:
     index country: 1
   end
 
+The shard key may also reference a field in an embedded document, by using
+the "." character to delimit the field names:
+
+.. code-block:: ruby
+
+  shard_key "location.x" => 1, "location.y" => 1
+
+  shard_key "location.x", "location.y"
+
+.. note::
+
+  Because the "." character is used to delimit fields in embedded documents,
+  Mongoid does not currently support shard key fields that themselves
+  literally contain the "." character.
+
 .. note::
 
   If a model declares a shard key, Mongoid expects the respective collection


comandeo (approved):

johnnyshields:
In your document:

> Mongoid does not currently support shard key fields that themselves literally contain the "." character.

This should be amended to say that MongoDB itself does not support '.' (or leading '$') in shard key names. Refer to: https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/

johnnyshields:
Does MongoDB allow `.` in shard key names? This page says that leading `$` is not allowed, but is ambiguous about `.`
https://www.mongodb.com/docs/v5.2/core/dot-dollar-considerations/

jamis:
I've not tried it, @johnnyshields, but my reading of that page suggests that only dollar-prefixed names are forbidden as shard key names. Either way, this ticket is concerned specifically with Mongoid, and we've made the decision to disallow "." characters in shard key names there.

johnnyshields:
Hmmm... wouldn't it make sense to check it? Are there tests which cover the documented behavior here? At first glance I couldn't find any tests that have a shard key containing `.`

Sharding is pretty dangerous functionality if it's done wrong, so it makes sense to be extra cautious here esp. if Mongoid docs advertising a certain behavior.

johnnyshields:
Ah oops sorry I see you added the tests in [MONGOID-5582](https://jira.mongodb.org/browse/MONGOID-5582)

Looking at the API of the server, if `.` is interpreted to mean embedded fields as per your tests, then it would seem that it would be impossible to use a literal dot in the field name. Server team should clarify this.

PR #5594, opened 2023-04-07
MONGOID-5595 Forward explain options to driver

naveg:
This allows callers to adjust any available `explain` options, as documented here:

https://www.mongodb.com/docs/ruby-driver/master/api/Mongo/Collection/View/Explainable.html

### DIFF
--- a/lib/mongoid/contextual/mongo.rb
+++ b/lib/mongoid/contextual/mongo.rb
@@ -38,6 +38,16 @@ module Mongoid
       # @attribute [r] view The Mongo collection view.
       attr_reader :view
 
+      # Run an explain on the criteria.
+      #
+      # @example Explain the criteria.
+      #   Band.where(name: "Depeche Mode").explain
+      #
+      # @param [ Hash ] options customizable options (See Mongo::Collection::View::Explainable)
+      #
+      # @return [ Hash ] The explain result.
+      def_delegator :view, :explain
+
       attr_reader :documents_loader
 
       # Get the number of documents matching the query.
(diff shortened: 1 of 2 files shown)


johnnyshields:
@naveg can you add specs which confirm the options are passed to the driver? Ideally should make an assertion on the output of explain.

lib/mongoid/contextual/mongo.rb
```diff
@@ -182,9 +182,11 @@ def exists?(id_or_conditions = :none)
       # @example Explain the criteria.
       #   Band.where(name: "Depeche Mode").explain
       #
+      # @param [ Hash ] options customizable options (See Mongo::Collection::View::Explainable)
+      #
       # @return [ Hash ] The explain result.
-      def explain
-        view.explain
+      def explain(options = {})
+        view.explain(**options)
```
johnnyshields, line 189: Should just use:

```ruby
delegate :explain, to: :view
```

naveg: since this class has `extend Forwardable`, it would need to be `def_delgator :view, :explain`

that work for you?

naveg: updated but happy to change approach

johnnyshields: I’d say leave in the code comment like you had it above the def_delegator. Would be good to add spec demonstrating the options can be passed thru as well.

Im not a maintainer of this repo FYI.

naveg: good call - done

anything else I should do to help get this merged?

johnnyshields: Looks good to me lets see what the MongoDB team says.

alexbevi:
Hi @naveg, thanks for submitting this PR. We'll review it shortly and get back to you, but based on a cursory review this looks to be in pretty good shape :)

comandeo:
@naveg Thank you for your contribution!

PR #5601, opened 2023-04-12
MONGOID-5554 Expand arrays in queries

comandeo:
Even though [the ticket](https://jira.mongodb.org/browse/MONGOID-5554) that initiates this PR describes a particular issue, the solution turned out to be global.

ActiveRecord expands arrays containing one element in queries:
```ruby
Person.where(id: [1])
#  Person Load (0.3ms)  SELECT "people".* FROM "people" WHERE "people"."id" = ?  [["id", 1]]
Person.where(id: [1, 2])
#  Person Load (0.5ms)  SELECT "people".* FROM "people" WHERE "people"."id" IN (?, ?)  [["id", 1], ["id", 2]]
```

Even though it seems to be a performance optimisation, It turned out that some Rails functionality depends on this behaviour. For example, `to_key` method returns an array of a single element (in case of a simple primary key). Therefore, the following find the document in ActiveRecord, but not in Mongoid:
```ruby
Person.where(id: person.to_key)
```

Even though this performance optimisation is not necessary in Mongoid (server expands arrays of one element automatically), we decided to copy the Rails behaviour. It ensures that third party gems that relies on this functionality of ActiveRecord work in Mongoid, too.

### DIFF
--- a/docs/release-notes/mongoid-9.0.txt
+++ b/docs/release-notes/mongoid-9.0.txt
@@ -17,6 +17,31 @@ The complete list of releases is available `on GitHub
 please consult GitHub releases for detailed release notes and JIRA for
 the complete list of issues fixed in each release, including bug fixes.
 
+Arrays containing a single element are now expanded in queries
+--------------------------------------------------------------
+
+**Breaking change** In Mongoid 8.1 and older if a query includes an array as
+a value, it is used in the criteria without changes:
+
+.. code-block:: ruby
+
+  # Mongoid 8.x behaviour
+  Person.where(name: ['Kate']).selector # { 'name' => [ 'Kate' ] }
+  Person.where(name: ['Kate', 'Anna']).selector # { 'name' => [ 'Kate', 'Anna' ] }
+
+In Mongoid 9.0 we change this behavior to bring it in accordance with
+ActiveRecord:
+
+.. code-block:: ruby
+
+  # Mongoid 9.x behaviour
+  Person.where(name: ['Kate']).selector # { 'name' => 'Kate' }
+  Person.where(name: ['Kate', 'Anna']).selector # { 'name' => [ 'Kate', 'Anna' ] }
+
+You can return to the old behavior by setting
+``expand_single_element_arrays_in_query`` to ``false`` in your
+``config/mongoid.yml`` file.
+
 Deprecated options removed
 --------------------------
 
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -109,6 +109,14 @@ module Mongoid
     # document might be ignored, or it might work, depending on the situation.
     option :immutable_ids, default: true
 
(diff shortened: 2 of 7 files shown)


jamis (approved):

johnnyshields:
**No, please DON'T do this.**

We already have `any_in(id: [1, 2])` method in Mongoid and the `$in` operate to do the behavior you are suggesting in this PR.

We need to be able to use `where` to query for Array literal values, and this behavior is in wide use today. For example:

```ruby
class Person
  include Mongoid::Document

  field :lucky_numbers, type: Array
end

bob = Person.create!(lucky_numbers: [7, 13])
jane = Person.create!(lucky_numbers: [7, 13, 19])
joe = Person.create!(lucky_numbers: [7])
jim= Person.create!(lucky_numbers: [9, 10])

Person.where(lucky_numbers: [7, 13]).to_a #=> returns only bob
```

**MQL is not SQL.**

johnnyshields:
Also, this implementation is *far beyond the scope* of the [original ticket](https://jira.mongodb.org/browse/MONGOID-5554). I'm sorry, but it's completely unacceptable that changes to a fundamental method like `.where` would be made in a haphazard manner without a proper JIRA ticket to thoroughly consider the change and its potential side effects.

comandeo:
Hi @johnnyshields, thank you for the technical aspect of your feedback! I agree that this is a very controversial change, and I anticipated some feedback here. Since our JIRA currently requires additional authentication, GitHub is the best place to gather feedback.

I'm not a big fan of the solution I came up with here. The reason why we even considered such a change is that there is an expectation for Mongoid to be as much of a drop-in replacement for AR as possible. As you quite rightly pointed out multiple times in our previous discussions, third-party gems written for AR should work with Mongoid. Being able to handle the to_key method in queries is the kind of thing expected to work out of the box. So the question is whether the cost of supporting compatibility is too high. **I agree with you that in this case, the undesirable side effects outweigh the potential benefits.** At the very least, this behavior should never be enabled by default.

Regarding patching Rails – the feedback you received is quite similar to what we [used to get](https://github.com/rails/rails/pull/33981#issuecomment-424824766).

As for the suggested patch to Devise – to_key may return a meaningful array in the case of compound primary keys. Therefore, .find_by(id: obj.id) is not equal to .find_by(id: obj.to_key).

johnnyshields:
@comandeo it seems like Rails is moving in the direction of supporting [composite primary keys](https://rubyonrails.org/2023/4/14/this-week-in-rails-a-week-dedicated-to-composite-primary-keys-32298f96) as we speak, so that seems to be the motivation for `.to_key` returning an Array.

Still, `find_by(_id: resource.to_key)` is non-sensical, because a composite key would need to be something like `find_by(_id1: resource.to_key.first, _id2: resource.to_key.second)`. Mongoid should look at what Rails is doing here, because likely Devise will need to change it's code anyway to support composite keys.

PR #5602, opened 2023-04-12
MONGOID-5592 Allow key vault client configuration

comandeo:


spec/mongoid/clients/factory_spec.rb
```diff
@@ -282,12 +265,60 @@
...
+                }
+              }
+            }
+          end
+
+          it "returns a client" do
+            expect(client).to be_a(Mongo::Client)
+          end
+
+          it 'sets schema_map for the client' do
```
jamis, line 319: Is this spec title supposed to say `key_vault_client` instead of `schema_map`?

jamis (approved):
Looks good! Just one question re: the title of a spec, possibly copy/paste error.

spec/mongoid/clients/factory_spec.rb
comandeo: Good catch, thank you! Fixed.

jamis (approved):

PR #5603, opened 2023-04-13
MONGOID-5590 - Clarify that shard key dot limitation comes from MongoDB server

johnnyshields:
As per [DOCS-15993](https://jira.mongodb.org/browse/DOCS-15993), the limitation of dot in shard keys comes from MongoDB server, not from Mongoid.

### DIFF
--- a/docs/reference/sharding.txt
+++ b/docs/reference/sharding.txt
@@ -84,7 +84,7 @@ configured in the association as the field name:
   end
 
 The shard key may also reference a field in an embedded document, by using
-the "." character to delimit the field names:
+the ``.`` (dot) character to delimit the field names:
 
 .. code-block:: ruby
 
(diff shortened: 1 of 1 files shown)


johnnyshields:
@jamis ready to merge.

johnnyshields:
@jamis any reason why this doc-only PR can't be merged?

jamis:
@johnnyshields -- I think this might not be true anymore. For v6 of the database, I think you're right (https://www.mongodb.com/docs/v6.0/core/dot-dollar-considerations/), but the current docs (https://www.mongodb.com/docs/manual/core/dot-dollar-considerations) make no mention of that restriction. The page you're linking to in the docs seems to flatly contradict the statement that dots and dollar signs are not supported.

If you can find another page for the current server version (v7) that says that dots and dollar signs are forbidden in shard keys at the server level, please do update the PR. Otherwise, I think this PR might be obsolete.

johnnyshields:
@jamis can you check DOCS-15993. It seems someone closed it, but they didn't clearly state what action they took (at least I can't see it, it might be MongoDB internal only)

I see that dollar fields are definitely not allowed as shard keys:
https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/dollar-prefix/
image

However, the dots page does not say either way.
https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/periods/#std-label-period-field-names

johnnyshields:
Moved to https://github.com/mongodb/docs-mongoid/pull/123

PR #5604, opened 2023-04-13
MONGOID-5599 - Fix broken tests for sharded clusters

johnnyshields:
This PR fixes specs for sharded clusters. The gist is that the "expect_queries" test macro doesn't work with shards, or tends to fail sporadically in some cases.

Strangely, these specs seem to pass on Evergreen. I have no visibility as to what's going on there, but I am skeptical that they would pass with a proper multi-shard cluster setup. You may want to check if Evergreen is actually configured properly.

### DIFF
--- a/spec/support/expectations.rb
+++ b/spec/support/expectations.rb
@@ -4,15 +4,14 @@ module Mongoid
   module Expectations
 
     def connection_class
-      if defined?(Mongo::Server::ConnectionBase)
-        Mongo::Server::ConnectionBase
-      else
-        # Pre-2.8 drivers
-        Mongo::Server::Connection
-      end
+      Mongo::Server::ConnectionBase
     end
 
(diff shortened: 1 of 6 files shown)


spec/support/expectations.rb
```diff
@@ -4,15 +4,14 @@ module Mongoid
   module Expectations
 
     def connection_class
-      if defined?(Mongo::Server::ConnectionBase)
-        Mongo::Server::ConnectionBase
-      else
-        # Pre-2.8 drivers
-        Mongo::Server::Connection
-      end
+      Mongo::Server::ConnectionBase
```
johnnyshields, line 7: Pre-2.8 drivers versions are no longer used as per gemspec.

alexbevi:
Closing this out as the evergreen matrix for sharded tests doesn't indicate this problem

PR #5606, opened 2023-04-13
MONGOID-5607: Rescue StandardError instead of Exception

johnnyshields:
Should never rescue Exception because its used for things like sigkill.


alexbevi:
Seems reasonable (based on ["Rescue StandardError, Not Exception"](https://thoughtbot.com/blog/rescue-standarderror-not-exception) at least)

jamis:
Thanks @johnnyshields! Good catch.

PR #5608, opened 2023-04-18
Enable all cops, but explicitly opt each file out

jamis:
Adding Rubocop to a mature project requires that you either drop everything and immediately bring *all* of your code into compliance with the mandated coding style, or you find a way to do it piecemeal. This PR presents one way to do it a bit at a time, instead of all at once.

It works by:
1. Enabling all cops in `.rubocop.yml`. 
2. Adding `# rubocop:todo all` at the top of every Ruby file in the project. This effectively opts each file out of the linter.

At this point, running `rubocop` will pass. Yay! 🎉 

From this point on, every time we change something in a file, we should also remove the `# rubocop:todo all` comment at the top, and address the linter violations that result.

My primary concern with this (or with any piecemeal approach) is that we risk overwhelming our PRs with linter-appeasement. We could address this by having two PRs for each task: the first one addresses linter-related changes, and the second one addresses the actual changes needed for the task itself. Eventually, as more and more files opt into rubocop, this will become less of an issue.


.rubocop.yml
```diff
@@ -10,46 +8,48 @@ AllCops:
   TargetRubyVersion: 2.6
   NewCops: enable
   Exclude:
-    - 'spec/shared/lib/mrss/**/*'
+    - 'spec/shared/**/*'
```
comandeo, line 11: I am not sure whether it makes sense to enable the linter for specs. Let's see how it goes, though.

comandeo (approved):
I like this approach a lot, great job!

I have just one remark regarding the following suggestion:

> From this point on, every time we change something in a file, we should also remove the # rubocop:todo all comment at the top, and address the linter violations that result.

Let's do style changes is separate commits. Being able to track history of changes is very important, especially when investigating some weird bug. If we use the suggested approach, there will be style changes with commit messages like `MONGOID-XYZ Add this feature`. I would try to avoid that.

.rubocop.yml
jamis: I think we should, but yeah, we can see how it goes and discuss further. This highlighted change, though, just excludes the spec/shared folder explicitly because its a submodule, and should be linted elsewhere (if at all).

jamis:
> Let's do style changes is separate commits. Being able to track history of changes is very important, especially when investigating some weird bug. If we use the suggested approach, there will be style changes with commit messages like `MONGOID-XYZ Add this feature`. I would try to avoid that.

100% agree. I was going to propose this as part of this PR, but figured we could discuss procedure separately; I'm glad we agree on this, that makes it easy. :)

PR #5610, opened 2023-04-20
Bug Fix: Fix using exists? with args on relations.

johnnyshields:
Fix using `exists?` with args on relations. For example:

```
person.posts.exists?({ title: "50 Ways to Leave Your Lover" })
```

This was missed in the original implementation [MONGOID-5100](https://jira.mongodb.org/browse/MONGOID-5100)

Also cleans up method docs which were not accurate.


johnnyshields:
Need to look at this issue in more detail. There is possibly a need to differentiate between `exists?` (meaning persisted) and `any? / blank? present?` for embedded models, which I think should include non-persisted.

PR #5611, opened 2023-04-24
MONGOID-5588 Add automatic FLE documentation

comandeo:


johnnyshields:
The word "Automatic" seems like a bit of a misnomer. It implies that Mongoid will automatically know which fields to encrypt.

What about calling it "Schema-based Field Level Encryption" instead, because it requires the user to define a schema in advance?

docs/tutorials/automatic-encryption.txt
```diff
@@ -0,0 +1,211 @@
...
+    }
+    # This field is encrypted using AEAD_AES_256_CBC_HMAC_SHA_512-Random
+    # algorithm.
+    field :blood_type, type: String, encrypt: {
+        deterministic: false
+    }
+    # This field is encrypted using AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic
+    # algorithm and using  a different data key.
+    field :ssn, type: Integer, encrypt: {
+        deterministic: true, data_key_id: 'Vxr5m+5cQISjDOruzZgE0w=='
```
johnnyshields, line 161:
```
encrypt_with key_id: ___

vs

field :ssn, type: Integer, encrypt: { deterministic: true, data_key_id: ___
```

It would be good to use either `key_id` or `data_key_id` consistently.

johnnyshields:
Several questions I have here, which should be part of the tutorial:

1. How do I migrate an existing collection with unencrypted data to use encryption?

2. How do I query for encrypted data in various fields? What sorts of queries are supported--only exact match? How about different data types (String, Integer, etc.)

3. If I want to encrypt a string or decrypt a BSON::Binary value, how can I do that? In my existing gem I can do `Mongoid::EncryptedString.encrypt("foobar") / .decrypt("____")`

docs/tutorials/automatic-encryption.txt
comandeo: Wow, I am wondering how I could missed that. Thank you!

alexbevi:
@comandeo first pass completed, let me know what you think

jamis (approved):

alexbevi (approved):

PR #5614, opened 2023-05-05
MONGOID-5613 Fix encryption schema generation

comandeo:


lib/mongoid/clients/factory.rb
```diff
@@ -57,6 +57,9 @@ def create_client(configuration)
         config = configuration.dup
         uri = config.delete(:uri)
         database = config.delete(:database)
+        if database.nil?
+          database = Mongo::URI.get(uri).database
+        end
```
jamis, line 62: Is it worth condensing these four lines into one?

```suggestion
        database = config.delete(:database) || Mongo::URI.get(uri).database
```

jamis (approved):
Nice! Just one possible tweak to consider, but it's a nitpick. 👍

lib/mongoid/clients/factory.rb
comandeo: Good point, thank you!

PR #5615, opened 2023-05-05
MONGOID-5614 Document ffi dependency

comandeo:


docs/tutorials/automatic-encryption.txt
```diff
@@ -97,6 +98,17 @@ need to be installed manually. ``mongocryptd`` comes pre-packaged with
 enterprise builds of the MongoDB server (versions 4.2 and newer).
 For installation instructions, see the `MongoDB manual <https://mongodb.com/docs/manual/reference/security-client-side-encryption-appendix/#installation>`_.
 
+Add ``ffi`` to your Gemfile
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Mongo Ruby driver uses `ffi gem <https://github.com/ffi/ffi>`_ to call
```
jamis, line 104: Looks like a stray underscore character here. Also, probably only need backticks around the word "ffi" itself.

Also, grammatically: I'd insert "the" before "Mongo" and after "uses" (e.g. "**The** Mongo Ruby driver uses **the** ffi gem...").

jamis (approved):
Just a formatting issue, and a grammar issue, but looks good otherwise!

alexbevi (approved):

docs/tutorials/automatic-encryption.txt
jamis: Ah! Right. I keep forgetting about that.

comandeo: Thank you for the cleanup!

PR #5617, opened 2023-05-09
MONGOID-5600 make sure build calls initialize

jamis:
When MONGOID-2586 landed, it made it so that a model's `initialize` method would not be called in certain circumstances, including when calling `build` on a child collection. This PR corrects that specific case (though there are other circumstances where `initialize` is still not called, typically via the `Factory` methods).

ref https://jira.mongodb.org/browse/MONGOID-5600

### DIFF
--- a/lib/mongoid/document.rb
+++ b/lib/mongoid/document.rb
@@ -288,8 +288,15 @@ module Mongoid
 
     module ClassMethods
 
-      def suppress_callbacks
-        saved, Threaded.execute_callbacks = Threaded.execute_callbacks?, false
+      # Suppress callbacks (by default) for documents within the associated
+      # block. Callbacks may still be explicitly invoked by passing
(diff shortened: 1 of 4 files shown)


lib/mongoid/document.rb
```diff
@@ -102,7 +102,7 @@ def identity
     #
     # @return [ Document ] A new document.
     def initialize(attrs = nil, &block)
-      construct_document(attrs, execute_callbacks: true, &block)
+      construct_document(attrs, execute_callbacks: Threaded.execute_callbacks?, &block)
```
sos4nt, line 105: Since the method uses this as the default value, you can just omit the argument:
```suggestion
      construct_document(attrs, &block)
```

jamis: Good catch! Thanks 👍

comandeo (changes requested):

PR #5627, opened 2023-05-17
MONGOID-5514: Improve forking guidance

alexbevi:
Update forking guidance examples to test if `Mongoid` is `defined?` to ensure these examples don't fail when run in environments where Mongoid may not have been preloaded (ex: development environments)

### DIFF
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -822,7 +822,7 @@ the ``before_fork`` hook to close clients in the parent process
         client.reconnect
       end
     else
-      raise "Mongoid is not loaded. You might forget to enable app preloading."
+      raise "Mongoid is not loaded. You may have forgotten to enable app preloading."
     end
   end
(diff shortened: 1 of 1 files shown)


jamis (approved):
Makes me wish Ruby had a safe operator for defined/undefined constants. :)

johnnyshields:
I'm not sure that this is needed because it may just be obscuring a bigger problem that Mongoid is not pre-loaded before forking. In general there is little benefit to forking if you aren't pre-loading; you fork because you want to take advantage of copy-on-write (COW) memory sharing.

(I did encounter such a `constant Mongoid not defined` issue recently while upgrading to Puma 6 but ultimately fixed it by enabling pre-loading)

comandeo (approved):

alexbevi:
@comandeo I just modified the error wording slightly. Will merge once checks complete.

PR #5631, opened 2023-05-30
Don't allow ruby to coerce field types when evaluating _matches?

cameronmoreau:


cameronmoreau:
@johnnyshields here's some context: https://jira.mongodb.org/browse/MONGOID-5619

johnnyshields:
@cameronmoreau it seems the main case you are taking issue with is implicit Time to String comparison:

```
Time.current > '2012-01-01' => true
Time.current < '2012-01-01' => false

# Note Date class doesn't do this
Date.today > '2012-01-01'
in `>': comparison of Date with String failed (ArgumentError)
```

(To be very precise, this is actually due to a monkey patch from Rails' ActiveSupport library, as just using Time class without AS will result in:)

```
Time.now < '2012-01-01'
in `<' ArgumentError (comparison of Time with String failed)
```

None of the other cases (e.g. String vs. Numeric) etc. behave differently than the database because *neither Ruby nor AS coerces them*.

Moreover, Mongoid actually does allow doing DB queries such as `User.gt(created_at: "2023-01-01")` and will coerce to the field type (i.e. inside Mongoid). The only cases where the behavior diverges is if you use undefined field types such as `Object`, `Hash`, `Array` field type and do a query on the members--these should be considered edge cases and not recommended to be avoided.

It is my opinion the current behavior is actually desirable here--it is a reasonable convenience that a time string is coerced to Time. In several other tickets such as [MONGOID-4500](https://jira.mongodb.org/browse/MONGOID-4500) we have establish that the "in-memory" behavior *should* diverge from the database where it provides additional convenience to the user.

Lastly, this patch is likely to break existing apps.

TLDR; my strong recommendation is that this PR be closed.

johnnyshields:
@jamis as per above comment I recommend this PR be closed.

jamis:
As discussed in [MONGOID-5619](https://jira.mongodb.org/browse/MONGOID-5619), we're closing this as "Won't Fix", due to the potential to break existing apps.

PR #5632, opened 2023-06-02
MONGOID-5560 delete_one on embeds_many is callable and is not consistent with delete

jamis:
This PR updates the method delegation for the `has_many`, `has_and_belongs_to_many`, and `embeds_many` proxies to prevent them from responding to the `delete_one` method.

Mongoid defines an Array extension method, `delete_one`, which is intended to remove only the first matching element. (Contrast this with `Array#delete`, which removes all matching elements, and the proxy `#delete` methods, which remove the given element _and persist the change_.) The problem is that this method is exposed on the association proxies via method delegation, and has been confusing to users. The method only removes the element from the in-memory collection; it does not persist changes to the database, and in virtually every case the caller _probably_ intends to call `delete` instead of `delete_one`.

I believe it is safe to remove access to this method without putting the change behind a feature flag. The `delete_one` 
method was never explicitly intended to be exposed on the proxies, but most importantly, it's not part of the public API. Anyone who has been using `delete_one` via an association proxy has been relying on a private API.

### DIFF
--- a/lib/mongoid/association/embedded/embeds_many/proxy.rb
+++ b/lib/mongoid/association/embedded/embeds_many/proxy.rb
@@ -437,6 +437,7 @@ module Mongoid
           #
           # @return [ Criteria | Object ] A Criteria or return value from the target.
           ruby2_keywords def method_missing(name, *args, &block)
+            enforce_forwarding_list!(name)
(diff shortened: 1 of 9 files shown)


jamis:
I'm closing this PR in favor of #5633.

PR #5633, opened 2023-06-06
MONGOID-5560 alias delete_one to delete for `Many` proxies

jamis:
This PR is an alternative to #5632, which attempted to solve the issue presented in MONGOID-5560 by removing the `delete_one` method from the `Many` proxy subclasses.

This PR takes a simpler, more intuitive approach: treat `delete_one` as an alias for `delete`.

ref: https://jira.mongodb.org/browse/MONGOID-5560

### DIFF
--- a/lib/mongoid/association/embedded/embeds_many/proxy.rb
+++ b/lib/mongoid/association/embedded/embeds_many/proxy.rb
@@ -155,6 +155,10 @@ module Mongoid
             end
           end
 
+          # Mongoid::Extensions::Array defines Array#delete_one, so we need
+          # to make sure that method behaves reasonably on proxies, too.
+          alias delete_one delete
+
           # Delete all the documents in the association without running callbacks.
           #
           # @example Delete all documents from the association.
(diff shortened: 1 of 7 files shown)


lib/mongoid/association/embedded/embeds_many/proxy.rb
```diff
@@ -155,6 +155,23 @@ def delete(document)
...
+          # to make sure that method behaves reasonably on proxies, too.
+          alias delete_one delete
+
+          # Removes a single document from the collection *in memory only*.
+          # It will *not* persist the change.
+          #
+          # @param [ Document ] document The document to delete.
+          #
+          # @api private
+          def _remove(document)
```
jamis, line 168: The private `EmbedsMany::Proxy#delete_one` method was being called in a couple other places, via `#send`. This bothered me because:

1. `delete_one` is a public method on Array
2. this `Proxy` class delegates to Array, and so this redefinition changes the original contract of the `Array#delete_one` method
4. making this private means that invoking `delete_one` on the proxy publicly goes through the delegation chain and invokes `Array#delete_one`, which is both surprising and non-functional on the proxy
5. the name `delete_one` doesn't clearly describe what this method is actually used for

I've opted to pull it out of private visibility, give it a better name, and make it clear that this is an internal API. (In general, I think we ought to avoid using `#send` to invoke methods on other classes, as much as possible.)

PR #5656, opened 2023-07-03
MONGOID-5076 Reset the atomic operation state on reload

jamis:
Reloading a document was not resetting the atomic operation state, which meant the reloaded might object might still believe there were changes to persist. This makes sure that state is reset on reload.

ref: https://jira.mongodb.org/browse/MONGOID-5076

### DIFF
--- a/lib/mongoid/reloadable.rb
+++ b/lib/mongoid/reloadable.rb
@@ -17,16 +17,14 @@ module Mongoid
     #
     # @return [ Document ] The document, reloaded.
     def reload
-      if @atomic_selector
-        # Clear atomic_selector cache for sharded clusters. MONGOID-5076
-        remove_instance_variable('@atomic_selector')
-      end
-
       reloaded = _reload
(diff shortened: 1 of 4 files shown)


lib/mongoid/reloadable.rb
```diff
@@ -17,16 +17,14 @@ module Reloadable
     #
     # @return [ Document ] The document, reloaded.
     def reload
-      if @atomic_selector
-        # Clear atomic_selector cache for sharded clusters. MONGOID-5076
-        remove_instance_variable('@atomic_selector')
```
comandeo, line 22: I am not sure I understand why removing `@atomic_selector` is not necessary now.

comandeo (approved):

lib/mongoid/reloadable.rb
jamis: I should have left a note here about that. Thanks for mentioning it. The reason is that it is dead code -- it was an artifact of a fix from the 7.x series, and is no longer needed. I removed it as part of some casual cleaning up here.

PR #5660, opened 2023-07-06
MONGOID-5643 Storable#add_field_expression add values with same operator keys using $and when value is a hash with symbol operator key

JohnMaguir:
Mongoid::Criteria::Queryable::Storable#add_field_expression when value is a hash with symbol operator key add values with same operator keys using $and

The keys of the hash being added are converted to a string but they keys of the corresponding hash in the selector are not converted to a string so they do not match and wrong result is given from `Mongoid::Criteria::Queryable::Storable#add_field_expression`
because `'$in' != :$in`

Currently using a symbol rather than a string will result in a different output e.g.
```
#Strings
Band.where("foo"=>{'$in' =>["bar"]}).where('foo' => {'$in' => ['zoom']})
=>  {"foo"=>{'$in'=>["bar"]}, "$and"=>[{"foo"=>{'$in'=>["zoom"]}}]}

#Symbols
Band.where("foo"=>{:$in=>["bar"]}).where('foo' => {:$in => ['zoom']})
=> {"foo"=>{:$in=>["zoom"]}}
```

With this fix we will see the correct output
```
#Symbols
Band.where("foo"=>{:$in=>["bar"]}).where('foo' => {:$in => ['zoom']})
=>  {"foo"=>{:$in=>["bar"]}, "$and"=>[{"foo"=>{:$in=>["zoom"]}}]}
```

### DIFF
--- a/lib/mongoid/criteria/queryable/storable.rb
+++ b/lib/mongoid/criteria/queryable/storable.rb
@@ -48,7 +48,7 @@ module Mongoid
             if value.is_a?(Hash) && selector[field].is_a?(Hash) &&
               value.keys.all? { |key|
                 key_s = key.to_s
-                key_s.start_with?('$') && !selector[field].key?(key_s)
+                key_s.start_with?('$') && !selector[field].keys.map(&:to_s).include?(key_s)
               }
             then
               # Multiple operators can be combined on the same field by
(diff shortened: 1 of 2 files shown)


johnnyshields:
@comandeo this looks good to merge.

comandeo (approved):

PR #5661, opened 2023-07-07
RUBY-2993 allow Mongoid with BSON 5 to have literal BSON::Decimal128 fields

jamis:
BSON 5 changes the default serialization of `BSON::Decimal128` values so that they return as native Ruby `BigDecimal` values. This change impacts Mongoid by making it impossible for fields declared as `BSON::Decimal128` to actually return `BSON::Decimal128` values.

This PR adds a configuration option (`allow_bson5_decimal128`) that re-registers the BSON::Decimal128 type handler so that these values are returned literally, rather than be translated into `BigDecimal` values.

*Caveat*: this changes the behavior of bson-ruby globally, for all requests running in the same process. If, for instance, you use Mongoid for some things, and the driver directly for others, changing this setting will impact all replies, regardless of how they were made.

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -87,7 +87,7 @@ module Mongoid
     #
     # @note this option only has effect when BSON 5+ is present. Otherwise,
     #   the setting is ignored.
-    option :allow_bson_decimal128, default: false, on_change: -> (allow) do
+    option :allow_bson5_decimal128, default: false, on_change: -> (allow) do
         if BSON::VERSION >= '5.0.0'
           if allow
             BSON::Registry.register(BSON::Decimal128::BSON_TYPE, BSON::Decimal128)
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -80,6 +80,23 @@ module Mongoid
     # Store BigDecimals as Decimal128s instead of strings in the db.
     option :map_big_decimal_to_decimal128, default: true
 
+    # Allow BSON::Decimal128 to be parsed and returned directly in
+    # field values. When BSON 5 is present and the this option is set to false
+    # (the default), BSON::Decimal128 values in the database will be returned
+    # as BigDecimal.
+    #
+    # @note this option only has effect when BSON 5+ is present. Otherwise,
+    #   the setting is ignored.
+    option :allow_bson_decimal128, default: false, on_change: -> (allow) do
+        if BSON::VERSION >= '5.0.0'
+          if allow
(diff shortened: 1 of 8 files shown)


lib/mongoid/config.rb
```diff
@@ -80,6 +80,23 @@ module Config
...
     option :map_big_decimal_to_decimal128, default: true
 
+    # Allow BSON::Decimal128 to be parsed and returned directly in
+    # field values. When BSON 5 is present and the this option is set to false
+    # (the default), BSON::Decimal128 values in the database will be returned
+    # as BigDecimal.
+    #
+    # @note this option only has effect when BSON 5+ is present. Otherwise,
+    #   the setting is ignored.
+    option :allow_bson5_decimal128, default: false, on_change: -> (allow) do
```
comandeo, line 90: :+1:

comandeo (approved):

johnnyshields:
@jamis I don't understand the motivation behind this PR.

Why not just allow the following:

```
class Person
  include Mongoid::Document

  field :age, type: BigDecimal
  field :salary, type: BSON::Decimal128
end
```

It would make sense that both age and salary fields are stored as Decimal128 in DB, however age is demongoized as a BigDecimal and salary is demongoized as a BSON::Decimal128. This avoids any awkward global configs and user confusion.

jamis:
Hello @johnnyshields -- FWIW, I agree with you that there is something awkward in this approach. It's an unfortunate consequence of the work that was done last year to make BSON::Decimal128 automatically convert to BigDecimal, at the driver level. There were a number of tickets related to that work (including RUBY-2954, which you filed).

Because the conversion happens at the driver level, Mongoid has no way to know whether or not a translation has occurred. It could, of course, translate the value _back_ to Decimal128 if it gets a BigDecimal for what it expects to be a Decimal128 field, but that operation has the potential to be non-idempotent. 

Another option could be to just have Mongoid always enable that `allow_bson5_decimal128` setting (or the equivalent; the setting would go away in that case) and handle the conversion itself...but then we wind up with the sort of issue you ran into in RUBY-2954 (where sometimes a conversion to BigDecimal is desired, and sometimes not).

Last year, the team decided to opt for a universal translation to BigDecimal, which will suffice for the majority of cases. This new flag is not expected to see much use, as the BSON::Decimal128 type is not very useful by itself, and most users will have been converting it to something like BigDecimal anyway. The flag allows the few users who might be impacted by this--who might have some legitimate cause for using BSON::Decimal128 directly--to retain that legacy functionality.

PR #5665, opened 2023-07-12
MONGOID-5615 Allow use key_alt_name

comandeo:


alexbevi (approved):

lib/mongoid/config/encryption.rb
```diff
@@ -192,13 +192,20 @@ def algorithm_for(field)
...
-      # @return [ Array<BSON::Binary> | nil ] The keyId encryption schema field,
-      #   or nil if the key id is nil.
-      def key_id_for(key_id_base64)
-        return nil if key_id_base64.nil?
+      # @return [ Array<BSON::Binary> | String | nil ] The keyId encryption schema field,
+      #   JSON pointer to the field that contains keyAltName,
+      #   or nil if both key_id_base64 and key_name_field are nil.
+      def key_id_for(key_id_base64, key_name_field)
+        return nil if key_id_base64.nil? && key_name_field.nil?
+        raise ArgumentError, 'FIXME' if !key_id_base64.nil? && !key_name_field.nil?
```
jamis, line 202: FIXME? :)

jamis (approved):
Great work! Just one place where an exception message needs to be fleshed out, but other than that, 👍 .  Marking it approved because I've got no material objections. :)

lib/mongoid/config/encryption.rb
comandeo: Oh my, shame on me :-( Thank you!

PR #5675, opened 2023-07-31
MONGOID-5624 Fix alias issue when store_as is different case

alexbeeken:
When storing a relation we want to reference it the same way we found it during attribute processing.

I believe this fixes this bug: https://jira.mongodb.org/browse/MONGOID-5624

Basically, this happened when an `embed_one` had a `store_as` alias with capitalization that was different than the original `name`. The error we were seeing was

```
NoMethodError:
       undefined method `nested_builder' for nil:NilClass
```

During a save attempt

### DIFF
--- a/lib/mongoid/association/macros.rb
+++ b/lib/mongoid/association/macros.rb
@@ -16,20 +16,7 @@ module Mongoid
 
         # A hash that maps aliases to their associations. This hash maps the
         # associations "in database name" to its "in code" name. This is used when
-        # associations specify the `store_as` option, or on a referenced association.
-        # On a referenced association, this is used to map the foreign key to
-        # the association's name. For example, if we had the following
-        # relationship:
-        #
-        #   User has_many Accounts
-        #
-        # User will have an entry in the aliased associations hash:
-        #
-        #   account_ids => accounts
-        #
-        # Note that on the belongs_to associations, the mapping from
-        # foreign key => name is not in the aliased_associations hash, but a
-        # mapping from name => foreign key is in the aliased_fields hash.
+        # associations specify the `store_as` option.
         #
         # @return [ Hash<String, String> ] The aliased associations hash.
         #
--- a/lib/mongoid/association/macros.rb
+++ b/lib/mongoid/association/macros.rb
@@ -36,10 +36,15 @@ module Mongoid
         # @api private
         class_attribute :aliased_associations
 
+        # @return [ Set<String> ] The set of associations that are configured 
+        #   with :store_as parameter.
+        class_attribute :stored_as_associations
+
         self.embedded = false
         self.embedded_relations = BSON::Document.new
         self.relations = BSON::Document.new
         self.aliased_associations = {}
+        self.stored_as_associations = Set.new
       end
 
       # This is convenience for libraries still on the old API.
@@ -220,6 +225,7 @@ module Mongoid
             self.relations = self.relations.merge(name => assoc)
             if assoc.embedded? && assoc.respond_to?(:store_as) && assoc.store_as != name
               self.aliased_associations[assoc.store_as] = name
+              self.stored_as_associations << assoc.store_as
             end
           end
         end
--- a/lib/mongoid/association/macros.rb
+++ b/lib/mongoid/association/macros.rb
@@ -16,7 +16,20 @@ module Mongoid
 
         # A hash that maps aliases to their associations. This hash maps the
         # associations "in database name" to its "in code" name. This is used when
-        # associations specify the `store_as` option.
+        # associations specify the `store_as` option, or on a referenced association.
+        # On a referenced association, this is used to map the foreign key to
+        # the association's name. For example, if we had the following
+        # relationship:
+        #
+        #   User has_many Accounts
+        #
+        # User will have an entry in the aliased associations hash:
+        #
+        #   account_ids => accounts
+        #
+        # Note that on the belongs_to associations, the mapping from
+        # foreign key => name is not in the aliased_associations hash, but a
+        # mapping from name => foreign key is in the aliased_fields hash.
         #
         # @return [ Hash<String, String> ] The aliased associations hash.
         #
--- a/lib/mongoid/attributes/processing.rb
+++ b/lib/mongoid/attributes/processing.rb
@@ -52,11 +52,11 @@ module Mongoid
         end
 
         if relations.has_key?(aliased)
-          pending_relations[name] = value
+          pending_relations[aliased.to_s] = value
           return true
         end
         if nested_attributes.has_key?(aliased)
-          pending_nested[name] = value
+          pending_nested[aliased.to_s] = value
           return true
         end
         return false
--- a/lib/mongoid/attributes/processing.rb
+++ b/lib/mongoid/attributes/processing.rb
@@ -52,11 +52,11 @@ module Mongoid
         end
 
         if relations.has_key?(aliased)
-          pending_relations[aliased.to_s] = value
+          pending_relations[aliased] = value
           return true
         end
         if nested_attributes.has_key?(aliased)
-          pending_nested[aliased.to_s] = value
+          pending_nested[aliased] = value
           return true
         end
         return false
--- a/lib/mongoid/attributes/processing.rb
+++ b/lib/mongoid/attributes/processing.rb
@@ -44,19 +44,17 @@ module Mongoid
       # @return [ true | false ] True if pending, false if not.
       def pending_attribute?(key, value)
         name = key.to_s
-
         aliased = if aliased_associations.key?(name)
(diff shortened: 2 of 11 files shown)


alexbeeken:
@johnnyshields is there anything more I need to do to get this ready for review? Thanks!

tahoemph (approved):

danhealy (approved):
👜 ✨

comandeo:
Hi @alexbeeken, thank you very much for your contribution! Proposed changes fix the issue; unfortunately, there are multiple tests that fail. I am looking into it right now.

comandeo:
The problem is that the `aliased_associations` hash contains not only aliases specified by `as` option, but also `*_ids` alias for associations. Therefore, a code like `User.update(address_ids: [address.id])` fails with the proposed fix. Mongoid tries to assign ids to `addresses` attribute instead.

alexbeeken:
@comandeo thanks for pointing this out. I'll take another look soon.

comandeo:
@alexbeeken We discussed the PR with the team, and I believe I know a possible solution. I'll commit some changes later today, so we can review them together. Thank you for your cooperation!

comandeo:
The problem was caused by fixes related to https://jira.mongodb.org/browse/MONGOID-5089 Our current issue seems to be more significant that the one in MONGOID-5089. Therefore, we should revert some changes done there.

jamis (approved):

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
```diff
@@ -3723,13 +3723,6 @@ class Distributor
     end
   end
 
-  context "when there is a foreign key in the aliased associations" do
-    it "has the correct aliases" do
-      expect(Dog.aliased_associations["breed_ids"]).to eq("breeds")
-      expect(Breed.aliased_associations["dog_ids"]).to eq("dogs")
-    end
-  end
```
johnnyshields, line 3731: This looks like a breaking change. Aren't some applications likely to depend on this syntax?

I think we should add a feature flag here and very clearly document the change in the release notes.

@jamis please hold on merging this until above is addressed.

spec/mongoid/attributes_spec.rb
```diff
@@ -2707,4 +2707,12 @@
       catalog.set_field.should == Set.new([ 1, 2 ])
     end
   end
+
+  context "when en embedded field has a capitalized store_as name" do
+    let(:person) { Person.new(Purse: { brand: "Gucci" }) }
+
+    it "saves successfully" do
+      expect(person.save!).to eq(true)
```
johnnyshields, line 2715: Shouldn't this assert that the Gucci brand purse is actually assigned, both initially and also on reload?

spec/mongoid/association/referenced/belongs_to/proxy_spec.rb
```diff
@@ -1383,6 +1383,8 @@ class C
     end
 
     it "uses the correct explicit id" do
+      pending "https://jira.mongodb.org/browse/MONGOID-5089"
```
johnnyshields, line 1386: MONGOID-5089 was fixed. Why are we reintroducing pending here? Was this perhaps an incorrectly resolved merge conflict?

comandeo: Changes in this PR un-fix [MONGOID-5089](https://jira.mongodb.org/browse/MONGOID-5089). We can reopen and re-triage it. Since it was for a very specific use case, it seems reasonable to sacrifice 5089 in favour of this one. What do you think?

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
comandeo: `aliased_associations` hash is marked as `@api private` - [here](https://github.com/mongodb/mongoid/blob/503fba5a6cb066d9ab2c774c9922349a4cb67be5/lib/mongoid/association/macros.rb#L36C11-L36C23). So, strictly speaking, an application shouldn't depend on its internals.

spec/mongoid/association/referenced/belongs_to/proxy_spec.rb
johnnyshields: No, it doesn't make sense to reintroduce a bug to fix another bug. Please deliver a proper fix that covers both.

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
comandeo: Well, if this used to work, it was definitely an unexpected side effect of MONGOID-5089. I can hardly imagine a use case where one may want to do something like this. It is the case when `Breed` class uses hashes as ids?

I'll check if something like this does work now in a reasonable situation, and document it and/or put it behind a feature flag this is the case.

spec/mongoid/association/referenced/belongs_to/proxy_spec.rb
comandeo: We were not aware that [MONGOID-5089](https://jira.mongodb.org/browse/MONGOID-5089) was reported by a customer, the ticket was created by a team member. Thank you for pointing out that you were affected by it, we will reconsider the proposed solution.

spec/mongoid/association/referenced/has_and_belongs_to_many/proxy_spec.rb
johnnyshields: @comandeo just to be clear I am not sure the example I gave used to work, **I didn't test it**, however I do vaguely recall seeing some _ids method work in cases I wouldn't expect it to in the past (I forget exactly what.)

spec/mongoid/association/referenced/belongs_to/proxy_spec.rb
johnnyshields: https://jira.mongodb.org/browse/MONGOID-5075 which is a dupe of https://jira.mongodb.org/browse/MONGOID-5089 was reported by a customer.

lib/mongoid/association/macros.rb
```diff
@@ -36,10 +36,15 @@ module Macros
         # @api private
         class_attribute :aliased_associations
 
+        # @return [ Set<String> ] The set of associations that are configured 
+        #   with :store_as parameter.
```
johnnyshields, line 40: Should add additional context which explains why this is needed and reference the ticket.

lib/mongoid/attributes/processing.rb
```diff
@@ -44,22 +44,46 @@ def process_attributes(attrs = nil)
...
         if nested_attributes.has_key?(aliased)
-          pending_nested[name] = value
+          set_pending_nested(name, aliased, value)
           return true
         end
-        return false
+        false
+      end
+
+      # Set value of the pending relation.
```
johnnyshields, line 63: Please improve this comment to explain what the method does and why, e.g. "checks if the name is a stored_as..."

lib/mongoid/attributes/processing.rb
```diff
@@ -44,22 +44,46 @@ def process_attributes(attrs = nil)
...
+      # @param [ Object ] value The value of the relation.
+      def set_pending_relation(name, aliased, value)
+        if stored_as_associations.include?(name)
+          pending_relations[aliased] = value
+        else
+          pending_relations[name] = value
+        end
+      end
+
+      # Set value of the pensing nested attribute.
```
johnnyshields, line 76: same here

spec/mongoid/attributes_spec.rb
```diff
@@ -2707,4 +2707,21 @@
...
+    end
+
+    it 'saves successfully' do
+      expect(person.save!).to eq(true)
+    end
+
+    it 'persists the value' do
+      person.save!
+      expect(person.reload.purse.brand).to eq('Gucci')
+    end
```
johnnyshields, line 2725: Should add one more test case that the raw data in MongoDB actually has capital "P" Purse. Theoretically the code could be storing with lowercase "P" in the DB and all the tests above would still pass.

spec/support/models/purse.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+# rubocop:todo all
```
johnnyshields, line 2: New files shouldn't have `rubocop:todo all`

lib/mongoid/attributes/processing.rb
jamis: FWIW, I think the comment is fine. Just a typo here: `pensing` => `pending`.

comandeo:
@alexbeeken Thank you very much for your contribution!

PR #5676, opened 2023-08-02
MONGOID-5611 Add ActiveJob serializer for ObjectId

comandeo:


jamis:
The test failures look like they might be related to adding activejob as a dependency -- a version conflict?

PR #5678, opened 2023-08-10
RUBY-3209 Add API docs generator and publisher

comandeo:


johnnyshields:
why not make this a rake task? There is already a `docs:yard` rake task, so why not fix it?

jamis (approved):

PR #5497, opened 2022-10-05
MONGOID-5391 - Add #pluck_each, extract out PluckEnumerator

johnnyshields:
I'd like to humbly request this feature for reconsideration to be merged. A certain Mongoid maintainer of xmas past rejected this on the grounds that there are plans in MONGOID-5411 to make a feature which returns query raw results as a hash--**which I fully support**--however the timeline to do MONGOID-5411 is unspecified, whereas this feature (`#pluck_each`) is ready to be used today and already yields a tremendous performance gain 🚀. I've gotten a ton of mileage out of using this in my app.

Moreover, I think the extraction of PluckEnumerator in this code is _probably_ a stepping stone implementing MONGOID-5411. I'd be happy to add looking at MONGOID-5411 to my queue.

PR #5689, opened 2023-08-26
MONGOID-5654 [Monkey Patch Removal] Move Hash#__consolidate__ into Mongoid::Contextual

johnnyshields:
Fixes MONGOID-5654

`Hash#__consolidate__` monkey-patch method is used only once in the entire Mongoid codebase, in a private capacity inside `Contextual::Mongo` class. There is no reason to monkey-patch the kernel Hash class for this.

This PR does the following:
- ~~Move the method into `Contextual::Mongo` as a private method.~~
- Move the associated `Hash#mongoize_for` private method, which is only called once inside `__consolidate__`
- Rename and refactor the methods.

* [ ] TODO: Move private method tests to cover a public method behavior.

The method name has `__` prefix so we can assume this is an internal only method that is safe to remove.

### DIFF
--- a/lib/mongoid/contextual/mongo.rb
+++ b/lib/mongoid/contextual/mongo.rb
@@ -811,7 +811,6 @@ module Mongoid
       def update_documents(attributes, method = :update_one, opts = {})
         return false unless attributes
 
-        attributes = Hash[attributes.map { |k, v| [klass.database_field_name(k.to_s), v] }]
         view.send(method, prepare_atomic_updates(klass, attributes), opts)
       end
 
@@ -1076,6 +1075,7 @@ module Mongoid
       # @return [ Hash ] The prepared atomic updates.
       def prepare_atomic_updates(klass, attributes)
         attributes.each_pair.with_object({}) do |(key, value), atomic_updates|
+          key = klass.database_field_name(key)
           if key.to_s.start_with?('$')
             value = value.each_with_object({}) do |(key2, value2), hash|
               key2 = klass.database_field_name(key2)
--- a/lib/mongoid/contextual/mongo.rb
+++ b/lib/mongoid/contextual/mongo.rb
@@ -810,6 +810,7 @@ module Mongoid
       # @return [ true | false ] If the update succeeded.
       def update_documents(attributes, method = :update_one, opts = {})
         return false unless attributes
+
         attributes = Hash[attributes.map { |k, v| [klass.database_field_name(k.to_s), v] }]
         view.send(method, prepare_atomic_updates(klass, attributes), opts)
       end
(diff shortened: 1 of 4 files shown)


johnnyshields:
The failing spec for ruby-2.7 is not related to this PR.

jamis (changes requested):
@johnnyshields - I love the idea of pulling this out of `Hash`. However, when it comes to testing a private method, it raises a bit of a red flag for me. I think that if the method needs to be tested in isolation, it should probably be pulled out into its own module, much like you did for `Embedded.traverse`. What do you think?

johnnyshields:
Hmm... in general I am like-minded re: not testing private methods. However, *in this case*, we already have the tests (coming from `Hash.__consolidate__`.) The most important thing IMHO is to keep the method private since it is entirely localized to `Contextual::Mongo`. The more methods that can be kept private, the easier refactoring becomes down the line. Making a method public solely because it is tested is _"the tail wagging the dog."_ Re: `Embedded.traverse`, the method needed to be public because it was called by multiple classes.

`Contextual::Mongo` is a bulky class (> 1000 lines) and I've always thought it should be refactored or split up--you can see a bit of that in [my pluck_each PR](https://github.com/mongodb/mongoid/pull/5497)--but I think it's best to add these private methods for now with specs as-is, and defer the question of `Contextual::Mongo` refactoring in a separate ticket where we look at it more holistically.

jamis:
I understand your point, but I still am strongly opposed to writing tests for methods with `private` visibility. I know there are instances of this in the Mongoid codebase, but when I encounter them, I've been rewriting them.

If you feel delegation is too heavy-handed for this feature, there are a couple of other options you might consider:

1. Make the method `public`, but prefix the name with an underscore and document it as `@api private`. The tests can then invoke it without resorting to `send`.
2. Instead of testing the method directly, test it indirectly by observing its effects in other, publicly exposed methods.

Also, just to clarify my original point: moving the method to a new module so it can be tested in isolation does not mean the new module and method must be (or even should be) part of Mongoid's public API. In many instances, I think it makes sense to document the new module and/or method with `@api private`. As you mentioned, it definitely makes refactoring easier.

There are probably other ways to approach this, as well, but adding new tests (or refactoring old tests) that directly invoke private methods is a non-starter for me.

johnnyshields:
@jamis would you accept tests for the public `update_documents` method which use RSpec's mock expectation syntax instead? Specifically I would be calling `update_documents` then asserting the `view` (which is a Mongo Ruby Driver object) is receiving a certain argument hash. I think this is fair game, what do you think?

jamis:
@johnnyshields -- yeah, that sounds reasonable. Thanks for meeting me halfway. :)

johnnyshields:
OK cool, will do. By the way, most of the monkey patch PRs do not include private method tests.

jamis:
@johnnyshields -- I've opened a new PR that builds on your work here (https://github.com/mongodb/mongoid/pull/5740), and I'm going to close this PR in favor of that one. I've described the reasoning for it on that PR. Thanks for your work on this!

PR #5690, opened 2023-08-27
Fix broken mongoid/tasks/database_rake_spec

johnnyshields:
I am pretty sure this spec is broken and Mongoid is not running to correctly.

You can see the failure here: https://github.com/tablecheck/mongoid-ultra/actions/runs/5989956616/job/16246745994

There are two problems:
1. The DB client for `encrypted` is not configured correctly.
2. `OptionParser#order!` is choking on RSpec's default `--pattern` option. This might be somewhat environment specific, but AFAIK RSpec injects this automatically.

### DIFF
--- a/lib/mongoid/tasks/encryption.rake
+++ b/lib/mongoid/tasks/encryption.rake
@@ -20,9 +20,8 @@ namespace :db do
             options[:key_alt_name] = v
           end
         end
-        # rubocop:disable Lint/EmptyBlock
-        parser.parse!(parser.order!(ARGV) {})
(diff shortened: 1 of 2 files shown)


johnnyshields:
@jamis @comandeo can you try running `rspec spec/mongoid/tasks/database_rake_spec.rb` with FLE options enabled locally? It appears to be breaking because it can't find the db client named `encrypted`. It appears that specs even like `db:setup` and `db:test:prepare` break when using FLE because they are trying to create indexes some model which uses client `encrypted`.

PR #5692, opened 2023-08-27
MONGOID-5653 - Move Hash#__nested__ monkey patch method to new module Mongoid::Attributes::Embedded.traverse

johnnyshields:
`Hash#__nested__` is an unnecessary monkey patch which is only used in one place. Let's remove it.

Moreover, it's logic duplicates the code in `Mongoid::Reloadable#extract_embedded_attributes`

This PR consolidates and refactors the logic, and fixes a number of edge-cases and adds lots of specs.

This PR also removes the `maybe_integer` Util method, because it doesn't do the correct thing. For example, a Hash may have a string literal key `"100"`, but `maybe_integer` will always convert this to numeric 100. (This case was actually covered in the pre-existing `__nested__` specs.)

### DIFF
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -3,6 +3,7 @@
 
 require "active_model/attribute_methods"
 require "mongoid/attributes/dynamic"
+require "mongoid/attributes/embedded"
 require "mongoid/attributes/nested"
 require "mongoid/attributes/processing"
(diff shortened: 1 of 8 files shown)


johnnyshields:
The failing spec for ruby-2.7 is not related to this PR.

jamis (approved):
Excellent, thank you @johnnyshields

PR #5693, opened 2023-08-28
MONGOID-5647 Allow #count to be used with #for_js

jamis:
Currently, trying to invoke `#count` after calling `#for_js` results in an error, because `#count` calls `view.count_documents` under the covers, and that method does not support the `$where` command. This PR puts a bandaid on this scenario by making `#count` call `view.count` if the filter contains a `$where` command.

### DIFF
--- a/lib/mongoid/contextual/mongo.rb
+++ b/lib/mongoid/contextual/mongo.rb
@@ -74,7 +74,12 @@ module Mongoid
       # @return [ Integer ] The number of matches.
       def count(options = {}, &block)
         return super(&block) if block_given?
-        view.count_documents(options)
+
+        if valid_for_count_documents?
+          view.count_documents(options)
+        else
+          view.count(options)
+        end
       end
 
       # Get the estimated number of documents matching the query.
@@ -1036,6 +1041,16 @@ module Mongoid
         end
         docs = eager_load(docs)
         limit ? docs : docs.first
+      end
+
+      # Queries whether the current context is valid for use with
+      # the #count_documents? predicate. A context is valid if it
+      # does not include a `$where` operator.
+      #
+      # @return [ true | false ] whether or not the current context
+      #   includes a `$where` operator.
+      def valid_for_count_documents?
+        view.filter.keys.none? { |k| k == '$where' }
       end
 
       def raise_document_not_found_error
--- a/spec/mongoid/contextual/mongo_spec.rb
+++ b/spec/mongoid/contextual/mongo_spec.rb
@@ -159,6 +159,16 @@ describe Mongoid::Contextual::Mongo do
         end
       end
     end
+
+    context 'when for_js is present' do
+      let(:context) do
+        Band.for_js('this.name == "Depeche Mode"')
+      end
+
+      it 'counts the expected records' do
+        expect(context.count).to eq(1)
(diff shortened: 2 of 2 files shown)


lib/mongoid/contextual/mongo.rb
```diff
@@ -1038,6 +1043,16 @@ def process_raw_docs(raw_docs, limit)
...
 
+      # Queries whether the current context is valid for use with
+      # the #count_documents? predicate. A context is valid if it
+      # does not include a `$where` operator.
+      #
+      # @return [ true | false ] whether or not the current context
+      #   includes a `$where` operator.
+      def valid_for_count_documents?
+        view.filter.keys.none? { |k| k == '$where' }
+      end
```
alcaeus, line 1054: This works for any `$where` on the top level. However, `$where` may be nested within an `$and`, `$or`, `$not`, or `$nor`. Consider:
```js
{
    "$and": [
        { "foo": { "$eq": 1 } },
        { "$where": function() { return (hex_md5(this.name) == "9b53e667f30cd329dca1ec9e6a83e994") } }
    ]
}
```

I'm not sure whether it's possible to chain `for_js` with other operators to create such a condition, but if so this method may need to traverse any of the logical operators and check if they are each valid.

jamis: Yeah, I wondered about that. I'll take another look.

@alexbevi -- I love that judo, but I worry about potential false positives (e.g. a string literal in the query containing `"$where"` ). Such false positives might not matter, though; the worst case is that it uses the deprecated `count` API.

alcaeus: Yeah, since we're erring on the side of caution with this and most likely aren't affected by the edge cases of `count` providing wrong results, I think we're fine using `count` when the query would also have worked with `countDocuments`.

lib/mongoid/contextual/mongo.rb
```diff
@@ -1038,6 +1043,24 @@ def process_raw_docs(raw_docs, limit)
...
+      #
+      # @return [ true | false ] whether or not the current context
+      #   excludes a `$where` operator.
+      def valid_for_count_documents?(hash = view.filter)
+        # Note that `view.filter` is a BSON::Document, and all keys in a
+        # BSON::Document are strings; we don't need to worry about symbol
+        # representations of `$where`.
+        hash.keys.each do |key|
+          return false if key == '$where'
+          return false if hash[key].is_a?(Hash) && !valid_for_count_documents(hash[key])
```
jamis, line 1058: I ran a quick benchmark on recursively traversing a `Hash` versus `hash.to_s.match?`, and it was quite a bit faster to traverse the hash, even in the pathological cases (deeply nested with no matching key). It wasn't that much harder to do the recursive traversal, so I opted for that.

/cc @alcaeus @alexbevi

alexbevi (approved):
LGTM!

lib/mongoid/contextual/mongo.rb
alexbevi, line 1058:
```suggestion
          return false if hash[key].is_a?(Hash) && !valid_for_count_documents?(hash[key])
```

I think there's a typo there

PR #5698, opened 2023-09-02
RUBY-2807 - Add "compressors" option to Mongoid.yml

johnnyshields:
This PR adds the "compressors" option to the auto-generated `mongoid.yml` template. Note that since it is a driver option, it is not covered by the Introspection feature. This is important, because I think many users look at the generated `mongoid.yml` as their primary reference for config options.

In addition, I've removed a line from the docs about setting compressors via connection string. Although this is *technically possible*, it is not something that should be recommended to users. (Virtually _all options_ can be set via connection string, and the Mongoid docs never recommend to use it, so it feels out-of-place in this specific instance.)


johnnyshields:
@jamis this is good to merge as well, only affects docs / comment in auto-generated code.

PR #5701, opened 2023-09-02
MONGOID-5670 [Monkey Patch Removal] Remove Hash#delete_id and Hash#extract_id

johnnyshields:
Fixes MONGOID-5670

Moved to Nested::NestedBuildable

These should first be marked deprecated, so I will raise a PR to 8.1 branch.

Also it seems there are no specs for these. Since they are private, it will stay that way.

### DIFF
--- a/lib/mongoid/criteria.rb
+++ b/lib/mongoid/criteria.rb
@@ -160,7 +160,7 @@ module Mongoid
     #
     # @return [ Object ] The id.
     def extract_id
-      selector.extract_id
+      selector['_id'] || selector[:_id] || selector['id'] || selector[:id]
     end
 
     # Adds a criterion to the +Criteria+ that specifies additional options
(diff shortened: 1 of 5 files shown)


lib/mongoid/criteria.rb
```diff
@@ -160,7 +160,7 @@ def embedded?
     #
     # @return [ Object ] The id.
     def extract_id
-      selector.extract_id
+      selector['_id'] || selector[:_id] || selector['id'] || selector[:id]
```
johnnyshields, line 163: The usage of `extract_id` for a query Criteria selector seems to be different than usage for extracting from a document Hash. For example, the code comment for this method says `Could be in an $and query`, which it would seem the current logic does not cover.

I think it makes sense to inline the code here, given the query vs. document usage difference, but would like to hear thoughts.

jamis: I don't mind seeing this inlined here, even though it does seem to duplicate the logic. It's better than the monkey patch, for sure. Also, yeah, this probably does only need to consider the string case, but I'm in favor of leaving it as-is simply because that's the literal implementation that existed before. 👍

PR #5705, opened 2023-09-03
Object monkey patch removal

johnnyshields:
Remove the following double-underscore monkey-patches:
- Object#__find_args__ (also defined in Array, Range)
- Object#__setter__ (also defined in NilClass)
- Object#__sortable__ (also defined in TrueClass, FalseClass)
- Object#__to_inc__ (also defined in BigDecimal)

These methods are not used anywhere in the codebase (nor do they have any tests.) Since their names have double-underscore we can assume they are safe to remove. I will explicitly mark them as deprecated in 8.1 for good measure.


johnnyshields:
Oops this was mistaken, these methods need refactoring in order to removed. PR closed.

PR #5708, opened 2023-09-03
MONGOID-5663 [Monkey Patch Removal] Remove Object#__sortable__

johnnyshields:
Fixes MONGOID-5663

`Object#__sortable__` is a kernel monkey patch which is only used in `Contextual::Memory`.

This PR moves it to a private method in `Contextual::Memory` with a minor amount of refactoring. Existing tests already cover all the functionality.

### DIFF
--- a/lib/mongoid/contextual/memory.rb
+++ b/lib/mongoid/contextual/memory.rb
@@ -647,25 +647,6 @@ module Mongoid
         compare_operand(a) <=> compare_operand(b)
       end
 
-      # Get the operand value to be used in comparison.
-      # Adds capability to sort boolean values.
-      #
-      # @api private
-      #
-      # @example Get the comparison operand.
-      #   compare_operand(true) #=> 1
-      #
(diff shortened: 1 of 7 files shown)


lib/mongoid/contextual/memory.rb
```diff
@@ -627,22 +627,42 @@ def apply_sorting
...
       #
       # @return [ Integer ] The comparison value.
       def compare(a, b)
-        case
-        when a.nil? then b.nil? ? 0 : 1
-        when b.nil? then -1
-        else a <=> b
+        return 0 if a.nil? && b.nil?
+        return 1 if a.nil?
+        return -1 if b.nil?
```
johnnyshields, line 645: This is refactored to be easier to read.

jamis:
Thank you Johnny! I took the liberty of tweaking it a bit so the methods are deprecated instead of removed (sure do wish these had been declared `@api private` to begin with) but otherwise kept things as you proposed. 👍

PR #5709, opened 2023-09-03
MONGOID-5662 [Monkey Patch Removal] Remove Object#__to_inc__

johnnyshields:
Fixes MONGOID-5662

`Object#__to_inc__` is a kernel monkey patch used only in the #inc and #mul atomic methods.

This PR removes the monkey patch and inlines its code to the two places it is used. The logic is covered by existing specs, so no new specs needed.


jamis:
Closing this in favor of https://github.com/mongodb/mongoid/pull/5741 so that we deprecate the monkey-patched methods, instead of removing them. Thanks for your work on this!

PR #5711, opened 2023-09-03
MONGOID-5661 [Monkey Patch Removal] Add test which checks for monkey patches

johnnyshields:
Fixes [MONGOID-5661](https://jira.mongodb.org/browse/MONGOID-5661)

This PR adds a spec to check which kernel monkey patches Mongoid has added. It will help the team to remove them overtime, and will help ensure Mongoid developers do not add new monkey patches going forward.

### DIFF
--- a/spec/mongoid/criteria/queryable/extensions/symbol_spec.rb
+++ b/spec/mongoid/criteria/queryable/extensions/symbol_spec.rb
@@ -13,6 +13,10 @@ describe Symbol do
       end
     end
 
+    after do
+      Symbol.undef_method(:fubar)
+    end
+
     let(:fubar) do
       :testing.fubar
     end
--- a/spec/mongoid/monkey_patches_spec.rb
+++ b/spec/mongoid/monkey_patches_spec.rb
@@ -4,7 +4,7 @@ require 'spec_helper'
 
 # @note This test ensures that we do not inadvertently introduce new monkey patches
 # to Mongoid. Existing monkey patch methods which are marked with +Mongoid.deprecated+
-# should not be listed in this test.
+# are excluded from this test.
 RSpec.describe('Do not add monkey patches') do # rubocop:disable RSpec/DescribeClass
   classes = [
     Object,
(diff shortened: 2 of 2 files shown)


spec/mongoid/criteria/queryable/extensions/symbol_spec.rb
```diff
@@ -13,6 +13,10 @@
       end
     end
 
+    after do
+      Symbol.undef_method(:fubar)
```
johnnyshields, line 17: This spec dynamically monkey-patches Symbol, so I need to clean it up after the spec runs.

jamis:
Hey Johnny, I like the idea behind this PR -- proactively looking for and reporting on new or removed monkey patches -- but it does not seem to play nice with the `Mongoid.deprecate` declarations. The `source_location` for deprecated methods are given a path to `active_support/deprecation/method_wrappers.rb`, where the wrapper methods are being declared, and thus are no longer reported as being under the mongoid source tree.

I'm not sure if there's a clean way to handle that case or not. Probably we can get by without this spec since the monkey patches in Mongoid seem to be encapsulated neatly in extension modules and are thus relatively easy to keep an eye on. Still, if you have any ideas for how to work around the `deprecate` issue, I'm interested.

johnnyshields:
We can just ignore deprecated methods  (i.e. not whitelist them in the spec).

The point of this spec is only to verify that "no new Monkey Patch methods are inadvertently added." The spec is **not** intended to assert that "certain Monkey Patch methods are present."

johnnyshields:
@jamis I have added a `@note` comment explaining the above and have remove the deprecated methods from the list. This is ready for final review.

jamis:
Thanks Johnny. I'm not sure why, but there's just something about this test--the test itself, not the implementation--that bothers me the more I think about it. I think I'm dubious about whether it will actually add value, or just get in the way...but I'm willing to give it the benefit of the doubt for now. I've merged it. We'll see what it does for us!

PR #5712, opened 2023-09-03
MONGOID-5666 Define #resizable instance and class methods for Set

johnnyshields:
Fixes MONGOID-5666


johnnyshields:
@jamis this one is ready to go. Looks like you already added the instance method for Set#resizable? but not the class method.

PR #5718, opened 2023-09-06
MONGOID-5656 Rails specs ought to enable FLE

jamis:
The Rails specs in the Mongoid evergreen suite do not enable FLE for the database rake tasks, which means those specs are consistently being skipped.

### DIFF
--- a/lib/mongoid/tasks/database.rb
+++ b/lib/mongoid/tasks/database.rb
@@ -27,7 +27,7 @@ module Mongoid
             logger.info("MONGOID: collection options ignored on: #{model}, please define in the root model.")
           end
         rescue Exception
-          puts "error while creating collection for #{model}"
+          logger.error "error while creating collection for #{model}"
           raise
         end
       end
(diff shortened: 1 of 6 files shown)


lib/mongoid/tasks/database.rb
```diff
@@ -26,6 +26,9 @@ def create_collections(models = ::Mongoid.models, force: false)
           else
             logger.info("MONGOID: collection options ignored on: #{model}, please define in the root model.")
           end
+        rescue Exception
+          puts "error while creating collection for #{model}"
```
comandeo, line 30: We use `logger` above, should we use it here, too?

comandeo (approved):
Great job, thank you!

lib/mongoid/tasks/database.rb
jamis: For whatever reason, the logger output doesn't display in CI, which is where I had to look to troubleshoot this initially. I can dig in and see why the logger output isn't being displayed; if I can figure that out, then I'll switch the `puts` to the logger here.

johnnyshields:
@jamis thank you for this 🎉

PR #5723, opened 2023-09-18
MONGOID-5601 Atlas Search Index Management

jamis:
Adds helpers for declaring, creating, and removing atlas search indexes for Mongoid document models.


spec/mongoid/search_indexable_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+  private
+
+  def timeboxed_wait(step: 5, max: 300)
+    start = Mongo::Utils.monotonic_time
+
+    loop do
+      yield
+
+      sleep step
+      raise Timeout::Error, 'wait took too long' if Mongo::Utils.monotonic_time - start > max
```
johnnyshields, line 47: A few points re: timeout.

- Large collections (100M+ docs) should be considered the _canonical use case_ for search indexes.
- First, indexing large collections is going to take much longer than 5 minutes. I would propose there should be _no timeout_, and instead, indicate the index building progress. (Atlas UI does this--you can see the % completion of index--so there is presumably some way Mongoid could as well.
   - Consider that is it is frustrating to make a user wait 5 minutes then fail :)
- Second, if you set socket_timeout in your Mongoid.yml--which also is a best practice--your index building here is going to timeout after 30 seconds. Index building should skip socket timeout entirely, since it happens in background anyway.
- Finally, whatever logic you add here should be backported to traditional index building as well, and possible extracted out to a generic class which applies to both.

lib/mongoid/search_indexable.rb
```diff
@@ -0,0 +1,166 @@
...
+        collection.search_indexes.create_many(search_index_specs)
+      end
+
+      # Waits for the named search indexes to be created.
+      #
+      # @param [ Integer ] interval the number of seconds to wait before
+      #   polling again (only used when a progress callback is given).
+      #
+      # @yield [ SearchIndexable::Status ] the status object
+      def wait_for_search_indexes(names, interval: 5)
```
comandeo, line 77: We should probably also document `names` param

comandeo (approved):

lib/mongoid/search_indexable.rb
jamis: wups, good catch!

PR #5725, opened 2023-09-27
Fix autosave for accepts_nested_attributes_for

kodawill:
#### Summary
This PR attempts to [automatically save](https://www.mongodb.com/docs/mongoid/master/reference/associations/#autosaving) associated documents for referenced (i.e., non-embedded) associations when the parent is saved even if `autosave` option set to `false` as long as associated documents are declared as nested attributes via `accepts_nested_attributes_for` and `autosave` is set to `true`.

```ruby
class Foo
  include Mongoid::Document
 
  has_many :bars, autosave: false

  accepts_nested_attributes_for :bars, autosave: true # This should autosave :bars association
                                                      # even if autosave is set to false in has_many
end
 
class Bar
  include Mongoid::Document
 
  belongs_to :foo
end
 
foo = Foo.new(bars: [Bar.new])
foo.save # must save associated bars documents


```

### DIFF
--- a/spec/mongoid/attributes/nested_spec.rb
+++ b/spec/mongoid/attributes/nested_spec.rb
@@ -66,6 +66,36 @@ describe Mongoid::Attributes::Nested do
         expect(association).to_not be_autosave
       end
     end
+
+    context "when autosave is explicitly false in association" do
+      context "when autosave is true as nested attributes" do
+
+        let(:post) do
+          NestedPost.new
+        end
+
+        before do
+          NestedPost.accepts_nested_attributes_for :likes, autosave: true
+        end
+
+        after do
+          NestedPost.send(:undef_method, :likes_attributes=)
+          NestedPost.nested_attributes.clear
+        end
+
+        let(:association) do
+          NestedPost.reflect_on_association(:likes)
+        end
+
+        it "sets autosave to true" do
+          expect(association).to be_autosave
+        end
+
+        it "autosaves if the association is not embedded" do
+          expect(post).to respond_to(:autosave_documents_for_likes)
+        end
+      end
+    end
   end
 
   describe "#initialize" do
--- a/spec/mongoid/attributes/nested_spec.rb
+++ b/spec/mongoid/attributes/nested_spec.rb
@@ -91,8 +91,14 @@ describe Mongoid::Attributes::Nested do
           expect(association).to be_autosave
         end
 
-        it "autosaves if the association is not embedded" do
+        it "defines autosave_documents_for_likes" do
           expect(post).to respond_to(:autosave_documents_for_likes)
+        end
+
+        it "saves associated NestedLike documents" do
+          like = NestedLike.new
+          post = NestedPost.new(likes: [like])
+          expect { post.save }.to change(NestedLike, :count).by(1)
         end
       end
     end
--- a/spec/mongoid/attributes/nested_spec_models.rb
+++ b/spec/mongoid/attributes/nested_spec_models.rb
@@ -16,6 +16,12 @@ class NestedComment
   belongs_to :post, class_name: "NestedPost"
 end
 
+class NestedLike
+  include Mongoid::Document
+
+  belongs_to :post, class_name: "NestedPost"
+end
+
 class NestedPost
   include Mongoid::Document
 
@@ -23,6 +29,8 @@ class NestedPost
   belongs_to :author, class_name: "NestedAuthor"
   has_many :comments, class_name: "NestedComment"
   accepts_nested_attributes_for :comments
+  has_many :likes, class_name: "NestedLike", autosave: false
+  accepts_nested_attributes_for :likes, autosave: true
 end
 
 class NestedBook


kodawill:
This is now ready for review.

johnnyshields:
@kodawill two points here.

First, please make a test case which demonstrates why this functionality is needed. For example:
1. Start with a test model (such as Person or Car)
2. Assign nested attributes of a has_many related model
3. Make a test which asserts the related model should be persisted. This test should fail on master, and succeed after applying your patch.

Second, (assuming this is a legitimate issue per #1) I don't think the presence of `accepts_nested_attributes_for` should set auto-save to true, because that would affect cases where nested attributes are not passed-in. Instead, if `autosave: false`, an ideal fix IMHO would only invoke auto-save logic if the user passed in nested attributes for the relation, but otherwise would not autosave.

kodawill:
@johnnyshields first of all thank you for your feedback. I have reverted previous changes and instead, as you suggested, wrote a failing spec demonstrating the possible bug for autosaving. If I understand you correctly, auto-save logic should only be invoked if `autosave: true` in `accepts_nested_attributes_for` even if association is set as `autosave: false`. Please confirm if that's what you have in mind. Please refer in description as I have updated the model definition in it for illustration.

johnnyshields:
@kodawill re: your comment:

> If I understand you correctly, auto-save logic should only be invoked if autosave: true in accepts_nested_attributes_for even if association is set as autosave: false

I think we are getting our wires crossed here. Let's leave `autosave` aside for now, and instead focus on the _behavior_ when setting nested attributes, i.e., is the nested model actually being in the database.

In other words, leveraging `autosave` is one of many possible ways to achieve our desired behavior, i.e. save the nested object.

kodawill:
@johnnyshields Added another test (failing) to describe the desired behavior.

kodawill:
And here is my proposal for the fix:
```ruby
--- a/lib/mongoid/attributes/nested.rb
+++ b/lib/mongoid/attributes/nested.rb
@@ -47,6 +47,7 @@ module Mongoid
         #   enable it.
         def accepts_nested_attributes_for(*args)
           options = args.extract_options!.dup
+          autosave_via_nested_attributes = options[:autosave]
           options[:autosave] = true if options[:autosave].nil?

           options[:reject_if] = REJECT_ALL_BLANK_PROC if options[:reject_if] == :all_blank
@@ -55,7 +56,7 @@ module Mongoid
             self.nested_attributes["#{name}_attributes"] = meth
             association = relations[name.to_s]
             raise Errors::NestedAttributesMetadataNotFound.new(self, name) unless association
-            autosave_nested_attributes(association) if options[:autosave]
+            autosave_nested_attributes(association, autosave_via_nested_attributes) if options[:autosave]

             re_define_method(meth) do |attrs|
               _assigning do
@@ -78,12 +79,12 @@ module Mongoid
         #   Person.autosave_nested_attributes(metadata)
         #
         # @param [ Mongoid::Association::Relatable ] association The existing association metadata.
-        def autosave_nested_attributes(association)
+        def autosave_nested_attributes(association, autosave_via_nested_attributes=false)
           # In order for the autosave functionality to work properly, the association needs to be
           # marked as autosave despite the fact that the option isn't present. Because the method
           # Association#autosave? is implemented by checking the autosave option, this is the most
           # straightforward way to mark it.
-          if association.autosave? || (association.options[:autosave].nil? && !association.embedded?)
+          if association.autosave? || ((association.options[:autosave].nil? || autosave_via_nested_attributes) && !association.embedded?)
             association.options[:autosave] = true
             Association::Referenced::AutoSave.define_autosave!(association)
           end
```

kodawill:
Another option by using another (additional) method:

```ruby
--- a/lib/mongoid/attributes/nested.rb
+++ b/lib/mongoid/attributes/nested.rb
@@ -46,6 +46,7 @@ module Mongoid
         #   association. Note that since the default is true, setting autosave to nil will still
         #   enable it.
         def accepts_nested_attributes_for(*args)
+          autosave_nested_attribute = autosave_nested_attribute?(args)
           options = args.extract_options!.dup
           options[:autosave] = true if options[:autosave].nil?

@@ -55,7 +56,7 @@ module Mongoid
             self.nested_attributes["#{name}_attributes"] = meth
             association = relations[name.to_s]
             raise Errors::NestedAttributesMetadataNotFound.new(self, name) unless association
-            autosave_nested_attributes(association) if options[:autosave]
+            autosave_nested_attributes(association, autosave_nested_attribute) if options[:autosave]

             re_define_method(meth) do |attrs|
               _assigning do
@@ -78,16 +79,33 @@ module Mongoid
         #   Person.autosave_nested_attributes(metadata)
         #
         # @param [ Mongoid::Association::Relatable ] association The existing association metadata.
-        def autosave_nested_attributes(association)
+        # @param [ true | false ] autosave_nested_attribute Autosave once options[:autosave] is true in accepts_nested_attributes_for
+        def autosave_nested_attributes(association, autosave_nested_attribute = false)
           # In order for the autosave functionality to work properly, the association needs to be
           # marked as autosave despite the fact that the option isn't present. Because the method
           # Association#autosave? is implemented by checking the autosave option, this is the most
           # straightforward way to mark it.
-          if association.autosave? || (association.options[:autosave].nil? && !association.embedded?)
+          if association.autosave? || ((association.options[:autosave].nil? || autosave_nested_attribute) && !association.embedded?)
             association.options[:autosave] = true
             Association::Referenced::AutoSave.define_autosave!(association)
           end
         end
+
+
+
+        # This method should be used to detect whether a nested attribute association
+        # is autosave or not when parent document is saved. Nested attribute association
+        # is autosave if `autosave` option in #accepts_nested_attributes_for is explicitly set to true.
+        #
+        # @param [ Hash ] args The options being passed in accepts_nested_attributes_for.
+        #
+        # @return [ true | false ] Whether nested attribute is to be autosave
+        # on parent document save or not.
+        def autosave_nested_attribute?(args)
+          args = args.dup
+          options = args.extract_options!
+          options[:autosave] == true
+        end
       end
     end
   end
```

johnnyshields:
@jamis it's fine to close this PR. I will take it over and re-raise.

PR #5726, opened 2023-09-27
MONGOID-5686 Fix array operations in update_all

comandeo:


jamis (approved):
Very nice!

lib/mongoid/extensions/hash.rb
```diff
@@ -166,6 +166,24 @@ def to_criteria
...
+      #
+      # This is necessary for special cases like $rename, $addToSet and $push.
+      #
+      # @param [ String ] operator The operator.
+      # @param [ Class ] klass The model class.
+      # @param [ String | Symbol ] key The field key.
+      # @param [ Object ] value The original value.
+      #
+      # @return [ Object ] Value prepared for the provided operator.
+      def value_for(operator, klass, key, value)
```
johnnyshields, line 179: I propose any new monkey patch method (private or otherwise) should have `__` prefix.

PR #5727, opened 2023-10-04
MONGOID-5658 Fix callbacks for embedded

comandeo:
This is a band-aid for [MONGOID-5658](https://jira.mongodb.org/browse/MONGOID-5658)

The existing implementation of callbacks guarantees a sane order of execution; however, it causes stack overflow if there are many embeddeds (even in a flat structure). To fix this issue we suggest ignoring around callbacks for embedded documents. This allows us to patch the existing implementation of the callbacks and keep the proper order of callbacks.

If something still needs to be done around a database operation for an embedded document, it can be done in corresponding callback of the parent document.

We should discuss a necessity of callbacks for embedded documents in general. If we still feel that we should keep this functionality, the callbacks implementation in general should be revised.

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -137,6 +137,14 @@ module Mongoid
     # See https://jira.mongodb.org/browse/MONGOID-5542
     option :prevent_multiple_calls_of_embedded_callbacks, default: true
 
+    # When this flag is true, callbacks for embedded documents will not be
+    # called. This is the default in 9.0.
+    #
+    # Setting this flag to false restores the pre-9.0 behavior, where callbacks
+    # for embedded documents are called. This may lead to stack overflow errors
+    # if there are more than cicrca 1000 embedded documents in the root
+    # document's dependencies graph.
+    # See https://jira.mongodb.org/browse/MONGOID-5658 for more details.
     option :around_callbacks_for_embeds, default: false
 
     # Returns the Config singleton, for use in the configure DSL.
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -137,10 +137,10 @@ module Mongoid
     # See https://jira.mongodb.org/browse/MONGOID-5542
     option :prevent_multiple_calls_of_embedded_callbacks, default: true
 
-    # When this flag is true, callbacks for embedded documents will not be
+    # When this flag is false, callbacks for embedded documents will not be
     # called. This is the default in 9.0.
     #
-    # Setting this flag to false restores the pre-9.0 behavior, where callbacks
+    # Setting this flag to true restores the pre-9.0 behavior, where callbacks
     # for embedded documents are called. This may lead to stack overflow errors
     # if there are more than cicrca 1000 embedded documents in the root
     # document's dependencies graph.
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -137,6 +137,8 @@ module Mongoid
     # See https://jira.mongodb.org/browse/MONGOID-5542
     option :prevent_multiple_calls_of_embedded_callbacks, default: true
 
+    option :around_callbacks_for_embeds, default: false
+
     # Returns the Config singleton, for use in the configure DSL.
     #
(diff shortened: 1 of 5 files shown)


lib/mongoid/config.rb
```diff
@@ -137,6 +137,16 @@ module Config
     # See https://jira.mongodb.org/browse/MONGOID-5542
     option :prevent_multiple_calls_of_embedded_callbacks, default: true
 
+    # When this flag is true, callbacks for embedded documents will not be
+    # called. This is the default in 9.0.
```
jamis, line 141: I feel like the name of the option might be backwards, if this is the case. I would expect `around_callbacks_for_embeds = true` to mean that around callbacks *are* invoked. If we want a true value to mean they are *not* invoked, perhaps the option should be called something like `omit_around_callbacks_for_embeds`?

lib/mongoid/interceptable.rb
```diff
@@ -149,6 +149,14 @@ def run_callbacks(kind, with_children: true, skip_if: nil, &block)
     #
     # @api private
     def _mongoid_run_child_callbacks(kind, children: nil, &block)
+      if Mongoid::Config.around_callbacks_for_embeds
```
jamis, line 152: The logic here seems to contradict the documentation for the `around_callbacks_for_embeds` option. Maybe the documentation just has the logic backwards?

lib/mongoid/interceptable.rb
```diff
@@ -162,6 +170,46 @@ def _mongoid_run_child_callbacks(kind, children: nil, &block)
       end
```
jamis, line 170: (GitHub won't let me leave a comment on a line of code that's not actually part of the PR, so I'm leaving it here)

A few lines above this, `_mongoid_run_child_callbacks` is invoked, which is fine. However, since we already know that around callbacks for embeds are enabled, we might as well just invoke `_mongoid_run_child_callbacks_with_around`, and avoid adding an extra call to the stack.

lib/mongoid/interceptable.rb
```diff
@@ -162,6 +170,46 @@ def _mongoid_run_child_callbacks(kind, children: nil, &block)
...
     end
 
+    def _mondoid_run_child_before_callbacks(kind, children: [], callback_list: [])
```
johnnyshields, line 173: mondoid --> mongoid

lib/mongoid/interceptable.rb
```diff
@@ -162,6 +170,46 @@ def _mongoid_run_child_callbacks(kind, children: nil, &block)
...
+        env.value = !env.halted
+        callback_list << [next_sequence, env]
+        if (grandchildren = child.send(:cascadable_children, kind))
+          _mondoid_run_child_before_callbacks(kind, children: grandchildren, callback_list: callback_list)
+        end
+      end
+      callback_list
+    end
+
+    def _mondoid_run_child_after_callbacks(callback_list: [])
```
johnnyshields, line 193: mondoid --> mongoid

lib/mongoid/interceptable.rb
comandeo: You are right, thank you. Adjusted the docs.

lib/mongoid/config.rb
comandeo: Updated, thank you!

lib/mongoid/interceptable.rb
comandeo: Good catch, thank you. Fixed.

lib/mongoid/interceptable.rb
comandeo: Fixed, thank you. Mondoid sounds kind of nice, though...

lib/mongoid/interceptable.rb
comandeo: Fixed, thanks.

jamis (approved):

lib/mongoid/interceptable.rb
adviti-mishra: Mondoid does sound kind of nice

PR #5729, opened 2023-10-10
MONGOID-5689 Add Rails 7.1 support

comandeo:


jamis (approved):
Looks good! Sorry about the test infrastructure changes I made (removing Docker). There was a time while you were on vacation recently when Docker images couldn't be downloaded, and it looked like it was going to be a long-term situation, so I "fixed" it. Sounds like things are back to normal now. Sorry I neglected to talk to you about that when you got back.

PR #5736, opened 2023-10-30
MONGOID-5608 allow using `#exists?` with args on relations

jamis:
Adds extended support for `#exists?` on `embeds_many` and `has_many` associations, so the API is the same as is present elsewhere in Mongoid, where the `#exists?` accepts conditions that can be used to narrow the scope of the query.

### DIFF
--- a/lib/mongoid/association/referenced/has_many/proxy.rb
+++ b/lib/mongoid/association/referenced/has_many/proxy.rb
@@ -217,9 +217,23 @@ module Mongoid
           # @example Are there persisted documents?
           #   person.posts.exists?
           #
+          # @param [ :none | nil | false | Hash | Object ] id_or_conditions
+          #   When :none (the default), returns true if any persisted
+          #   documents exist in the association. When nil or false, this
+          #   will always return false. When a Hash is given, this queries
+          #   the documents in the association for those that match the given
+          #   conditions, and returns true if any match. Any other argument is
+          #   interpreted as an id, and queries for the existence of documents
+          #   in the association with a matching _id.
+          #
           # @return [ true | false ] True is persisted documents exist, false if not.
-          def exists?
-            criteria.exists?
+          def exists?(id_or_conditions = :none)
+            case id_or_conditions
+            when :none then criteria.exists?
+            when nil, false then false
+            when Hash then criteria.where(id_or_conditions).exists?
+            else criteria.where(_id: id_or_conditions).exists?
+            end
           end
 
           # Find the matching document on the association, either based on id or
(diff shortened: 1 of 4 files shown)


lib/mongoid/association/referenced/has_many/proxy.rb
```diff
@@ -217,9 +217,23 @@ def each(&block)
...
           # @return [ true | false ] True is persisted documents exist, false if not.
-          def exists?
-            criteria.exists?
+          def exists?(id_or_conditions = :none)
+            case id_or_conditions
+            when :none then criteria.exists?
+            when nil, false then false
+            when Hash then criteria.where(id_or_conditions).exists?
+            else criteria.where(_id: id_or_conditions).exists?
+            end
```
johnnyshields, line 236: why can't this delegate to criteria#exists?

jamis: Good catch! It certainly could. I'll fix.

lib/mongoid/association/referenced/has_many/proxy.rb
```diff
@@ -217,9 +217,18 @@ def each(&block)
...
+          #   conditions, and returns true if any match. Any other argument is
+          #   interpreted as an id, and queries for the existence of documents
+          #   in the association with a matching _id.
+          #
           # @return [ true | false ] True is persisted documents exist, false if not.
-          def exists?
-            criteria.exists?
+          def exists?(id_or_conditions = :none)
+            criteria.exists?(id_or_conditions)
           end
```
johnnyshields, line 232: Why not:

```
delegate :exists?, to: :criteria
```

lib/mongoid/association/referenced/has_many/proxy.rb
```diff
@@ -217,9 +217,18 @@ def each(&block)
...
+          #   will always return false. When a Hash is given, this queries
+          #   the documents in the association for those that match the given
+          #   conditions, and returns true if any match. Any other argument is
+          #   interpreted as an id, and queries for the existence of documents
+          #   in the association with a matching _id.
+          #
           # @return [ true | false ] True is persisted documents exist, false if not.
-          def exists?
-            criteria.exists?
+          def exists?(id_or_conditions = :none)
```
johnnyshields, line 230: Why not:

```
delegate :exists?, to: :criteria
```

jamis: That would work, too. In general I'm not a fan of that style. It tends to stymie investigation into execution flow, at least for me. I'll think about it, though.

johnnyshields: It is recommended by Rubocop: https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsdelegate

comandeo (approved):
LGTM!
We have now tests for `:none` and `Hash`. Should we also add tests for `nil` and `false` ?

comandeo (approved):

PR #5740, opened 2023-11-06
MONGOID-5654: refactor and deprecate Hash#__consolidate__

jamis:
This PR takes the work by @johnnyshields  (here: https://github.com/mongodb/mongoid/pull/5689) and reworks it so that the `Hash#__consolidate__` method is deprecated, rather than removed. (In an ideal world, we would indeed just remove the method, but even though the naming of the method strongly suggests it is not part of the public API, the method itself was never explicitly declared to be `@api private`, and so out of an abundance of caution we are opting to deprecate it until the next major release.)

### DIFF
--- a/lib/mongoid/atomic_update_preparer.rb
+++ b/lib/mongoid/atomic_update_preparer.rb
@@ -20,6 +20,8 @@ module Mongoid
       # @return [ Hash ] The prepared atomic updates.
       def prepare(attributes, klass)
         attributes.each_pair.with_object({}) do |(key, value), atomic_updates|
+          key = klass.database_field_name(key.to_s)
+
           if key.to_s.start_with?('$')
             (atomic_updates[key] ||= {}).update(prepare_operation(klass, key, value))
           else
--- a/lib/mongoid/atomic_update_preparer.rb
+++ b/lib/mongoid/atomic_update_preparer.rb
@@ -66,11 +66,6 @@ module Mongoid
 
       # Mongoize for the klass, key and value.
       #
-      # @api private
-      #
-      # @example Mongoize for the klass, field and value.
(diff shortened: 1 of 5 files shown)


johnnyshields:
Awesome, thanks 🎉

lib/mongoid/atomic_update_preparer.rb
```diff
@@ -0,0 +1,91 @@
...
+        else mongoize_for(operator, klass, operator, value)
+        end
+      end
+
+      # Mongoize for the klass, key and value.
+      #
+      # @api private
+      #
+      # @example Mongoize for the klass, field and value.
+      #   {}.mongoize_for("$push", Band, "name", "test")
```
comandeo, line 70: Looks like a leftover.

comandeo (approved):
:+1:

lib/mongoid/atomic_update_preparer.rb
```diff
@@ -0,0 +1,91 @@
...
+    class << self
+      # Convert the key/values in the attributes into a hash of atomic updates.
+      # Non-operator keys are assumed to use $set operation.
+      #
+      # @param [ Class ] klass The model class.
+      # @param [ Hash ] attributes The attributes to convert.
+      #
+      # @return [ Hash ] The prepared atomic updates.
+      def prepare(attributes, klass)
+        attributes.each_pair.with_object({}) do |(key, value), atomic_updates|
```
johnnyshields, line 22: I believe you need to insert `key = klass.database_field_name(key)` immediately after this line. Please double check.

jamis: Ah, took a minute, thought you were referring to something else. :) Right, I like that. I'll refactor.

PR #5741, opened 2023-11-06
MONGOID-5662 Deprecate Object#__to_inc__ and BigDecimal#__to_inc__

jamis:
This extends @johnnyshields work in https://github.com/mongodb/mongoid/pull/5709 and deprecates `Object#__to_inc__` and `BigDecimal#__to_inc__`.

Though these are implicitly private API methods, they were never explicitly declared as such, and so we're opting to deprecate them for now and remove them in a later major release.

### DIFF
--- a/lib/mongoid/extensions/big_decimal.rb
+++ b/lib/mongoid/extensions/big_decimal.rb
@@ -3,9 +3,16 @@
 
 module Mongoid
   module Extensions
-
     # Adds type-casting behavior to BigDecimal class.
     module BigDecimal
+      # Behavior to be invoked when the module is included.
+      #
+      # @param [ Module ] base the class or module doing the including
+      #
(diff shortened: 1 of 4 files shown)


johnnyshields:
Jamis, thanks for actioning these.

I'm happy to spend some time on the weekend updating my other PRs to preserve existing methods and deprecate them if it would help.

jamis:
@johnnyshields -- right. Dmitry and I discussed it and agree that unless a method was *explicitly* declared private or `@api private`, it must be implicitly public, and thus must be deprecated before being removed. It's annoying red tape, but we're preferring to be cautious.

MONGOID-5660 is my focus this week, and I'm optimistic about getting through them all by the end of the week. The groundwork you've laid for these is greatly appreciated!

PR #5742, opened 2023-11-07
MONGOID-5694 Fix validations on associations - DO NOT MERGE

comandeo:


lib/mongoid/persistable/updatable.rb
```diff
@@ -38,8 +38,16 @@ def update_attribute(name, value)
...
-        assign_attributes(attributes)
-        save
+        self.class.with_session do |session|
+          session.with_transaction do
+            assign_attributes(attributes)
+            save.tap do |result|
+              session.abort_transaction unless result
+            end
+          end
+        end
```
johnnyshields, line 48: Does this work on versions of MongoDB which don't support transactions?

comandeo: No, it doesn't. It also does not work on standalones, which is a supported topology by Mongoid. I am just looking for possible solutions in this draft; it won't be merged.

PR #5744, opened 2023-11-12
MONGOID-5701: Rubocop: Disable empty line cops

johnnyshields:
This disables several rubocop cops which most of Mongoid's existing codebase intentionally doesn't follow:

- Layout/EmptyLinesAroundBlockBody
- Layout/EmptyLinesAroundClassBody
- Layout/EmptyLinesAroundModuleBody

Here is an example of what most Mongoid code does today:

```ruby
module Mongoid

  # Some comment
  class SomeClass

    # Some comment
    def my_method
    end
  end
end
```

I think this is aesthetically pleasing, however, the two empty lines will trigger Rubocop. Rather than updating all the existing code, I think we should just disable the cops.

Notes:
- The BlockBody one is used for `class_methods do` etc. on certain blocks that look like module definitions. Also in RSpecs
- For Class/ModuleBody, there is a `EnforcedStyle: beginning_only` option which we could consider, however, I think it is better to just keep this one flexible.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -58,9 +58,7 @@ Gemspec/OrderedDependencies:
   Enabled: false
 
 Layout/EmptyLinesAroundBlockBody:
-  Exclude:
-    - '**/*.rake'
-    - 'spec/**/*'
+  Enabled: false
 
 Layout/EmptyLinesAroundClassBody:
   Enabled: false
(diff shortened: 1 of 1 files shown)


jamis:
Well, beauty is in the eye of the beholder, I guess. :) I find the blank line awkward and lopsided, personally, and I've been removing them when I encounter them. I'm not sure I want to disable those cops, honestly.

johnnyshields:
Either way is fine by me, but it's busy-work to be removing those lines in every PR and lookout for failing Rubcop. If no-lines is the rule, then should make a PR to remove them everywhere--can be easily done with a regex find/replace.

PR #5746, opened 2023-11-13
MONGOID-5702: Monkey Patch Removal: Remove __numeric__ class method from Numeric module

johnnyshields:
It is now inlined into its only usage. It was previously declared @api private so assume it is safe to delete.


johnnyshields:
This should be a straightforward merge.

jamis:
@johnnyshields thanks for this one, yeah, it should be totally safe to apply, thanks to that `private` flag.

However, I feel like this is a step backward in readability. The `evolve` method is harder to grok at a glance. I understand wanting to remove monkeypatched methods on standard classes (three cheers for fixing that, hurray!) but I feel like there's still value in isolating the responsibility that `__numeric__` currently holds.

`Mongoid::Utils` might be a reasonable host for such functionality, if the goal is primarily to remove that monkeypatch.

johnnyshields:
@alcaeus, Jamis' suggestion of moving `__numeric__` method to Utils makes the code *less* readable: it would mean taking logic that neatly fits in a single module today, and turning it into a yardsale across multiple modules.

It's fine to close this, but the bottom line is that removing monkey patches is table-stakes for any serious Ruby gem--except Rails' ActiveSupport gem of course to which the community has granted a pass. Mongoid is not ActiveSupport.

PR #5747, opened 2023-11-14
[ON HOLD] MONGOID-5703: [Monkey Patch Removal] Extract symbol query macros into a gem

johnnyshields:
Fixes MONGOID-5703.

This PR extracts Symbol query-related monkey patches such as `:symbol.eq`, `:symbol.in`, etc. into a separate directory, which will be planned to be moved to a gem in the future.

This removes 35 different Monkey Patches on the Symbol class in total.

- [ ] TODO: Make separate PR which deprecates the query operators. Including the gem should override (i.e. undo) the deprecations.
- [ ] TODO: Tests (unit and integration)
- [ ] TODO: Remove usages of Symbol monkey patches from the main app tests.

Reasons:
- Refer to MONGOID-5660 for a litany of reasons why Monkey Patches are bad (conflicts, security, magic, user surprise, etc.)
- In addition, these symbols are a sort of query DSL which is not native to MongoDB itself. It is much better for users to become used to writing MQL in Ruby (e.g. `{ name: { '$in' => %w[Alice Bob] } }` rather than this Symbol hack `{ :name.in => %w[Alice Bob] }`
- The extraction itself is straightforward and low-risk.
- There is no equivalent in ActiveRecord to this Symbol DSL.
- This can be moved to a gem (we can host it in Github Mongoid org where other previously-extracted gems such as Mongoid::Paranoia are) so it won't incur any migration cost for users who still need it when upgrading.


johnnyshields:
@jamis @alexbevi please confirm this approach is acceptable--moving these methods to a gem--and I will work to complete this PR.

jamis:
Please hold off on that, Johnny. I don't think we have appetite right now for removing the Symbol monkeypatches.

alexbevi:
Closing this out as this is not the path we're choosing to pursue presently. As this is a draft PR the decision can be revisited if conditions change.

PR #5748, opened 2023-11-14
MONGOID-5691 Ensure that all Mongoid config options are mentioned in the documentation

jamis:
Refreshes the documentation for the configuration options so it mirrors the current mongoid.yml template.

### DIFF
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -174,7 +174,7 @@ for details on driver options.
           # Change the default authentication mechanism. Valid options are: :scram,
           # :mongodb_cr, :mongodb_x509, and :plain. Note that all authentication
           # mechanisms require username and password, with the exception of :mongodb_x509.
-          # Default on mongoDB 3.0 is :scram, default on 2.4 and 2.6 is :plain.
+          # Default is :scram since MongoDB 3.0; earlier versions defaulted to :plain.
           # auth_mech: :scram
 
           # The database or source to authenticate the user against.
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -131,100 +131,84 @@ for details on driver options.
   development:
     # Configure available database clients. (required)
     clients:
-      # Define the default client. (required)
+      # Defines the default client. (required)
       default:
-        # A uri may be defined for a client:
-        # uri: 'mongodb://user:password@myhost1.mydomain.com:27017/my_db'
-        # Please see driver documentation for details. Alternatively, you can
-        # define the following:
-        #
-        # Define the name of the default database that Mongoid can connect to.
+        # Mongoid can connect to a URI accepted by the driver:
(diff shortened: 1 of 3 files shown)


docs/reference/configuration.txt
```diff
@@ -131,100 +131,84 @@ for details on driver options.
...
-          #connect: :direct
-
-          # Change the default time in seconds the server monitors refresh their status
+          # roles:
+          #   - 'dbOwner'
+
+          # Change the default authentication mechanism. Valid options are: :scram,
+          # :mongodb_cr, :mongodb_x509, and :plain. Note that all authentication
+          # mechanisms require username and password, with the exception of :mongodb_x509.
+          # Default on mongoDB 3.0 is :scram, default on 2.4 and 2.6 is :plain.
```
johnnyshields, line 177: MongoDB 2.4/2.6 should not be supported by latest Mongoid. Anyone using them at this point is gonna have a bad time...

docs/reference/configuration.txt
```diff
@@ -131,100 +131,84 @@ for details on driver options.
...
-          - myhost3.mydomain.com:27017
+          - localhost:27017
         options:
-          # These options are Ruby driver options, documented in
-          # https://mongodb.com/docs/ruby-driver/current/reference/create-client/
-
+          # Note that all options listed below are Ruby driver client options (the mongo gem).
+          # Please refer to the driver documentation of the version of the mongo gem you are using
+          # for the most up-to-date list of options.
+          #
```
johnnyshields, line 151: World remove the `#` on this line to separate the above general comment from write concern below.

lib/rails/generators/mongoid/config/templates/mongoid.yml
```diff
@@ -118,17 +118,32 @@ development:
...
+    # broken_view_aggregate: true
+
+    # When this flag is set to false, the view options will be correctly
+    # propagated to readable methods.
+    # broken_view_options: true
+
+    # When this flag is set to true, the update and replace methods will
+    # validate the paramters and raise an error if they are invalid.
+    # validate_update_replace: false
+
```
johnnyshields, line 146: Would make this a single whitespace line rather than double

docs/reference/configuration.txt
alexbevi: Maybe something like:
```
# Default is :scram (since MongoDB 3.0). Note the previous default for MongoDB 2.4 and 2.6 was :plain.
```

lib/rails/generators/mongoid/config/templates/mongoid.yml
jamis: It probably comes down to personal preference at this point, here. I tend to prefer a stronger delineation between environment sections in these config files, especially if one of the sections is large (like the `development` section in the template). A wider gap between sections makes it easier to see at a glance where one section ends and the other begins.

PR #5750, opened 2023-11-16
MONGOID-5472 Save models using the overridden storage options used for their loading

jamis:
Makes the storage options writable for individual records, and incorporates those individualized storage options into the persistence context so that subsequent operations on that document object will then fall back to those storage options, if others are not overriding them.

See MONGOID-5472.

### DIFF
--- a/spec/mongoid/document_persistence_context_spec.rb
+++ b/spec/mongoid/document_persistence_context_spec.rb
@@ -34,30 +34,64 @@ describe Mongoid::Document do
   context 'when loaded with an overridden persistence context' do
     let(:options) { { collection: 'extra_people' } }
     let(:person) { Person.with(options) { Person.create username: 'zyg14' } }
-    let(:context) { Mongoid::PersistenceContext.get(person) }
 
-    it 'remembers its persistence context' do
-      expect(person.collection_name).to be == :extra_people
+    # Mongoid 9+ default persistence behavior
+    context 'when Mongoid.legacy_persistence_context_behavior is false' do
+      config_override :legacy_persistence_context_behavior, false
+
+      it 'remembers its persistence context when created' do
+        expect(person.collection_name).to be == :extra_people
+      end
+
+      it 'remembers its context when queried specifically' do
+        person_by_id = Person.with(options) { Person.find(_id: person._id) }
+        expect(person_by_id.collection_name).to be == :extra_people
+      end
+
+      it 'remembers its context when queried generally' do
+        person # force the person to be created
+        person_generally = Person.with(options) { Person.all[0] }
+        expect(person_generally.collection_name).to be == :extra_people
+      end
+
+      it 'can be reloaded without specifying the context' do
+        expect { person.reload }.not_to raise_error
+        expect(person.collection_name).to be == :extra_people
+      end
+
+      it 'can be updated without specifying the context' do
+        person.update username: 'zyg15'
+        expect(Person.with(options) { Person.first.username }).to be == 'zyg15'
+      end
     end
 
-    it 'can be reloaded without specifying the context' do
-      expect { person.reload }.not_to raise_error
-    end
+    # pre-9.0 default persistence behavior
+    context 'when Mongoid.legacy_persistence_context_behavior is true' do
+      config_override :legacy_persistence_context_behavior, true
 
-    it 'remembers its context when queried specifically' do
-      person_by_id = Person.with(options) { Person.find(_id: person._id) }
-      expect(person_by_id.collection_name).to be == :extra_people
-    end
+      it 'does not remember its persistence context when created' do
+        expect(person.collection_name).to be == :people
+      end
 
-    it 'remembers its context when queried generally' do
-      person # force the person to be created
-      person_generally = Person.with(options) { Person.all[0] }
-      expect(person_generally.collection_name).to be == :extra_people
-    end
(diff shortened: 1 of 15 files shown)


johnnyshields:
@jamis I'm concerned here that PersistenceContext includes not only client/database/collection but also read mode, write concern, and roughly ~70 other options from `Mongo::Client::VALID_OPTIONS`. I'm concerned this will introduce unintended side-effects--are you sure that all ~70 of those options are safe to "remember" on the model instance? Wouldn't it be better to more narrowly-scope the context to just the client/database/collection that are the focus of [MONGOID-5472](https://jira.mongodb.org/browse/MONGOID-5472)?

jamis:
Hey Johnny, thanks for raising this concern. Can you share a specific side-effect that might be introduced with this? I'd really rather not have to black list or white list these options, as that could complicate the documentation and use of this feature.

johnnyshields:
@jamis, your PR would have the affect of persisting all of the following options at the model level, including even things like `logger`, `ssl_cert` etc.

I think it is much more sensible if only `client/database/collection` are stored in the model level state, and for the others are derived from the `client` as base along with any overrides. The user expectation here is that if they modify the client at runtime--for example rotating an ssl cert--that model instances would be subject to such modifications EXCEPT for narrowly remembering their `client/database/collection`.

I cannot predict all possible edges which may arise from persisting there--and that's exactly the point! 😄 

```
      :app_name,
      :auth_mech,
      :auth_mech_properties,
      :auth_source,
      :auto_encryption_options,
      :bg_error_backtrace,
      :cleanup,
      :compressors,
      :direct_connection,
      :connect,
      :connect_timeout,
      :database,
      :heartbeat_frequency,
      :id_generator,
      :load_balanced,
      :local_threshold,
      :logger,
      :log_prefix,
      :max_idle_time,
      :max_pool_size,
      :max_read_retries,
      :max_write_retries,
      :min_pool_size,
      :monitoring,
      :monitoring_io,
      :password,
      :platform,
      :read,
      :read_concern,
      :read_retry_interval,
      :replica_set,
      :resolv_options,
      :retry_reads,
      :retry_writes,
      :scan,
      :sdam_proc,
      :server_api,
      :server_selection_timeout,
      :socket_timeout,
      :srv_max_hosts,
      :srv_service_name,
      :ssl,
      :ssl_ca_cert,
      :ssl_ca_cert_object,
      :ssl_ca_cert_string,
      :ssl_cert,
      :ssl_cert_object,
      :ssl_cert_string,
      :ssl_key,
      :ssl_key_object,
      :ssl_key_pass_phrase,
      :ssl_key_string,
      :ssl_verify,
      :ssl_verify_certificate,
      :ssl_verify_hostname,
      :ssl_verify_ocsp_endpoint,
      :truncate_logs,
      :user,
      :wait_queue_timeout,
      :wrapping_libraries,
      :write,
      :write_concern,
      :zlib_compression_level,
  ```

**Comment by jamis** [comment]
@johnnyshields -- Dmitry and I have taken a closer look at this, and while I'm not concerned about the extra options, it does turn out that hanging onto the persistence contexts like that has the potential to leak client instances, which is not acceptable. Further, it appears as though the intent of the ticket should have been less about remembering the persistence context (as it stated), and more about just remembering the storage options. I've updated the PR, with a stab at that.

File: spec/mongoid/document_persistence_context_spec.rb [hunk 1]
```diff
@@ -30,4 +30,68 @@
...
+
+      it 'can be reloaded without specifying the context' do
+        expect { person.reload }.not_to raise_error
+        expect(person.collection_name).to be == :extra_people
+      end
+
+      it 'can be updated without specifying the context' do
+        person.update username: 'zyg15'
+        expect(Person.with(options) { Person.first.username }).to be == 'zyg15'
+      end
```
johnnyshields, line 65: Please add a test which covers the case of overriding the context when setting/reloading

```
      it 'allows further overriding context when persisting' do
        person.username = 'bob'
        Person.with(collection: 'other') { person.save! }
        expect { person.reload.username }.to eq 'zyg14'
      end

      it 'allows further overriding context when reloading' do
        expect { Person.with(collection: 'other') { person.reload } }.to raise_error(Mongoid::Errors::DocumentNotFound)
      end
```

spec/mongoid/document_persistence_context_spec.rb
```diff
@@ -63,6 +63,18 @@
         person.update username: 'zyg15'
         expect(Person.with(options) { Person.first.username }).to be == 'zyg15'
       end
+
+      it 'an explicit context takes precedence over a remembered context when persisting' do
+        person.username = 'bob'
+        # should not actually save -- the person does not exist in the
+        # `other` collection and so cannot be updated.
```
johnnyshields, line 70: Doesn't this do an upsert, so it will create it as a new record in the 'other' collection?

johnnyshields: Hmm... that is surprising to me. Calling `.save!` and *silently not persisting* does not seem like acceptable behavior to me. As a user I'd prefer it either to upsert, or if that's not possible, have `.save!` throw an error and `.save` return false.

jamis: I agree, it's not great behavior. I opened https://jira.mongodb.org/browse/MONGOID-5712 last week to investigate this more.

comandeo (approved):

PR #5751, opened 2023-11-20
MONGOID-5502: Rails framework compatibility docs

alexbevi:
image


docs/reference/compatibility.txt
```diff
@@ -491,3 +491,58 @@ are supported by Mongoid.
...
+
+.. [#rails-activerecord-dependency] Depends on ``ActiveRecord``
+
+.. [#rails-activejob-dependency] Serialization of BSON & Mongoid objects is
+   fraught, works best if you explicitly send ``ObjectId``'s as strings, and
+   reconstitute them in the job
+
+.. [#rails-actioncable-dependency] not all subscription adapters are compatible
+   with Mongoid (e.g. the PostgreSQL adapter) but most if not all other adapters
+   are compatible.
```
alexbevi, line 548: Looking for feedback on how best to phrase these :)

jamis: Feels like the activejob footnote might be best given as an example somewhere, e.g.

```ruby
record = Model.find(...)
MyJob.perform_later(record._id.to_s)

class MyJob < ApplicationJob
  def perform(id_as_string)
     id = BSON::ObjectId.from_string(id_as_string)
     record = Model.find(id)
     # ...
  end
end
```

For the ActionCable footnote, I'm wondering now if that's even necessary to say. I can't imagine anyone would expect the PostgreSQL adapter to work with MongoDB. Could perhaps just say something like, "There is no MongoDB adapter for ActionCable. Redis is a robust backend for ActionCable, and we recommend using it instead." As Basecamp/37signals, we used Redis extensively, despite using MySQL for all of our persistent data. Seems like a pretty common pattern.

johnnyshields:
@alexbevi I think for ActiveRecord what users want to know is whether Mongoid and ActiveRecord can be used side-by-side in the same app. The "X" here would seem to indicate that they cannot be. If its safe, I would indicate it as a checkmark,

Similar thing re: ActiveJob Postgres adapter, when you say "Not compatible with Mongoid" do you mean there is a code conflict where one breaks the other?

alexbevi:
> @alexbevi I think for ActiveRecord what users want to know is whether Mongoid and ActiveRecord can be used side-by-side in the same app. The "X" here would seem to indicate that they cannot be. If its safe, I would indicate it as a checkmark,
 
I was going for something a bit more "literal" (Mongoid doesn't "directly" support ActiveRecord) - but this is likely not really necessary or meaningful. I'm going to remove ActiveRecord from the frameworks list and expand the introduction a bit more to call out frameworks that are dependent on AR can't use Mongoid as a drop-in replacement.


> (To echo Jamis' point, no one expects AR to talk to MongoDB nor Mongoid to SQL.)
> 
> Similar thing re: ActiveJob Postgres adapter, when you say "Not compatible with Mongoid" do you mean there is a code conflict where one breaks the other?

Will expand on this more.

docs/reference/compatibility.txt
```diff
@@ -499,6 +499,15 @@ Rails Frameworks Support
 Ruby on Rails is comprised of a number of frameworks, which Mongoid attempts to
 provided compatibility with wherever possible.
 
+Though Mongoid attempts to offer API compatibility with `Active Record <https://guides.rubyonrails.org/active_record_basics.html>`_,
+frameworks that depend directly on Active Record may not work as expected when
```
johnnyshields, line 503: suggest "libraries" (or "gems") instead of "frameworks"

docs/reference/compatibility.txt
```diff
@@ -499,6 +499,15 @@ Rails Frameworks Support
...
+Mongoid is used as a drop-in replacement.
+
+.. note::
+
+  Mongoid can be used alongside other Active Record adapters within the same
```
johnnyshields, line 508: "Mongoid can be used alongside Active Record in the same application without issue"

docs/reference/compatibility.txt
```diff
@@ -531,18 +540,26 @@ provided compatibility with wherever possible.
...
 
-   * - ``ActiveRecord``
-     - |x|
-
    * - ``ActiveStorage``
      - |x| [#rails-activerecord-dependency]_
 
-.. [#rails-activerecord-dependency] Depends on ``ActiveRecord``
+.. [#rails-actioncable-dependency] There is currently no MongoDB adapter for
+   ``ActionCable``, however any existing adapter can be used successfully in conjunction
```
johnnyshields, line 547: I'd suggest to add "any existing adapter _such as Redis_" to clarify the meaning here

docs/reference/compatibility.txt
```diff
@@ -531,18 +540,26 @@ provided compatibility with wherever possible.
...
+    record = Model.find(...)
+    MyJob.perform_later(record._id.to_s)
 
-.. [#rails-actioncable-dependency] not all subscription adapters are compatible
-   with Mongoid (e.g. the PostgreSQL adapter) but most if not all other adapters
-   are compatible.
+    class MyJob < ApplicationJob
+      def perform(id_as_string)
+        id = BSON::ObjectId.from_string(id_as_string)
+        record = Model.find(id) # ...
```
johnnyshields, line 563: You can pass `id` a string directly into `Model.find`, so the `id = BSON...` line is not needed.

docs/reference/compatibility.txt
```diff
@@ -491,3 +491,74 @@ are supported by Mongoid.
...
+     - |x| [#rails-activerecord-dependency]_
+
+   * - ``ActionView``
+     - |checkmark|
+
+   * - ``ActiveJob``
+     - |checkmark| [#rails-activejob-dependency]_
+
+   * - ``ActiveModel``
+     - |checkmark|
```
johnnyshields, line 540: Should add a footnote that Mongoid::Document is actually built on top of ActiveModel (e.g. for validations, callbacks, etc.)

docs/reference/compatibility.txt
```diff
@@ -491,3 +491,74 @@ are supported by Mongoid.
...
+   * - ``ActiveStorage``
+     - |x| [#rails-activerecord-dependency]_
+
+.. [#rails-actioncable-dependency] There is currently no MongoDB adapter for
+   ``ActionCable``, however any existing adapter (such as `Redis <https://guides.rubyonrails.org/action_cable_overview.html#redis-adapter>`_)
+   can be used successfully in conjunction with Mongoid models
+
+.. [#rails-activerecord-dependency] Depends directly on ``ActiveRecord``
+
+.. [#rails-activejob-dependency] Serialization of BSON & Mongoid objects  works best
```
johnnyshields, line 551: Double-space here between objects and works

docs/reference/compatibility.txt
```diff
@@ -491,3 +491,74 @@ are supported by Mongoid.
...
+
+   * - ``ActionView``
+     - |checkmark|
+
+   * - ``ActiveJob``
+     - |checkmark| [#rails-activejob-dependency]_
+
+   * - ``ActiveModel``
+     - |checkmark|
+
```
johnnyshields, line 541: It's worth adding ActiveSupport to this framework list and mention that Mongoid requires ActiveSupport and uses it heavily. (This is obvious to us, but folks unfamiliar with Mongoid may not be aware.)

docs/reference/compatibility.txt
```diff
@@ -491,3 +491,74 @@ are supported by Mongoid.
...
+     - |x| [#rails-activerecord-dependency]_
+
+.. [#rails-actioncable-dependency] There is currently no MongoDB adapter for
+   ``ActionCable``, however any existing adapter (such as `Redis <https://guides.rubyonrails.org/action_cable_overview.html#redis-adapter>`_)
+   can be used successfully in conjunction with Mongoid models
+
+.. [#rails-activerecord-dependency] Depends directly on ``ActiveRecord``
+
+.. [#rails-activejob-dependency] Serialization of BSON & Mongoid objects  works best
+   if you explicitly send ``ObjectId``'s as strings, and reconstitute them in the job:
```
johnnyshields, line 552: would write this as ``BSON::ObjectId`` for avoidance of doubt.

docs/reference/compatibility.txt
```diff
@@ -537,19 +537,29 @@ Mongoid is used as a drop-in replacement.
...
    can be used successfully in conjunction with Mongoid models
 
 .. [#rails-activerecord-dependency] Depends directly on ``ActiveRecord``
 
-.. [#rails-activejob-dependency] Serialization of BSON & Mongoid objects  works best
-   if you explicitly send ``ObjectId``'s as strings, and reconstitute them in the job:
+.. [#rails-activemodel-dependency] ``Mongoid::Document`` includes ``ActiveModel::Model``
+   and leverages ``ActiveModel::Validations`` for validations
+
+.. [#rails-activesupport-dependency] ``Mongoid::Document`` extends ``ActiveSupport::Concern``. Mongoid
```
johnnyshields, line 557: Would suggest to remove "`Mongoid::Document`` extends ``ActiveSupport::Concern``" here. I think this can just be:

> Mongoid requires ``ActiveSupport`` and uses it extensively, including ``ActiveSupport::TimeWithZone`` for time handling.

johnnyshields (approved):

comandeo (approved):

PR #5569, opened 2023-03-05
MONGOID-5579: Mongoid 9.0: Ruby, JRuby, Rails version support drops

johnnyshields:
This commit does the following for Mongoid 9.0:
- Drop support for Ruby 2.6, Rails 5.2
- Support JRuby 9.4, drop JRuby 9.3 (both will be supported on Mongoid 8.1, so this is fine.)
- Deprecation of Ruby 2.7, Rails 6.0
- CI changes congruent with the above
- Remove deprecated #as_json :compact option
- Remove all tests/hacks related to use of I18n 1.0, since it is not longer usable with Rails 6.0
- Workaround JRuby issue with "Logger::#{level}".constantize to get specs to pass
- Fixes a few kwarg-related specs for Ruby 3.2
- Fix a Float-precision spec for JRuby
- Remove the rspec-mocks-diag hack gem in favor of standard rspec. See MONGOID-5581.

PR #5754, opened 2023-11-23
MONGOID-5575 Add and remove supported rubies

comandeo:


comandeo:
Closing this in favor of #5569

PR #5758, opened 2023-11-24
MONGOID-5717 Fix some cases of missing **kwargs pass-thru

johnnyshields:
This is in preparation for ultimately getting rid of Ruby2Keywords when we drop support for Rails 2.7


comandeo:
Tests fails on JRuby 9.3 with proposed changes.

PR #5764, opened 2023-11-29
MONGOID-5710 Ensure env vars are set in docker

comandeo:


spec/mongoid/railties/bson_object_id_serializer_spec.rb
```diff
@@ -2,11 +2,14 @@
...
-require 'active_job'
-require 'mongoid/railties/bson_object_id_serializer'
 
-describe Mongoid::Railties::ActiveJobSerializers::BsonObjectIdSerializer do
-  let(:serializer) { described_class.instance }
+describe 'Mongoid::Railties::ActiveJobSerializers::BsonObjectIdSerializer' do
+  min_rails_version '6.0'
+
+  require 'active_job'
+  require 'mongoid/railties/bson_object_id_serializer'
```
jamis, line 10: it looks like Rails 5.2 specs are still choking on the `ActiveJob::Serializers` reference. I wonder if these require statements need to be in a `before` block? Perhaps at least the serializer one?

jamis (changes requested):

spec/mongoid/railties/bson_object_id_serializer_spec.rb
comandeo: Since we are dropping support for Rails 5.2, we should probably just remove corresponding configuration.

jamis (approved):

PR #5766, opened 2023-12-06
MONGOID-5706 Add railsmdb as a dependency

jamis:
This creates a new gem, `mongoid-odm`. The `mongoid` gem becomes a "meta" gem that simply depends on `mongoid-odm` and `railsmdb`.

This mirrors how Rails does it, with the top-level `rails` meta-gem, which simply depends on the different Rails frameworks.

Note: the `railsmdb` gem will need to be updated as well, so it depends on `mongoid-odm` instead of `mongoid`.

Note 2: if we accept this change, we'll probably want to backport it to 8.0 and 8.1, so that railsmdb can be used by those Mongoid versions as well (because it depends on `mongoid-odm` now).


jamis:
(Note: failing specs because of the dependency on the `mongoid-odm` gem, which doesn't exist yet.)

johnnyshields:
**I strongly feel this change is unnecessary**, and is just going to be a headache for the user community.

1. Who (which user/customer) is actually requesting this change?
1. Mongoid is **not** analogous to Rails, it is analogous to ActiveRecord, and AR is a single gem.
2. Why not just merge the contents of Railsmdb into Mongoid itself? railsmdb is just some simple generators, and Mongoid has several generals already.
3. Moreover, Railsmdb now introduces a new top level namespace and a new CLI process name, confuses newbie users as to what gem does what. Why not just extend the existing `rails` cli process with new tasks, or else make `mongoid-cli`? The best would be able to do `rails --mongoid --skip-active-record`, perhaps it's even possible to do this by adding a Rails extension in Mongoid.

alexbevi:
@johnnyshields:

> Who (which user/customer) is actually requesting this change?

This is a new feature we're introducing to allow CLI functionality to be rolled out that makes it easier to work with MongoDB and Ruby/Rails. Currently the feature set is limited, but we anticipate growing this to application generators or more complex generators to lower the barrier to adoption of features such as Atlas Search, CSFLE, Vector Search and more.

> Mongoid is not analogous to Rails, it is analogous to ActiveRecord, and AR is a singular gem (i.e. it doesn't have generators in a separate dependency.)
> Why not just merge the contents of Railsmdb into Mongoid itself? railsmdb is just some simple generators, and Mongoid has several generators already.

I believe the reasoning here was due to `railsmdb` being more tightly coupled to Rails and we want to ensure we can manage releases independently of Mongoid itself. 

> Railsmdb introduces a new top level namespace and a new CLI process name, confuses newbie users as to what gem does what. Why not just extend the existing rails CLI process with new tasks, or else make mongoid-cli? The best would be able to do rails --mongoid --skip-active-record, perhaps it's even possible to do this by adding a Rails extension in Mongoid.

We explored this (extending `rails`) but ultimately landed on a feature we had more control over.

> (If the goal here is rename Mongoid to something else, why not just do that outright?)

That is not the goal.

Ultimately we want the `gem install mongoid` or `gem "mongoid'` experience to remain the same for users. We're introducing new tooling and looking to introduce this in the most seamless way possible.

johnnyshields:
@alexbevi if I am a new user, after this PR is merged, I now need to learn what 4 things are:

- Rails
- Mongoid
- RailsMDB
- Mongo Ruby Driver

As a new user, I have read in my Rails Getting Started guide that I make a new Rails app using `rails new my_app`. But RailsMDB is telling me that I need run `railsmdb new my_app`. Will this include Mongoid? RailsMdb and Mongoid are different? or the same? And what about that Driver??! It's so confusing 😖 I'm gonna cry then go learn Python instead.

Have you explored being able to do the following?

```
rails new --mongoid --skip-active-record
```

> Due to railsmdb being more tightly coupled to Rails

Mongoid already has Railties, etc. and hooks into Rails. So I really don't understand this logic.

alexbevi:
> As a new user, I have read in my Rails Getting Started guide that I make a new Rails app using rails new my_app. But RailsMDB is telling me that I need run railsmdb new my_app. Will this include Mongoid? RailsMdb and Mongoid are different? or the same? And what about that Driver??! It's so confusing 😖 I'm gonna cry then go learn Python instead.

@johnnyshields with all due respect, you're potentially conflating a number of things here to try and prove a point. Mongoid exists to smooth out the Ruby on Rails experience with MongoDB - not to promote an abstraction that you need to learn _as well_ as the underlying driver.

There are definitely feature that require "dropping down to the driver", which we evaluate based on popularity/utility and expose via Mongoid APIs whenever possible/appropriate.

We're exploring introducing new tooling and functionality. If there are opportunities to obviate confusing concepts via better documentation we will address those as well, but it's important to note here this feature is not _exclusively_ meant to be a replacement for the `rails` CLI - it's meant to enable us to roll out new functionality over time.

johnnyshields:
@alexbevi certainly we all (the present audience) understand what Railsmdb/Mongoid/Driver are, but new users won't. New users are already overwhelmed with component names and concepts, and adding _yet another_ one, while yes, is _intended to help_, actually adds to this burden.

Again, I strongly suggest the team consider:
1. Adding these generators into Mongoid, which already hooks into Rails, and avoid installing a new gem.
2. Explore hooking directly into the standard `rails` CLI (`rails new --mongoid`), to reduce cognitive load on all users, and especially new ones.

johnnyshields:
> that lowering the barrier to new users effectively just doing rails new --mongoid is not the goal of the product long-term

The team ought to consider what the goal of the product is. Is MongoDB (the company) trying to win over ActiveRecord users? (or nah?)

johnnyshields:
I urge you again not to do this change. Introducing a new `mongoid-odm` gem is both silly and going to cause problems for complex app setups.

I still have not heard a good argument for not including these generators into the core Mongoid gem.

alexbevi:
While testing the integration of `railsmdb` with Mongoid we identified an obstacle due to `railsmdb` requiring Rails 7, whereas Mongoid still supports Rails 6. We're not looking to change our Rails version support for Mongoid at this time, so though we could invest in reconciling this disparity, for the moment we're going to postpone these integration efforts and continue with an independent release of `railsmdb`.

Our goal is to deliver the best MongoDB experience for Ruby on Rails developers, which we will continue to strive towards. Integration of `railsmdb` and Mongoid to bring the two solutions together is something we're still committed to however we will defer this for the moment.

PR #5769, opened 2023-12-15
Fix typo/leftover in mongoid.yml file generator template

redross:
If my English is not failing me, this seems like a typo/leftover.


jamis (approved):
Thanks for catching that!

PR #5777, opened 2024-01-23
MONGOID-5709 Fix atomic_path calculation

comandeo:


jamis (approved):

lib/mongoid/atomic.rb
```diff
@@ -178,13 +178,15 @@ def atomic_position
...
-      end
+      return @atomic_paths if @atomic_paths
+
+      paths = if _association
+                _association.path(self)
+              else
+                Atomic::Paths::Root.new(self)
+              end
+
+      paths.tap { @atomic_paths = paths unless new_record? }
```
johnnyshields, line 189: This PR introduces new behavior based on whether the subject model is `new_record?` or not. I would expect to see specs which cover both the new record and existing record cases, but it seems there is only one test added.

johnnyshields: …or MongoDB can ignore this comment and merge the PR.

PR #5779, opened 2024-01-24
MONGOID-5735 Undo deprecation of sum/avg/min/max methods

johnnyshields:
The `sum/avg/min/max` methods are in widespread use and cannot be deprecated. Note they have [equivalents in ActiveRecord](https://apidock.com/rails/ActiveRecord/Calculations/average).

It is ok to deprecate the `#aggregates` method as it largely an implementation detail.

(It is also fine to give a suitable alternative with identical functionality *first, then* deprecate these.)


johnnyshields:
@jamis can you please take a look at this?

alexbevi:
@johnnyshields we're reviewing this and will circle back.

alexbevi:
This is being addressed in #5802

PR #5786, opened 2024-02-09
MONGOID-5552 warn on session/client mismatch

jamis:


johnnyshields:
Hmm... shouldn't models work across clients? It would be useful if we could load a model from one client and save it on another.

comandeo-mongo (approved):

PR #5787, opened 2024-02-21
MONGOID-5704 Fix performance regression on belongs_to validation

jamis:
When validating a belongs_to association, Mongoid was loading the associated record and validating it, which could result in a recursive chain of records being loaded and validated, even though they had been persisted and had not changed.

This PR fixes the regression by only validating the associated record if it is (1) already loaded, and (2) either new (unpersisted), or changed.

### DIFF
--- a/lib/mongoid/validatable/associated.rb
+++ b/lib/mongoid/validatable/associated.rb
@@ -17,19 +17,6 @@ module Mongoid
     #     validates_associated :name, :addresses
     #   end
     class AssociatedValidator < ActiveModel::Validator
-      attr_reader :attributes
-
-      # Create a new validator with the given options.
-      #
-      # @params [ Hash ] options the options to use to
-      #   initialize the validator.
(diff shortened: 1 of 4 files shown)


lib/mongoid/validatable/associated.rb
```diff
@@ -16,32 +16,110 @@ module Validatable
...
+      def attributes
+        options[:attributes]
+      end
 
-      # Validates that the associations provided are either all nil or all
-      # valid. If neither is true then the appropriate errors will be added to
-      # the parent document.
+      # Checks that the named associations of the given record
+      # (`attributes`) are valid. This does NOT load the associations
+      # from memory, and will only validate records that are dirty
```
comandeo-mongo, line 28:
```suggestion
      # from the database, and will only validate records that are dirty
```

comandeo-mongo (approved):

PR #5789, opened 2024-02-23
Remove dependency on ruby2_keywords

Earlopain:
Now that Ruby 2.7 is required (#5768) this is not needed anymore


jamis:
Thank you!

PR #5801, opened 2024-03-06
MONGOID-5739 Support BSON 5

jamis:
BSON 5 is released, and Mongoid ought to support it.

### DIFF
--- a/mongoid.gemspec
+++ b/mongoid.gemspec
@@ -42,7 +42,7 @@ Gem::Specification.new do |s|
   s.add_dependency("mongo", ['>=2.18.0', '<3.0.0'])
   s.add_dependency("concurrent-ruby", ['>= 1.0.5', '< 2.0'])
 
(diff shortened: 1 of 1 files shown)


mongoid.gemspec
```diff
@@ -42,7 +42,7 @@ Gem::Specification.new do |s|
   s.add_dependency("mongo", ['>=2.18.0', '<3.0.0'])
   s.add_dependency("concurrent-ruby", ['>= 1.0.5', '< 2.0'])
 
-  s.add_development_dependency("bson", ['>=4.14.0', '<5.0.0'])
+  s.add_development_dependency("bson", '>=4.14.0')
```
johnnyshields, line 45: '< 6.0.0' is prudent here, unless you can absolutely guarantee that all future version of BSON will be compatible with this version of Mongoid.

comandeo-mongo (approved):

PR #5806, opened 2024-03-28
MONGOID-5754 Making sure Mongoid works right with latest driver

jamis:
Mongoid's test suite exhibits some new failures when run against the latest version of the driver (2.20).


comandeo-mongo (approved):
👏

PR #5808, opened 2024-04-04
MONGOID-5758: Add Mongoid.reconnect_clients and improve forking webserver documentation

johnnyshields:
This PR does the following:
- Add `Mongoid.reconnect_clients` method analogous to `Mongoid.disconnect_clients`. The reason for adding this is to simplify web server hooks (see added docs.)
- Corrects the `@return` in the docs for `disconnect_clients`. Also added specs which cover the existing behavior.
- Updates documentation related to web server forking.

This will hopefully help others avoid [issues we saw in our production](https://github.com/puma/puma/issues/3134).

See also: https://github.com/mongodb/mongo-ruby-driver/pull/2856

### DIFF
--- a/lib/mongoid.rb
+++ b/lib/mongoid.rb
@@ -100,6 +100,16 @@ module Mongoid
     Clients.disconnect
   end
 
+  # Reconnect all active clients.
+  #
+  # @example Reconnect all active clients.
+  #   Mongoid.reconnect_clients
+  #
+  # @return [ true ] True.
+  def reconnect_clients
+    Clients.reconnect
+  end
+
   # Convenience method for getting a named client.
   #
   # @example Get a named client.
--- a/lib/mongoid/clients.rb
+++ b/lib/mongoid/clients.rb
@@ -47,9 +47,19 @@ module Mongoid
       #
       # @return [ true ] True.
       def disconnect
-        clients.values.each do |client|
-          client.close
-        end
+        clients.each_value(&:close)
+        true
+      end
+
+      # Reconnect all active clients.
+      #
+      # @example Reconnect all active clients.
+      #   Mongoid::Clients.reconnect
+      #
+      # @return [ true ] True.
+      def reconnect
+        clients.each_value(&:reconnect)
+        true
       end
 
       # Get a stored client with the provided name. If no client exists
--- a/docs/reference/configuration.txt
+++ b/docs/reference/configuration.txt
@@ -863,58 +863,57 @@ as the following example shows:
 Usage with Forking Servers
 ==========================
 
-When using Mongoid with a forking web server such as Puma, Unicorn or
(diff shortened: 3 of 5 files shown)


johnnyshields:
FYI I've verified the documentation and changes here in my production Puma app.

lib/mongoid.rb
```diff
@@ -95,11 +95,21 @@ def default_client
...
   def disconnect_clients
     Clients.disconnect
   end
 
+  # Reconnect all active clients.
+  #
+  # @example Reconnect all active clients.
+  #   Mongoid.reconnect_clients
+  #
+  # @return [ Array<Mongo::Client> ] The driver clients.
```
jamis, line 108: Hey @johnnyshields -- here and elsewhere, do you think there's value in actually returning the list of clients? I don't believe the current return value is intentional (or even useful). I think I'd rather leave the documented return value as `true` (or at least "truthy"), rather than be committed to always returning a list of clients here.

lib/mongoid/clients.rb
```diff
@@ -45,11 +45,19 @@ def default
       # @example Disconnect all active clients.
       #   Mongoid::Clients.disconnect
       #
-      # @return [ true ] True.
+      # @return [ Array<Mongo::Client> ] The driver clients.
       def disconnect
-        clients.values.each do |client|
-          client.close
-        end
+        clients.values.each(&:close)
```
jamis, line 50: 👍 thank you, much better.

jamis (changes requested):

lib/mongoid.rb
johnnyshields: @jamis the documentation should match what the method does, and today the method returns the list of clients. It is certainly fine to raise another PR and change the return value, I doubt anyone is relying on this behavior today.

jamis: You're not wrong, Johnny. However, it's also fair to say that the method returns truthy, which is what the documentation today (poorly) implies. I much prefer a truthy return value, than telling the world to expect it to return a list of clients. There are other ways to get the list of clients, and I'd really rather this not be one of them.

johnnyshields: OK so would like me to change this method behavior of both `disconnect` (existing) and `reconnect` (new) in this PR?

jamis: `reconnect` should mirror the behavior of `disconnect`. I'm honestly more worried about the documentation (and the contract it exposes) than the actual implementation, but as long as we're not promising clients more than we want to deliver (e.g. exposing the client list as a return value here), I'm okay.

johnnyshields: @jamis this is now done--all methods in this PR return true.

jamis (approved):
Thank you Johnny. Great work on cleaning up the docs, too. Much appreciated.

PR #5810, opened 2024-04-11
Add pretty print support

stenlarsson:
This adds the `pretty_print` method, which pretty prints the same information as the `inspect` method. This is meant to be called by the default 'pp' gem, e.g. by calling `pp(person)` or `person.pretty_inspect`.

The specs are an almost identical to the specs for the `inspect` method to ensure the same information is included.


jamis:
Hey @stenlarsson, thank you so much for the PR! We really do appreciate you taking the time to contribute to Mongoid.

However, we're not comfortable coupling Mongoid to a third-party gem like `pp`. Including it would require us to add a test-time dependency on the `pp` gem, and to ensure that we maintain this new method through any updates to that gem. That's not an obligation we're willing to take on right now.

Rather, it might make more sense to release this integration as a separate gem, which patches the `pretty_print` method into Mongoid as necessary.

stenlarsson:
@jamis What do you mean "third-party gem"? PP is part of Ruby's standard library.

jamis:
Apologies. The word "gem" threw me; I guess I don't associate "gems" with the standard library! I'll reopen this and discuss with the team. Thank you for clarifying.

lib/mongoid/inspectable.rb
```diff
@@ -19,6 +19,38 @@ def inspect
...
+    def pretty_print(pretty_printer)
+      keys = fields.keys | attributes.keys
+      pretty_printer.group(1, "#<#{self.class.name}", '>') do
+        sep = lambda { pretty_printer.text(',') }
+        pretty_printer.seplist(keys, sep) do |key|
+          pretty_printer.breakable
+          as = if (field = fields[key])
+            field.options[:as]
+          end
+          pretty_printer.text("#{key}#{as ? "(#{as})" : nil}")
```
jamis, line 40: Maybe something like this?

```suggestion
          as = "(#{fields[key].options[:as]})" if fields[key]
          pretty_printer.text("#{key}#{as}")
```

Mostly, the assignment to a trivial `if` statement feels awkward to me; I think it feels more natural to use the `if` statement modifier in that case, even though you wind up duplicating the `fields[key]` query.

lib/mongoid/inspectable.rb
```diff
@@ -19,6 +19,38 @@ def inspect
       "#<#{self.class.name} _id: #{_id}, #{inspection * ', '}>"
     end
 
+    # This pretty prints the same information as the inspect method. This is
+    # meant to be called by the default 'pp' gem.
```
jamis, line 23:
```suggestion
    # meant to be called by the standard 'pp' library.
```

lib/mongoid/inspectable.rb
stenlarsson: This doesn't quite work because `fields[key]` can be nil. I created a variation of your suggestion.

stenlarsson:
Thanks for the review!

I don't know if there is any official documentation explaining what "default gem" means, but this page gives a pretty good explanation: https://stdgems.org/

Should I squash the commits, or is that done when merging the PR?

jamis:
Thank you, @stenlarsson! I wasn't aware of any official terminology around gems/libraries for Ruby; in my brain, "gem" has always meant a third-party thing. TIL. :)

No need to squash the commits; I'll take care of that when I merge. 👍

jamis (approved):

PR #5815, opened 2024-05-10
Mongoize is not called on update_all, when $set operator is used

dem:
Backport to `8.1-stable` branch solution to mogoize values https://github.com/mongodb/mongoid/pull/5814


jamis (approved):
Thank you again for fixing this, and for including a solution for 8.1! 👍

PR #5823, opened 2024-05-20
Fix mongoize update all array operators master

JohnMaguir:
Issue when using `$pull` or `$pop` with `update_all`
Atomic operators that deal with arrays or sets: `$addToSet`, `$push`, `$pull`, `$pop`, `$pullAll`
Only `$addToSet` and `$push` are covered by current tests and `$pull`, `$pop` functionality is broken as `Mongoid::AtomicUpdatePreparer.mongoize_for` returns `nil` for query such as `update_all("$pull" => { genres: "electronic" })`

This became an issue after fixes from https://github.com/mongodb/mongoid/pull/5814 as correct `key` is now being passed to `Mongoid::AtomicUpdatePreparer.mongoize_for`

Fix is to `mongoize` value the same way it is for `$addToSet` or `$push`.
Also added tests for `$pullAll` even though it is currently working but good to have all operators covered.

Corresponding fix for 8.1 branch - https://github.com/mongodb/mongoid/pull/5824


spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -3640,6 +3640,75 @@
...
+            end
+          end
+
+          context "when operation is $pullAll" do
+
+            before do
+              depeche_mode.update_attribute(:genres, ["pop", "electronic", "dance" ])
+              new_order.update_attribute(:genres, ["electronic", "pop", "dance"])
+              context.update_all("$pullAll" => { genres: ["pop", "electronic"] })
+            end
```
dem, line 3702: @JohnMaguir for $pullAll it is better to prepare multiple entries in the data to be pulled from the array, just to cover it fully:
```
              depeche_mode.update_attribute(:genres, ["pop", "electronic", "dance", "pop" ])
              new_order.update_attribute(:genres, ["electronic", "pop", ''electronic","dance"])
```

JohnMaguir: Thank you @dem updated now

jamis (approved):
Good catch! Thank you 🙇

PR #5824, opened 2024-05-20
Fix mongoize update all array operators 8.1

JohnMaguir:
Corresponding 8.1 fix for https://github.com/mongodb/mongoid/pull/5823/files


lib/mongoid/extensions/hash.rb
```diff
@@ -198,7 +198,7 @@ def to_criteria
       def value_for(operator, klass, key, value)
         case operator
         when "$rename" then value.to_s
-        when "$addToSet", "$push" then value.mongoize
+        when "$addToSet", "$push", '$pull', '$pop' then value.mongoize
```
johnnyshields, line 201: `$pullAll`?

jamis: Thanks @johnnyshields. I looked close at this exact question when I was evaluating these PRs, and it's not necessary to specify `$pullAll` here. In the case of `$push`, `$pull`, etc, the value of the operation may different than the declared value of the field itself (because the field is array-valued), and so must be handled specially. For `$pullAll`, the value is an array, for an array-valued field, and so it is handled fine by the default case.

dem: @jamis any chances to see this changes in 8.1-stable and next release 8.1.x?

jamis: Thanks for the nudge, @dem. I'm not sure how this fell through the cracks, but I'll get it merged today.

jamis (approved):

PR #5829, opened 2024-06-10
5508 touch on custom field

adviti-mishra:
This branch adds a spec to touchable_spec.rb to test whether the symbol representing the field to touch provided to the touch option is indeed touched. Two extra models - Band and Label - were also added to touchable_spec_model.rb for this. See https://jira.mongodb.org/browse/MONGOID-5508 for more.


jamis (approved):
Looks great!

PR #5831, opened 2024-06-15
MONGOID-5786: Fix some compatibility issues with Enumerable API

cperezabo:
Hi, in accordance with Ruby's enumerable API, I've added support for initial value when a block is given to `Enumerable#sum`.

Tell me if the provided changes are ok and I could replicate the same changes to `min` and `max` methods.

Cheers!

### DIFF
--- a/lib/mongoid/contextual/aggregable/memory.rb
+++ b/lib/mongoid/contextual/aggregable/memory.rb
@@ -94,7 +94,7 @@ module Mongoid
         #
         # @return [ Numeric ] The sum value.
         def sum(field = nil)
-          return super() if block_given?
+          return super(field || 0) if block_given?
 
           aggregate_by(field, :sum) || 0
         end
--- a/lib/mongoid/contextual/aggregable/memory.rb
+++ b/lib/mongoid/contextual/aggregable/memory.rb
@@ -90,7 +90,8 @@ module Mongoid
         # @example Get the sum for the provided block.
         #   aggregable.sum(&:likes)
         #
-        # @param [ Symbol ] field The field to sum.
+        # @param [ Symbol | Numeric ] field The field to sum, or the intial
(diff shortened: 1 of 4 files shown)


cperezabo:
Hi @jamis, how correct is to specify the `Numeric` type to the parameter? Because in fact, you can pass any kind of object that implements `:+` as initial value, not only `Numeric`. Should it simply be `Object`?

jamis (approved):
Thank you! Good catch, and thanks too for the tests. 👍

cperezabo:
Hi @jamis, how correct is to specify the Numeric type to the parameter? Because in fact, you can pass any kind of object that implements :+ as initial value, not only Numeric. Should it simply be Object?

jamis:
@cperezabo -- good question! You're correct that the nature of Ruby means that anything that quacks like a number (or which responds to `:+`, as you said) could work as a parameter here. The point of the documentation here is less to strictly specify the types, than to imply use cases. Here, we say "Numeric", because of what it suggests about the interface, and less because we strictly require numbers.

If we specified `Object`, instead, it would communicate very little about the expected contract.

In a perfect world, we would document argument's *interface*, which is possible, but gets verbose quickly. For now, specifying `[ Symbol | Numeric ]` is succinct, and suggests the allowed interfaces.

cperezabo:
One more thing, this bug is also present in `min` and `max` methods. Shouldn't have be done in this PR? 
Now that this one is merged, are you going to implement the changes in those methods or do you need me to do that?

jamis:
Fixing them ourselves is not a high priority; unless you intend to address those methods sooner, we'll just add a ticket and put it in our backlog.

cperezabo:
I will do it, I want the API to be correct so I don't have to think "what's going on?" the next time I need to pass an initial value 🥲

PR #5832, opened 2024-06-17
MONGOID-5785 allow named scopes to remove a default scope

jamis:
As currently implemented, named scopes are merged with the default scope (see [here](https://github.com/mongodb/mongoid/blob/a5657dc6044c1f870dc36db3aed3cfdea66e7b66/lib/mongoid/scopable.rb#L297)). This means you cannot declare a named scope that unsets any part of the default scope; if you try, the default scope will simply be merged back into it.

Merging the default scope is not necessary, because any scope you pass in will already be based on the default scope:

```ruby
class Person
  include Mongoid::Document

  default_scope { where(deleted_at: nil) }
  scope :germans, -> { where(country: 'DE') }
end
```

Here, `Person.germans` will always use the default scope, even without it being merged in, because calling `where` will initialize a new `Criteria` object using that default scope. Merging the default scope is redundant.

However, because this has been this way for a very long time (basically since the feature was first introduced prior to 2014), we risk breaking code that inadvertently depends on this behavior. Thus, I've opted to put the fix behind a flag, which we can flip when we eventually release Mongoid 10.

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -171,6 +171,15 @@ module Mongoid
     # See https://jira.mongodb.org/browse/MONGOID-5658 for more details.
     option :around_callbacks_for_embeds, default: false
 
+    # When this flag is false, named scopes cannot unset a default scope.
+    # This is the traditional (and default) behavior in Mongoid 9 and earlier.
+    #
+    # Setting this flag to true will allow named scopes to unset the default
+    # scope. This will be the default in Mongoid 10.
+    #
+    # See https://jira.mongodb.org/browse/MONGOID-5785 for more details.
+    option :allow_scopes_to_unset_default_scope, default: false
+
     # Returns the Config singleton, for use in the configure DSL.
     #
     # @return [ self ] The Config singleton.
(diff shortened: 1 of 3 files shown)


comandeo-mongo (approved):

johnnyshields:
@jamis @comandeo-mongo please roll this back. This behavior should stay as is with no feature flag. It is completely counter-intuitive that using a named scope would magically cause the default scope to disappear.

https://jira.mongodb.org/browse/MONGOID-5785?focusedCommentId=6461746&focusedId=6461746&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6461746

johnnyshields:
Correction: I misunderstood the change here. Since it's only calling ".unscoped" within a named scope it should be fine.

stanley90:
Hi, I'm a bit late to the party (was on vacation), just want to pinpoint that using `unscoped` does **not** solve the problem I described, because it removes everything, whereas I only want to remove a specific part of the default scope (paranoia, precisely). I'll take a closer look at this when I'm back at work. Thanks

jamis:
@stanley90 -- thanks for chiming in. The PR includes tests that confirm correct behavior using `unscoped`, as well as with `remove_scoping` (see https://github.com/mongodb/mongoid/pull/5832/files#diff-d374f75265fac28314512810734ae94770db90759b5b05d5f76e546e01acecfaR1258-R1259). I'm confident it works correctly in both cases.

stanley90:
@jamis I patched the `Scopable.define_scope_method` in my project and confirm it does solve the problem. I'll keep the patch until I'm able to upgrade Mongoid 😀 Thanks

PR #5838, opened 2024-07-12
MONGOID-5789 Allow nil attribute access (backport of #5836)

jamis:
Allow `nil` attribute access (returning `nil` instead of raising an obscure exception). See #5836.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -197,6 +197,8 @@ jobs:
       with:
         ruby-version: "${{matrix.ruby}}"
         bundler: 2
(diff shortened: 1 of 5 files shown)


jamis:
The evergreen failures are due to rockylinux8 not being a fully-compatible replacement for rhel80 (and we can't use rhel80 because we're running the tests in docker, and there is no docker image available for rhel80).

The GH action failures are almost certainly due to this: https://github.com/actions/runner-images/issues/10215

comandeo-mongo (approved):

PR #5845, opened 2024-07-29
MONGOID-5734 Custom polymorphic types

jamis:
This PR adds support for custom polymorphic types via a global registry. While polymorphic type keys still default to the fully qualified name of the class, it is now possible to specify different aliases by which a polymorphic type might be identified in the database.

For example, consider a simple data model involving a manager, a department, and a team:

```ruby
class Department
  include Mongoid::Document
  has_many :managers, as: :unit
end

class Team
  include Mongoid::Document
  has_one :manager, as: :unit
end

class Manager
  include Mongoid::Document
  belongs_to :unit, polymorphic: true
end
```

Each document in the `managers` collection will include two fields, `unit_id` and `unit_type`, where the `unit_type` field names the class (and thus, indirectly, the collection) of the unit that the manager is in charge of. For example:

```ruby
team = Team.create
tina = Manager.new(unit: team)
```

Inspecting `tina.unit_type` would return the string `"Team"`.

With this PR, programmers may now specify alternative keys to represent different classes, thus decoupling the code from the data. For example:

```ruby
class Department
  include Mongoid::Document
  identify_as 'dept'
  has_many :managers, as: :unit
end
```

The `identify_as 'dept'` directive says that this class should be identified by the string `"dept"` in the database. You can specify multiple aliases, too (e.g. `identify_as 'dept', 'div', 'agency'`), in which case the first key is the "default", and the others are only used for looking up records. This lets you refactor your code without breaking the associations in your data.

It is important to note that these aliases are _global_. The keys you specify must be unique across your entire code base. However, it is possible to register alternate _resolvers_, which may be used for different subsets of your models. In this case, the keys must only be unique for each resolver. For example:

```ruby
Mongoid::ModelResolver.register_resolver Mongoid::ModelResolver.new, :eng
Mongoid::ModelResolver.register_resolver Mongoid::ModelResolver.new, :purch

module Engineering
  class Department
    include Mongoid::Document
    identify_as 'dept', resolver: :eng
  end

  class Manager
    include Mongoid::Document
    belongs_to :unit, polymorphic: :eng
  end
end

module Purchasing
  class Department
    include Mongoid::Document
    identify_as 'dept', resolver: :purch
  end

  class Manager
    include Mongoid::Document
    belongs_to :unit, polymorphic: :purch
  end
end
```

Both `Engineering::Department` and `Purchasing::Department` are aliased as `"dept"`, but use their own resolver, so there is no conflict.


alexbevi (approved):
API and developer experience LGTM

comandeo-mongo (approved):

PR #5848, opened 2024-07-31
MONGOID-5734 Custom polymorphic types (backport to 8.1-stable)

jamis:
This is a backport of #5845 to 8.1-stable. Please see that PR for full information.

### DIFF
--- a/lib/mongoid/association/nested/one.rb
+++ b/lib/mongoid/association/nested/one.rb
@@ -50,11 +50,24 @@ module Mongoid
           @attributes = attributes.with_indifferent_access
           @association = association
           @options = options
-          @class_name = options[:class_name] ? options[:class_name].constantize : association.klass
+          @class_name = class_from(options[:class_name])
           @destroy = @attributes.delete(:_destroy)
         end
 
         private
+
+        # Coerces the argument into a class, or defaults to the association's class.
+        #
+        # @param [ String | Mongoid::Document | nil ] name_or_class the value to coerce
+        #
+        # @return [ Mongoid::Document ] the resulting class
+        def class_from(name_or_class)
+          case name_or_class
+          when nil, false then association.klass
+          when String then name_or_class.constantize
+          else name_or_class
+          end
+        end
 
         # Extracts and converts the id to the expected type.
         #
--- a/lib/mongoid/association/referenced/belongs_to.rb
+++ b/lib/mongoid/association/referenced/belongs_to.rb
@@ -102,6 +102,21 @@ module Mongoid
           @polymorphic ||= !!@options[:polymorphic]
         end
 
+        # Returns the object responsible for converting polymorphic type references into
+        # class objects, and vice versa. This is obtained via the `:polymorphic` option
+        # that was given when the association was defined.
+        #
+        # See Mongoid::ModelResolver.resolver for how the `:polymorphic` option is 
+        # interpreted here.
+        #
+        # @raise KeyError if no such resolver has been registered under the given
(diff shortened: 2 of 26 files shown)


lib/mongoid/errors/unrecognized_model_alias.rb
```diff
@@ -0,0 +1,53 @@
...
+    #       identify_as 'Department::Engineering'
+    #
+    #       # ...
+    #     end
+    #   end
+    #
+    # Better practice would be to use unique strings instead of class names to
+    # identify these polymorphic types in the database (e.g. 'dept' instead of
+    # 'Department::Engineering').
+    class UnrecognizedModelAlias < MongoidError
```
johnnyshields, line 42: It might be nice to rename these new errors to include the word `Polymorphic` in their name, e.g. `UnrecognizedPolypmorphicAlias`. From their names they sound more generic than they actually are--the code docs indicate they are very specific to polymorphism.

lib/mongoid/identifiable.rb
```diff
@@ -0,0 +1,28 @@
...
+      # by their class names, but alternative aliases may be used instead,
+      # if desired.
+      #
+      # @param [ Array<String | Symbol> ] aliases the list of aliases to
+      #   assign to this class.
+      # @param [ Mongoid::ModelResolver::Interface | Symbol | :default ] resolver the
+      #   resolver instance to use when registering the type. If :default, the default
+      #   `ModelResolver` instance will be used. If any other symbol, it must identify a
+      #   previously registered ModelResolver instance.
+      def identify_as(*aliases, resolver: :default)
```
johnnyshields, line 23: How about calling this method `polymorphic_as` instead? `indentify_as` is quite generic, and may make users think think it has something to do with the `_id` field (which it does not.)

lib/mongoid/association/nested/one.rb
```diff
@@ -50,12 +50,25 @@ def initialize(association, attributes, options)
           @attributes = attributes.with_indifferent_access
           @association = association
           @options = options
-          @class_name = options[:class_name] ? options[:class_name].constantize : association.klass
+          @class_name = class_from(options[:class_name])
```
johnnyshields, line 53: This variable should be refactored to be called `@klass` instead of `@class_name`, because it's not a name String--its an actual `Class` object.

lib/mongoid/association/referenced/belongs_to.rb
```diff
@@ -102,6 +102,21 @@ def polymorphic?
...
+        #
+        # See Mongoid::ModelResolver.resolver for how the `:polymorphic` option is 
+        # interpreted here.
+        #
+        # @raise KeyError if no such resolver has been registered under the given
+        #   identifier.
+        #
+        # @return [ nil | Mongoid::ModelResolver ] the resolver to use
+        def resolver
+          @resolver ||= Mongoid::ModelResolver.resolver(@options[:polymorphic])
```
johnnyshields, line 117: Same here `ModelResolver` --> `PolymorphicModelResolver`

johnnyshields:
Jamis this a great feature to have!! Thank you for it, something I've wanted for a long time.

Please see comments, I think clear/specific naming could greatly enhance the developer experience and reduce potential for confusion/conflicts/heartbreak down the line.

johnnyshields:
Hmmm... STI is another form of polymorphism, so the name `polymorphic_*` still works I think.

I would assert that the names you have today are so generic as to be easily confused with things completely unrelated to polymorphic relations or collections.

PR #5874, opened 2024-10-02
MONGOID-4680 implement cache_version for better Rails caching support

jamis:
This PR adds a `#cache_version` method to `Mongoid::Document` instances, as well as to the `has_many` and `embeds_many` proxies, which brings Mongoid closer to API parity with ActiveRecord.

This also modifies `#cache_key` to better parallel how it works in ActiveRecord: if `cache_version` returns non-nil, `cache_key` will *not* embed the timestamp in the key. Instead, the timestamp is used as the version for the entry (which reduces key churn).

**Note**: this PR does slightly change the behavior of Mongoid when used with Rails caching (`#cache_key` will not embed the timestamp when `#cache_version` is non-nil, as mentioned). We discussed this as a team and do not believe this constitutes a breaking change; in the worst case, systems that depend on caching may find at least some of their caches invalidated the first time they deploy this change.

For systems where the new cache key format is a concern, you can "revert" to the original cache key format by defining a `#cache_version` of `nil`:

```ruby
def cache_version = nil

# or, for older, less glamorous, more boring and verbose rubies:
def cache_version; nil; end
```

### DIFF
--- a/lib/mongoid/association/many.rb
+++ b/lib/mongoid/association/many.rb
@@ -178,6 +178,19 @@ module Mongoid
         criteria.unscoped
       end
 
+      # For compatibility with Rails' caching. Returns a string based on the
+      # given timestamp, and includes the number of records in the relation
+      # in the version.
+      #
+      # @param [ String | Symbol ] timestamp_column the timestamp column to
+      #   use when constructing the key.
+      #
+      # @return [ String ] the cache version string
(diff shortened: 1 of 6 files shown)


jamis:
Well, TIL that `$count` as a group aggregation operator was only introduced in server 5.0. I need to find a workaround for earlier server versions before this can be merged.

PR #5877, opened 2024-10-09
MONGOID-5411 allow results to be returned as demongoized hashes

jamis:
Mongoid now supports a new `raw` directive when building queries. When present, the resulting query will be returned as hashes, rather than instantiated document models. The hashes will not be "demongoized" -- the values will be returned exactly as provided by the database, directly.

If you specify `typed: true`, the hashes will be returned with the values translated from the raw value stored in the database, into the corresponding type defined by the `field` definitions on the model. (That is to say, they will be "demongoized".)

For example:

```ruby
result = Person.raw.first
p result  #=> {"_id"=>BSON::ObjectId(...), "name" => "...", "birth_date" => 2002-01-01 00:00:00 -0600, ... }

result = Person.raw(typed: true).first
p result  #=> {"_id"=>BSON::ObjectId(...), "name" => "...", "birth_date" => 2002-01-01, ... }

results = Person.where(...).limit(...).raw
p results.to_a #=> [ { "_id"=>BSON::ObjectId(...), ... }, { "_id"=>BSON::ObjectId(...), ... }, ... ]
```

The `typed: true` option will also honor embedded documents, correctly demongoizing the embedded hashes according to their declared types.

### DIFF
--- a/lib/mongoid/contextual/mongo.rb
+++ b/lib/mongoid/contextual/mongo.rb
@@ -883,7 +883,11 @@ module Mongoid
         doc = if document.respond_to?(:_id)
                 document
               elsif criteria.raw_results?
-                demongoize_hash(klass, document)
+                if criteria.typecast_results?
+                  demongoize_hash(klass, document)
+                else
+                  document
+                end
               else
                 Factory.from_db(klass, document, criteria)
               end
@@ -1062,7 +1066,11 @@ module Mongoid
       #   single document.
       def process_raw_docs(raw_docs, limit)
         docs = if criteria.raw_results?
-                 raw_docs.map { |doc| demongoize_hash(klass, doc) }
+                 if criteria.typecast_results?
+                   raw_docs.map { |doc| demongoize_hash(klass, doc) }
+                 else
+                   raw_docs
+                 end
                else
                  mapped = raw_docs.map { |doc| Factory.from_db(klass, doc, criteria) }
                  eager_load(mapped)
--- a/lib/mongoid/contextual/mongo.rb
+++ b/lib/mongoid/contextual/mongo.rb
@@ -989,6 +989,10 @@ module Mongoid
       # of the given hash as if it were the raw representation of a document of
       # the given klass.
       #
+      # @note this method will modify the given hash, in-place, for performance
+      # reasons. If you wish to preserve the original hash, duplicate it before
+      # passing it to this method.
+      #
       # @param [ Document ] klass the Document class that the given hash ought
       #   to represent
       # @param [ Hash | nil ] hash the Hash instance containing the values to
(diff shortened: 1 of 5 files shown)


johnnyshields:
Jamis, thank you, this is super useful.

Please kindly check the following:
1. Combining with .only/.exclude projections, eg Customer.only(:name, :age).raw
2. Consider making this .format(:raw) in case there are other potential formats to return. I think it may be useful to have options for demongoized raw vs non-demongized raw (what is actually in DB), as the latter is the best performance strictly speaking.
3. Please check that all enumerable methods like .each, etc work in a progressive loading fashion (using GETMORE)
4. Please check that embedded models are also demongized.

jamis:
> 1. Combining with .only/.exclude projections, eg Customer.only(:name, :age).raw

Good suggestion. I've added a few tests for that [here](https://github.com/mongodb/mongoid/pull/5877/files#diff-56ee0ff139cc683178654f38c4899e8b2c716c60c79c8c0a681ed76fa997526aR2327-R2350).

> 2. Consider making this .format(:raw) in case there are other potential formats to return. I think it may be useful to have options for demongoized raw vs non-demongized raw (what is actually in DB), as the latter is the best performance strictly speaking.

I think `#raw` reads better from a DSL perspective. If, eventually, we want the mongoized values directly, we can add that as a parameter to `#raw` (e.g. `raw(:db)` or some such). For now, I think taking that next step is out of scope for this feature, though.

> 3. Please check that all enumerable methods like .each, etc work in a progressive loading fashion (using GETMORE)

There are already tests for this `mongoid/contextual/mongo_spec.rb`, but I [tweaked them a bit](https://github.com/mongodb/mongoid/pull/5877/files#diff-684e0fef7b9f01808bbf37c8984046259173f50e0a6c51e42c5ba97a5d84ec4bL1240-L1253) to make it clearer that `#each` plays nicely with the underlying cursor.

> 4. Please check that embedded models are also demongized.

This is [already done](https://github.com/mongodb/mongoid/pull/5877/files#diff-56ee0ff139cc683178654f38c4899e8b2c716c60c79c8c0a681ed76fa997526aR2307-R2325), as well.

Thanks for the feedback, @johnnyshields!

jamis:
I did a couple of simple benchmarks on this, to get some idea of just how much difference this makes. The benchmarks were:

1. "Gizmo." Loading a thousand models ("gizmos") with no embedded records.
2. "Room." Loading a thousand models ("rooms") with 30 embedded records each ("furnishings"). Each room also made sure to instantiate its embedded furnishings.

Each benchmark was performed a thousand times, and the elapsed time reported.

The results:

```
Gizmo: Instances :: 10.239385
Gizmo: Raw       ::  9.867614

Room: Instances :: 758.110451
Room: Raw       :: 144.602256
```

Note that if the furnishings were not requested, the "Room: Instances" benchmark ran much quicker, in under 80 seconds or so.

## Conclusions

Returning just the demongoized hashes makes very little difference in performance for records with no embedded children.

Returning demongoized hashes for records with embedded children *hurts* performance, if you do not intend to access the embedded children.

Returning demongoized hashes is significantly *better* when querying records with embedded children, when you intend to access those embedded children.

johnnyshields:
For benchmarks, try:
- Adding 100 fields to objects (Gizmo / Room)
- Do benchmarks with `.only` projection on 3 fields.
- Try looking at memory and object allocations as well.

One of my main use cases for this is reporting, where I'm getting 1,000,000+ objects and putting them in a CSV. Currently I'm using the `#pluck_each` logic in this PR heavily for it: https://github.com/mongodb/mongoid/pull/5497 which is definitely faster than loading objects, takes roughly 1/10 the time to run the full report (assuming I'm only plucking the fields I need)

jamis:
@johnnyshields -- I've updated my previous comment to include some basic memory profiling.

comandeo-mongo (approved):

PR #5878, opened 2024-10-10
MONGOID-5820 Fix the session object leak when using transactions

Virus-X:
## The issue
I found the memory leak inside `Threaded` class, when saving documents within transaction: session object is retained forever inside thread-local variable.

> Affected version: Mongoid **9.0.2** / Ruby **3.3.4**

## Explanation
* Mongoid uses thread-local variables to temporarily hold `modified_documents` until transaction is committed. 
* The `[mongoid]:modified-documents` contains a hash with `session` as key and Set of documents as value. 
* During transaction commit phase, the `Mongoid::Threaded::clear_modified_documents` is called, which clears documents from the set, however, the hash inside `[mongoid]:modified-documents` still contains a session object as a key.
* This session object is then persists in memory, even after the session was finished by the `Mongoid::Threaded::clear_session` call.

## Testing
I used the following code to reproduce the issue:
``` ruby
require_relative 'spec_helper'
require 'memory_profiler'

describe "Mongoid test" do
  let(:model) do
    Class.new do
      include Mongoid::Document

      store_in collection: 'tests'
      field :name, type: String
    end
  end

  context 'memory leak test' do
    it 'store in transaction' do
      run_memory_profiling do
        model.transaction do
          model.create!(name: 'test')
        end
      end
    end
  end

  def run_memory_profiling(&block)
    # Warmup
    yield

    report = MemoryProfiler.report do
      2000.times(&block)
      GC.start(full_mark: true, immediate_sweep: true)
    end
    p "Total retained: #{report.total_retained_memsize / 1024} KB (#{report.total_retained} objects). "
  end
end
```

### Initial results:
```
Total retained: 3430 KB (36053 objects). 
```
Report shows a number of retained objects, which corresponds to 2000 iterations of test:
image

### Results after modification:
```
Total retained: 6 KB (61 objects).
```

### DIFF
--- a/spec/lite_spec_helper.rb
+++ b/spec/lite_spec_helper.rb
@@ -82,13 +82,13 @@ RSpec.configure do |config|
     config.add_formatter(RSpec::Core::Formatters::JsonFormatter, File.join(File.dirname(__FILE__), '../tmp/rspec.json'))
   end
 
-  if SpecConfig.instance.ci? && !%w(1 true yes).include?(ENV['INTERACTIVE']&.downcase)
-    config.around(:each) do |example|
-      TimeoutInterrupt.timeout(example_timeout_seconds) do
-        example.run
-      end
-    end
-  end
+  # if SpecConfig.instance.ci? && !%w(1 true yes).include?(ENV['INTERACTIVE']&.downcase)
+  #   config.around(:each) do |example|
+  #     TimeoutInterrupt.timeout(example_timeout_seconds) do
+  #       example.run
+  #     end
+  #   end
+  # end
 
   def local_env(env = nil, &block)
     around do |example|
--- a/spec/lite_spec_helper.rb
+++ b/spec/lite_spec_helper.rb
@@ -82,13 +82,13 @@ RSpec.configure do |config|
     config.add_formatter(RSpec::Core::Formatters::JsonFormatter, File.join(File.dirname(__FILE__), '../tmp/rspec.json'))
   end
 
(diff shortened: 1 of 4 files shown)


spec/lite_spec_helper.rb
```diff
@@ -82,13 +82,13 @@ def example_timeout_seconds
...
-      end
-    end
-  end
+  # if SpecConfig.instance.ci? && !%w(1 true yes).include?(ENV['INTERACTIVE']&.downcase)
+  #   config.around(:each) do |example|
+  #     TimeoutInterrupt.timeout(example_timeout_seconds) do
+  #       example.run
+  #     end
+  #   end
+  # end
```
jamis, line 91: Are these lines commented out because they might come back? Should we just delete them and rely on git to remember them if we need to resurrect them?

jamis (approved):
Looks good! Just one question about the commented-out bit in lite_spec_helper.rb, but I've got no material concerns.

spec/lite_spec_helper.rb
comandeo-mongo: Oh, good catch! This should be removed eventually, I believe, but not in scope of this PR. I'll revert.

comandeo-mongo:
Hi @Virus-X , thank you very much for your contribution! I am sorry it takes me so much time to merge it. The reason was that we needed a test for the change; and while adding the test I discovered one more related issue that was also fixed.

Virus-X:
Thanks! Was glad to help!

PR #5900, opened 2024-11-12
Bump version to 8.1.7

comandeo-mongo:


alcaeus (approved):
NB: it would be nice if we could somehow automate this. We have similar logic in [the PHP extension](https://github.com/mongodb/mongo-php-driver/blob/v1.x/bin/update-release-version.php) that we invoke during the release process.

PR #5908, opened 2024-12-02
MONGOID-5831 Remove usage of Rails private API

comandeo-mongo:


jamis (approved):
Nice! Is it worth declaring `__redefine` as `@api private`? Or is the double underscore sufficient to indicate that?

PR #5911, opened 2024-12-05
MONGOID-5780 Fix chaining nots resulting in incorrect negation state

DarshanaVenkatesh:
MONGOID-5780
Ensure that chaining "not" results in the correct negating state.


jamis (approved):
Nice! 👍

PR #5912, opened 2024-12-06
MONGOID-5782 Ensure embedded documents are also marked timeless

DarshanaVenkatesh:
[MONGOID-5782](https://jira.mongodb.org/browse/MONGOID-5782)

When a parent doc is set to timeless, this should also keep the embedded document as timeless. This is a bug fix as previously, the updated_at time would get set to current.


jamis (approved):
Great sleuthing! This looks good. 👍

jamis (changes requested):
(Updating this to officially revoke my approval of the PR, per our discussions.)

PR #5918, opened 2024-12-11
MONGOID-5816: attr_readonly leaks into sibling classes

DarshanaVenkatesh:
MONGOID-5816

When we have two classes A and B that are subclasses from Base, declaring attr_readonly in the context of A or B leaks that declaration into the entire class hierarchy. Currently, all classses share the same readonly attributes at the end of declaration. Fixed so that A and B inherit the read only attributes of Base but including additional readonly attributes in the subclasses is instance specific.


jamis (approved):
💪

PR #5921, opened 2024-12-13
MONGOID-5809: remove ruby -w warnings

DarshanaVenkatesh:
When run with ruby -w, Mongoid is fairly noisy and provides complaints about method redefinitions and overrides. Remove warnings by preventing redefinition.


lib/mongoid/serializable.rb
```diff
@@ -13,6 +13,8 @@ module Serializable
     included do
 
       class << self
+         remove_method :include_root_in_json if method_defined?(:include_root_in_json)
+         remove_method :include_root_in_json= if method_defined?(:include_root_in_json=)
```
jamis, line 17: Nice! I had to dig a bit to understand that these methods are defined in ActiveModel, and we're just re-defining them here to account for our default value `Mongoid.include_root_in_json`. Maybe add a comment to that effect?

Also, there's an extra space before these lines -- we've standardized on indentations in 2-space increments.

lib/mongoid/traversable.rb
```diff
@@ -179,8 +179,8 @@ def discriminator_value
 
     included do
       class_attribute :discriminator_key, instance_accessor: false
-
       class << self
+        remove_method :discriminator_key if method_defined?(:discriminator_key)
```
jamis, line 183: Same here -- please add a comment explaining that the preceding `class_attribute` creates both a getter and a setter, but we need to reimplement the getter (by delegating).

Also, the `if method_defined?(...)` is unnecessary here, since we know that the preceding `class_attribute` ensures the method *is* defined. 👍

jamis (changes requested):

jamis (approved):
💯

PR #5922, opened 2024-12-18
MONGOID-5822: Embedded association validations break when association depends on parent update

DarshanaVenkatesh:
MONGOID-5822

In the case where there is an association with validation dependent on a parent condition, the document is not marked as invalid if the parent update results in an invalid association unless there was also a change in the child.


jamis (approved):
👍 very nice!

PR #5924, opened 2024-12-18
MONGOID-5822: Embedded association validations break when association depends on parent update (backport for 8.1)

DarshanaVenkatesh:
[MONGOID-5822](https://jira.mongodb.org/browse/MONGOID-5822)

In the case where there is an association with validation dependent on a parent condition, the document is not marked as invalid if the parent update results in an invalid association unless there was also a change in the child.


jamis (approved):

EduardoRSeifert:
just so you guys know: i tried to monkey patch this in and i'm getting an infinite loop. i still haven't been able to track down why, exactly, however

PR #5934, opened 2025-01-15
MONGOID-5839 Fix eager-loading from STI subclasses

jamis:
If the root of a query is an STI subclass (e.g. `Subclass.all`) AND the query tries to eager load (`includes`) another association, the eager load was failing because it was looking for inverse associations using the STI subclass name, instead of the class at the root of the hierarchy.

### DIFF
--- a/lib/mongoid/association/eager_loadable.rb
+++ b/lib/mongoid/association/eager_loadable.rb
@@ -42,6 +42,9 @@ module Mongoid
         docs_map = {}
         queue = [ klass.to_s ]
 
+        # account for single-collection inheritance
+        queue.push(klass.root_class.to_s) if klass != klass.root_class
+
         while klass = queue.shift
           if as = assoc_map.delete(klass)
             as.each do |assoc|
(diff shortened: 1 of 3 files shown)


GCorbel:
I just tried and the fix work with Mongoid 8.0.8. Thanks.

I had to change because the model I used didn't include `Traversable` so I replaced `queue.push(klass.root_class.to_s) if klass != klass.root_class` by

```
      root_class = self.klass
      root_class = root_class.superclass while root_class.hereditary?

      queue.push(root_class.to_s) if klass != root_class
```

jamis:
@stanley90 sorry, some illness and a long weekend left this one in a bit of limbo. I hope to push it across the finish line (and backport it as necessary) in the next couple of days.

comandeo-mongo (approved):

stanley90:
Hi, me again. It's been fixed for a while, but could we get a release? It's biting me quite a bit. Thx

jamis:
8.0.9 has been released. Patch releases for 8.1.x and 9.0.x are forthcoming (probably tomorrow).

jamis:
8.1.9 and 9.0.5 are released now, as well:  https://www.mongodb.com/community/forums/t/mongoid-maintenance-releases-now-available-9-0-5-8-1-9-8-0-9/311172

stanley90:
Thanks!

PR #5939, opened 2025-01-21
MONGOID-5825 Do not update timestamp on destroyed

comandeo-mongo:


jamis (approved):

johnnyshields:
This is likely going to break apps which use `Mongoid::Paranoia` (which my app does heavily). I think it makes sense to consider destroying a form of update; many times you want to use the `updated_at` value to check if there has been any form of update on an object, including destroying it.

jamis:
@johnnyshields, can you clarify how this will affect `Mongoid::Paranoia`? All this PR does is prevent the timestamps from being applied to a frozen object. Is there a valid situation where timestamping a frozen object is an acceptable action?

comandeo-mongo:
@johnnyshields do you use this one - https://github.com/simi/mongoid_paranoia ?

johnnyshields:
Yep, I use that one. Come to think of it, I just remembered I monkey patched my app to set updated_at when things are deleted, so maybe this wont break anything after all…

Still, I wonder if there are other possible solutions to the original ticket?

comandeo-mongo:
I ran `mongoid_paranoia` tests agains this branch, and they all pass. I understand that this is not a 100% guarantee, but still a good sign.

As to the change itself - I agree with @jamis , I do not see why it can be destructive. Without this PR an attempt to set timestamps raises `can't modify frozen Hash` error. With this PR there is no attempt to set timestamps. So, `created_at` and `updated_at` are not set in both cases. I can hardly imagine someone depends on this particular error.

johnnyshields:
@comandeo-mongo but why does the hash become frozen in the first place? Is it not possible to modify a "deleted" object in Mongoid?

jamis:
> @comandeo-mongo but why does the hash become frozen in the first place? Is it not possible to modify a "deleted" object in Mongoid?

The freeze occurs here: https://github.com/mongodb/mongoid/blob/master/lib/mongoid/persistable/deletable.rb#L111

It's been this way for at least 12 years, since before 682bcd9c53 . It should also be noted that ActiveRecord behaves similarly (https://github.com/rails/rails/blob/fc906ea0b9aef363d3443e631108d027bb850092/activerecord/lib/active_record/persistence.rb#L459).

This isn't something we have any appetite to change right now, and it will take a lot to convince us to break parity with Rails here, but if you feel strongly about it, please do research it and see what the implications are for leaving a destroyed object unfrozen.

johnnyshields:
@jamis why do you think Mongoid and Rails freeze the inner hash on destroy?

jamis:
My guess is that it's to prevent bugs caused by modifying a destroyed object, and perhaps even re-persisting it by accident. A destroyed object is, by definition, deleted, and marking it read-only makes a lot of sense to me. If I needed to propagate additional information about a deleted object, I'd wrap the deleted object in a hash alongside the new metadata, keeping that deleted object isolated and untouched.

johnnyshields:
Hmmm... I was thinking it was probably primarily to conserve process memory.

Theoretically there would be no "harm" in modifying a deleted object in Ruby after you've persisted it... it's not a standard use case, but I wouldn't expect the ORM/ODM to prevent it. If that's the intended behavior, then I'd think there should be a proper `Mongoid::Error::CannotModifyDeletedModel` rather than letting the `Cannot modify frozen hash` native Ruby error bubble up in a manner which is confusing/unintuitive to users.

comandeo-mongo:
It's hard to say now whether it was an intentional behaviour from the beginning, but I think it might be. A modification of a destroyed object probably contradicts the idea of Active Record pattern itself. An instance of an Active Record object represents a record in the database. As long as the object is destroyed, the corresponding record is removed, and there is nothing to "represent". The object is out of responsibility of the Active Record machinery, even though it is still accessible. So, I believe that this constraint makes sense.

PR #5964, opened 2025-03-11
MONGOID-5843 Ensure BSON::Decimal128 is considered to be numeric (Backport to 8.1)

jamis:
Backport to 8.1-stable:

Because ActiveModel's `validate_numericality_of` validator looks at the value of a field _before_ it was typecast, it fails when the underlying type is a `BSON::Decimal128` (which is not a true numeric type). This PR specializes ActiveModel's `NumericalityValidator` so that it recognizes `BSON::Decimal128` as a numeric type.


GCorbel:
I confirm this works with Rails 7.0 and Mongoid 8.1

comandeo-mongo (approved):

PR #5968, opened 2025-03-20
MONGOID-5852 Fix after_action_commit callbacks

ellnix:
None of them worked due to a syntax error where
`set_options_for_callbacks` did not accept a second argument.

I thought this fix was relatively straightforward enough to be a quick PR without going through support. Apologies if I did something incorrectly, I'm new to Mongo in general.

### DIFF
--- a/lib/mongoid/clients/sessions.rb
+++ b/lib/mongoid/clients/sessions.rb
@@ -213,8 +213,8 @@ module Mongoid
 
         # Transforms custom options for after_commit and after_rollback callbacks
         # into options for +set_callback+.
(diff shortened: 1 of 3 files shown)


ellnix:
The change to `lib/` is pretty much the same as Rails' version:

https://github.com/rails/rails/blob/e9d7ca5e0b834ff2ce3d43965deac670ff8ac790/activerecord/lib/active_record/transactions.rb#L330

comandeo-mongo (approved):

jamis (approved):
Good catch! Thanks for the PR. :+1:

PR #5969, opened 2025-03-21
MONGOID-5852 Fix after_action_commit callbacks (#5968)

comandeo-mongo:
This is the backport of MONGOID-5852.


johnnyshields:
Any reason this is not merged?

PR #5975, opened 2025-04-15
MONGOID-5863 `#last` overrides `#skip`

jamis:
Prior to Mongoid 8.1, `Model.skip(1).last` would return the next-to-last document. A refactoring in 8.1 resulted in this breaking, due to an extraneous call to `#skip` that would override any other invocation of `#skip`.

## Summary

Fixes `Model.skip(1).last` so it again returns the next-to-last document, instead of just the last document.

### DIFF
--- a/spec/mongoid/contextual/mongo_spec.rb
+++ b/spec/mongoid/contextual/mongo_spec.rb
@@ -3292,6 +3292,12 @@ describe Mongoid::Contextual::Mongo do
     it "limits the results" do
       expect(context.skip(1).entries).to eq([ new_order ])
     end
+
+    context "with #last" do
+      it "returns the nth from last element" do
+        expect(context.skip(1).last).to eq(depeche_mode)
+      end
+    end
   end
 
   describe "#sort" do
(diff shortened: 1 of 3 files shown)


alcaeus (approved):
For the record, I'm unsure why this is expected to return the nth-from-last element, but I agree with fixing the inadvertent BC break and look at this separately.

johnnyshields:
By the way, I still haven't upgraded from Mongoid 7 due to concern over bugs like this. I am thinking to try upgrading soon, I will likely find more things.

johnnyshields:
Regression found, as expected: https://jira.mongodb.org/browse/MONGOID-5865

spec/mongoid/contextual/mongo_spec.rb
```diff
@@ -3292,6 +3292,12 @@
     it "limits the results" do
       expect(context.skip(1).entries).to eq([ new_order ])
     end
+
+    context "with #last" do
+      it "returns the nth from last element" do
+        expect(context.skip(1).last).to eq(depeche_mode)
+      end
+    end
```
johnnyshields, line 3300: The method has `sort(inverse_sorting)` and then later `v.to_a.reverse`, which ensure the ordering in the case hat multiple docs are returned.

Are there cases for the following, which ensure correct document ordering the result for multiple docs?
- combining `skip(n)` with `limit(m)` (where n, m > 1)
- `last(n)` (where n > 1)

PR #5976, opened 2025-05-05
[WIP] MONGOID-5865 - BSON::Document should be correctly demongoized into a Hash

johnnyshields:
This PR fixes a regression from Mongoid 7.

```ruby
def MyModel
  include Mongoid::Document
  field :my_hash, type: Hash
end

doc = MyModel.new(my_hash: { foo: { bar: 1 } }

doc = MyModel.all.first

# Mongoid 7.x:
doc.my_hash.class #=> Hash -- CORRECT
doc.my_hash['foo'].class #=> Hash -- CORRECT 

# Mongoid 9.x:
doc.my_hash.class #=> Mongoid::Document -- BAD
doc.my_hash['foo'].class #=> Mongoid::Document -- BAD
```

Three cases to consider:
1. When `field :foo, type: Hash` is explicitly declared.
2. When using a dynamic Hash field
3. Other cases, e.g. Range which is represented as a Hash { min: 1, max: 7 }

The code here currently only fixes case #1. Need to check the behavior of the other two cases on Mongoid 7.

### DIFF
--- a/lib/mongoid/extensions/hash.rb
+++ b/lib/mongoid/extensions/hash.rb
@@ -121,12 +121,39 @@ module Mongoid
         #
         # @return [ Hash | nil ] The object mongoized or nil.
         def mongoize(object)
-          return if object.nil?
           case object
           when BSON::Document
+            # TODO: This can call object.transform_values after RUBY-3664
+            # is included in Mongoid's minimum BSON version.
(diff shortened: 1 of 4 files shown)


johnnyshields:
Hmm... this will not be a good solution

PR #5979, opened 2025-05-09
Add 'tzinfo-data' for Windows development

johnnyshields:
This is needed to run Mongoid tests (and any Rails gem test) on Windows. Rails adds it automatically when you do rails new on windows. Since it is in the **Gemfile** (**not** the gemspec) it doesn't affect anyone's actual code, it is purely for development purposes.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,3 +17,7 @@ if RUBY_PLATFORM =~ /java/
 end
 
 gem 'i18n', *i18n_versions
+
+platform :windows do
+  gem 'tzinfo-data'
+end


Gemfile
```diff
@@ -17,3 +17,7 @@ if RUBY_PLATFORM =~ /java/
 end
 
 gem 'i18n', *i18n_versions
+
+platform :windows do
+  gem 'tzinfo-data'
+end
```
alexbevi, line 23:
```suggestion
# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
gem "tzinfo-data", platforms: %i[ windows jruby ]
```

Maybe we should mirror what the Rails generator produces? I'm not opposed to only doing this for Windows but figure it couldn't hurt.

johnnyshields: Sure, that's fine. Thanks!

comandeo-mongo (approved):

PR #5980, opened 2025-05-09
MONGOID-5872 - [🛠️ Let's Fix Mongoid's Broken CI] Actually test Rails 6.0.0, 7.0.0, 7.1.0, and released version of 8.0.0

johnnyshields:
Fun fact: Mongoid's CI is not testing the following versions of Rails at all:
- 6.0.x
- 7.0.x
- 7.1.x

The reason is that `gem 'actionpack', '~> 6.0'` resolves to `6.Y.Z` where Y and Z are the highest possible version. To enforce it is actually testing `6.0.Z`, one must declare it as `~> 6.0.0`

In addition, Rails 8.0 is pinned at 8.0.0-rc2, i.e. not the latest 8.0.Z release.

Please kindly backport this to any branches under stable maintenance to ensure that future patches to those branches also have proper test coverage.

### DIFF
--- a/gemfiles/rails-6.0.gemfile
+++ b/gemfiles/rails-6.0.gemfile
@@ -1,11 +1,11 @@
 # rubocop:todo all
 source 'https://rubygems.org'
 
-gem 'actionpack', '~> 6.0'
-gem 'activemodel', '~> 6.0'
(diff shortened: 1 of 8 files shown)


johnnyshields:
Any reason this is not merged yet?

PR #5982, opened 2025-05-10
MONGOID-5865 - Revert 9.0 Hash behavior

johnnyshields:
In Mongoid 9.0, an unannounced breaking change was introduced that causes `field type: Hash` to become `BSON::Document`.

This change is not desirable and should be reverted.

### DIFF
--- a/lib/mongoid/extensions/hash.rb
+++ b/lib/mongoid/extensions/hash.rb
@@ -122,12 +122,8 @@ module Mongoid
         # @return [ Hash | nil ] The object mongoized or nil.
         def mongoize(object)
           return if object.nil?
-          case object
-          when BSON::Document
-            object.dup.transform_values!(&:mongoize)
-          when Hash
-            BSON::Document.new(object.transform_values(&:mongoize))
-          end
+
+          object.dup.transform_values!(&:mongoize)
         end
 
         # Can the size of this object change?


johnnyshields:
@jamis may we get a comment on this? This is a blocker for Mongoid 9.0 upgrade for us.

johnnyshields:
@jamis thank you, I really appreciate that giving users from 7.5 an upgrade path should be a top priority for the mongoid team.

On this topic, I would really like to see the `will_change` behavior feature flagged for upgrade--it is inexplicable to me that the Mongoid team adds feature flags for the most miniscule of buggy behaviors that no real-world user would care about, yet misses feature flags for obvious-as-day critical app-breaking changes.

I have long offered to volunteer TableCheck's entire app code and its 30,000+ tests as a proving ground for Mongoid releases. MongoDB has never taken me up on it. If you would like to try upgrading to Mongoid 10.0 yourself on my current 7.5 app and see the breakage yourself, I am happy to share access.

jmarsh-gaia:
I'm also running into this issue when upgrading from Mongoid 8 to 9. What is the advice for those of us who are trying to stay up to date? Should we convert our code to use stringified keys or just wait until mongoid 10?

Jibola:
This project has been queued up for future quarters; as this is stale. We'll delete it for now and circle back with it as reference.

PR #5999, opened 2025-06-23
MONGOID-5841 Don't auto-build if the parent is frozen

jamis:
If a record is frozen (perhaps because it has just been destroyed), attempts to auto-build associations on it should return nil, instead of raising a mysterious `FrozenError`.


johnnyshields:
@jamis this is a bad idea.

jamis:
MONGOID-5841 has been closed as "works as designed;" this PR is being closed without merging.

PR #6002, opened 2025-06-26
RUBY-3643 New release process (backport to 9.0)

jamis:
This backports #5994 (the release process changes) to 9.0-stable.


jamis:
Added in #6016

PR #6009, opened 2025-07-08
MONGOID-5882 Isolation state

jamis:
## Summary

Some applications will use threads for concurrency. Others will use fibers. Prior to this PR, Mongoid worked fine with threads, but it's internal state could get into odd states when run under fibers.

This PR allows applications to indicate which *isolation level* they wish to use, and Mongoid's state will be isolated to that scope.

```ruby
# the default -- inherit the isolation level from `ActiveSupport::IsolatedExecutionState` if possible
Mongoid.isolation_level = :rails

# The following two explicit settings are supported:
Mongoid.isolation_level = :thread # the classic behavior
Mongoid.isolation_level = :fiber  # NEW!
```

When using the `:fiber` isolation level, Mongoid's internal state will be inherited from any parent fiber. If you want to make sure a fiber begins with a clean slate, you can wipe the isolation state with `Mongoid::Threaded.reset!`.

### DIFF
--- a/lib/config/locales/en.yml
+++ b/lib/config/locales/en.yml
@@ -712,6 +712,16 @@ en:
             the expression %{javascript} is not allowed."
           resolution: "Please provide a standard hash to #where when the criteria
             is for an embedded association."
+        unsupported_isolation_level:
+          message: "The isolation level '%{level}' is not supported."
+          summary: >
+            You requested an isolation level of '%{level}', which is not
+            supported. Only `:thread` and `:fiber` isolation levels are
+            currently supported; note that the `:fiber` level is only
+            supported on Ruby versions 3.2 and higher.
+          resolution: >
+            Use `:thread` as the isolation level. If you are using Ruby 3.2
+            or higher, you may also use `:fiber`.
         validations:
           message: "Validation of %{document} failed."
           summary: "The following errors were found: %{errors}"
--- a/lib/config/locales/en.yml
+++ b/lib/config/locales/en.yml
@@ -716,12 +716,13 @@ en:
           message: "The isolation level '%{level}' is not supported."
           summary: >
             You requested an isolation level of '%{level}', which is not
-            supported. Only `:thread` and `:fiber` isolation levels are
-            currently supported; note that the `:fiber` level is only
-            supported on Ruby versions 3.2 and higher.
+            supported. Only `:rails`, `:thread` and `:fiber` isolation
+            levels are currently supported; note that the `:fiber` level is
+            only supported on Ruby versions 3.2 and higher.
           resolution: >
             Use `:thread` as the isolation level. If you are using Ruby 3.2
-            or higher, you may also use `:fiber`.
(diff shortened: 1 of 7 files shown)


johnnyshields:
Does `:fiber` cause weird side effects with embedded callbacks?

jamis:
> Does `:fiber` cause weird side effects with embedded callbacks?

It shouldn't. That's what most of the tests in this PR are designed to ensure: that nested fibers inherit state from their parent fiber. The entire point of this PR is to make sure Mongoid can be compatible with fiber-based libraries like [Falcon](https://github.com/socketry/falcon).

jamis:
@johnnyshields :

> 1. Is it correct to assume that "fiber" isolation also implies "thread" isolation, as fibers exist on threads (and the "main fiber" of each thread is a separate fiber?) Are there tests for this?

See https://github.com/mongodb/mongoid/pull/6009/files#diff-fc964a96de10aff651e8e1d4a35745a2a33f357819002b11af3cdacfe3e56979R58-R78

> 2. Aside from not being on the latest Ruby, is there any reason that a user using Puma (not Falcon) wouldn't want to use :fiber isolation? If no (i.e. fiber is safe) then how about making it the default in Mongoid 10?

It's worth considering. Again, "aside from not being on the latest Ruby" is the sticking point. As long as we support Ruby prior to 3.2, we really can't make `:fiber` the default.

travisbell:
@jamis I've been running this branch in prod for the past day or so and everything is looking great. I haven't seen the previous issue with mis-matched queries so I believe this is looking good. With these changes I think we can officially say Falcon is supported! 💪🏼

jamis:
@travisbell That's wonderful! Thanks for giving it a spin and letting me know. That gives me confidence to move forward with this.

comandeo-mongo (approved):

travisbell:
Knowing how it's failing in our environment might make it a bit easier for to create a reproduction. Let me take a stab at it...

jamis:
I created https://jira.mongodb.org/browse/MONGOID-5915 to track this. Thanks again for your work on a reproduction for it! That's most of the battle, right there. 👍

travisbell:
@johnnyshields There's two parts to this as far as my understanding goes:

1. The actual IO calls: since Ruby's IO library is compatible with Fibers (and will yield and resume properly as required automatically) we don't have anything to worry here. Ruby does the right thing out of the box.
2. Connection pool usage: since the Mongo driver implements a thread based connection pool, we are not contention free here. This doesn't mean it doesn't work, it just means there can be contention on the main thread. In a default Falcon environment each process only has a single thread, so that means all of your requests (fibers) are checking out/in connections from the same single pool per process.

This all works fine though, and to get around this difference, I just bumped the size of my connection pool a bit. Eventually it would be nice to see this change but it really hasn't been a problem for us, so I haven't worried about it. The Mongoid issue is the only real issue we've come across in ~2 years.

PR #6019, opened 2025-07-23
MONGOID-5887 Update mongo.rb to add hint in the aggregates

akshay58538:
MongoID aggregates dont propagate the hint passed in the criteria, into the pipeline aggregate.

This causes the query to ignore the hint and use an index decided by the planner, which is not ideal for the cases where the user intentionally passed a certain hint.


johnnyshields:
Oh wow. I had seen this behavior but naively assumed hints weren't supported by the aggregation pipeline 🤦‍♂️ Truly great catch!

PR #6031, opened 2025-08-26
MONGOID-5892 - Treat serialize option only with values of nil and [] differently

bpardee:
When calling serializable_hash(only: []) should suppress all fields.  Currently it's treated the same as nil where it gets ignored.

## Summary

Previously, passing `only: []` to `serializable_hash` was treated the same as passing `only: nil`, meaning all fields were included. The correct behavior is to treat an empty array as "include no fields," suppressing all output.

This fix is gated behind the `Mongoid.serializable_hash_with_legacy_only` feature flag. In 9.1 and later, the flag defaults to `true` (legacy behavior); set it to `false` to opt into the new behavior. The default will flip to `false` in Mongoid 10.0, and the flag will be removed in 11.0.

### DIFF
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -259,15 +259,15 @@ module Mongoid
     # consequences of duplicate index checking.
     option :allow_duplicate_index_declarations, default: false
 
-    # When this flag is false, performing a serializable_hash(only: []) will
+    # When this flag is true, performing a serializable_hash(only: []) will
     # result in all fields being included. This is the traditional (and default)
     # behavior in Mongoid 9 and earlier.
     #
-    # Setting this flag to true will change the behavior to suppress all fields
+    # Setting this flag to false will change the behavior to suppress all fields
     # in the serialized hash. This will be the default in Mongoid 10.
     #
     # See https://jira.mongodb.org/browse/MONGOID-5892 for more details.
-    option :serializable_hash_with_legacy_only, default: false
+    option :serializable_hash_with_legacy_only, default: true
 
     # Returns the Config singleton, for use in the configure DSL.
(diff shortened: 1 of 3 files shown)


jamis:
@bpardee -- thank you so much for this PR, we really appreciate the time you've taken to dig into Mongoid and contribute a fix.

Because this PR fixes behavior that has been in place for so long (since 2012!), we're going to play it cautious, here, and move forward with the assumption that changing this behavior may potentially break existing applications.

Our procedure in this case is:

1. Introduce a feature flag (e.g. `Mongoid::Config.serializable_hash_with_legacy_only`) in the next minor version (i.e. 9.1) that allows the new behavior to be opted into. The flag should default to `false` when first introduced (requiring the new behavior to be explicitly opted into.)
2. In the next major version (i.e. 10.0), the default for that flag is changed to `true`, so that this becomes the default behavior, but people can still opt into the old behavior if they need to. (This makes it easier for people to upgrade without fear of their app breaking.) We also deprecate the flag in this version.
3. In the next major version after that (i.e. 11.0), we remove the flag and the legacy behavior.

We've created a ticket in Jira to track this (https://jira.mongodb.org/browse/MONGOID-5892). If you are willing to add the feature flag and corresponding specs to this PR (see `Mongoid::Config` for examples of other flags), I'd be happy to merge it. I know that's more than you signed up for, though, so if you don't have the time, we'll get to it as part of finalizing the eventual 9.1.0 release.

Thanks again!

johnnyshields:
Great catch @bpardee.

jamis (changes requested):

jamis (approved):
Thank you! This looks great.

PR #6033, opened 2025-09-08
Port the new release infrastructure to 7.5-stable

jamis:
In order to cut a new release of the 7.5 series, we need to port the release infrastructure to 7.5-stable.


comandeo-mongo (approved):
Test failures will be fixed in a follow-up PR

PR #6034, opened 2025-09-09
MONGOID-5895 Make Reload Properly Update new_record

mmrlivingstone:
We've run into an issue when reloading a document which has had it's ID set to one that matches the value of an existing document.

Right now, when `.reload` is called in these circumstances, the data is pulled from the database, but the document does not update it's `new_record` value to `false`. This results in errors when trying to save, as Mongoid attempts to insert the document instead of updating the existing one.

It's easy to reproduce with something like the pseudo-code below (the new specs also show this)
```ruby
doc1 = SomeDocument.create!

doc2 = SomeDocument.new(id: doc1.id)
doc2.save! # This will error
```

### DIFF
--- a/lib/mongoid/reloadable.rb
+++ b/lib/mongoid/reloadable.rb
@@ -17,6 +17,12 @@ module Mongoid
       reloaded = _reload
       check_for_deleted_document!(reloaded)
 
+      # In an instance where we create a new document, but set the ID to an existing one,
+      #   when the document is reloaded, we want to set new_record to false.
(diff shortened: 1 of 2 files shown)


jamis (approved):

jamis:
@mmrlivingstone -- Mongoid 9.0.8 has been released with this fix. Thank you for the patch, and for your patience!

mmrlivingstone:
@jamis Thanks for the incredibly quick turn around on this!

PR #6042, opened 2025-09-22
MONGOID-5890 Handle BSON::Decimal128 specially when checking for changes

jamis:
With the `map_big_decimal_to_decimal128` option enabled, Mongoid will translate between `BSON::Decimal128` and `BigDecimal` when records are loaded and saved. However, if you assign an identical floating point value to the field, it will mark the record as changed, even though it (functionally) did not. This is because `BSON::Decimal128` is not an actual numeric type, and equality comparisons with numeric types will return false.

This PR fixes the dirty tracking with regard to `BigDecimal` and `BSON::Decimal128` by adding a normalization step to the comparison and handling the `Decimal128` type specially.

### DIFF
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -371,7 +371,7 @@ module Mongoid
     # comparison purposes. This is necessary because `BSON::Decimal128` does
     # not implement `#==` in a way that is compatible with `BigDecimal`.
     def normalize_value(value)
-      value.is_a?(BSON::Decimal128) ? value.to_d : value
+      value.is_a?(BSON::Decimal128) ? BigDecimal(value.to_s) : value
     end
 
     # Determine if the attribute will not change, by comparing the current
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -175,7 +175,7 @@ module Mongoid
           localized = fields[field_name].try(:localized?)
           attributes_before_type_cast[name.to_s] = value
           typed_value = typed_value_for(field_name, value)
-          unless attributes[field_name] == typed_value || attribute_changed?(field_name)
+          unless attribute_will_not_change?(field_name, typed_value) || attribute_changed?(field_name)
             attribute_will_change!(field_name)
           end
           if localized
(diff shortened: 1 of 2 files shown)


comandeo-mongo (approved):

lib/mongoid/attributes.rb
```diff
@@ -371,7 +371,7 @@ def lookup_attribute_presence(name, value)
     # comparison purposes. This is necessary because `BSON::Decimal128` does
     # not implement `#==` in a way that is compatible with `BigDecimal`.
     def normalize_value(value)
-      value.is_a?(BSON::Decimal128) ? value.to_d : value
```
johnnyshields: This should maybe detect `.respond_to?(:to_d)` and use it if available, as there may be differences between `.to_d` and `BigDecimal(value.to_s)`, and `.to_d` should be considered canonical.

jamis: Canonical, yes, but whether we use `#to_d` here or not, we _still_ have to do the `BigDecimal(value.to_s)` bit as well, because of the case where `#to_d` is not available. In the interest of keeping the code simpler and easier to read, I'm opting to use the lowest common denominator, here.

Thanks for your input, Johnny. 👍

johnnyshields: @jamis I mean this very politely, but this line of thinking has consistently gotten Mongoid into trouble over the years.

"Easy to read" is *not* the highest virtue of code--nor is DRY, etc. The goal should be **"intuitive/correct behavior for users"**--in this case the user is the "Mongoid developer".

Sometimes it takes an extra line or two to achieve both backwards compatibility (`BigDecimal(value.to_s)`) and forward behavioral correctness (`#to_d`). This is ok--a terse code comment `#to_d was added in mongo-ruby 5.0.0` is enough to help with readability.

By doing things your way, you are burying a *technical debt landmine* in the code that future users will walk across. Granted, not using `#to_d` in-and-of-itself is not catastrophic, but since you are aware there is an opportunity to delegate to the driver--why not do it now? Your way is to wait until MONGOID-19328 is raised "Mongoid does not normalize BigDecimals properly, but driver does.", and someone will see the blame "Oh yeah, the developer back then, some guy named @jamis thought it would be *easier to read* if he masked the driver behavior."

And the karmic cycle continues...

PR #6054, opened 2025-10-22
Rails 8.1 Support

AlexKovynev:


alcaeus:
@AlexKovynev thank you for the contribution. I see that this adds a number of config files for GitHub Copilot. While these will certainly be helpful going forward, I think we should split them off from this PR as they have nothing to do with Rails 8.1 support

AlexKovynev:
> @AlexKovynev thank you for the contribution. I see that this adds a number of config files for GitHub Copilot. While these will certainly be helpful going forward, I think we should split them off from this PR as they have nothing to do with Rails 8.1 support

yes sure. Will remove it now. Thnx. Autocreation :(

comandeo-mongo (approved):

jamis (approved):

comandeo-mongo:
Thank you for your contribution @AlexKovynev !

PR #6069, opened 2025-12-05
MONGOID-5910 Implement enumerable methods on top level models (any?, many?, and one?)

kyleburgess2025:
This PR adds the enumerable methods `any?`, `many?`, and `one?` to the `Criteria` class. It also delegates calls to these functions at the class level to the methods defined on `Criteria`. For example, `User.any?` should return if any documents exist in the `users` table.

## Summary

You may now invoke the enumerable methods `#any?`, `#many?`, and `#one?` on `Criteria` instances. Invoking them at the class level will delegate to those methods, as well.

```ruby
User.any? #-> returns true if any (1 or more) user records exist
User.where(admin: true).one? #-> returns true if there is exactly one admin record
User.where(banned: true).many? #-> returns true if there are "many" (2 or more) banned users
```


spec/mongoid/criteria/findable_spec.rb
```diff
@@ -1302,6 +1302,212 @@
...
+          Band.where(name: "Nonexistent Band")
+        end
+
+        it "returns false" do
+          expect(criteria.one?).to be false
+        end
+      end
+    end
+
+    context "when called on class" do
```
kyleburgess2025: let me know if this needs to be moved elsewhere or isnt needed

lib/mongoid/findable.rb
```diff
@@ -116,6 +116,36 @@ def exists?(id_or_conditions = :none)
       with_default_scope.exists?(id_or_conditions)
     end
 
+    # Return true if any documents exist in the criteria.
+    #
+    # @example Determine if any documents exist
+    #   criteria.any?
+    #   
+    # @return [ true | false ] If any documents exist.
+    def any?(*args, &block)
```
jamis: It's safe to leave off the `*args` and `&block` here, since they aren't used in the method body anymore.

Ditto for the other two methods you added. 👍

lib/mongoid/findable.rb
```diff
@@ -116,6 +116,36 @@ def exists?(id_or_conditions = :none)
       with_default_scope.exists?(id_or_conditions)
     end
 
+    # Return true if any documents exist in the criteria.
+    #
+    # @example Determine if any documents exist
+    #   criteria.any?
```
jamis: I know this is an artifact from when you had this on `Mongoid::Criteria`, but for documentation purposes you can just reference `Model`, generically.

```suggestion
    #   Model.any?
```

Ditto for the other new methods you added.

jamis (approved):

PR #6075, opened 2025-12-12
MONGOID-5800 When initializing an embedded association with an invalid hash, got undefined method `with_indifferent_access' error

kyleburgess2025:
Currently, if a document with an `embeds_many` relationship is created with the embedded field set to a single hash, instead of an array of hashes, an error is raised with `undefined method 'with_indifferent_access'`. This behavior is new in 8.0. This PR allows backwards compatibility by checking if a value has the method `with_indifferent_access`, and, if not, we can assume it is an array that should be a hash. We convert it to a hash and call `with_indifferent_access` on the result.

### DIFF
--- a/lib/mongoid/association/nested/many.rb
+++ b/lib/mongoid/association/nested/many.rb
@@ -48,7 +48,11 @@ module Mongoid
         # @param [ Hash ] attributes The attributes hash to attempt to set.
         # @param [ Hash ] options The options defined.
         def initialize(association, attributes, options = {})
-          if attributes.respond_to?(:with_indifferent_access)
+          # If a hash is passed, and the first key is not an integer,
+          # we assume it's a single document being passed in.
+          if attributes.is_a?(Hash) && !attributes.keys.first.is_a?(Integer)
+            @attributes = [attributes]
+          elsif attributes.respond_to?(:with_indifferent_access)
             @attributes = attributes.with_indifferent_access.sort do |a, b|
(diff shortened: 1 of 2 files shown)


lib/mongoid/association/nested/many.rb
```diff
@@ -32,8 +32,10 @@ def build(parent, options = {})
           attributes.each do |attrs|
             if attrs.is_a?(::Hash)
               process_attributes(parent, attrs.with_indifferent_access)
-            else
+            elsif attrs[1].respond_to?(:with_indifferent_access)
               process_attributes(parent, attrs[1].with_indifferent_access)
+            else
+              process_attributes(parent, Hash[*attrs].with_indifferent_access)
```
kyleburgess2025: do we need more guards here in case it can't be cast to a hash?

jamis: I think it's probably smart to check that array has a non-zero and even number of elements, before creating a hash from it.

We should probably also raise an exception if that's not the case, so that we never have data being silently ignored. A simple `ArgumentError` would probably suffice, indicating that the arguments aren't supported or something.

jamis (changes requested):
This is a good solution, handling an edge case that we hadn't considered. As the OP noted in the ticket, this is technically not valid data, but if someone ran into this, and if we can handle the situation gracefully, we might as well.

jamis (approved):

PR #6078, opened 2025-12-19
MONGOID-5881 Setting an embeds_many association to empty array before save does not persist it

kyleburgess2025:


Jibola:
Superseded by https://github.com/mongodb/mongoid/pull/6157

PR #6080, opened 2026-01-09
MONGOID-5878 Object#=~ removed in Ruby 3.2

kyleburgess2025:
Object#=~ removed in Ruby 3.2, but we were doing Regex matches on fields that could potentially be objects. This PR ensures that if =~ is not defined, nil is returned.

### DIFF
--- a/lib/mongoid/matcher/eq_impl_with_regexp.rb
+++ b/lib/mongoid/matcher/eq_impl_with_regexp.rb
@@ -20,13 +20,11 @@ module Mongoid
       #
       # @api private
       module_function def matches?(original_operator, value, condition)
-        return false if value.nil? && (condition.is_a?(Regexp) || condition.is_a?(::BSON::Regexp::Raw))
-
         case condition
         when Regexp
-          value.to_s =~ condition
+          value.respond_to?(:=~) && value =~ condition
         when ::BSON::Regexp::Raw
-          value.to_s =~ condition.compile
+          value.respond_to?(:=~) && value =~ condition.compile
         else
           if value.kind_of?(Time) && condition.kind_of?(Time)
             EqImpl.time_eq?(value, condition)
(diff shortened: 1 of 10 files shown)


spec/integration/matcher_operator_data/in.yml
```diff
@@ -114,16 +114,16 @@
   document:
     name: [Boris]
   query:
-    count:
+    name:
       $in:
         - !ruby/regexp /..ris/
-  matches: false
+  matches: true
```
kyleburgess2025: Changed to `matches: true` because the name implies this should match (in contrast to the next test, which should not match and says so in the name)

spec/integration/matcher_operator_data/in.yml
```diff
@@ -132,18 +132,18 @@
...
     name: [Boris]
   query:
-    count:
+    name:
       $in:
         - !ruby/object:BSON::Regexp::Raw
           pattern: ..ris
           options: ''
-  matches: false
+  matches: true
```
kyleburgess2025: Same as above

lib/mongoid/matcher/eq_impl_with_regexp.rb
```diff
@@ -22,9 +22,9 @@ module EqImplWithRegexp
       module_function def matches?(original_operator, value, condition)
         case condition
         when Regexp
-          value =~ condition
+          value.to_s =~ condition
```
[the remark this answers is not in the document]
kyleburgess2025: I've thought about this a bit... the `nil` thing is a good point, as this would be a change from our current behavior. I'll add a check for nil. The other potentially problematic types are arrays and hashes, but testing with them doesn't cause any unexpected behavior.

jamis (changes requested):
Thank you for your work on this! Just a question for you, and an alternative implementation to consider.

kyleburgess2025:
@jamis Investigation shows that =~ used to work on arrays, but on objects, it just returns nil no matter what. I'm going to try to preserve that behavior.

jamis (approved):
Good job catching that those tests were trying to match on the wrong field name!

PR #6128, opened 2026-04-06
MONGOID-5867 Merge touch updates into embedded document insert

evanqi:
## Summary

When inserting an embedded document with a touchable parent (the default for `embedded_in`), Mongoid issues two separate `update_one` calls:

1. `$push` — inserts the embedded document (`insert_as_embedded`)
2. `$set` — updates `updated_at` on the parent chain (`after_save` touch callback)

This commit merges both into a single `update_one`, eliminating the redundant round-trip.

- `insert_as_embedded` gathers touch updates via `_gather_touch_updates` and merges the `$set` into the same operation as the `$push`
- The `after_save` touch callback checks a `Threaded` flag and skips the now-redundant persistence, while still running `:touch` callbacks and cleaning up dirty tracking
- The `Threaded` flag follows the existing `begin_autosave`/`exit_autosave`/`autosaved?` pattern

## Relationship to positionally()

[Jamis noted in MONGOID-5867](https://jira.mongodb.org/browse/MONGOID-5867?focusedId=8027497#comment-8027497) that consolidating the parent touch with the insert is tricky because of the `positionally` utility method, which replaces explicit array indexes with the `$` positional update operator.

This implementation avoids that concern: the merged `$set` touch paths are included in the same `operations` hash that `positionally` already processes for the `$push`. Both operators get consistent index-to-`$` replacement in the same call. For example, with `Building > floors[0] > sofas`:

```ruby
# Before positionally:
{ "$push" => { "floors.0.sofas" => doc },
  "$set"  => { "floors.0.updated_at" => time, "updated_at" => time } }

# After positionally (selector includes floors._id):
{ "$push" => { "floors.$.sofas" => doc },
  "$set"  => { "floors.$.updated_at" => time, "updated_at" => time } }
```

No new race conditions are introduced — the stale-index concern is mitigated by the existing `positionally` mechanism identically for both operators.

## Test plan

- [x] Single `update_one` when pushing into a touchable parent (`expect_query(1)`)
- [x] `updated_at` propagated to all ancestors after reload
- [x] Embedded document is persisted correctly
- [x] Touch callbacks still fire on the parent chain
- [x] No behavior change with `touch: false`
- [x] Full `touchable_spec.rb` suite passes (240 examples, 0 failures)
- [x] Rubocop clean

### DIFF
--- a/lib/mongoid/persistable/creatable.rb
+++ b/lib/mongoid/persistable/creatable.rb
@@ -41,6 +41,12 @@ module Mongoid
 
       # Insert the embedded document.
       #
+      # When the parent association is touchable (which is the default for
+      # +embedded_in+), the touch timestamp updates are merged into the
+      # same +update_one+ call that performs the insert. This avoids a
+      # second round-trip that the +after_save+ touch callback would
+      # otherwise issue.
+      #
       # @api private
       #
       # @example Insert the document as embedded.
@@ -54,8 +60,18 @@ module Mongoid
           _parent.insert
         else
           selector = _parent.atomic_selector
+          operations = atomic_inserts
+
+          if _touchable_parent?
+            touches = _parent._gather_touch_updates(Time.current)
+            if touches.present?
+              operations['$set'] = touches
+              Threaded.begin_touch_merged(self)
+            end
+          end
+
           _root.collection.find(selector).update_one(
-            positionally(selector, atomic_inserts),
+            positionally(selector, operations),
             session: _session
           )
         end
--- a/lib/mongoid/threaded.rb
+++ b/lib/mongoid/threaded.rb
@@ -22,6 +22,8 @@ module Mongoid
 
     VALIDATIONS_KEY = 'validations'
 
+    TOUCH_MERGED_KEY = 'touch-merged'
+
     STACK_KEYS = Hash.new do |hash, key|
       hash[key] = "#{key}-stack"
     end
@@ -202,6 +204,17 @@ module Mongoid
       validations_for(document.class).push(document._id)
     end
 
+    # Mark that a document's touch updates have been merged into
+    # an atomic insert on the current thread.
+    #
+    # @example Begin touch merged.
+    #   Threaded.begin_touch_merged(doc)
+    #
+    # @param [ Document ] document The embedded document being inserted.
+    def begin_touch_merged(document)
+      touch_merged_for(document.class).push(document._id)
+    end
+
     # Exit autosaving a document on the current thread.
     #
     # @example Exit autosave.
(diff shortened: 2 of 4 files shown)


jamis (approved):
Thank you! This works great. 👍

Greg-Myers-SB:
This change contains a bug that removes records using embeds_one
Reproduction
```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

# Reproduction for mongoid 9.1.0 embeds_one silent data loss bug.
# https://github.com/mongodb/mongoid/issues/6128
#
# When both parent and child include Mongoid::Timestamps, calling
# create_<child> on an embeds_one association silently drops the
# embedded document data — the $set for the document gets replaced
# by the touch timestamps instead of being merged.
#
# Usage: ruby mongoid_9_1_embeds_one_bug.rb
# Requires a running MongoDB on localhost:27017.

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "mongoid", "9.1.0" # working on 9.0.11
end

Mongoid.configure do |config|
  config.clients.default = {
    uri: "mongodb://localhost:27017/mongoid_embeds_one_bug_repro"
  }
end

class Parent
  include Mongoid::Document
  include Mongoid::Timestamps

  field :name, type: String
  embeds_one :child
end

class Child
  include Mongoid::Document
  include Mongoid::Timestamps

  field :value, type: String
  embedded_in :parent
end

parent = Parent.create!(name: "test")
parent.create_child(value: "important_data")

reloaded = Parent.find(parent.id)

if reloaded.child.nil?
  puts "BUG CONFIRMED: child is nil after reload — data was silently dropped"
  exit 1
else
  puts "OK: child persisted with value=#{reloaded.child.value}"
  exit 0
end

```

evanqi:
> This change contains a bug that removes records using embeds_one Reproduction
> 
> ```ruby
> #!/usr/bin/env ruby
> # frozen_string_literal: true
> 
> # Reproduction for mongoid 9.1.0 embeds_one silent data loss bug.
> # https://github.com/mongodb/mongoid/issues/6128
> #
> # When both parent and child include Mongoid::Timestamps, calling
> # create_<child> on an embeds_one association silently drops the
> # embedded document data — the $set for the document gets replaced
> # by the touch timestamps instead of being merged.
> #
> # Usage: ruby mongoid_9_1_embeds_one_bug.rb
> # Requires a running MongoDB on localhost:27017.
> 
> require "bundler/inline"
> 
> gemfile(true) do
>   source "https://rubygems.org"
>   gem "mongoid", "9.1.0" # working on 9.0.11
> end
> 
> Mongoid.configure do |config|
>   config.clients.default = {
>     uri: "mongodb://localhost:27017/mongoid_embeds_one_bug_repro"
>   }
> end
> 
> class Parent
>   include Mongoid::Document
>   include Mongoid::Timestamps
> 
>   field :name, type: String
>   embeds_one :child
> end
> 
> class Child
>   include Mongoid::Document
>   include Mongoid::Timestamps
> 
>   field :value, type: String
>   embedded_in :parent
> end
> 
> parent = Parent.create!(name: "test")
> parent.create_child(value: "important_data")
> 
> reloaded = Parent.find(parent.id)
> 
> if reloaded.child.nil?
>   puts "BUG CONFIRMED: child is nil after reload — data was silently dropped"
>   exit 1
> else
>   puts "OK: child persisted with value=#{reloaded.child.value}"
>   exit 0
> end
> ```

Good find, sorry about introducing the bug. I created a fix PR: https://github.com/mongodb/mongoid/pull/6150

PR #6141, opened 2026-04-27
MONGOID-5684 Ensure hash-typed fields return Hash

jamis:
Fixes [MONGOID-5684](https://jira.mongodb.org/browse/MONGOID-5684).

### Problem

When a document with a `field :name, type: Hash` is saved and reloaded, the field value is a `BSON::Document` rather than a plain `Hash`:

```ruby
class Foo
  include Mongoid::Document
  field :feature, type: Hash
end

foo = Foo.create!(feature: { 'x' => 1 })
foo = Foo.find(foo.id)
foo.feature.class  #=> BSON::Document  (should be Hash)
```

This happens because `BSON::Document` is a subclass of `Hash`, so Mongoid's type check passes — but `Hash::ClassMethods` had no `demongoize` method, so the call fell through to `Object.demongoize`, which returns its argument unchanged.

A secondary symptom: after modifying the field, `field_was` returned a plain `Hash` whose string keys weren't accessible via symbol — inconsistent with the `BSON::Document` returned by the getter.

### Fix

**`lib/mongoid/extensions/hash.rb`** — Add `Hash::ClassMethods#demongoize`. When `legacy_hash_fields` is `false`, it converts `BSON::Document` → `Hash` recursively. Plain `Hash` inputs are returned as the same object (no allocation), so subsequent attribute reads are O(1).

**`lib/mongoid/attributes.rb`** — After demongoizing a `BSON::Document` to a plain `Hash`, write the plain `Hash` back into `attributes`. This ensures in-place mutations on the returned value are visible to dirty tracking. The write-back is scoped precisely to the `BSON::Document` → `Hash` conversion and does not affect other field types (Date, Time, BigDecimal, Set, etc.).

**`lib/mongoid/config.rb`** — Add the `legacy_hash_fields` option (default: `true`).

### Feature flag

The fix is opt-in to avoid breaking applications that depend on the current behavior:

```ruby
Mongoid.configure do |config|
  config.legacy_hash_fields = false  # opt in to correct behavior
end
```

The default (`true`) preserves the existing behavior unchanged.

### Performance

`Hash.demongoize` is called on every attribute read. The key question is whether the conversion adds unacceptable overhead.

**Scenario 1 — `legacy_hash_fields: true` (current behavior, BSON::Document returned as-is)**

| Input | Throughput |
|---|---|
| Empty | 7.67M i/s |
| Flat, 5 keys | 7.60M i/s |
| Nested, 3 levels × 10 keys | 7.57M i/s |

O(1), no allocation — just a config check and an early return.

**Scenario 2 — `legacy_hash_fields: false`, first read (BSON::Document input, conversion)**

| Input | Throughput | vs. legacy |
|---|---|---|
| Empty | 5.10M i/s | 1.5× slower |
| Flat, 5 keys | 1.22M i/s | 4.2× slower |
| Nested, 3 levels × 10 keys | 6.87k i/s | **742× slower** |

This is the one-time cost paid when a document is first loaded from the database. For the nested case, 111 `Hash` objects are allocated and 1,110 nodes are visited per call (145 µs on the benchmark machine). This cost is proportional to the size and depth of the stored value.

**Scenario 3 — `legacy_hash_fields: false`, subsequent reads (plain Hash input)**

After the first read, `attributes` holds the converted plain `Hash`. Subsequent reads return it as-is:

| Input | Throughput | vs. legacy |
|---|---|---|
| Empty | 27.2M i/s | 3.5× faster |
| Flat, 5 keys | 27.1M i/s | 3.6× faster |
| Nested, 3 levels × 10 keys | 27.2M i/s | 3.6× faster |

O(1), size-independent — two type checks and an early return.

**Summary:** with `legacy_hash_fields: false`, each Hash-typed attribute read pays a one-time conversion cost proportional to the hash's size. All subsequent reads of that field are faster than the current legacy path. Applications with large, deeply-nested Hash fields loaded frequently from the database should measure the first-read cost against their own document shapes before enabling the flag.

### DIFF
--- a/lib/mongoid/attributes.rb
+++ b/lib/mongoid/attributes.rb
@@ -101,6 +101,11 @@ module Mongoid
     # @api private
     def process_raw_attribute(name, raw, field)
       value = field ? field.demongoize(raw) : raw
+      # When demongoize converts a BSON::Document to a plain Hash (i.e.
+      # legacy_hash_fields is false), store the plain Hash back into
+      # attributes so that in-place mutations on the returned value are
+      # visible to dirty tracking.
+      attributes[name] = value if raw.is_a?(BSON::Document) && !value.is_a?(BSON::Document)
       is_relation = relations.key?(name)
       attribute_will_change!(name) if value.resizable? && !is_relation
       value
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -78,6 +78,11 @@ module Mongoid
     # Store BigDecimals as Decimal128s instead of strings in the db.
     option :map_big_decimal_to_decimal128, default: true
 
+    # When true (the default), preserve legacy behavior where Hash-typed fields
+    # return BSON::Document after loading from the database. Set to false to
+    # have Hash-typed fields return plain Hash, matching the declared field type.
+    option :legacy_hash_fields, default: true
+
     # Allow BSON::Decimal128 to be parsed and returned directly in
     # field values. When BSON 5 is present and this option is set to false
     # (the default), BSON::Decimal128 values in the database will be returned
--- a/lib/mongoid/extensions/hash.rb
+++ b/lib/mongoid/extensions/hash.rb
@@ -105,6 +105,27 @@ module Mongoid
       Mongoid.deprecate(self, :to_criteria)
 
       module ClassMethods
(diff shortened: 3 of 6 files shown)


jamis:
This implementation imposes an unacceptable overhead with the recursive copy. Closing this in favor of #6142, which addresses only the issue raised in the ticket.

PR #6143, opened 2026-04-30
MONGOID-5915 avoid async state corruption by forcing child fibers to copy parent state

jamis:
MONGOID-5882 added the ability to specify isolation state, so that Mongoid could run under either threads or fibers. However, the fiber isolation mode was insufficient and allowed sibling fibers to clobber shared state. This change still allows fibers to inherit state from their parent, but keeps the shared state more isolated.

### DIFF
--- a/lib/mongoid/threaded.rb
+++ b/lib/mongoid/threaded.rb
@@ -10,6 +10,10 @@ module Mongoid
     # symbol because keys for fiber-local storage must be symbols.
     STORAGE_KEY = :'[mongoid]'
 
+    # Tracks which fiber owns the storage hash, to detect when a fiber has
+    # inherited (rather than created) its storage from a parent fiber.
+    STORAGE_OWNER_KEY = :'[mongoid]:owner'
+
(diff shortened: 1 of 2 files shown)


travisbell:
WOOHOO!

jamis:
❤️ thanks for noticing, @travisbell! I was going to ping you and ask if you could try this out -- it resolves (for me?) the problem with that reproduction script you provided. We hope to release Mongoid 9.1.0 next week sometime, but I freely acknowledge that our deadlines are not your deadlines. :)

travisbell:
👍🏼

I don't mind testing this out at all. I will and report back.

comandeo-mongo (approved):

travisbell:
@jamis I've been running your branch since Friday afternoon and have not witnessed a single query corruption error all weekend. Things are looking very good.

I'll probably keep running this branch until 9.1.0 is released so I'll be sure to report back if I see anything pop up in the meantime but it's looking good.

jamis:
@travisbell that's wonderful to hear. Thank you for the feedback! 👍

PR #6151, opened 2026-05-26
MONGOID-5057 Add support for `has_many :through`

jamis:
## Summary

Implements read-only `:through` associations for `has_one` and `has_many`, matching ActiveRecord's API:

```ruby
class Customer
  has_one :franchise
  has_one :store, through: :franchise
end

class Physician
  has_many :appointments
  has_many :patients, through: :appointments
end
```

Both association types support `:source` to override the inferred source name, and `:class_name` for non-conventional class names. `has_many :through` additionally accepts `:order`. All through associations are read-only; write attempts raise `Mongoid::Errors::ReadonlyAssociation`. Eager loading via includes is supported for both types.

The two query patterns supported for `has_many :through` are the join-model pattern (FK on the intermediate, e.g. `belongs_to :patient` on `Appointment`) and the reverse-FK pattern (FK on the target, e.g. `has_many :readers` on `Book`).

## Out of Scope

- write operations
- polymorphic through associations
- custom `:scope`
- embedded through associations
- chaining through a `has_many :through` that is itself a through association
- `has_and_belongs_to_many` as the source association.


comandeo-mongo (approved):

sandstrom:
I was wondering just now, does mongoid support this -- lo and behold! :) 

Two small questions:

- Are there any docs for this yet? Did find them in the diff, but maybe it's coming later or is in another repo?

- Will the docs elaborate on query performance implications, because I'm guessing the underlying queries will either be more expensive than what you'd usually have in a SQL database, or need to use aggregation-queries?

jamis:
There will be docs, but we have a separate documentation team and they use an internal repository. There is nothing available yet.

The docs will probably not going into great depth on performance implications for this feature; the current implementation is fairly naive, though, and optimized more for convenience than performance right now. This is minimal "v1". There is definitely more that could be done, both for performance optimization, and for extending what is supported.

sandstrom:
Great! 🎉 

Not trying to be annoying about performance. I think this is a great start. 

As long as the docs just explains what queries will be made underr the hood, the reader can deduce what that means for them in terms of perf.

PR #6153, opened 2026-05-27
Fix search (and vector search) index errors

jamis:
Fixes a few errors in tests with search and vector search tests; also fixes an issue with `vector_search` and `auto_embed_search` due to filter being used instead of a separate `$match` stage.


comandeo-mongo (approved):
🥇

PR #6158, opened 2026-06-15
Enhancements to the aggregation-based eager load mechanism

cperezabo:
While trying these features out in a real project I also ran into two pre-existing `eager_load` bugs, fixed here in their own commits:

- `Model.eager_load([]).first` used to raise instead of just returning the documents; now it falls back to the normal path.
- a `has_many` whose target shares its collection with sibling subclasses pulled those siblings in too (crashing while materializing one whose discriminator didn't resolve); it now filters by the target subclass, the way a normal query does.

_Edit:_ PR description has been reformatted/edited by @jamis to accommodate the team's automatic release note builder.

## Summary

This closes a few gaps in `eager_load`. Each case below either wasn't supported or crashed before but now they resolve in the same single aggregation as the rest, except the polymorphic case, which adds exactly one faceted query.

### 1. Associations defined only on a subclass

When you query through a superclass, `eager_load` now finds associations declared only on a subclass.

```ruby
class Gadget
  include Mongoid::Document
end

class Speaker < Gadget
  has_many :cables
end

Gadget.eager_load(:cables) # resolves :cables through Speaker
```

Works for `has_many`, `has_one`, `has_and_belongs_to_many` and `belongs_to`.

### 2. References that live inside embedded documents

A `belongs_to` sitting in an embedded document is now eager-loaded through its embedding parent, with no extra query:

```ruby
class Device
  include Mongoid::Document
end

class Port
  include Mongoid::Document
  embedded_in :computer
  belongs_to :device
end

class Computer
  include Mongoid::Document
  embeds_one :port
end

Computer.eager_load(port: :device).first.port.device # no extra query
```

This holds true regardless of how deep the embedding goes, even when the embedded chain is attached to a referenced parent.

```ruby
Desk.eager_load(computer: { port: { pin: :device } }) # referenced → embedded → embedded → referenced
```

`has_and_belongs_to_many` inside an embedded document works too — each element keeps its own matches:

```ruby
class Port
  include Mongoid::Document
  embedded_in :computer
  has_and_belongs_to_many :devices
end

Computer.eager_load(port: :devices)
```

### 3. Polymorphic `belongs_to`

A polymorphic `belongs_to` can't be a single `$lookup` (its target collection changes per row), so it's resolved with one extra aggregation that fans out per type:

```ruby
class Cartridge
  include Mongoid::Document
  belongs_to :device, polymorphic: true # Printer, Scanner, ...
end

Cartridge.eager_load(:device) # root query + one faceted query, regardless of how many types
```

Thanks to @cperezabo for the PR!

### DIFF
--- a/lib/mongoid/association/eager.rb
+++ b/lib/mongoid/association/eager.rb
@@ -12,18 +12,12 @@ module Mongoid
       # @param [ Array<Mongoid::Association::Relatable> ] associations
       #   Associations to eager load
       # @param [ Array<Document> ] docs Documents to preload the associations
-      # @param [ Boolean ] use_lookup Whether to use $lookup aggregation
-      #   for eager loading. This is used in Criteria#eager_load.
-      # @param [ Array<Hash> ] pipeline The aggregation pipeline to use
-      #   when using $lookup for eager loading.
       #
       # @return [ Base ] The eager load preloader
-      def initialize(associations, docs, use_lookup = false, pipeline = [])
+      def initialize(associations, docs)
         @associations = associations
         @docs = docs
         @grouped_docs = {}
-        @use_lookup = use_lookup
-        @pipeline = pipeline
       end
 
       # Run the preloader.
@@ -34,12 +28,6 @@ module Mongoid
       # @return [ Array ] The list of documents given.
       def run
         @loaded = []
-
-        if @use_lookup
-          preload_with_lookup
-          @loaded = @docs
-          return @loaded.flatten
-        end
 
         while shift_association
           preload
@@ -58,21 +46,6 @@ module Mongoid
       #   loader.preload
       def preload
         raise NotImplementedError
-      end
-
-      # Preload the current association using $lookup aggregation.
-      # This method executes the aggregation pipeline
-      # and instantiates the documents.
-      # @example Preload the current association using $lookup.
-      #   loader.preload_with_lookup
-      def preload_with_lookup
-        # For $lookup aggregation, execute pipeline and instantiate documents
-        owner_class = @associations.first.owner_class
-        aggregated_docs = owner_class.collection.aggregate(@pipeline)
-        aggregated_docs.each do |doc|
-          parsed_doc = Factory.from_db(owner_class, doc)
-          @docs << parsed_doc
-        end
       end
 
       # Retrieves the documents referenced by the association, and
--- a/lib/mongoid/association/eager.rb
+++ b/lib/mongoid/association/eager.rb
@@ -4,6 +4,15 @@ module Mongoid
   module Association
     # Base class for eager load preload functions.
     class Eager
+      # Build a preloader for the given arguments and run it.
+      #
+      # @param (see #initialize)
+      #
+      # @return [ Array ] The list of documents given.
+      def self.run(associations, docs, use_lookup = false, pipeline = [])
+        new(associations, docs, use_lookup, pipeline).run
+      end
+
       # Instantiate the eager load class.
       #
       # @example Create the new belongs to eager load preloader.
@@ -12,12 +21,18 @@ module Mongoid
       # @param [ Array<Mongoid::Association::Relatable> ] associations
       #   Associations to eager load
       # @param [ Array<Document> ] docs Documents to preload the associations
+      # @param [ Boolean ] use_lookup Whether to use $lookup aggregation
+      #   for eager loading. This is used in Criteria#eager_load.
+      # @param [ Array<Hash> ] pipeline The aggregation pipeline to use
+      #   when using $lookup for eager loading.
       #
       # @return [ Base ] The eager load preloader
-      def initialize(associations, docs)
+      def initialize(associations, docs, use_lookup = false, pipeline = [])
         @associations = associations
         @docs = docs
         @grouped_docs = {}
+        @use_lookup = use_lookup
+        @pipeline = pipeline
       end
 
       # Run the preloader.
@@ -28,6 +43,12 @@ module Mongoid
       # @return [ Array ] The list of documents given.
       def run
         @loaded = []
+
+        if @use_lookup
+          preload_with_lookup
+          @loaded = @docs
+          return @loaded.flatten
+        end
 
         while shift_association
           preload
@@ -46,6 +67,21 @@ module Mongoid
       #   loader.preload
       def preload
         raise NotImplementedError
+      end
+
+      # Preload the current association using $lookup aggregation.
+      # This method executes the aggregation pipeline
+      # and instantiates the documents.
+      # @example Preload the current association using $lookup.
+      #   loader.preload_with_lookup
+      def preload_with_lookup
+        # For $lookup aggregation, execute pipeline and instantiate documents
+        owner_class = @associations.first.owner_class
+        aggregated_docs = owner_class.collection.aggregate(@pipeline)
+        aggregated_docs.each do |doc|
+          parsed_doc = Factory.from_db(owner_class, doc)
+          @docs << parsed_doc
+        end
       end
 
       # Retrieves the documents referenced by the association, and
--- a/lib/mongoid/association/eager_load/inclusion.rb
+++ b/lib/mongoid/association/eager_load/inclusion.rb
@@ -3,11 +3,27 @@
 module Mongoid
   module Association
     module EagerLoad
-      # An inclusion of an eager load, in the role it plays while the pipeline is
-      # built. Each kind knows how it contributes; the LookupPipeline holds the
-      # stage-building helpers they lean on. A node carries its own children, so
-      # the pipeline is built by recursion from the roots downward.
+      # Something an eager load contributes to the pipeline, in the role it plays
+      # while the pipeline is built. A root is asked to contribute and the whole
+      # tree follows by recursion. AssociationInclusion stands for a single
+      # association; DiscriminatedInclusion stands for a name several subclasses
+      # share.
       class Inclusion
+        # Add this inclusion's stages to the destination.
+        #
+        # @param [ Array<Hash> ] destination The pipeline (or sub-pipeline) the
+        #   stages are appended to.
+        # @param [ Array<Mongoid::Association::Relatable> ] chain The embedded path
+        #   accumulated from the ancestors above this inclusion (empty at the top).
+        def contribute(destination, chain)
+          raise NotImplementedError
+        end
+      end
+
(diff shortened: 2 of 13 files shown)


lib/mongoid/association/eager_loadable.rb
```diff
@@ -104,31 +110,211 @@ def preload(associations, docs)
...
+        return pipeline << create_pipeline(assoc, assoc_map.dup) if chains.empty?
+
+        # The same inclusion can sit under several embedded parents (e.g. two
+        # embeds_one of the same class); graft a fresh $lookup onto each path.
+        chains.each do |chain|
+          graft_embedded_lookup(pipeline, create_pipeline(assoc, assoc_map.dup), chain, assoc)
         end
+      end
+
+      def graft_embedded_lookup(pipeline, stage, chain, association)
```
jamis: While private methods technically don't need YARD documentation, in practice it is helpful. Especially for methods that are non-trivial, or whose return values and arguments have types that are not immediately obvious. Could you please add some documentation for these methods, to help future maintainers out?

(I will say -- the inline comments are nicely done, and much appreciated!)

lib/mongoid/association/eager_loadable.rb
```diff
@@ -104,31 +110,211 @@ def preload(associations, docs)
...
+      def polymorphic_keys_by_type(assoc, docs)
+        docs.each_with_object({}) do |doc, keys_by_type|
+          type, key = polymorphic_reference(assoc, doc)
+          (keys_by_type[type] ||= []) << key if type
+        end
+      end
+
+      # Fetch every target in one aggregation: a $facet with a $lookup branch per
+      # type. Returns the targets as { type => { primary_key => document } }.
+      def fetch_polymorphic_targets(assoc, keys_by_type)
```
jamis: The `cross_cluster_inclusions` method now (correctly) skips polymorphic associations, but I think that just means this method (`fetch_polymorphic_targets`) needs to make sure a single generated pipeline at this end does not attempt to reference collections in disparate clusters. The risk here is that someone might issue an `eager_load` on polymorphic collections that exist in different clusters, and the fetch at this point will then fail (silently or otherwise).

jamis (changes requested):
This is elegant! Definitely a challenge to reason about cleanly, which gives me a bit of (maintenance-related) concern, but I think adding some documentation to the private methods will help with that.

cperezabo:
My intention is, once this PR is merged, refactor and reificate some abstractions so this can be easier to maintain and extend. I haven't do it because I didn't want to make the PR longer. I decided to keep the already existing god-class and its imperative code as-is and move forward. 
So, this is the first step: tests + implementation that works.
Then, refactoring.

cperezabo:
The good part is that everything is covered with easy-to-read tests. So they are well defined guardrails to make refactoring.

jamis:
> My intention is, once this PR is merged, refactor and reificate some abstractions so this can be easier to maintain and extend. I haven't do it because I didn't want to make the PR longer. I decided to keep the already existing god-class and its imperative code as-is and move forward. So, this is the first step: tests + implementation that works. Then, refactoring.

I can appreciate that. However, if the refactoring and other work relates directly to this feature, I'd rather see a single complete PR than a series of PR's that continue iterating on the same theme. On the other hand, if the refactoring is broader in scope and does _not_ directly relate to this feature, then it makes sense to do it in a separate PR.

The goal (on my end) is to make sure any code that lands in master is _production ready_. If I were planning to cut a new release of Mongoid tomorrow, would this PR be of sufficient quality? If (heaven forbid!) you were to be hit by a bus tomorrow, and your proposed refactorings never land, would the PR as it currently stands be "good enough"? I think it's almost there; if you make at least the changes I proposed, I'd be happy to merge it at that point.

cperezabo:
I already did the changes you mentioned. You can see the commits just above your last comment haha. I'm now working in a full refactoring by the way.

jamis:
Sorry for the delay -- there was a very recent change to the i18n gem that broke our CI runs with Ruby 3.1, and I was trying to get it fixed. :) This looks great, thank you! Could you pull in the latest changes from master, so I can get the tests here green?

cperezabo:
Hi @jamis, can you take a look at the changes?

lib/mongoid/association/eager_load/inclusion.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+
+module Mongoid
+  module Association
+    module EagerLoad
+      # An inclusion of an eager load, in the role it plays while the pipeline is
+      # built. Each kind knows how it contributes; the LookupPipeline holds the
+      # stage-building helpers they lean on. A node carries its own children, so
+      # the pipeline is built by recursion from the roots downward.
+      class Inclusion
```
jamis, line 13: Let's limit the surface area of the public API by marking these classes with `@api private`. They represent an implementation detail and aren't intended for public consumption.

lib/mongoid/association/eager_load/inclusion.rb
```diff
@@ -0,0 +1,120 @@
...
+      class JoinedInclusion < Inclusion
+        class << self
+          # The default kind: a referenced, non-polymorphic association, i.e. the
+          # one no sibling kind claims.
+          def for?(association)
+            (superclass.subclasses - [ self ]).none? { |kind| kind.for?(association) }
+          end
+        end
+
+        def contribute(destination, chain)
```
jamis, line 106: A YARD comment would be helpful here. I know it probably partly duplicates the YARD comment on the superclass `#contribute` method, but it's nice to have them in place.

lib/mongoid/association/eager_load/inclusion_tree.rb
```diff
@@ -0,0 +1,70 @@
...
+  module Association
+    module EagerLoad
+      # The tree of nested inclusions an eager load asks to load. Built from the
+      # criteria's inclusions, it contributes each root node's stages to the
+      # pipeline, with each node already carrying its own children.
+      #
+      # Each root branch is built from its own copy of the inclusions, and an
+      # inclusion is removed as it is placed, so it lands once per branch even if
+      # more than one parent in that branch points at it, and a circular chain of
+      # inclusions can't loop forever.
```
jamis, line 16: Probably worth marking this as `@api private` as well.

lib/mongoid/association/eager_load/lookup_pipeline.rb
```diff
@@ -0,0 +1,119 @@
...
+      #         { '$lookup' => {             # JoinedInclusion(:tracks), nested
+      #           'as' => 'tracks',
+      #           'pipeline' => [
+      #             { '$sort' => {
+      #               '_id' => 1
+      #             } }
+      #           ]
+      #         } }
+      #       ]
+      #     } } ]
```
jamis, line 38: `@api private` here, too

lib/mongoid/association/eager_load/polymorphic_preloader.rb
```diff
@@ -0,0 +1,59 @@
...
+module Mongoid
+  module Association
+    module EagerLoad
+      # Resolves a polymorphic belongs_to onto already-materialized root documents.
+      #
+      # A polymorphic belongs_to can't be expressed as a $lookup: its target
+      # collection varies per document. So once the roots are materialized, the
+      # foreign keys are grouped by type, PolymorphicTargets resolves the documents
+      # for those keys, and the result is set on each document.
+      class PolymorphicPreloader
```
jamis, line 16: `@api private`

(and so forth -- probably most of these new classes should be for internal use only)

lib/mongoid/association/eager_load/polymorphic_targets.rb
```diff
@@ -0,0 +1,138 @@
...
+      # The targets of a polymorphic belongs_to, indexed as
+      # { type => { primary_key => document } }. Each subclass reaches the types that
+      # live in one place (the root's database or elsewhere); .for resolves the whole
+      # set, routing each type to the subclass that can reach it.
+      class PolymorphicTargets
+        class << self
+          # Resolve every polymorphic target for the foreign keys grouped by type.
+          # The types whose documents share the root's database are fetched together
+          # in one $facet; those living elsewhere are read through their own models.
+          # Returns them indexed as { type => { primary_key => document } }.
```
jamis: Any public method should have YARD documentation, even if the class itself is private. (For internal classes, the consistent documentation is a big help to the maintainers.)

lib/mongoid/association/eager.rb
```diff
@@ -12,18 +12,12 @@ class Eager
...
       #   Associations to eager load
       # @param [ Array<Document> ] docs Documents to preload the associations
-      # @param [ Boolean ] use_lookup Whether to use $lookup aggregation
-      #   for eager loading. This is used in Criteria#eager_load.
-      # @param [ Array<Hash> ] pipeline The aggregation pipeline to use
-      #   when using $lookup for eager loading.
       #
       # @return [ Base ] The eager load preloader
-      def initialize(associations, docs, use_lookup = false, pipeline = [])
+      def initialize(associations, docs)
```
jamis: Note that removing parameters from a public method changes the (unfortunately) public API, and we have to consider it a breaking change. (I know that there is little chance anyone is directly using this class in their own code, but because we didn't flag the class as `@api private`, we have to assume that someone _might_ be.)

Instead, let's keep the parameters, and move the bit in EagerLoadable (line 119) back into `Eager#run` (branching based on the value of `use_lookup`). That will preserve the public API.

If having that code extracted is preferable, we can then mark this class as `@api private`, and add a deprecation warning in this initializer when `use_lookup` is true. I'm not sure what we'd direct people to do instead, but that's where we'd do it. In a later release, when we bump the major version next, we could do the proper refactoring.

jamis (changes requested):
I like this refactoring -- it cleans a lot of things up nicely. There are a few things I've noted, and Copilot raises some good points as well (though I think some of them are probably pre-existing issues that Copilot is only surfacing because they were moved around in the refactoring; if so, feel free to indicate that and we can address them in a different PR).

cperezabo:
Well, I've addressed:
- `@api private` on the new classes (@jamis)
- YARD where it was missing (@jamis)
- restored Eager signature to prevent breaking change (@jamis)
- cross-database fallback (Copilot)
- embeds_one synthesis (Copilot)
- same-name-across-subclasses overwrite (Copilot)

Regarding [this](https://github.com/mongodb/mongoid/pull/6158#discussion_r3455439469), I couldn't reproduce, so left it

jamis (approved):
Thank you for your work on this! It looks great. 👍

PR #6161, opened 2026-06-26
MONGOID-5939 Guard against NoSQL operator injection in where/find_by

comandeo-mongo:
## What

Adds a guard in `expr_query` that rejects top-level `$` operators not in an explicit allowlist when `allow_unsafe_query_operators` is set to `false`. The default is `true` (existing behavior preserved — no BC break).

To opt into the strict mode:

```ruby
Mongoid.configure do |config|
  config.allow_unsafe_query_operators = false
end
```

With strict mode on, operators like `$where` and `$function` raise `Errors::InvalidQuery`. Allowed operators: `$and`, `$or`, `$nor`, `$not`, `$text`, `$comment`, `$expr`, `$jsonSchema`, `$alwaysFalse`, `$alwaysTrue`.

## Why

When user-supplied params are passed directly to `where`/`find_by` (e.g. via `to_unsafe_h`), an attacker can inject top-level operators like `$where` to execute arbitrary JavaScript on the MongoDB server.

## Files changed

- `lib/mongoid/config.rb` — new `allow_unsafe_query_operators` option (default: `true`)
- `lib/mongoid/criteria/queryable/selectable.rb` — `ALLOWED_QUERY_OPERATORS` constant and guard in `expr_query`
- `spec/mongoid/criteria/queryable/selectable_where_spec.rb` — 14 new examples covering the guard
- `Gemfile` — add `ostruct` for Ruby 4.0+ compatibility

## Test plan

- [x] 14 new examples in `selectable_where_spec.rb` covering blocked operators, allowlisted operators, and opt-in strict mode
- [x] All 1681 queryable specs pass with no regressions
- [x] RuboCop reports no offenses

Jira: https://jira.mongodb.org/browse/MONGOID-5939

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,3 +25,5 @@ gem 'i18n', *i18n_versions
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem "tzinfo-data", platforms: %i[ windows jruby ]
+
+gem 'ostruct' if RUBY_VERSION >= '4.0'
--- a/lib/mongoid/config.rb
+++ b/lib/mongoid/config.rb
@@ -294,6 +294,15 @@ module Mongoid
     # See https://jira.mongodb.org/browse/MONGOID-5892 for more details.
     option :serializable_hash_with_legacy_only, default: true
 
+    # When true (default), all top-level query operators are passed through
+    # to MongoDB without restriction when using +where+/+find_by+. Set to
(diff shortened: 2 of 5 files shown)


jamis (changes requested):
There are several rubocop complaints from the new code, but otherwise it looks good 👍

jamis (approved):

